PR #16, opened 2019-01-02
Forward any body of the request to the mock route handler

### LINKED ISSUE #13: Forward any body of the request to the mock route handler
I ran into a usecase whereby I needed to personalize mock responses based on the request's body for `POST` requests:

```
msw.post(
  "http://localhost:3000/cats",
  (req, res, { status, set, delay, json }) => {

    if(req.body) {
        // do something with body
    }

    return res(status(404), delay(10000), json({ errorMessage: `no cats :(` }));
  }
);
```

------

I can send a quick PR for this if there's interest for this.

dashed:
Closes https://github.com/kettanaito/msw/issues/13

The Fetch API for `Body` has several methods to resolve the body contents: https://developer.mozilla.org/en-US/docs/Web/API/Body

Only `Body.text()` and `Body.json()` are useful since they can be serialized into JSON. 

If the contents of the body cannot be resolved either as text (e.g. it's empty), or as JSON (e.g. parsing error), then their respective values will be `undefined` (or `void 0`).

--------

@kettanaito This PR was something similar I had to augment for the `mockServiceWorker.js` file I used.

### DIFF
--- a/mockServiceWorker.js
+++ b/mockServiceWorker.js
@@ -64,6 +64,13 @@ self.addEventListener('fetch', async (event) => {
         reqHeaders[name] = value
       })
 
+      /**
+       * If the body cannot be resolved (either as JSON or to text/string),
+       * the default value will be undefined.
+       */
+      const json = await req.json().catch(() => void 0)
+      const text = await req.text().catch(() => void 0)
+
       const clientResponse = await messageClient(client, {
         url: req.url,
         method: req.method,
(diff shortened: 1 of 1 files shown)


mockServiceWorker.js
```diff
@@ -74,6 +81,8 @@ self.addEventListener('fetch', async (event) => {
         redirect: req.redirect,
         referrer: req.referrer,
         referrerPolicy: req.referrerPolicy,
+        json: json,
+        text: text,
```
kettanaito:
```diff
-  json: json,
-  text: text,
+  body: json || text,
```

Shouldn't this be available under uniform `req.body`?

kettanaito:
Thank you for the pull request!

Please, I had only one question regarding how a request's body should be referenced. Could you give it a look?

dashed:
@kettanaito Yep makes sense 👍 I've amended the commit with your feedback.

kettanaito (approved):
Great! Approved.

PR #41, opened 2020-03-01
fix: example project mock api request params not found

### LINKED ISSUE #43: "req.params" does not contain my URL path parameter
## Environment

- `msw: 0.4.2`

## Current behavior

When defining the following request handler:

```js
rest.get('/users/:userId', (req, res, ctx) => {
  return res(ctx.text(req.params.userId)
})
```

The `req.params.userId` reference returns `undefined`. I can see that all the parameters are nested in the `req.params.params` key, which is not the correct behavior.

## Expected behavior

Request URL path parameters are listed on the `req.params` object, containing only key-value pairs of parameter name and parameter value. There must be no other information (like `matches`).

hehehai:
- Fixes #43 

request params 
```javascript
{
	matches: true
	isExact: true
	params: {username: "xxx"}
}
```

### DIFF
--- a/example/src/mocks/index.js
+++ b/example/src/mocks/index.js
@@ -4,7 +4,7 @@ import userMocks from './user'
 const { start } = composeMocks(
   ...userMocks,
   rest.get('https://api.github.com/users/:username', (req, res, { json }) => {
-    const { username } = req.params
+    const { username } = req.params.params
 
     return res(
       json({
--- a/example/src/mocks/user.js
+++ b/example/src/mocks/user.js
@@ -6,7 +6,7 @@ import { rest } from 'msw'
  */
 export default [
   rest.get('https://api.backend.com/user/:username', (req, res, { json }) => {
-    return res(json({ username: req.params.username }))
+    return res(json({ username: req.params.params.username }))
   }),
   rest.post(/user\/create/, (req, res, { json }) => {
     return res(json({ message: 'User has been created!' }))


kettanaito:
Hello, @hehehai. Thank you for submitting this pull request!

It looks like the library assigns the wrong value to the `requestWithParams.params` object during request handling:

https://github.com/open-draft/msw/blob/1f212a1c8f6383a4e9f7d33a939843db74f8d0b7/src/handleIncomingRequest.ts#L30-L38

The call to `matchPath()` returns the Object you've described above (), but only the `params` key needs to be passed as a value to `requestWithParams.params`. I think this is what should happen:

```js
 // Retrieve request URL parameters based on the provided mask 
 const match = relevantRequestHandler.mask 
   ? matchPath(req.url, { path: relevantRequestHandler.mask }) 
   : {} 
  
 const requestWithParams = { 
   ...req, 
   params: match.params,
 } 
```

Could you please fix that? That'd be awesome.

kettanaito (changes requested):
Please see my explanation in the comment above. In case of any question please ask, I'd be happy to help.

hehehai:
```ts
const match = relevantRequestHandler.mask
    ? matchPath(req.url, { path: relevantRequestHandler.mask })
    : { params: {} }

const requestWithParams = {
	...req,
    params: match.params,
}
```

match: FullMath | {}

so `match.params` maybe `undefined`

hehehai:
```ts
const { params = {} } = relevantRequestHandler.mask
      ? matchPath(req.url, { path: relevantRequestHandler.mask })
      : {}
```
or
use `&&` 
There will be type conversions. I don’t know much about the types that xx may return. I only know whether it is `RegExp` or `string`, but the type conversion of `&&` may have unexpected values
```ts
const { params = {} } = relevantRequestHandler.mask
      && matchPath(req.url, { path: relevantRequestHandler.mask })
```

kettanaito:
Yeah, that's a valid concern. I'll update with more details on how `matchPath` operates below.

When called, `matchPath` will _always_ return an object of the following structure:

https://github.com/open-draft/msw/blob/1f212a1c8f6383a4e9f7d33a939843db74f8d0b7/src/utils/matchPath.ts#L51-L57

This means that when we check whether there is a URI mask to match against the actual request URL the call to `matchPath` always resolves to an Object. In case there is no mask provided, we explicitly set a fallback to an empty object, so `{}.params` in that case produces `undefined` instead of an exception. Here's how it's implemented at the moment:

https://github.com/open-draft/msw/blob/1f212a1c8f6383a4e9f7d33a939843db74f8d0b7/src/handleIncomingRequest.ts#L30-L38

Since the return payload of `matchPath` is not the parameters, but a slightly complicated structure, I suggested to rename that variable from `params` to `match` (suggestions welcome). Preserving the same value as there is now we ensure it's always an object:

```js
const match = relevantRequestHandler.mask 
   ? matchPath(req.url, { path: relevantRequestHandler.mask }) 
   : {}
```

I would not complicate this value assignment at this point. 

Now our match will contain `match.params` in case there is a URI mask present. That's what we want to pass to the `requestWithParams` object down the line. You make a good point regarding `matchPath()` function and that is may not return any `params`. This would mean that `match.params` may be `undefined`, and referencing that object in your response resolver would throw an error (i.e. `req.params.foo`). 

To prevent that, let's have a fallback for the parameters as well:

```js
const requestWithParams = { 
  ...req, 
  params: match.params || {}
} 
```

@hehehai, what do you think about this suggestion? Does it make sense to you?

hehehai:
That's good. The semantics of this implementation will be clearer, making `math` assignment less complicated.

hehehai:
image

ts error

kettanaito:
@hehehai could you please screenshot the error after hovering over the highlighted word?

I have a suspicion that's TS compaining on `params` property not existing on an empty object, but let's check to be safe. Thanks.

hehehai:
yes，ts think `match` type is `{}`, not has `params`

kettanaito:
Got it. What do you say if we explicitly add the `params` property to the empty fallback object?

```ts
const match = relevantRequestHandler.mask
  ? matchPath(req.url, { path: relevantRequestHandler.mask })
  : { params: null }
```

I wonder what do you think about this approach. Please let me know.

hehehai:
image

`{params: null}` vs `{params: {}}`
image

Which one do you think would be more suitable?

kettanaito:
@hehehai, that's a great catch! You're right, I think that having `{ params: {} }` instead of `{ params: null }` makes much more sense. Absolutely agree on that decision.

kettanaito:
Fixed by #46

PR #45, opened 2020-03-02
Adds GraphQL request handlers

### LINKED ISSUE #39: Support mocking GraphQL API
I suggest that MSW would provide a standard set of request handlers for mocking a GraphQL API.

## Motivation

- GraphQL is an adopted technology, mocking which would be beneficial for users
- Mock-driven GraphQL development may be a good try-out for the technology, as you don't have to write any server-side code before deciding to fully adopt GraphQL

## Features

- Ability to target GraphQL operations based on type (query/mutation) and name
- Ability to mock response data and errors according to GraphQL specification

### Mocking a query

```js
import { composeMocks, graphql } from 'msw'

const { start } = composeMocks(
  graphql.query({ operation: 'GetProjectDetail' }, (req, res, ctx) => {
    const { projectPath } = req.variables

    return res(
      ctx.data({
        data: {
          name: 'mocked-project'
        },
      }),
    )
  }),
)

start()
```

### Mocking a mutation

```js
import { composeMocks, graphql } from 'msw'

const { start } = composeMocks(
  graphql.mutation({ operation: 'GetProjectDetail' }, (req, res, ctx) => {
    return res(
      ctx.data({
        data: {
          name: 'mocked-project'
        },
      }),
    )
  }),
)

start()
```

kettanaito:
- Closes #39 

## Roadmap

- [x] Add the `graphql` request handlers to the standard request handlers
- [x] Add the `graphql.query()` request handler
- [x] Expose `req.variables` to the response resolver
- [x] Set `req.variables` to an empty object, if no variables are set
- [x] Make context utilities request handler-specific
- [x] Add `ctx.errors()` utility for mocking GraphQL errors
- [x] Add the `graphql.mutation()` request handler
- [x] Support RegExp as `selector.operation` to handle multiple GraphQL operations matching the expression
- [x] Parse the query to determine operation name (don't rely on Apollo, not everybody uses it)
- [x] Add integration tests
  - [x] Query
  - [x] Mutation
  - [x] Variables
  - [x] Errors


kettanaito:
GraphQL subscriptions will be introduced later, when the necessary research is done on the topic. Query and mutation handling is a meaningful functionality for a developer.

Issue #21, opened 2019-05-12
Handle multiple headers with the same name


kettanaito:
https://github.com/kettanaito/msw/blob/fb1e53cf9cb719ea03c7a3f964461cfb84aa54e2/mockServiceWorker.js#L59-L65

Need to ensure multiple headers with the same name can be handled by MockServiceWorker.

This may also mean to adjust the current algorithm for handling headers between client/sw.

## Surface

- [ ] Adjust serialization/deserialization of `headers` in MSW and `interceptRequest`
- [ ] Adjust public API of `context.set()` to support setting of multiple headers with the same name
- [ ] Add tests

kettanaito:
> Noting down our productive discussion with @redraushan below.

## Data structure

I suggest to store the multiple headers in a list of tuples:

```js
// mockServiceWorker.js
// This would need to be constructed dynamically from "req.headers"
const headers = [['name', 'value'], ['name', 'value']]
```

Current implementation mirrors the `req.headers` into a new object instance:

https://github.com/open-draft/msw/blob/6b4242845bb6d301252fbd41c0f4818ce4883e5b/mockServiceWorker.js#L62-L65

## De-serialization

When accepted from the MSW on the client side in `interceptRequest()`, the custom reviver function during JSON parsing should be adjusted to handle the list of tuples. Now it passes the object to `new Headers()` right away.

https://github.com/open-draft/msw/blob/6b4242845bb6d301252fbd41c0f4818ce4883e5b/src/utils/interceptRequest.ts#L15-L17

kettanaito:
I was going through the open issues and this is one of the most needed things at the moment. I was wondering if you have the time/desire to implement this, @hehehai? Feel free to say no, I'd try to do this some time next week in that case.

redraushan:
@kettanaito this was the issue that I discussed with you, it’s my bad I literally forgot to implement the changes. If anyone is not taking in 2-3 days, it’s mine 😅

hehehai:
@kettanaito  I'm sorry, I don't have any ideas at the moment😐

PR #46, opened 2020-03-03
Fixes request params incorrectly return

### LINKED ISSUE #43: "req.params" does not contain my URL path parameter

hehehai:
- Fixes #43
- Successor to #41

### DIFF
--- a/src/handleIncomingRequest.ts
+++ b/src/handleIncomingRequest.ts
@@ -28,13 +28,13 @@ export const createIncomingRequestHandler = (
     }
 
     // Retrieve request URL parameters based on the provided mask
-    const params = relevantRequestHandler.mask
+    const match = relevantRequestHandler.mask
       ? matchPath(req.url, { path: relevantRequestHandler.mask })
-      : {}
+      : { params: {} }
 
     const requestWithParams = {
       ...req,
-      params,
+      params: match.params,
     }
 
     const mockedResponse:


kettanaito:
Awesome! Welcome to contributors! :tada:

hehehai:
Thanks, this is my first pr on github, this is great!

kettanaito:
@hehehai way to go! It's been a pleasure discussing things with you.
If you've got a wish to continue, I'd appreciate your feedback on #39. There's also a #21 to do, which shouldn't be difficult after some short discussion. Anyways, much thanks!

hehehai:
```ts
const params = relevantRequestHandler.mask
	? matchPath(req.url, { path: relevantRequestHandler.mask }).params
    : {}

const requestWithParams = {
    ...req,
    params,
}
```
@kettanaito  Sorry, I am too entangled in the assignment, it will be clearer to take out the params directly from the match.

There is no type error

kettanaito:
@hehehai, hm. Well, it seems that full match is not used for anything but parameters, so go ahead, sounds reasonable to me. Could you please raise an issue about this suggestion first? 🙏

hehehai:
@kettanaito ok

PR #53, opened 2020-03-04
Prepends host origin to the routes starting with a leading slash

### LINKED ISSUE #52: Forward leading slash to match the page's hostname
Current behavior expects the developer to either specify an entire hostname (`http://backend.com/users`) or use a wildcard (`*/users`) to handle request URI.

## Motivation

It's familiar to developers to specify routes with a leading slash and expecting the hostname to be matched implicitly.

## Expected behavior

When provided the route `/users` MSW would match any `/users` requests that are fired from the _current page's hostname_:

- `http://localhost:8080/users`

**However, it won't match the same paths from the different hostname:**

- `https://api.dev.backend/users`

kettanaito:
## GitHub

- Fixes #52

### DIFF
--- a/src/handlers/rest.ts
+++ b/src/handlers/rest.ts
@@ -8,7 +8,17 @@ const createRESTHandler = (method: RESTMethods) => {
       mask,
       predicate(req) {
         const hasSameMethod = method === req.method
-        const urlMatch = match(mask, req.url)
+
+        // Prepends a host origin to the routes that start
+        // with the slash ("/"). This way such routes will match
+        // the respective hostname's routes, while bypassing
+        // the routes from different hosts.
+        const resolvedMask =
+          typeof mask === 'string' && mask.startsWith('/')
+            ? `${location.origin}${mask}`
+            : mask
+
+        const urlMatch = match(resolvedMask, req.url)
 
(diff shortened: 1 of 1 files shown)


src/handlers/rest.ts
```diff
@@ -8,7 +8,17 @@ const createRESTHandler = (method: RESTMethods) => {
...
         const hasSameMethod = method === req.method
-        const urlMatch = match(mask, req.url)
+
+        // Prepends a host origin to the routes that start
+        // with the slash ("/"). This way such routes will match
+        // the respective hostname's routes, while bypassing
+        // the routes from different hosts.
+        const resolvedMask =
+          typeof mask === 'string' && mask.startsWith('/')
+            ? `${location.origin}${mask}`
```
kettanaito, line 18: I'm not tackling this on the `node-match-path` level, as such a low-level library must not care about the hosts, origins, etc. Therefore, I'm checking if dealing with the route that starts with a slash, and prepending an origin address (protocol + hostname + port, without routes).

kettanaito: I've chosen to prepend this information here for a purpose of preserving the routes in the mocking schema as-is. It would be easier for developers to navigate through the schema if their input is left intact.

PR #56, opened 2020-03-07
Adds integration tests using WDS and Puppeteer

### LINKED ISSUE #7: Add integration tests
## What:

Need to add integration tests.

## Why:

Self-explanatory.

## Roadmap:

- [x] Define minimum integration scenario setup (a story, a SPA, other)
- [x] Prepare the list of integration scenarios
- [x] Choose the testing framework (cypress, other)
- [x] Write integration tests

kettanaito:
## GitHub

- Closes #7 

## Changes

- Uses WDS and Puppeteer for integration testing
- Integrates new set of tests into CI
- Adds integration tests for currently present features (REST, request masking)
- Adds command line utility to run a focused test using WDS (for debugging)

## Roadmap

- [x] Remove `*.client.tsx` modules and use `page.evalate()` for dispatching requests on the page
- [x] `*.mocks.ts` cannot properly resolve the `msw` import on the TypeScript side (webpack compilation is valid)
- [x] Resolve the `spawnServer` promise when webpack compilation finishes, not when the server is established (this results into browser window spawning sooner than the test scenario is built, awaiting the compilation)
- [x] Configure Puppeteer to launch the browser on local in a head-full mode (on CI it must remain headless)

## Test suites

- [x] REST method-based handling of requests (GET matches, while POST doesn't)
- [x] String / Mask / RegExp request URI matching
- [x] All `rest` utilities (may be used in a single mock definition)
- [x] ~~Assert the responses are real when MSW is disabled~~


kettanaito:
## Why WDS?

I've chosen `webpack-dev-server` (WDS) to grab a client-side example code and run a temporary server with it. This way I can keep isolated examples that also represent test suites. WDS is much more handy than an express server, as it deals with the compilation process automatically. It also accepts a webpack compiler, which I can configure to handle my client-side examples as I wish (for example, by aliasing the imports to `msw` module to the local built module).

## Why Puppeteer?

I've considered the following technologies:

- WebdriverIO (Selenium)
- Cypress
- Puppeteer

### WebdriverIO

Felt like it's too low-level and I was writing **a ton of boilerplate** to get the trivial operations done. From my previous experience with Webdriver on another project I can conclude that it's quite flaky in certain scenarios.

### Cypress

I love Cypress. I've been using it for almost all my client-side libraries at this point. I usually build a Storybook and navigate through it with Cypress, which proves to be a great experience.

For MSW, however, Cypress is **too much setup for the value**. Cypress also requires extra tweaks to support TypeScript in both its configuration and loaded modules, which Jest does in one line of configuration (`ts-jest`).

### Puppeteer

I found Puppeteer being in a middle of Webdriver and Cypress: it ships with a useful high-level API and doesn't require to have _any setup_ at all. It also comes with the built-in `res.fromServiceWorker()` that determines if a specific response was served by a Service Worker, which is what MSW needs to assert.

kettanaito:
> Resolve the spawnServer promise when webpack compilation finishes, not when the server is established (this results into browser window spawning sooner than the test scenario is built, awaiting the compilation)

Not going to resolve that, as it seems to create some race conditions and overall works okay with Puppeteer awaiting the build.

> Configure Puppeteer to launch the browser on local in a head-full mode (on CI it must remain headless)

Should always run headless.

kettanaito:
> Assert the responses are real when MSW is disabled

Currently not possible to test reliably due to #58

PR #60, opened 2020-03-09
graphql: Accepts the operation name as a plain string/RegExp

kettanaito:
> This change introduced a **BREAKING CHANGE**. However, since the API of the library is not final, any version update should be considered carefully, as it may contain backward-incompatible changes.

- Shortens the declaration of `graphql.query()` and `graphql.mutation()` request handlers by accepting a string/RegExp directly (without the `{ operation }` wrapper).


kettanaito:
To preserve a proper semver this change should result into a minor version bump.

Issue #61, opened 2020-03-09
Success message of "msw init" contains a misleading API example


kettanaito:
## Current behavior

Current success message of running `msw init` command is:

https://github.com/open-draft/msw/blob/1fc124726f2d6b1e76e5cb7bfe306d94b3061acd/cli/init.js#L29-L34

The API example illustrated in the message is misleading, as the library does not export the `msw` namespace, neither that namespace has the `start()` method.

## Expected behavior

The message of successfully running the `msw init` command should link the developer to the documentation.

hehehai:
```js
console.log(`
${chalk.green('Service Worker successfully created!')}
Continue by creating a mocking module and starting the Service Worker：

  Define mocks: ${chalk.cyan.bold(
    'https://redd.gitbook.io/msw/getting-started#define-mocks',
  )}
`)
```
![image](https://user-images.githubusercontent.com/12692552/76284014-118a7a00-62d7-11ea-9fc0-54b98059af95.png)

Link to short link should be better, but short links on the Internet may not be very stable.

@kettanaito msw cli can feedback more help information, including information such as examples or documentation

now cli
```bash
$ npx msw
```
![image](https://user-images.githubusercontent.com/12692552/76283148-ccfddf00-62d4-11ea-82c2-dce291ee6cf9.png)

command check
```js
yargs
	// ......
	.help()
	.demandCommand().argv
```

```bash
$ npx msw
```

![image](https://user-images.githubusercontent.com/12692552/76283100-a93a9900-62d4-11ea-9dd4-d06e7223f16b.png)

command example
```js
(yargs) => {
      // builder
      yargs
        .positional('publicDir', {
          type: 'string',
          description: 'Relative path to server public directory',
          required: true,
          normalize: true,
        },
        .example('init', 'msw init public')
        .help()
    },
```
![image](https://user-images.githubusercontent.com/12692552/76283220-0fbfb700-62d5-11ea-82b1-31863b7f680b.png)

kettanaito:
@hehehai, that anchor to `#define-mocks` is a great idea! Agree, shortlinks take less space, but unless it's self-hosted it won't be reliable. Documentation's link structure may also change, but I see no bulletproof way of binding docs and this message. I'd go with the link you've suggested.

Maybe let's omit the `Define mocks:` part? The message above already explains what the link has to do.

```
$ npx msw init

Service Worker successfully created!
Continue by creating a mocking definition module and staring the Service Worker:

<LINK HERE>
```

[...]

PR #63, opened 2020-03-13
Fix cli feedback, enhancement  cli error params check and example

hehehai:
fix: cli successed feedback #61 
enhancement: https://github.com/open-draft/msw/issues/61#issuecomment-596919400


kettanaito (approved):
Awesome! Thank you.

PR #67, opened 2020-03-19
Supports response patching

### LINKED ISSUE #66: Response Patching (following the documentation) does not work
I am using the create-react-app in this project.
Following the documentation (https://redd.gitbook.io/msw/recipes/response-patching) and through various combinations of attempts to get it working, I either end up with errors like:
Uncaught TypeError: Cannot read property 'entries' of undefined
    at ServiceWorkerContainer.<anonymous> (index.js:1828)
or if I reconstruct the fetch call properly, it creates an infinite loop and still reports errors.

  "devDependencies": {
    "msw": "^0.8.2"
  }

kettanaito:
## GitHub

- Fixes #66 

## Changes

- Adds new `ctx.fetch()` utility to perform a bypassed request
- Supports bypassing of requests that have `x-msw-bypass` header set
- Sets the following new properties on the mocked `req` object:
  - `bodyUsed`
  - `integrity`
  - `destination`
  - `keepalive`

### DIFF
--- a/mockServiceWorker.js
+++ b/mockServiceWorker.js
@@ -52,18 +52,32 @@ self.addEventListener('fetch', async (event) => {
     new Promise(async (resolve) => {
       const client = await event.target.clients.get(clientId)
 
+      // Bypass requests while MSW is not enabled
       if (!client || !self.__isMswEnabled) {
         return resolve(getOriginalResponse())
+      }
+
+      // Bypass requests with the explicit bypass header
+      if (req.headers.get('x-msw-bypass') === 'true') {
+        const modifiedHeaders = serializeHeaders(req.headers)
+        // Remove the bypass header to comply with the CORS preflight check
+        delete modifiedHeaders['x-msw-bypass']
+
+        return resolve(
+          fetch(
+            new Request(req.url, {
+              ...req,
+              headers: new Headers(modifiedHeaders),
+            }),
+          ),
+        )
       }
 
       /**
        * Converts "Headers" to the plain Object to be stringified.
        * @todo See how this handles multipe headers with the same name.
        */
-      const reqHeaders = {}
-      req.headers.forEach((value, name) => {
-        reqHeaders[name] = value
-      })
+      const reqHeaders = serializeHeaders(req.headers)
 
       /**
        * If the body cannot be resolved (either as JSON or to text/string),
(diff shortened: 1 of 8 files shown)


kettanaito:
@hehehai could you please review this change? I'd be very thankful.

src/handleIncomingRequest.ts
```diff
@@ -27,9 +39,11 @@ export const createIncomingRequestHandler = (
     }
 
     // Retrieve request URL parameters based on the provided mask
-    const params = relevantRequestHandler.mask
-      ? match(relevantRequestHandler.mask, req.url).params
-      : {}
+    const params =
+      (relevantRequestHandler.mask &&
+        match(resolveRequestMask(relevantRequestHandler.mask), req.url)
```
kettanaito, line 44: With the latest release of `node-match-path` it returns explicit `null` in case there are no parameters. Handling that by introducing this change.

src/handleIncomingRequest.ts
```diff
@@ -40,7 +54,7 @@ export const createIncomingRequestHandler = (
 
     const mockedResponse:
       | MockedResponse
-      | undefined = relevantRequestHandler.resolver(
+      | undefined = await relevantRequestHandler.resolver(
```
kettanaito, line 57: Response resolvers are now async, allowing to perform async side-effects inside them, such as `ctx.fetch()`.

mockServiceWorker.js
```diff
@@ -52,18 +52,32 @@ self.addEventListener('fetch', async (event) => {
...
+              ...req,
+              headers: new Headers(modifiedHeaders),
+            }),
+          ),
+        )
+      }
+
       /**
        * Converts "Headers" to the plain Object to be stringified.
        * @todo See how this handles multipe headers with the same name.
```
hehehai, line 78: Spelling may be wrong：multipe，multiple

mockServiceWorker.js
```diff
@@ -79,10 +93,14 @@ self.addEventListener('fetch', async (event) => {
...
         mode: req.mode,
         credentials: req.credentials,
+        destination: req.destination,
+        integrity: req.integrity,
         redirect: req.redirect,
         referrer: req.referrer,
         referrerPolicy: req.referrerPolicy,
         body: json || text,
+        bodyUsed: req.bodyUsed,
+        keepalive: req.keepalive,
```
hehehai, line 103: Get keepAlive is more appropriate, just feel the style is unified some

hehehai (approved):
Great modification

mockServiceWorker.js
kettanaito: In terms of naming consistency I agree, but this is a property that is spelled that way (see the "Browser compatibility" table in [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) documentation).

kettanaito:
Thanks for the operative review, @hehehai! Much appreciated.

PR #78, opened 2020-03-22
Add Integration test that exposes response patching on a POST

skipdavis1117:
Running the 2 commands below shows the expected responses for a POST vs a GET to the same endpoint.  The test properly does response patching for a POST, but the data that is returned is from a GET, and so the test fails.

`curl --request POST --data '{"title":"foo","body":"bar","userId":"1"}' -H 'Content-Type: application/json' https://jsonplaceholder.typicode.com/posts`

`curl --request GET https://jsonplaceholder.typicode.com/posts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "msw",
-  "version": "0.10.0",
+  "version": "0.11.0",
   "description": "Deviation-less client-side runtime API mocking using Service Workers.",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",


test/rest-api/response-patching.test.ts
```diff
@@ -51,4 +51,42 @@ describe('REST: Response patching', () => {
       })
     })
   })
+
+  describe('given a post request to be patched', () => {
+    it('should be able to properly request and patch a post', async () => {
+      const REQUEST_URL = 'https://jsonplaceholder.typicode.com/posts'
+
+      const data = {
+        url: REQUEST_URL,
```
kettanaito: There's no need to provide the `url` property to the `RequestInit` object if you supply it as the first argument to `fetch()` (which you do):

```js
api.page.evaluate((url, init) => fetch(url, init), REQUEST_URL, data)
```

test/rest-api/response-patching.mocks.ts
```diff
@@ -28,6 +28,19 @@ const { start } = composeMocks(
       )
     },
   ),
+
+  rest.post(
+    'https://jsonplaceholder.typicode.com/posts',
```
kettanaito: I understand this is a placeholder response from a third-party service. Do you know if there's any end of life period for such responses? I'm worried that this placeholder will auto-close after some timeout and tests would start to fail.

For the cases when interaction with third-party APIs is unavoidable (like response patching) I usually pick a public API that's unlikely to fail (i.e. GitHub API). I'd advise to do the same for this test.

kettanaito (changes requested):
Thank you for adding this! I've got a few comments, but otherwise looks good.

I've also noticed that the commits you have are not connected to your GitHub account. It would be great if your contribution was associated with you, so I highly recommend to configure Git in your `msw` fork. You can do so by:

```bash
$ cd msw # checkout to your forked repository directory
$ git config user.email "<GITHUB_ACCOUNT_EMAIL>"
$ git config user.name "<GITHUB_USERNAME>"
```
> Provide the values _without_ the "<" and ">" symbols around.

Once done, you can re-author those commits you have by running:

```bash
$ git rebase -i b2c10e623a099e64a7fe3b33aab3a00da4007f17 # latest commit on original master
# mark all your new commits with "e" (stands for "edit") and close the Vim prompt.
# Git will go commit by commit, and you'd have to do:
$ git commit --amend --author="YOUR_NAME <EMAIL@FOO.COM>"
$ git rebase --continue
```

Verify the new commits not belong to you by doing `git log` and observing the author of the commits. Once this is done you can force push to your remote fork and the changes will take place.

> If you feel this manipulation is too troublesome you can leave things as-is, or I can try to re-author commits to point to your account for you.

test/rest-api/response-patching.test.ts
skipdavis1117: done.

kettanaito:
@skipdavis1117 just wanted to let you know in case you need any help, or have questions, I'm here to help. Will have some time to look into this issue about tomorrow in the evening.

skipdavis1117:
I was able to make the changes that you suggested with regards to the url
and authoring commits.  Finding a public POST API that is less likely to
fail, returns a 200, a common data set consistently, and can properly
expose the problem is proving more difficult.

Thanks for your patience.  I am new to JS and new to the opensource world.
Rest assured that I will not make the same mistakes twice.

kettanaito:
Please, no worries at all. I'm extremely thankful for your help in tackling this. Some things take time, so please take yours and we'll solve this issue together.

I've performed `POST https://jsonplaceholder.typicode.com/posts` request and can see it returning:

```json
{
  "id": 101
}
```

> I like your choice of JSON Placeholder, just would need to verify if those placeholders ever get pruned. Thinking about this, the most reliable way to test response patching would be to spawn a local server with certain endpoints and mock against those. Let's keep this thought for the future improvements, one thing at a time. I'll keep the progress on this suggestion in #79).

I presume that's the expected behavior. The test run on CI illustrates that the actual body received contains a list of hundreds of posts, which proves the issue reproducible.

skipdavis1117:
The jsonplaceholder does not actually create anything.  It is just a mock
that will always return the same response for a given request.  I'll look
into spinning up a super simple express server for the purposes of the
response patching integration tests.

Thanks!

kettanaito:
@skipdavis1117, oh I see. Maybe for simplicity's sake let's use JSON Placeholder for now? We can always improve in the future, but what we use for the actual server is not that important at the moment, I think. We'd better channel our effort into figuring out why `ctx.fetch()` seems to ignore the `init` part that contains `method` and other request parameters.

test/rest-api/response-patching.mocks.ts
skipdavis1117: We will stick with jsonplaceholder for now and add an express server for testing response patching later for the purposes resiliency.

skipdavis1117: We are going to leave this in place for now until we implement an express server for more robust healthchecks.

kettanaito:
My initial though is that this problem may be related to how Request-compliant object is created:

https://github.com/open-draft/msw/blob/b2c10e623a099e64a7fe3b33aab3a00da4007f17/src/context/fetch.ts#L36-L39

## Insights

The `req` object that MSW exposes to the end user does not fully implement the `Request` interface accepted by `fetch()`. MSW exposes a request-like object constructed from scratch in order to have already parsed body, parameters, query, and so on.

In order to have a `ctx.fetch(req)` call signature the `ctx.fetch` function must be able to augment the `MockedRequest` and compose a Request-compliant object from it to pass to `window.fetch()`. This is the area closest to the fetch invocation, makes sense to investigate it first.

If the area above does not prove resultful, the next step would be to see how Service Worker constructs the actual request. This is the surface where `ctx.fetch()` calls bubble to:

https://github.com/open-draft/msw/blob/b2c10e623a099e64a7fe3b33aab3a00da4007f17/mockServiceWorker.js#L66-L73

Issue #80, opened 2020-03-23
Handle thrown errors


kentcdodds:
I'd love it if errors my resolvers throw would get converted to rejected responses with a status code of 500. That way the app doesn't come crashing down if I hit something unexpected.

kettanaito:
Makes sense. I'd treat such errors as 500, but without the `X-Powered-By: "msw"` header set to distinguish between mocked 500 response and library's error.

@kentcdodds, could you please describe what use case led to an error? That way I will make sure I account for all scenarios. Thanks.

kentcdodds:
I have some utilities I use in my resolvers. For these utilities it's just easier to throw an error (because we don't have access to the context):

```
throw new Error('username must be provided')
```

I realize that I'm missing the status code for this case and could probably make something work for that.

Also, there are sometimes regular runtime errors for things I haven't considered and it would be nice if the whole app didn't just crash when I hit those.

kettanaito:
That's reasonable, thanks for sharing. 

I find the safest way of error handling in this case to be the `mockServiceWorker.js` itself. That brings up the question of how to ensure that the installed library operates with the last Service Worker file (#81). Since the Service Worker is served by the application, it's possible to have an old file served, while the client-side code expects a different behavior from it.

I'm going to tackle Service Worker's integrity firsts, and introduce internal error handling as the next step. Meanwhile, I suppose you can handle such errors by doing this:

```js
const { start } = composeMocks(/* your request handlers */)
start.catch(console.error)
```

> This will not take into account any exceptions raised by `mockServiceWorker.js`, but rather cover the errors in the client-side library.

kentcdodds:
Hmmm.. But the error is thrown by the resolver which is called by the client-side code, not the service worker right?

—
Kent C. Dodds

--- original message ---
On March 24, 2020, 3:28 PM MDT notifications@github.com wrote:



That's reasonable, thanks for sharing.

I find the safest way of error handling in this case to be the mockServiceWorker.js itself. That brings up the question of how to ensure that the installed library operates with the last Service Worker file (#81). Since the Service Worker is served by the application, it's possible to have an old file served, while the client-side code expects a different behavior from it.

I'm going to tackle Service Worker's integrity firsts, and introduce internal error handling as the next step. Meanwhile, I suppose you can handle such errors by doing this:

const { start } = composeMocks(/* your request handlers */)
start.catch(console.error)


[...]

PR #82, opened 2020-03-25
Adds Service Worker integrity check

### LINKED ISSUE #81: Assert "mockServiceWorker" integrity
## Motivation

The library must assert the integrity of the `mockServiceWorker.js` file for multiple reasons:

- Ensure the file is up-to-date, as it lives in the user's public directory and is not updated during new installations of `msw` package
- Ensure the file is actually a valid MSW file, to prevent malware masking as MSW

### Publication

1. Minimize
1. Remove comments
1. Generate a checksum out of the `mockServiceWorker.js` file content

### Client-side

1. Signal to Service Worker upon `start()` invocation to get the integrity information.
1. Assert the integrity on the client-side.
1. Display an error in case integrity fails. Suggest to run `npx init <PUBLIC_DIR>` to resolve the issue.

kettanaito:
## GitHub

- Closes #81 
- Relates to #80 

## Changes

- `mockServiceWorker.js` file is now modified and copied into the distribution directory. Its modification involve injection of an integrity checksum based on the file's minified and comment-free content. This is necessary to conclude whether the client-side library communicates with the running Service Worker that can understand it.
- Adds nicer activation logging into browser's console for better developer's experience. Activation message is now always visible and contains links to Documentation and repository.

**Rolling out these changes would result into all users of MSW to be requested to update their Service Worker file in the public directory.**

## Roadmap

- [x] Assert Service Worker integrity
- [x] Add integration test for integrity assertion
- [x] Sign Service Worker module during the publishing and in CI
- [x] Deal with changed files (`integrity.json`) after Service Worker signing
- [x] Fix the return statement of `composeMocks().start()`:

```
API: composeMocks / start › given a custom Promise chain handler to start() › should return instance of ServiceWorkerRegistration
```

- [x] Consider writing a small Webpack plugin that would copy the SW, inject checksum, set checksum as a `webpack.DefinePlugin()` variable for the library's code


.gitignore
```diff
@@ -5,3 +5,4 @@ lib
 tmp
 *-error.log
 ./package-lock.json
+integrity.json
```
kettanaito, line 8: Integrity file is never committed to the repository, but is published to NPM in order to do the integrity diffing.

Issue #55, opened 2020-03-07
Log matched requests in a browser's console


kettanaito:
Those requests which have matched the request handlers defined in the `composeMocks()` call should be logged into a browser's console.

## Motivation

- Easier and faster for the developer to see if something has been mocked

## Specification

1. Each request matched against the defined mock definition is logged into the console.
1. The log entry of the matched request looks as follows:

```
[MSW] 10:10:02 GET /users
```

3. The log entry is represented by `console.group`, clicking on which reveals the following information:

    3.1. Intercepted request.
    3.2. Request handler used.
    3.3. Resolved mocked response object.

hehehai:
@kettanaito I think it would be nice to have a configuration to turn it on. The mock's print message would be mixed with the application's print message, and while we can do some printing styles, it would be better for the developer to choose whether to turn it on or not. What do you think?

kettanaito:
@hehehai, a good call. It would be good accept such logging preference via an argument. I wonder, where should it be provided? 

I though of providing it to the `start()` function, but it accepts two optional arguments that control Service Worker registration, and may still be useful. Providing them both in order to set the third argument with the logging level feels redundant.

kettanaito:
Since Mock Service Worker is recommended for development only, I see little worry over seeing too many messages. At the least you can always filter the console messages based on type and RegExp to find what you were looking for.

On the other hand, logging matched requests would improve the developer's experience dramatically when working with the library.

PR #84, opened 2020-03-27
Fixes response patching of POST request

### LINKED ISSUE #77: Response patching for POST does not work
Using ctx.fetch inside the response resolver for a POST results in a GET being sent (with no body) instead of a POST (with a body).

Notes:
- The req before sending the original request showed the correct body and bodyUsed:true, but looking at the req in mockServiceWorker showed the body gone.  
I suspect the problem is in context/fetch.ts, but initial scan did not show an obvious issue to me (JavaScript beginner)

I wrote an integration test to PR into the repo that replicated the issue, but I did not have permission to push up a branch.

## Tests:

```ts
// response-mocking.mocks.ts:
 rest.post(
  'https://jsonplaceholder.typicode.com/posts',
  async (req, res, ctx) => {
    const originalResponse = await ctx.fetch(req)
    return res(
      ctx.json({
        ...originalResponse,
        mocked: true,
      }),
    )
  },
),
```
```ts
// response-patching.test.ts:
describe('given a post request to be patched', () => {
  it('should be able to properly request and patch a post', async () => {
    const REQUEST_URL = 'https://jsonplaceholder.typicode.com/posts'

    api.page.evaluate(
      (url) =>
        fetch(url, {
          method: 'POST',
          headers: {
          },
          body: JSON.stringify({
            title: 'foo',
            body: 'bar',
            userId: 1,
          }),
        }),
      REQUEST_URL,
    )
    const res = await api.page.waitForResponse((res) => {
      return (
        // Await for the response from MSW, so that original response
        // from the same URL would not interfere.
        res.url() === REQUEST_URL && res.headers()['x-powered-by'] === 'msw'
      )
    })
    const body = await res.json()

    expect(res.status()).toBe(200)
    expect(body).toEqual({
      id: 101,
      title: 'foo',
      body: 'bar',
      userId: 1,
      mocked: true,
    })
  })
})

kettanaito:
> Original pull request: #78 

## GitHub

- Fixes #77

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -92,19 +92,16 @@ self.addEventListener('fetch', async function(event) {
       }
 
       // Bypass requests with the explicit bypass header
-      if (request.headers.get(bypassHeaderName) === 'true') {
-        const modifiedHeaders = serializeHeaders(request.headers)
+      if (requestClone.headers.get(bypassHeaderName) === 'true') {
+        const modifiedHeaders = serializeHeaders(requestClone.headers)
         // Remove the bypass header to comply with the CORS preflight check
         delete modifiedHeaders[bypassHeaderName]
 
-        return resolve(
-          fetch(
-            new Request(request.url, {
-              ...request,
-              headers: new Headers(modifiedHeaders),
-            }),
-          ),
-        )
+        const originalRequest = new Request(requestClone, {
+          headers: new Headers(modifiedHeaders),
+        })
+
+        return resolve(fetch(originalRequest))
       }
 
       /**
--- a/test/rest-api/response-patching.mocks.ts
+++ b/test/rest-api/response-patching.mocks.ts
@@ -28,6 +28,19 @@ const { start } = composeMocks(
       )
     },
   ),
+
+  rest.post(
+    'https://jsonplaceholder.typicode.com/posts',
+    async (req, res, ctx) => {
+      const originalResponse = await ctx.fetch(req)
+      return res(
+        ctx.json({
+          ...originalResponse,
+          mocked: true,
+        }),
+      )
+    },
+  ),
 )
 
 start()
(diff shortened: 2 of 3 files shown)


kettanaito:
Hi, @skipdavis1117. Sorry for closing your original pull request (#78), I've got stuck with rebasing your remote branch and couldn't push the new state of the commits tree.

> My guess is that I couldn't push to your remote because the changes were residing on `master`. It hasn't happened before when pushing to remote feature branches.

If you don't mind, I've opened this follow-up with your commits preserved. Thanks for understanding.

kettanaito:
## Solution

I've looked into this issue and spotted that on the Service Worker level right before dispatching a bypassed request it operated on the `GET` request, despite the original `request` being of method `POST`. This was related to the destructuring of the `request` Object. Apparently, if you destructure a POST request it produces a GET request.

https://github.com/open-draft/msw/blob/16e536a22de1be266cd038c2811dae2d7fb71ecd/src/mockServiceWorker.js#L102-L105

When the request object is passed to the `Request` constructor directly, it inherits the method and other properties accordingly.

I've also used the `requestClone` in the bypassed request handling to prevent "request already sent" and similar errors. I think it's generally a good practice in terms of intermittent request handling.

kettanaito:
@skipdavis1117, could you please review this change so we could merge it? Thanks.

skipdavis1117:
No worries, you could just ask me to do it.  I'd be happy to instead of
making more work for you.  I am more interested in working product than my
feelings. I have not had much time to troubleshoot this week, but I think
you were right that the problem is likely in the Request creation and not
in the fetch.  I'll have a little more time this weekend.

Thanks

kettanaito:
@skipdavis1117, hey, absolutely no worries! You add no extra work for me. I've discovered a fix while debugging, so I decided to push it and not trouble you. The tests you've contributed are great, I'm extremely thankful!

There are a few open issues, if you have time take a look at any of them and voice your opinion. Ones I would recommend to look are #21, #55 (discussion mainly).

skipdavis1117:
Great!  I'll look at those

skipdavis1117 (approved):
Looks good.

PR #89, opened 2020-03-29
Fixes memory leak in integration tests

### LINKED ISSUE #83: Memory leak in integration tests
Current setup of integration tests does not handle certain asynchronous operations in its setup well, which results into memory leaks often happening when testing locally.

### Jest warning

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

> This was caused by the missing `afterAll(() => api.cleanup())` hook.

### Memory leak exception

```
<--- Last few GCs --->

[7874:0x102812000]    85649 ms: Mark-sweep 1381.6 (1455.2) -> 1371.0 (1455.2) MB, 742.3 / 0.0 ms  (average mu = 0.160, current mu = 0.072) allocation failure scavenge might not succeed
[7874:0x102812000]    86412 ms: Mark-sweep 1378.7 (1455.7) -> 1375.3 (1436.7) MB, 635.4 / 0.0 ms  (+ 78.4 ms in 60 steps since start of marking, biggest step 5.1 ms, walltime since start of marking 739 ms) (average mu = 0.111, current mu = 0.064) allocati

<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0xaae3f75fd61]
Security context: 0x0317e0f1e6e9 <JSObject>
    1: stringSlice(aka stringSlice) [0x317ad8136c1] [buffer.js:~589] [pc=0xaae4065e42f](this=0x0317141826f1 <undefined>,buf=0x0317639bbb89 <Uint8Array map = 0x317d50d5df1>,encoding=0x0317e0f3e981 <String[4]: utf8>,start=0,end=840098)
    2: toString [0x3170a7a0819] [buffer.js:~643] [pc=0xaae3f68c2fb](this=0x0317639bbb89 <Uint8Array map = 0x317d50d5df1>,encodi...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x10003d035 node::Abort() [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 2: 0x10003d23f node::OnFatalError(char const*, char const*) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 3: 0x1001b8e15 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 4: 0x100586d72 v8::internal::Heap::FatalProcessOutOfMemory(char const*) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 5: 0x100589845 v8::internal::Heap::CheckIneffectiveMarkCompact(unsigned long, double) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 6: 0x1005856ef v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 7: 0x1005838c4 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 8: 0x100590188 v8::internal::Heap::AllocateRawWithLigthRetry(int, v8::internal::AllocationSpace, v8::internal::AllocationAlignment) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
 9: 0x1005901df v8::internal::Heap::AllocateRawWithRetryOrFail(int, v8::internal::AllocationSpace, v8::internal::AllocationAlignment) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
10: 0x100562064 v8::internal::Factory::NewRawTwoByteString(int, v8::internal::PretenureFlag) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
11: 0x100561ca9 v8::internal::Factory::NewStringFromUtf8(v8::internal::Vector<char const>, v8::internal::PretenureFlag) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
12: 0x1001db1b8 v8::String::NewFromUtf8(v8::Isolate*, char const*, v8::NewStringType, int) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
13: 0x1000e8822 node::StringBytes::Encode(v8::Isolate*, char const*, unsigned long, node::encoding, v8::Local<v8::Value>*) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
14: 0x100056889 void node::Buffer::(anonymous namespace)::StringSlice<(node::encoding)1>(v8::FunctionCallbackInfo<v8::Value> const&) [/Users/kettanaito/.nvm/versions/node/v10.16.3/bin/node]
15: 0xaae3f75fd61
```

## Expected behavior

- Integration test cleans up all the side-effects its establishes before the next integration test starts.
- `runBrowserWith` and `spawnServer` utilities have tests that assert they clean up after themselves (i.e. asserting no process is running at the port previously occupied by the server after closing)

## Area of effect

- `runBrowserWith`
- `spawnServer`

kettanaito:
## GitHub

- Fixes #83

### DIFF
--- a/config/IntegrityWebpackPlugin/utils/copyServiceWorker.js
+++ b/config/IntegrityWebpackPlugin/utils/copyServiceWorker.js
@@ -24,11 +24,6 @@ module.exports = async function copyServiceWorker(
   }
 
   const destFileDirectory = path.dirname(destFilePath)
-  console.log(
-    'Checking if "%s" path exists...',
-    destFileDirectory,
-    fs.existsSync(destFileDirectory),
-  )
 
   if (!fs.existsSync(destFileDirectory)) {
     console.log('Destination directory does not exist, creating...')
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,9 +1,9 @@
 module.exports = {
   roots: ['<rootDir>/src'],
   transform: {
-    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.ts$': 'ts-jest',
   },
-  testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.(j|t)sx?$',
+  testRegex: '(.+)\\.(test|spec)\\.(j|t)s$',
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
   setupFiles: ['./jest.setup.js'],
 }
--- a/test/jest.config.js
+++ b/test/jest.config.js
@@ -8,6 +8,6 @@ module.exports = {
   },
   moduleFileExtensions: ['ts', 'js', 'json'],
   transform: {
-    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.ts$': 'ts-jest',
   },
 }
--- a/test/support/runBrowserWith.ts
+++ b/test/support/runBrowserWith.ts
@@ -7,13 +7,13 @@ export interface TestAPI {
   origin: string
   browser: puppeteer.Browser
   page: puppeteer.Page
-  cleanup: () => Promise<unknown>
+  cleanup: () => Promise<void>
 }
 
 export const runBrowserWith = async (
   mockDefinitionPath: string,
 ): Promise<TestAPI> => {
-  const { server, origin } = await spawnServer(mockDefinitionPath)
+  const { server, origin, closeServer } = await spawnServer(mockDefinitionPath)
 
   const browser = await puppeteer.launch({
     headless: true,
@@ -25,17 +25,10 @@ export const runBrowserWith = async (
   })
 
   const cleanup = () => {
-    return new Promise((resolve, reject) => {
-      browser
-        .close()
-        .then(() => {
-          server.close(resolve)
-        })
-        .catch(reject)
-    })
+    return browser.close().then(closeServer)
   }
 
-  process.on('exit', cleanup)
+  // process.on('exit', cleanup)
 
   return {
     server,
--- a/test/support/spawnServer.ts
+++ b/test/support/spawnServer.ts
@@ -7,12 +7,13 @@ import * as WebpackDevServer from 'webpack-dev-server'
 import * as HtmlWebpackPlugin from 'html-webpack-plugin'
 import * as packageJson from '../../package.json'
 
-interface Payload {
+export interface SpawnServerAPI {
   server: WebpackDevServer
   origin: string
+  closeServer: () => Promise<void>
 }
 
-export const spawnServer = (mockDefs: string): Promise<Payload> => {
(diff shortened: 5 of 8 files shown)


kettanaito:
The first thing I'm trying is removing the `--runInBand` flag for Jest test runner. Without it, Jest would create a separate child process for each test and run the test in the respective child process.

Observable behavior is such that without that flag Jest:

1. Serializes all tests matching the pattern.
1. Runs all tests simultaneously (with the flag I observe a one-by-one test runs)

Perhaps, due to the allocated child processes and simultaneous run there's less memory allocation/cleanup work necessary for the tests to perform.

## Resolution

Removing the `--runInBand` option does not help. On the contrary, it makes tests fail on CI, which never happened before. Segmentation fault and memory leak problems were experienced on local only.

kettanaito:
Increasing the amount of allocated memory using this command to `8000` (8GB) removes the problem, but it's not a solution:

```bash
$ node --max_old_space_size=8000 node_modules/.bin/jest ...
```

> The default memory limit of V8 engine is 1.5GB and is, apparently, exceeded when running test locally. This may be the diverging point to explain why tests don't fail on CI: most likely a CI machine has a higher memory limit set, which bypasses the out of memory exception.

kettanaito:
Switching to `ts-loader` provided a performance boost as compared to `awesome-typescript-loader`. The memory leak also didn't happen, however, the heap size kept growing between the tests and reached `~1GB`, which still indicates a smaller, but a memory leak.

kettanaito:
I conclude that the TypeScript compilation is what leaks the memory between test suites. It appears that closing the WDS via `wds.close()` does not clean up the memory allocated for TypeScript compiler to run. I also suspect that it runs in a watch mode, regardless whether the webpack compilation has `mode: 'production'` and `watch: true`, or not. 

As the next step I may try to compile the mock file manually via `tsc` and then hand it over to WDS. I'd like to refrain from running a server manually as much as possible.

> One option may also be to have a custom webpack plugin that compiles a given TS module into JavaScript and exits.

kettanaito:
Closing, since disabling hot module replacement in WDS resolved this issue (https://github.com/mswjs/msw/issues/83). 

We can migrate from the `awesome-typescript-loader` when the urge arises. So far I cannot notice any significant performance decrease whatsoever.

PR #91, opened 2020-03-31
Handles internal exception as a 500 response

### LINKED ISSUE #80: Handle thrown errors

kettanaito:
> This change bumps the integrity checksum of the Service Worker. Users would have to update the Service Worker, following the instructions in the integrity failure error message.

## GitHub

- Fixes #80 

## Changes

- `handleRequestWith` now executes in a `try`/`catch` block
- Any exceptions originating from a request handler are signaled to the Service Worker and it responds with a 500 response, including a suggestion error message to handle the exception, and the detailed original exception in the response's body

## TODO

- [x] Rewrite the successful mock signal from `channel.send(body)` to `channel.send({ type: '...', payload: body })` to be aligned with all the other signals

### DIFF
--- a/src/handleRequest.ts
+++ b/src/handleRequest.ts
@@ -14,79 +14,96 @@ import {
 export const handleRequestWith = (requestHandlers: RequestHandler[]) => {
   return async (event: MessageEvent) => {
     const channel = createBroadcastChannel(event)
-    const message: ServiceWorkerMessage<MockedRequest> = JSON.parse(
-      event.data,
-      (key, value) => {
-        // Serialize headers
-        if (key === 'headers') {
-          return new Headers(value)
-        }
 
-        // Prevent empty fields from presering an empty value.
-        // It's invalid to perform a GET request with { body: "" }
-        if (value === '') {
-          return undefined
-        }
+    try {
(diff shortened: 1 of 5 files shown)


src/handleRequest.ts
```diff
@@ -14,79 +14,97 @@ import {
...
+
+      channel.send({
+        type: 'MOCK_SUCCESS',
+        payload: responseWithHeaders,
+      })
+    } catch (error) {
+      channel.send({
+        type: 'INTERNAL_ERROR',
+        payload: {
+          status: 500,
```
kentcdodds, line 100: Could this be: `error.status || 500`?

kettanaito: Hm. Could you please give me an example of such error?

Please note that throwing in request handlers is not the recommended way of mocking server errors. I'd strongly recommend mocking server errors according to [this guide](https://redd.gitbook.io/msw/recipes/mocking-error-responses), which implies composing an error response via `res()`.

PR #99, opened 2020-04-13
Handles unregistering of the Service Worker

### LINKED ISSUE #79: Support local server for complex testing scenarios
> This issue originates from #77 

## What:

I suggest to add a support for spawning a local server for the purpose of testing. 

## Why:

There are complex test scenarios (such as "response patching") that assert against an actual server implementation. Relying on third-party servers brings a factor of instability to the CI pipeline of MSW. Having a local server for the sake of testing sounds like a viable option.

> The local server can also replace any external dependencies and be used as an actual running server. This is something to consider.

## How:

- [ ] Write a utility function that spawns an Express server, accepts the routes and returns a Promise that resolves whenever the server is up and running (spawn at port `0` for the server to always have a dedicated port)
- [ ] Return utility functions to perform any request against the current instance of the server (not to bind test to any specific ports)
- [ ] Adjust the `response-matching.test.ts` and any other tests that require an actual server to use the introduced local server implementation
- [ ] Close the server `afterAll()` test suites are finished

### LINKED ISSUE #96: Unregister self when not activated?
I have MSW working for a project on localhost:3000, then I close that project and open a new one which does *not* have MSW and that starts up on localhost:3000 (the default for create-react-app based apps) and MSW hangs around, logs in the console, and generally confuses people.

Can we have MSW's service worker unregister itself when it's not activated when the app starts up? I'd be fine if this is opt-in.

### LINKED ISSUE #98: Hard Reload (⌘+⇧+R) makes the service worker fail to handle fetches
If I hit <kbd>⌘</kbd>+<kbd>⇧</kbd>+<kbd>R</kbd> on a page where the service worker is working, then after the reload I'll see that the service worker is activated, but it doesn't handle fetch requests.

I made a video to demonstrate the issue: https://twitter.com/kentcdodds/status/1248642870413680641

Maybe this is related to this: https://github.com/w3c/ServiceWorker/issues/952

kettanaito:
### Service Worker instance handling

- Worker instance now keeps an internal `clients` map that holds all controlled clients.
- Worker instance provisions conditional mocking based on the `MOCK_ACTIVATE` event dispatched by a client. All clients have mocking disabled by default.
- Worker now bypasses requests with `mode: "navigate"` to prevent loading initial page load assets
- Client **no longer deactivates mocking** on `beforeunload` event. Instead, it dispatches the `CLIENT_CLOSED` event to notify the worker that the client has closed.
- Client now reloads the worker whenever a new worker instance has been installed.
- Calling `start()` now looks up for already running instances of the worker and updates them, instead of creating a new instance.
- Calling `stop()` now disables the mocking state for the client that called it. This allows different clients (tabs) to have different mocking state.

### Internal changes

- Adds `test.request()` utility that encapsulates a `fetch()` call in a page's context, and awaits for that request to finish, before resolving the promise.
- Adds a way to customize WDS routes by `withRoutes(app)`. This way WDS can be used as a local server for testing more complex server interactions scenarios (#79)

```ts
runBrowserWith(path.resolve('path/to/example.mocks.ts', {
  withRoutes(app) {
    app.post('/users', (req, res) => {
      res.status(200).json({ real: 'data' })
    })
  }
})
```

## GitHub

- Closes #79
- Closes #96 
- Potentially closes #98 

## Roadmap

- [x] Add a test suite on how Service Worker unregistration behaves with multiple clients (tabs) opened
- [x] Can there be a `mode: "navigate"` request that users may want to mock?
- [x] Declare the `clients` map on the worker instance (`self`) upon activation

### DIFF
--- a/src/composeMocks.ts
+++ b/src/composeMocks.ts
@@ -2,14 +2,12 @@ import { until } from '@open-draft/until'
 import { RequestHandler } from './handlers/requestHandler'
 import { handleRequestWith } from './handleRequest'
 import { requestIntegrityCheck } from './requestIntegrityCheck'
-import { addMessageListener } from './utils/createBroadcastChannel'
+import {
+  addMessageListener,
+  createBroadcastChannel,
+} from './utils/createBroadcastChannel'
 
 export type Mask = RegExp | string
-
-interface InternalContext {
-  worker: ServiceWorker
-  registration: ServiceWorkerRegistration
-}
 
 interface PublicAPI {
   start(
@@ -17,6 +15,11 @@ interface PublicAPI {
     options?: RegistrationOptions,
   ): Promise<ServiceWorkerRegistration>
   stop(): Promise<void>
+}
+
+interface InternalContext {
+  worker: ServiceWorker
+  registration: ServiceWorkerRegistration
 }
 
 const activateMocking = (worker: ServiceWorker) => {
@@ -50,6 +53,10 @@ const getWorkerByRegistration = (registration: ServiceWorkerRegistration) => {
   return registration.active || registration.installing || registration.waiting
 }
 
+/**
+ * Returns an active Service Worker instance.
+ * When not found, registers a new Service Worker.
+ */
 const getWorkerInstance = async (
   url: string,
   options: RegistrationOptions,
@@ -113,15 +120,13 @@ const createStart = (context: InternalContext): PublicAPI['start'] => {
       })
     })
 
-    // Deactivate mocking and unregister the active Service Worker
-    // when the window unloads.
     window.addEventListener('beforeunload', () => {
       if (worker?.state !== 'redundant') {
-        // Deactivating the mocking prevents the Service Worker
-        // from serving page assets on initial load. Otherwise
-        // it would expect the MSW to resolve a mock for assets,
-        // which would result into an infinite promise.
-        worker.postMessage('MOCK_DEACTIVATE')
+        // Notify the Service Worker that this client has closed.
+        // Internally, it's similar to disabling the mocking, only
+        // client close event has a handler that self-terminates
+        // the Service Worker when there are no open clients.
+        worker.postMessage('CLIENT_CLOSED')
       }
     })
 
@@ -158,6 +163,9 @@ const createStop = (context: InternalContext): PublicAPI['stop'] => {
    * Stop the active running instance of the Service Worker.
    */
   return async () => {
+    // Signal the Service Worker to disable mocking for this client.
+    // Use this an an explicit way to stop the mocking, while preserving
+    // the worker-client relation. Does not affect the worker's lifecycle.
     context.worker.postMessage('MOCK_DEACTIVATE')
   }
 }
--- a/src/composeMocks.ts
+++ b/src/composeMocks.ts
@@ -122,11 +122,6 @@ const createStart = (context: InternalContext): PublicAPI['start'] => {
         // it would expect the MSW to resolve a mock for assets,
         // which would result into an infinite promise.
         worker.postMessage('MOCK_DEACTIVATE')
-
-        // Unregister the Service Worker.
-        // This way it doesn't persist between page sessions, as well as
-        // doesn't affect other apps running on the same address.
-        registration.unregister()
       }
     })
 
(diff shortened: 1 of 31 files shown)


src/composeMocks.ts
```diff
@@ -14,53 +19,119 @@ interface PublicAPI {
...
-): PublicAPI['start'] => {
+const activateMocking = (worker: ServiceWorker) => {
+  worker.postMessage('MOCK_ACTIVATE')
+
+  return new Promise((resolve, reject) => {
+    // Wait until the mocking is enabled to resolve the start Promise
+    addMessageListener(
+      'MOCKING_ENABLED',
+      () => {
+        console.groupCollapsed(
```
kettanaito, line 33: Moving the successful activation message to the client-side, so the service worker is not subjected to integrity changes whenever the message's content is changed (contains potentially dynamic links to docs).

kettanaito:
## Client-specific mocking state

Service Worker can control multiple clients (think of it as browser tabs). With that, each individual client must have control over whether mocking should be enabled (via `start()` and `stop()` methods of MSW).

This means that the mocking state is de facto client-specific, while Service Worker instance _is one_. It cannot keep a global flag like `__mswEnabled`, as it does now. 

**Solution:** Keep a hash map of `{ [clientId: string]: boolean }` that would act as the source of truth regarding the mocking state. Service Worker can look up the id of the client that performs fetch (`event.clientId`) and decide on the mocking.

> Interestingly, each page refresh produces a client _with another ID_. This makes the previous written state connected to the same client obsolete in the hash map. The only issue this brings is that the hash map will grow in size between refreshes. Can be solved by pruning the hash map to contain only those keys which are currently active Service Worker clients (via `self.clients.matchAll()`).

## Accepted solution

The worker instance keeps an internal hash map of `clients` that represents all the controlled clients (transient mirror of `self.clients.matchAll()`. The entry in that map is created upon `MOCK_ACTIVATE` event, and removed upon `CLIENT_CLOSED` event. The values in that map are boolean flags that represent whether respective client should have mocking enabled. The values are changed in `MOCK_ACTIVATE`/`MOCK_DEACTIVATE` events.

kettanaito:
## Unexpected persistency on the same host

Since Service Worker controls a client on the given host, when you open another project on the same host, it will be controlled by the irrelevant Service Worker.

### Self termination

It's possible to self terminate the Service Worker from within its scope:

```js
// mockServiceWorker.js
self.registration.unregister()
```

However, it's not clear how to catch the event when there are no more active clients. 

### Suggested solutions

1. Ping the library (when?), and unregister if no response
1. Add a client-side listener on `beforeunload`, and check how many active clients controlled by the worker is left. If it's only one, unregister the service worker.

```js
// composeMocks.ts
window.addEventListener('unload', async () => {
  const allClients = await worker.clients.matchAll()

  if (allClients.length === 0) {
    registration.unregsiter()
  }
})
```

> One cannot get the list of clients from outside the `ServiceWorkerGlobalScope`.

## Accepted solution

Signal a custom `CLIENT_CLOSED` event upon `beforeunload` window event to the worker. In the worker, delete the `event.clientId` entry in the local `clients` hash map. If after the deletion there are no more clients left, unregister itself, as it means the last client window has been closed.

> This should also handle unexpected persistency (#96), as for the client to be set in the `clients` map it needs to dispatch the `MOCK_ACTIVATE` event. Once you close the last client that did so, the worker will unregister itself, even if there is another project running on the same host (which is impossible to have at the same time, so there is always a fact of closing/opening a new tab).

kettanaito:
## Can `navigate` request be mocked?

Per MDN:

> `navigate` — A mode for supporting navigation. The navigate value is intended to be used only by HTML navigation. A navigate request is created only while navigating between documents.

I conclude it to be very unlikely for developers to mock a navigation request between documents via `fetch()`. Instead:

- If you mock the _result_ of navigate, mock the resource you get.
- If you mock the _fact of navigation_, use `location.href`, or other methods to dispatch a proper navigation event.

src/mockServiceWorker.js
```diff
@@ -8,20 +8,23 @@
 /* tslint:disable */
 
 const INTEGRITY_CHECKSUM = '<INTEGRITY_CHECKSUM>'
-const bannerStyle = 'color:orangered;font-weight:bold;'
 const bypassHeaderName = 'x-msw-bypass'
 
+let clients = {}
```
kettanaito, line 13: Code declared in the global scope of the worker can be executed multiple times during the worker's lifecycle. This declaration should be moved to the worker's instance (`self`) to guarantee that the worker doesn't clear up its clients map unexpectedly.

kettanaito: I've tried setting it within `activate`, or keeping the declaration only in the global scope, yet both options yield broken worker logic. I'm leaving this as it is until further notice.

PR #102, opened 2020-04-17
Logs matched requests into browser console

### LINKED ISSUE #55: Log matched requests in a browser's console

kettanaito:
## Changes

- Logs matched requests into browser console
- Each log entry contains a timestamp that indicates when the mocking has happened
- Strips the hostname of the local requests
- Always strips query parameters from the logged request URL
- Status code is colored depending on the value (green, yellow, or red)

## Preview

Screen Shot 2020-04-17 at 20 19 41

## GitHub

- Closes #55

## Roadmap

- [x] Consider delayed responses

### DIFF
--- a/src/handleRequest.ts
+++ b/src/handleRequest.ts
@@ -10,6 +10,7 @@ import {
   ServiceWorkerMessage,
   createBroadcastChannel,
 } from './utils/createBroadcastChannel'
+import { log } from './logger'
 
 export const handleRequestWith = (requestHandlers: RequestHandler[]) => {
   return async (event: MessageEvent) => {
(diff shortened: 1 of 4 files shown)


src/handleRequest.ts
```diff
@@ -83,12 +84,14 @@ export const handleRequestWith = (requestHandlers: RequestHandler[]) => {
...
       // Transform Headers into a list to be stringified preserving multiple
       // header keys. Stringified list is then parsed inside the ServiceWorker.
-      const responseWithHeaders = {
+      const responseWithHeaders: MockedResponse = {
         ...mockedResponse,
         // @ts-ignore
         headers: Array.from(mockedResponse.headers.entries()),
       }
 
+      log(req, responseWithHeaders, relevantRequestHandler)
```
charlieforward9, line 93: For debugging tests in the IDE, where grouping is not built in and all the text prints into the console, it would be very helpful to disable this logging.

PR #109, opened 2020-04-20
Fix command

coryhouse:
Must use `npx` or the command won't work since `msw` isn't installed globally at step one.


kettanaito (approved):
Yet another eagle-eye catch! Thanks!

PR #126, opened 2020-04-29
Reloads the page with registrations but without controller

### LINKED ISSUE #122: Hard reload fails to use the service worker if there's another client connected
Remember the hard reload issue? #98 well, it looks like I get that exact same issue with the latest msw (0.14.1) if I have two browser tabs open on the site.

kettanaito:
## Changes

During the retrieving of an active worker instance (or creation of a new worker) the library checks if there are worker registrations associated with the client and no `navigatior.serviceWorker.controller` present. In that case it determines that a hard reload has happened, and triggers a regular reload following it.

## GitHub

- Relates to #98 
- Fixes #122


kettanaito:
I've performed manual testing to verify this fix works. Despite my attempts to automate this, the test becomes flaky and passes much more often than not (in case when there is no fix).

PR #127, opened 2020-04-29
rollup-integrity-check-plugin: Sets up watcher on "transform" hook

### LINKED ISSUE #86: Integrity plugin does not update runtime global variable
> Successor to #81 

Current setup of the `IntegrityWebpackPlugin` does not update the global variable injected to contain the latest Service Worker's integrity checksum.

## Reasons

First, the `webpack.DefinePlugin()` cannot be used with a simple string value for the injected variable, as it caches the modules by default, so the modules that reference the global variable are cached and won't access the updated checksum.

> I've migrated to using [`webpack.DefinePlugin.runtimeValue()`](https://github.com/webpack/webpack/issues/7717#issuecomment-604936270) instead, which is supposed to skip modules caching and result into updated runtime global variable.

Using `runtimeValue()` didn't solve the issue, as the changes in `src/mockServiceWorker.js` do not trigger the bundle recompilation. When forcing the recompilation the issue is still not fixed.

## Expected behavior

The global Service Worker integrity checksum variable must have the latest value in the code in the watch mode of compilation (i.e. using `yarn start`). This way the local development always has the Service Worker with the latest integrity.

kettanaito:
> Completely internal changes, no release intended.

## Changes

- Moves the watcher (`this.addWatchFile`) call to the `transform` hook of rollup. 
- Removes providing of a value in `renderChunk` hook, as it doesn't seem to have access to a plugin-wide variable `this.checksum` anyway.

## GitHub

- Closes #86

### DIFF
--- a/config/plugins/rollup-integrity-check-plugin/index.js
+++ b/config/plugins/rollup-integrity-check-plugin/index.js
@@ -15,15 +15,11 @@ module.exports = function integrityCheck(options) {
 
   return {
     name: 'integrity-check',
-    renderChunk(...args) {
-      return replace(injectChecksum(this.checksum)).renderChunk(...args)
-    },
     transform(...args) {
+      this.addWatchFile(input)
       return replace(injectChecksum(this.checksum)).transform(...args)
     },
     buildStart() {
-      this.addWatchFile(input)
-
       if (!fs.existsSync(input)) {
         this.error(`Failed to locate the Service Worker file at: ${input}`)
       }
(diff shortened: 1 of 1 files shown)


kettanaito:
While this change does result into the generated `cjs`/`umd` modules having a proper integrity checksum on incremental builds, there's still an integrity error during the runtime when working with test scenarios preview.

It appears that WDS serves the outdated version of the library's alias (`packageJson.main`), while the changes in the worker file propagate to the browser as a regular worker update (thus, failing the integrity handshake).

kettanaito:
Since integrity mismatch does not disable the mocking, I don't find the WDS serving outdated build module a critical issue to solve at the moment.

PR #138, opened 2020-05-05
Support tree shaking

### LINKED ISSUE #110: Tree shake graphql?
If I'm not using graphql, it'd be cool if that didn't end up in my client. Do you think it's possible to make the graphql dep tree-shakeable?

Sorry, I don't think the template applies here.

kettanaito:
## Changes

- Renames `composeMocks` directory to `setupWorker`.
- Moves `rest.ts` and `graphql.ts` to the root of the `src`, to resemble the import structure.
- Adds multiple build entries, so they are properly cross-referenced and can be tree-shaken.

## GitHub

- Fixes #110 

## Roadmap

- [x] Verify that `/*#__PURE__*/` comment affects tree shaking. If not, remove them.
- [x] Build a single `mockServiceWorker.js` file. Let all the tools reference that file in the `lib` root.
- [x] Consider arranging the split chunks better, `index2` may be confusing during stack tracing.


kettanaito:
## Chunk names

```diff
// rollup.config.js
  output: [
    {
+      chunkFileNames: '[name]-deps.js',
    },
  ],
```

kettanaito:
## PURE comment

Verified to not affect the build, removed.

kettanaito:
## Single `mockServiceWorker.js` file

I think introducing a custom step for the sake of moving the file up is unreasonable. At the moment `lib/esm` and `lib/umd` are independent, complete builds. I'd prefer them to remain that way.

Issue #145, opened 2020-05-06
ctx.fetch() not keeping original headers


aquerrec:
## Describe the bug
When mocking a REST API using `rest.get` i have a request containing headers, specifically "Authorization" header, and when i call `ctx.fetch()` with this request, in order to get the original response, i have an unauthorized response.
This is due to the fact that my "Authorization" header is missing.

After taking a look at `src/context/fetch.ts`, specifically the `augmentRequestInit()` function, it seems that the type of my request headers is `Headers`, and not `string[][]` nor `Record<string, string>` (see `HeadersInit`type).
I think that the spread operator is not working for the type `Headers`, therefore the headers are not kept :
```typescript
const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
  return {
    ...requestInit,
    headers: {
      ...requestInit.headers,
      'x-msw-bypass': 'true',
    },
  }
}
```

## Environment

- `msw: 0.15.5`
- `nodejs: 12.16.1`
- `npm: 6.13.4`

- browser version: Chrome `81.0.4044.129`

## To Reproduce
1. make a request with "Authorization" header in an authent context
2. call `ctx.fetch()` function with the original request
3. you will see an "401 Unauthorized (from ServiceWorker)" request in your network

My mock :
```typescript
rest.get('/path/to/my/resource', async (req, res, ctx) => {
        const mockedResponse = [
            "Val 1",
            "Val 2",
            "Val 3"
        ];

        const originalResponse = await ctx.fetch(req)

        return res(
            ctx.status(200),
            ctx.json(originalResponse && originalResponse.length ? originalResponse : mockedResponse)
        );
    })
```

## Expected behavior
Headers are kept.

## Temporary Solution
I implemented my own `fetch()` duplicating the existing one, and i change the implementation of the `augmentRequestInit()` function.
My working version :
```typescript
const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
    const headersCopy: Record<string, string> = {}
    if (requestInit.headers instanceof Headers) {
        requestInit.headers.forEach((value: string, name: string) => {
            headersCopy[name] = value
        });
    }
    
    // TODO handle the other types : string[][] and Record<string, string>

    return {
        ...requestInit,
        headers: {
            ...headersCopy,
            'x-msw-bypass': 'true',
        },
    }
}
```

kettanaito:
[...]

PR #146, opened 2020-05-08
Supports running mocks in Node

### LINKED ISSUE #104: Question: How to reuse the same mocked APIs to run Jest tests?
It would be very helpful to add some documentation and one example to show how to possibly reuse the same mocked APIs code for running Jest tests.

How to set up the Jest test environment so you can leverage the same mocked APIs?

The expectation is that in the node environment, mocks will respond with a shorter latency, compared to when called from the browser.

### LINKED ISSUE #154: Worker cannot serialize request headers with multiple values
## Describe the bug
When dispatched a request that has a header with multiple values, only the latest header value will be set on the intercepted `req` object sent to the library.

## Environment

- `msw: 0.15.8`

## Expected behavior
Request headers with multiple values are converted to `Record<string, string | string[]>` instance, preserving the multiplicity of values in a list.

## Screenshots

https://github.com/mswjs/msw/blob/6a1182d8847efa26a201fc6a37cf3e1f3ea593dc/src/mockServiceWorker.js#L179-L185

Technically, this is caused by the serialization function above, as it constructs a _shallow_ object, with the matched header name overriding the previously set one, instead of joining the values. 

I believe it should be implemented like this:

```js
function serializeHeaders(headers) {
  const reqHeaders = {}
  headers.forEach((value, name) => {
    const existingValue = reqHeaders[name]
    reqHeaders[name] = existingValue ? [].concat([existingValue, value]) : value
  })
  return reqHeaders
}
```

### LINKED ISSUE #158: Use URL instance in "req.url"
> This suggests a **breaking change**.

I suggest to store an intercepted request's URL as a `URL` instance, as opposed to `string`.

**Benefits**

- `URL` class is supported in both browser and Node environments;
- Allows to unify internal implementation in the underlying libraries, leading to fewer URL transformations;
- Gives more information to the developer, such as `req.url.protocol`, `req.url.port`, etc. This information is _already_ available, just not exposed to the developer, which is an experience flaw;

**Additional context**
There are multiple points of conversion happening, which transform a string representation of a request's URL into a `URL` instance anyway:

- `node-request-interceptor` converts it to `URL` in order to provide a clean URL in the middleware function;
- `msw@getResponse` converts to a `URL` instance to get a clean URL. With #146 it will also use that URL instance to resolve a relative URL against the current `window.location`.

kettanaito:
## Changes

- **Adds support for running the same mock definition in Node environment**.
- Breaking: `req.url` is now an instance of `URL`, not a string.
- Breaking: `req.query` is removed in favor of `req.url.searchParams`.

## GitHub

- Closes #104
- Closes #158 
- Fixes #154 

## Roadmap

- [x] Configure build to produce node-oriented module with `setupServer`
- [x] Ensure the node-oriented part is TypeScript-ready
- [x] Convert intercepted request to mocked request (headers, params, query, etc.)
- [x] Deal with `res()` function, that uses `new Headers()` instance, which does not exist in Node (also affecting `ctx.fetch()`: #145, #149)
- [x] Provide a basic set of integration tests (fetch/http/https/xhr)
- [x] Configure node bundle to properly rely on system modules (`https`, `tty, etc.`).
- [x] Resolve the circular `EventEmitter` dependency issue:

```
(!) Circular dependencies
node_modules/util/util.js -> node_modules/util/node_modules/inherits/inherits.js -> node_modules/util/util.js
node_modules/util/util.js -> node_modules/util/node_modules/inherits/inherits.js -> /Users/kettanaito/Projects/mswjs/msw/node_modules/util/util.js?commonjs-proxy -> node_modules/util/util.js
[!] Error: 'EventEmitter' is not exported by node_modules/events/events.js, imported by node_modules/node-request-interceptor/lib/http/ClientRequest/Socket.js
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
node_modules/node-request-interceptor/lib/http/ClientRequest/Socket.js (1:9)
1: import { EventEmitter } from 'events';
            ^
2: export class Socket extends EventEmitter {
3:     constructor(options, socketOptions) {
Error: 'EventEmitter' is not exported by node_modules/events/events.js, imported by node_modules/node-request-interceptor/lib/http/ClientRequest/Socket.js
```

- [x] Remove Node-related modules from `rest-deps.js` chunk.

### CRA + @testing-library/react

- [x] Relative request URLs fail during interception (https://github.com/mswjs/node-request-interceptor/issues/7, **resolved in** https://github.com/mswjs/node-request-interceptor/pull/8):

```
    TypeError: Invalid URL: /login

      2152 |             this.readyState = this.LOADING;
      2153 |             this.data = data || '';
    > 2154 |             const url = new URL(this.url);
           |                         ^
      2155 |             const req = {
      2156 |                 url: cleanUrl(url),
      2157 |                 method: this.method,
```


kettanaito:
After adopting the `headers-utils` library the headers of the mocked response are joined with a `","`, not `", "` (missing space). That is a minor difference, and shouldn't affect anything, but I'd like to debug why exactly that happened.

src/context/body.test.ts
```diff
@@ -10,7 +10,7 @@ describe('body', () => {
     })
 
     it('should not have any "Content-Type" header set', () => {
-      expect(result.headers.get('Content-Type')).toEqual(null)
+      expect(result.headers.get('content-type')).toBeUndefined()
```
kettanaito: This violates backwards compatibility with the `Headers` spec. This should return `null`.

src/mockServiceWorker.js
```diff
@@ -181,7 +181,9 @@ If you wish to mock an error response, please refer to this guide: https://redd.
 function serializeHeaders(headers) {
   const reqHeaders = {}
   headers.forEach((value, name) => {
-    reqHeaders[name] = value
+    reqHeaders[name] = reqHeaders[name]
+      ? [].concat(reqHeaders[name]).concat(value)
```
kettanaito, line 185: Does this need to join the headers in a list?

src/node/setupServer.ts
```diff
@@ -0,0 +1,57 @@
...
+    open() {
+      interceptor = new RequestInterceptor()
+      interceptor.use(async (req) => {
+        const mockedRequest: MockedRequest = {
+          url: req.url,
+          method: req.method,
+          body: req.body || '',
+          query: req.query,
+          headers: new Headers(flattenHeadersObject(req.headers || {})),
+          params: {},
```
kettanaito, line 27: > Note: `params` are populated as a part of `getRespnse()` call.

src/setupWorker/glossary.ts
```diff
@@ -20,3 +22,7 @@ export interface StartOptions {
   }
   quiet?: boolean
 }
+
+export type ResponseWithHeaders = Omit<MockedResponse, 'headers'> & {
```
kettanaito: Check where this is used. There's a chance that introduction of a custom `Headers` polyfill negates the need to use this override.

src/utils/getResponse.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (!mockedResponse) {
+    return {
+      response: null,
+      handler: relevantHandler,
+    }
+  }
+
+  const responseWithHeaders: MockedResponse = {
+    ...mockedResponse,
+    // @ts-ignore
```
kettanaito: I believe there is no errors left to ignore here. This pragma comment must be removed.

src/utils/handleRequestWith.ts
```diff
@@ -42,63 +36,37 @@ export const handleRequestWith = (
...
 
-      // Transform Headers into a list to be stringified preserving multiple
-      // header keys. Stringified list is then parsed inside the ServiceWorker.
-      const responseWithHeaders: MockedResponse = {
-        ...mockedResponse,
-        // @ts-ignore
-        headers: Array.from(mockedResponse.headers.entries()),
+      const responseWithHeaders: ResponseWithHeaders = {
+        ...response,
+        headers: headersToList(response.headers),
```
kettanaito, line 72: This is already done in `getResponse()`. Remove?

kettanaito: Needs to be removed from `getResponse()`. Here it serializes the `Headers` instance into a list, so it can be sent to the worker.

src/mockServiceWorker.js
kettanaito: This is fine, since the worker serializes the `request.headers` in order to send them to the client. On the client's side, the serialized instance is used to create a polyfilled `new Headers(request.headers)` instance, which supports `Record<string, string | string[]>` as an input, and joins a header's multiple values with a comma.

kettanaito:
Need to make sure this is the expected behavior from Rollup:

```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
http (imported by node_modules/node-request-interceptor/lib/http/override.js, http?commonjs-external, node_modules/node-request-interceptor/lib/http/ClientRequest/ClientRequestOverride.js)
https (imported by node_modules/node-request-interceptor/lib/http/override.js, https?commonjs-external)
util (imported by node_modules/node-request-interceptor/lib/http/ClientRequest/ClientRequestOverride.js, node_modules/debug/src/node.js, util?commonjs-external)
tty (imported by node_modules/debug/src/node.js, tty?commonjs-external, node_modules/supports-color/index.js)
events (imported by node_modules/node-request-interceptor/lib/http/ClientRequest/Socket.js)
os (imported by node_modules/supports-color/index.js, os?commonjs-external)
(!) `this` has been rewritten to `undefined`
https://rollupjs.org/guide/en/#error-this-is-undefined
node_modules/node-request-interceptor/lib/index.js
3: import require$$0 from 'debug?commonjs-proxy';
4:
5: var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
                    ^
6:     function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
7:     return new (P || (P = Promise))(function (resolve, reject) {
...and 1 other occurrence
node_modules/node-request-interceptor/lib/http/ClientRequest/ClientRequestOverride.js
3: import require$$0 from 'debug?commonjs-proxy';
4:
5: var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
                    ^
6:     function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
7:     return new (P || (P = Promise))(function (resolve, reject) {
...and 1 other occurrence
node_modules/node-request-interceptor/lib/XMLHttpRequest/XMLHttpRequest/XMLHttpRequestOverride.js
3: import require$$0 from 'debug?commonjs-proxy';
4:
5: var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
                    ^
6:     function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
7:     return new (P || (P = Promise))(function (resolve, reject) {
...and 1 other occurrence
```

**Edit:** This error doesn't occur anymore after `node-requrest-interceptor` is bundled to `ES5` format (https://github.com/mswjs/node-request-interceptor/pull/11).

kettanaito:
I've tried to create a unit test example that asserts a component's behavior that uses `axios`, but it turns out axios cannot run in Node. Examples I find suggest to use an adapter, or to mock the axios' implementation. Either of those look greatly unreliable to me.

PR #153, opened 2020-05-11
Default mockServiceWorker.js URL

### LINKED ISSUE #152: Is there a way to submit edits to the website documentation?
When I load my app at a route other than the root, MSW would fail to load the service worker. For example, if I load my app from this URL `http://localhost:3000/customers/2dd641c3-4069-424e-ad3d-8b2563358dd8`, I get the following error.

```
[MSW] Failed to register Service Worker (./mockServiceWorker.js). DOMException: Failed to register a ServiceWorker for scope ('http://localhost:3000/customers/') with script ('http://localhost:3000/customers/mockServiceWorker.js'): The script has an unsupported MIME type ('text/html').
```

I figured out that this is because it was looking for the service worker relative to the initial URL, so I need to change my start method to be absolute.

```javascript
// old code
return setupWorker(...handlers).start();

// new code
return setupWorker(...handlers).start({
  serviceWorker: {
    url: '/mockServiceWorker.js',
  },
});
```

I was thinking to update the [custom URL](https://redd.gitbook.io/msw/recipes/custom-service-worker-url) page in the docs to indicate that the default setting is to look for the service mocks at `<HOSTNAME>/mockServiceWorker.js` but actually it looks for it at `<initialURL>/mockServiceWorker.js`.

Perhaps I still don't understand correctly but I believe that is the behavior.

An alternative would be to [update the default](https://github.com/mswjs/msw/blob/master/src/setupWorker/start/createStart.ts#L14) to

```javascript
const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
  serviceWorker: {
    url: '/mockServiceWorker.js', // remove the initial "." on this url value
    options: null as any,
  },
  quiet: false,
}
```

mitchconquer:
- Fixes #152 

Modify the default url the mockServiceWorker.js file is served from to be the domain root rather than relative to the starting url

### DIFF
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -11,7 +11,7 @@ import { requestIntegrityCheck } from '../../utils/requestIntegrityCheck'
 
 const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
   serviceWorker: {
-    url: './mockServiceWorker.js',
+    url: '/mockServiceWorker.js',
     options: null as any,
   },
   quiet: false,


kettanaito (approved):
Looks great! Thank you, and welcome to contributors! :tada:

PR #168, opened 2020-05-18
Parses a JSON request body of the intercepted request

### LINKED ISSUE #159: "application/json" req.body is not parsed
## Describe the bug
When referencing the `req.body` that has been issued via `JSON.stringify({...})`, it's exposed as a stringified JSON.

## Environment

- `msw: 0.15.9`

## Expected behavior

Whenever `req.headers['content-type']` is JSON, MSW should parse the request body into a JavaScript Object. The `Content-Type` header should be a source of truth.

## Roadmap

- [ ] Add an integration test to `test/rest-api/body` asserting multiple body content types (text/xml/json/blob).

kettanaito:
## Changes

- Parses a JSON request body of the intercepted request, so now `req.body` is contextually an Object or a string, with no need to explicitly parse it.

## GitHub

- Fixes #159 

## Roadmap

- [x] Parse a response JSON body when viewing the logs

### DIFF
--- a/src/utils/getJsonBody.test.ts
+++ b/src/utils/getJsonBody.test.ts
@@ -1,0 +1,15 @@
+import { getJsonBody } from './getJsonBody'
+
+describe('getJsonBody', () => {
+  describe('given a valid JSON', () => {
+    it('should return a parsed object', () => {
+      expect(getJsonBody('{"prop":"value"}')).toEqual({ prop: 'value' })
+    })
+  })
+
+  describe('given an invalid JSON', () => {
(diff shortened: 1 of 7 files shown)


kettanaito:
Encountering this exception during `response-patching` test:

```
TypeError: Failed to execute 'fetch' on 'Window': Request with GET/HEAD method cannot have body.
```

## Reason

Caused by a non set `req.body` received from the worker being treated as an empty string. 

## Solution

Handles request body outside of the JSON parsing of the worker's message.

PR #172, opened 2020-05-20
setupServer: Parses a request JSON body

### LINKED ISSUE #171: Request JSON body is not parsed in Node
## Describe the bug
When mocking a request in NodeJS and performing a request with `Content-Type: application/json`, the `req.body` reference in a request handle is stringified body, instead of a parsed one (see #159).

## Environment

- `msw: 0.16.1`
- `nodejs: v10.18.1`
- `npm: 6.13.4`

## To Reproduce

```js
const server = setupServer(
  rest.post('https://test.msw.io', (req, res, ctx) => {
    return res(
      ctx.status(403),
      ctx.set('x-header', 'yes'),
      ctx.json(req.body as Record<string, any>),
    )
  }),
)

server.listen()

fetch('https://test.msw.io', {
  method: 'POST',
  headers: {
  },
  body: JSON.stringify({
    payload: 'info',
  }),
})
```

## Expected behavior
Request's JSON body is always parsed to a JavaScript object, regardless if running in browser or node.

## Screenshots
```diff
+  Expected: {"payload": "info"}
-  Received: "{\"payload\":\"info\"}"
```

kettanaito:
## Changes

- Isolates a request body parsing functionality into `parseRequestBody` function.
- Reuses `parseRequestBody` function for both client-side (`handleRequestWith`), and Node (`setupServer`).
- Adds an integration test that asserts `Content-Type: application/json` request body to be parsed in the `req.body` reference of the request handler.

## GitHub

- Fixes #171 
- Follow up to #159, #168

### DIFF
--- a/src/utils/parseRequestBody.ts
+++ b/src/utils/parseRequestBody.ts
@@ -1,0 +1,17 @@
+import { getJsonBody } from './getJsonBody'
+import { MockedRequest } from '../handlers/requestHandler'
+
+type Request = Partial<Pick<MockedRequest, 'headers' | 'body'>>
+
+export function parseRequestBody(
+  body?: MockedRequest['body'],
+  headers?: MockedRequest['headers'],
+) {
+  if (body) {
+    // If the intercepted request's body has a JSON Content-Type
+    // parse it into an object, otherwise leave as-is.
+    const isJsonBody = headers?.get('content-type')?.includes('json')
+
+    return isJsonBody && typeof body !== 'object' ? getJsonBody(body) : body
+  }
+}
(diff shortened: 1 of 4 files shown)


src/utils/parseRequestBody.ts
```diff
@@ -0,0 +1,17 @@
...
+type Request = Partial<Pick<MockedRequest, 'headers' | 'body'>>
+
+export function parseRequestBody(
+  body?: MockedRequest['body'],
+  headers?: MockedRequest['headers'],
+) {
+  if (body) {
+    // If the intercepted request's body has a JSON Content-Type
+    // parse it into an object, otherwise leave as-is.
+    const isJsonBody = headers?.get('content-type')?.includes('json')
```
whyDontI, line 13: This won't match for `Appliction/JSON` content-type.
Can we first `toLowerCase` before `includes`?

It's case insensitive as per the [W3 RFC doc](https://www.w3.org/Protocols/rfc1341/4_Content-Type.html#:~:text=The%20type%2C%20subtype,not%20case%2Dsensitive.)

Let me know if this change is valid, I can raise a quick PR

PR #179, opened 2020-05-22
Parses empty GET request body as undefined

### LINKED ISSUE #175: Empty request JSON body throws exception in a request handler
## Describe the bug
When attempting to parse a JSON body, which is empty, an exception seems to happen.

## Environment

- `msw: 0.16.2`

## Expected behavior
Empty body is not attempted to be parsed, even if request has `Content-Type: application/json`.

## Screenshots

- https://github.com/kentcdodds/bookshelf/pull/58#issuecomment-631571131
- https://github.com/kentcdodds/bookshelf/commit/25eddf6ace83671b8fbee2afc4a07dcfed93e666

kettanaito:
## Changes

- Treats _only_ `GET` requests with empty body as `undefined` during the parsing of the message received from the service worker.

> The captured request's `req.body` is an empty string (`""`) even if not set.

- Removes unused type definitions.
- Adds integration tests for `GET` request with no body.

## GitHub

- Fixes #175

### DIFF
--- a/src/context/text.ts
+++ b/src/context/text.ts
@@ -1,7 +1,7 @@
 import { ResponseTransformer } from '../response'
 
 /**
- * Sets the given text as the body of the response.
+ * Sets a given text as a "Cotent-Type: text/plain" body of the response.
  * @example
  * res(text('Message'))
  */
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -23,6 +23,7 @@ import { prepareRequest } from './utils/logger/prepareRequest'
 import { prepareResponse } from './utils/logger/prepareResponse'
 import { getTimestamp } from './utils/logger/getTimestamp'
 import { styleStatusCode } from './utils/logger/styleStatusCode'
+import { isStringEqual } from './utils/isStringEqual'
 
 export enum RESTMethods {
   GET = 'GET',
@@ -54,7 +55,7 @@ const createRestHandler = (method: RESTMethods) => {
       predicate(req) {
         // Ignore query parameters and hash when matching requests URI
         const cleanUrl = getCleanUrl(req.url)
-        const hasSameMethod = method.toUpperCase() === req.method.toUpperCase()
+        const hasSameMethod = isStringEqual(method, req.method)
         const urlMatch = match(resolveRelativeUrl(mask), cleanUrl)
 
         return hasSameMethod && urlMatch.matches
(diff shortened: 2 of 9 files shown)


kettanaito:
Experiencing this exception when running `test/rest/body.test.ts`:

```
Protocol error (Network.getResponseBody): No data found for resource with given identifier

      19 |           url: `${test.origin}/get-body`,
      20 |         })
    > 21 |         const body = await res.text()
         |                                ^
      22 |
      23 |         expect(body).toBeUndefined()
      24 |       })

      at Promise (../node_modules/puppeteer/lib/Connection.js:154:63)
      at CDPSession.send (../node_modules/puppeteer/lib/Connection.js:153:16)
      at _contentPromise._bodyLoadedPromise.then (../node_modules/puppeteer/lib/NetworkManager.js:557:53)
        -- ASYNC --
      at Response.<anonymous> (../node_modules/puppeteer/lib/helper.js:82:19)
      at rest-api/body.test.ts:21:32
      at fulfilled (rest-api/body.test.ts:5:58)
```

Based on [this Puppeteer issue](https://github.com/puppeteer/puppeteer/issues/2258) this may mean that the navigation happened and destroyed the pending request, however, I don't think that's the case. I've also tried issuing a test request to a different path (/`get-body`), which produced the exact exception (not the "same path" issue).

Most likely has something to do with those requests/handlers. It's peculiar, because both GET and POST requests with no and empty (`""`) body respectively, fail, while they produce a different mocked response (in case of `GET` `req.body: undefined`, when `POST` `req.body: ''`).

kettanaito:
The `No data found for resource with given identifier` error is caused by the mocked response having `body: undefined` when attempting to call `await res.text()` (the same for `res.buffer`). It means that either Puppeteer or an HTTP spec somehow forbids setting an explicit `undefined` **or** an empty string as a response body. I don't think it's likely, but it's what's happening.

> When the `ctx.text()` has a fallback value to a non-empty string, `res.text()` works as expected.

kettanaito:
## Resolution

There is an issue with Puppeteer when calling `res.text` on a response that has `undefined` or `""` as its `body`. To circumvent it, I've made the test request handler always return a JSON body, where a _key_ of that body equals the actual `req.body`.

**Setting `undefined` or `""` as `ctx.text()` is still perfectly valid, it just doesn't work with the current testing setup.**

PR #186, opened 2020-05-27
fix: move cookie types to dependencies

### LINKED ISSUE #185: Exposed types are missing
## Describe the bug

All `@types` are missing which are exposed through the api.
Since the types of msw are only `devDependencies` they are not installed on using sides, but since they are exposed through the API and therefore the `.d.ts` files for msw, they should be `dependencies`. Then they are installed as well when installing msw.

`cookie` is an example here:

```sh
[runtime:tsc] ../../node_modules/msw/lib/types/rest.d.ts:142:64 - error TS7016: Could not find a declaration file for module 'cookie'. '.../node_modules/msw/node_modules/cookie/index.js' implicitly has an 'any' type.
[runtime:tsc]   Try `npm install @types/cookie` if it exists or add a new declaration (.d.ts) file containing `declare module 'cookie';`
[runtime:tsc] 
[runtime:tsc] 142         cookie: (name: string, value: string, options?: import("cookie").CookieSerializeOptions | undefined) => import("./response").ResponseTransformer;
[runtime:tsc]                                                                    ~~~~~~~~
[runtime:tsc] 
```

## Environment

- `msw: 0.17.2`
- `nodejs: v13.12.0`
- `yarn: 1.22.4`

KnisterPeter:
Since we expose the types through our typings we need them to be
installed by using libraries. Therefore they are added to the
dependencies.

Closes #185

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
   "sideEffects": false,
   "dependencies": {
     "@open-draft/until": "^1.0.0",
+    "@types/cookie": "^0.3.3",
     "chalk": "^4.0.0",
     "cookie": "^0.4.1",
     "graphql": "^15.0.0",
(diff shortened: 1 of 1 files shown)


kettanaito (approved):
Looks great! Thank you for this, and welcome to contributors! :tada:

PR #187, opened 2020-05-28
Supports runtime request handler declaration

### LINKED ISSUE #128: Consider a pattern to override responses after setup
I like the look of this library but to be able to use it I'd need the ability to override a response from within a test. Something like:

```
it('should handle some weird edge case response', () => {
  mockedHttpClient.onGet('/endpoint').replyOnce(() => ({
    // ...
  }));
  const { getByText } = render(....);
  // ...
});
```

**Additional context**
Jest handles this very well within their mock functions using `mockImplementationOnce` which I think is a good source of inspiration

https://jestjs.io/docs/en/mock-function-api.html

kettanaito:
## Changes

- Each request handler now has a `isUsed: boolean` property. Whenever it's set to `true`, this request handler is skipped when looking up the relevant request handler for the captured request.
- The `res` composition function now has a static method called `once`. When used, it responds with the given mocked response and marks the current request handler is used.
- See the API section below on new methods on both `setupServer` and `setupWorker`.

### `setupServer`

```js
const server = setupServer(
  // Initial "GET /books" handler, always there.
  rest.get('/books', (req, res, ctx) => res(ctx.json('mocked body'))
)

server.use(
  // Adding a new POST /login request handler on runtime, overrides nothing.
  // This handler stays active forever, until you call `server.resetHandlers`
  // or stops the mocking in general.
  rest.post('/login', (req, res, ctx) => {...})
)

server.use(
  // The next call to "GET /books" will respond with 500,
  // any subsequent calls will respond with 200 and mocked body,
  // because this request handler override uses `res.once()`.
  rest.get('/books', (req, res, ctx) => res.once(ctx.status(500))
)

// Restores the used state of all one-time request handlers
// (those that used `res.once()`), making them effective again.
server.restoreHandlers()

// Removes all request handlers that were added on runtime.
// Leaves only the initial "GET /books" 200 handler.
server.resetHandlers()

// Removes all request handlers that were added on runtime.
// Additionally accepts the list of request handlers that will
// replace the initial handlers from this point on.
server.resetHandlers(
  rest.get('/products', (req, res, ctx) => {...})
)
```

### `setupWorker`

```js
const worker = setupWorker(
  rest.get('/book/:bookId', (req, res, ctx) => {...})
)

// Adds a runtime request handler for "POST /login" requests.
worker.use(
  rest.post('/login', (req, res, ctx) => {...})
)

// Adds a one-time request handler override.
// Once this responds, it won't affect network anymore.
worker.use(
  rest.get('/book/:bookId', (req, res, ctx) => {
    return res.once(...)
  })
)

// Marks all one-time request handlers as unused,
// making them affect traffic again.
worker.restoreHandlers()

// Removes all request handlers that were added on runtime,
// meaning since the `setupWorker` call.
worker.resetHandlers()

// Removes all runtime request handler, and replaces the initial handlers
// passed to the `setupWorker` call with the given request handlers.
worker.resetHandlers(
  rest.get('/foo', (req, res, ctx) => {...})
)
```

## GitHub

- Closes #128 

## Roadmap

- [x] Support runtime definition in `setupServer`
- [x] Add a method that sets all runtime request as `isUsed: false` in `setupServer`
- [x] Add a method to completely replace the handlers forever
- [x] Support runtime definition in `setupWorker`
- [x] ~~Consider abstracting widely used `runtime.origin + url` into a `runtime.createRelativeUrl()` utility function~~ (reasonable, but out of scope)


.eslintrc.js
```diff
@@ -13,5 +13,11 @@ module.exports = {
     '@typescript-eslint/no-explicit-any': 0,
     '@typescript-eslint/explicit-module-boundary-types': 0,
     '@typescript-eslint/ban-ts-comment': 0,
+    '@typescript-eslint/no-namespace': [
+      2,
+      {
+        allowDeclarations: true,
```
kettanaito, line 19: Enable `allowDeclarations` so it's possible to augment `window` object as a part of some tests. Those tests usually transfer the `worker` reference by setting it on the global `window` object, so some worker methods could be called from within the Puppeteer's environment during tests.

src/index.ts
```diff
@@ -1,6 +1,10 @@
 import * as context from './context'
 
-export { setupWorker, composeMocks } from './setupWorker/setupWorker'
+export {
+  SetupWorkerApi,
```
kettanaito: I think this has no purpose in exposing it publicly. Instead, move it to `glossary` and reference from there internally, or as a part of tests.

src/node/setupServer.ts
```diff
@@ -7,18 +7,23 @@ import { RequestHandler, MockedRequest } from '../handlers/requestHandler'
 import { getResponse } from '../utils/getResponse'
 import { parseRequestBody } from '../utils/parseRequestBody'
 
+type RequestHandlersList = RequestHandler<any, any>[]
```
kettanaito: Similar type definition is present in `setupWorker`. Consider moving to `glossary` and reusing.

src/node/setupServer.ts
```diff
@@ -62,6 +69,32 @@ export const setupServer = (...handlers: RequestHandler<any, any>[]) => {
...
+     */
+    use(...handlers: RequestHandlersList) {
+      currentHandlers.unshift(...handlers)
+    },
+
+    /**
+     * Marks all request handlers that respond using `res.once()` as unused.
+     */
+    restoreHandlers() {
+      currentHandlers.forEach((handler) => {
```
kettanaito: Both server and client code share the same logic in use/restoreHandlers/resetHandlers. Can this logic be abstracted into utility functions, so it's guaranteed to be the same for client and server?

PR #196, opened 2020-06-04
Defers network requests until Service Worker is ready

### LINKED ISSUE #190: Defer actual requests until mocks are ready by default
Here'a an example @kentcdodds is using to ensure no requests are performed before the mock service worker is ready:

```js
const originalFetch = window.fetch

window.fetch = async (...args) => {
  await worker.start()
  window.fetch = originalFetch()
  return window.fetch(...args)
}
```

The library may ship this as a default behavior for _client-side_ execution. Most likely, working with `XMLHttpRequest`, since it's often used by `fetch`. This would grant a better user experience than having to ask users to wrap their app mounting logic until the service worker is ready.

## Cons

- Stubbing `XMLHttpRequest`

## Pros

- The stub is transparent, restores to the original implementation once mocks are ready.
- No need to defer app's rendering. App would now know absolutely nothing about mocking.

kettanaito:
> This is a breaking change.

## Changes

- Adds the `waitUntilReady: boolean` option to the `worker.start()` function.
- By default, starting a Service Worker via `worker.start()` now **defers all network requests until worker is ready**. That is powered by stubbing `XMLHttpRequest.prototype.send` and `window.fetch`. Once the worker is ready, the stubs are removed.

## GitHub

- Fixes #190

### DIFF
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -29,7 +29,7 @@ export interface StartOptions {
 
   /**
    * Defer any network requests until the Service Worker
-   * instance is ready.
+   * instance is ready. Defaults to `true`.
    */
   waitUntilReady?: boolean
 }
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -20,7 +20,18 @@ export interface StartOptions {
     url?: string
     options?: RegistrationOptions
   }
+
+  /**
+   * Disable the logging of captured requests
+   * into browser's console.
+   */
   quiet?: boolean
+
+  /**
+   * Defer any network requests until the Service Worker
+   * instance is ready.
+   */
+  waitUntilReady?: boolean
 }
 
 export type RequestHandlersList = RequestHandler<any, any>[]
--- a/src/utils/deferNetworkRequests.ts
+++ b/src/utils/deferNetworkRequests.ts
@@ -1,3 +1,5 @@
+import { until } from '@open-draft/until'
+
 /**
  * Intercepts and defers any requests on the page
  * until the Service Worker instance is ready.
@@ -7,32 +9,22 @@ export function deferNetworkRequests(
 ) {
   // Defer `XMLHttpRequest` until the Service Worker is ready.
   const originalXhrSend = window.XMLHttpRequest.prototype.send
-  const restoreXhrSend = function (
-    this: XMLHttpRequest,
-    ...args: Parameters<XMLHttpRequest['send']>
-  ) {
-    window.XMLHttpRequest.prototype.send = originalXhrSend
-    this.send(...args)
-  }
 
   window.XMLHttpRequest.prototype.send = function (
     ...args: Parameters<XMLHttpRequest['send']>
   ) {
-    workerReady
-      .then(() => restoreXhrSend.call(this, ...args))
-      .catch(() => restoreXhrSend.call(this, ...args))
+    until(() => workerReady).then(() => {
+      window.XMLHttpRequest.prototype.send = originalXhrSend
+      this.send(...args)
+    })
   }
 
   // Defer `fetch` requests until the Service Worker is ready.
   const originalFetch = window.fetch
-  const restoreFetch = (...args: Parameters<typeof window.fetch>) => {
-    window.fetch = originalFetch
-    return window.fetch(...args)
-  }
 
   window.fetch = async (...args) => {
-    return workerReady
-      .then(() => restoreFetch(...args))
-      .catch(() => restoreFetch(...args))
+    await until(() => workerReady)
+    window.fetch = originalFetch
+    return window.fetch(...args)
   }
 }
--- a/src/utils/deferNetworkRequests.ts
+++ b/src/utils/deferNetworkRequests.ts
@@ -1,0 +1,38 @@
+/**
+ * Intercepts and defers any requests on the page
+ * until the Service Worker instance is ready.
+ */
+export function deferNetworkRequests(
+  workerReady: Promise<ServiceWorkerRegistration | null>,
+) {
+  // Defer `XMLHttpRequest` until the Service Worker is ready.
+  const originalXhrSend = window.XMLHttpRequest.prototype.send
+  const restoreXhrSend = function (
+    this: XMLHttpRequest,
+    ...args: Parameters<XMLHttpRequest['send']>
+  ) {
+    window.XMLHttpRequest.prototype.send = originalXhrSend
+    this.send(...args)
+  }
+
+  window.XMLHttpRequest.prototype.send = function (
+    ...args: Parameters<XMLHttpRequest['send']>
+  ) {
+    workerReady
+      .then(() => restoreXhrSend.call(this, ...args))
+      .catch(() => restoreXhrSend.call(this, ...args))
+  }
+
+  // Defer `fetch` requests until the Service Worker is ready.
+  const originalFetch = window.fetch
+  const restoreFetch = (...args: Parameters<typeof window.fetch>) => {
+    window.fetch = originalFetch
+    return window.fetch(...args)
+  }
+
+  window.fetch = async (...args) => {
+    return workerReady
+      .then(() => restoreFetch(...args))
+      .catch(() => restoreFetch(...args))
+  }
+}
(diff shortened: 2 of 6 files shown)


src/utils/deferNetworkRequests.ts
```diff
@@ -0,0 +1,38 @@
...
+    window.XMLHttpRequest.prototype.send = originalXhrSend
+    this.send(...args)
+  }
+
+  window.XMLHttpRequest.prototype.send = function (
+    ...args: Parameters<XMLHttpRequest['send']>
+  ) {
+    workerReady
+      .then(() => restoreXhrSend.call(this, ...args))
+      .catch(() => restoreXhrSend.call(this, ...args))
```
kettanaito: Restore stubs even if the worker registration fails. Page must be operational in either way.

src/setupWorker/glossary.ts
```diff
@@ -20,7 +20,18 @@ export interface StartOptions {
...
+   * Disable the logging of captured requests
+   * into browser's console.
+   */
   quiet?: boolean
+
+  /**
+   * Defer any network requests until the Service Worker
+   * instance is ready.
+   */
+  waitUntilReady?: boolean
```
kimroen, line 34: Maybe we should mention what the default value is in these comments? Very convenient for tooltips!

src/utils/deferNetworkRequests.ts
```diff
@@ -0,0 +1,38 @@
...
+) {
+  // Defer `XMLHttpRequest` until the Service Worker is ready.
+  const originalXhrSend = window.XMLHttpRequest.prototype.send
+  const restoreXhrSend = function (
+    this: XMLHttpRequest,
+    ...args: Parameters<XMLHttpRequest['send']>
+  ) {
+    window.XMLHttpRequest.prototype.send = originalXhrSend
+    this.send(...args)
+  }
```
kimroen: These `restoreX` functions both restore *and* call the function - maybe they could be named in a way where that's clearer?

src/utils/deferNetworkRequests.ts
```diff
@@ -0,0 +1,38 @@
...
+  workerReady: Promise<ServiceWorkerRegistration | null>,
+) {
+  // Defer `XMLHttpRequest` until the Service Worker is ready.
+  const originalXhrSend = window.XMLHttpRequest.prototype.send
+  const restoreXhrSend = function (
+    this: XMLHttpRequest,
+    ...args: Parameters<XMLHttpRequest['send']>
+  ) {
+    window.XMLHttpRequest.prototype.send = originalXhrSend
+    this.send(...args)
```
kimroen: I'm not sure if this is a problem or not, but here we're first restoring the `send` function, and then we're calling `send` on `this` (which would be the function we just restored). I find this confusing.

Is it possible that using `call` would be cleaner for this case?

```diff
- this.send(...args)
+ window.XMLHttpRequest.prototype.send.call(window.XMLHttpRequest, ...args)
```

src/utils/deferNetworkRequests.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  // Defer `fetch` requests until the Service Worker is ready.
+  const originalFetch = window.fetch
+  const restoreFetch = (...args: Parameters<typeof window.fetch>) => {
+    window.fetch = originalFetch
+    return window.fetch(...args)
+  }
+
+  window.fetch = async (...args) => {
```
kimroen, line 25: `async` isn't necessary here?

test/msw-api/setup-worker/start/wait-until-ready.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  const res = await runtime.page.waitForResponse((res) => {
+    return res.url().includes('/numbers')
+  })
+
+  const status = res.status()
+
+  expect(status).toBe(404)
+
+  await runtime.cleanup()
+})
```
kimroen, line 58: We might want to include similar tests for using `XMLHttpRequest` as well - that stubbing could break right now without us knowing.

test/msw-api/setup-worker/start/wait-until-ready.test.ts
```diff
@@ -0,0 +1,42 @@
+import * as path from 'path'
+import { runBrowserWith } from '../../../support/runBrowserWith'
+
+test('defers network requests until the worker is ready', async () => {
+  const runtime = await runBrowserWith(
+    path.resolve(__dirname, 'wait-until-ready.mocks.ts'),
+  )
+
+  // Reload is necessary, because `runBrowserWith` has `{ waitUntil: 'networkidle0' }` predicate
+  // when openning a test scenario mock. Reload doesn't affect the worker state, and without
```
kimroen:
```suggestion
  // when opening a test scenario mock. Reload doesn't affect the worker state, and without
```

test/msw-api/setup-worker/start/wait-until-ready.mocks.ts
```diff
@@ -0,0 +1,16 @@
...
+  rest.get('/numbers', (req, res, ctx) => {
+    return res(ctx.json([1, 2, 3]))
+  }),
+)
+
+// By default, starting the worker defers the network requests
+// until the worker is ready to intercept them.
+worker.start()
+
+// Although this request is performed alognside an asynchronous
```
kimroen:
```suggestion
// Although this request is performed alongside an asynchronous
```

kimroen (approved):
Really nice work! I'm assuming you handle docs separately, but if not we should make sure to document this 📝 

All my comments are suggestions, so I'll approve and let you consider them ✨

src/utils/deferNetworkRequests.ts
kettanaito: Actually, I think I will use `@open-draft/until` to handle both resolved and rejected Promise state. That way I don't have to repeat the logic in `.then` and `.catch`. 

~~That won't work for `XMLHttpRequest`, though, since it cannot be async.~~This will also work for XHR, just need to be cautious not to return a Promise.

src/utils/deferNetworkRequests.ts
kettanaito: `this` in the current context is a reference to a created `XMLHttpRequest` instance, while binding it to `window.XMLHttpRequest` would bind it to the _class_.

I'm stubbing the _class method_, so it affects all instances that inherit that class and call that method. But when restoring, I restore the class method (what I stubbed), and call the `this.send()`, which means performing this current deferred request.

src/utils/deferNetworkRequests.ts
kettanaito: I'll adopt `performActualX`:

- `performActualXMLHttpRequest`
- `performActualFetch`

src/utils/deferNetworkRequests.ts
kimroen: Makes sense! Thanks for clarifying 💡

kimroen:
In the original issue you wrote this:

> Most likely, working with XMLHttpRequest, since it's often used by fetch.

But here you've stubbed both. Would you mind expanding on why you ended up going that way?

kettanaito:
@kimroen, I've assumed that `window.fetch` works similar to the `fetch` from JSDOM (that is calling `XMLHttpRequest` internally). However, in a browser, `fetch` doesn't rely on `XMLHttpRequest` at all (tested in Chrome). That's why two separate stubs.

src/utils/deferNetworkRequests.ts
kettanaito: Removed the restoration functions at all, used `until` that helped to resolve the Promise, regardless whether it resolves/rejects.

kimroen:
I was wondering about another thing:

What happens to the order of requests here when dealing with multiple requests before the worker is ready? As far as I can tell there's nothing here to make sure things happen in the order they happened originally. It will also clean up multiple times (not that that should be a problem).

kettanaito:
@kimroen, since each request is an async operation, there is no order of requests at place. The way this stub works is that it intercepts when a request is performed _per-instance_, so as soon as the worker is ready, each instance will continue with the actual request.

In regards to multiple clean-ups, I think that's not the case. The `workerReady` promise resolves once per runtime, so no matter how many requests you issue, when `workerReady` resolves (once), all of them become unfrozen. 

Since the worker unregisters itself on `beforeunload`, if you reload the page while `workerReady` Promise is pending, it will be thrown away, and your app's logic will trigger again, call the `worker.start()` again, and create a new Promise to await.

kimroen (approved):

PR #198, opened 2020-06-04
add realistic server response delay to ctx.delay()

### LINKED ISSUE #192: Using "ctx.delay" without arguments should add a realistic server response delay
Quite often you would want a mocked response to behave similar to the actual response in terms of response time. Real servers rarely reply to requests instantly, while using `res()` would do so.

Calling `ctx.delay()` without arguments should implicitly create a random realistic response delay, similar to the one of the actual server.

```js
rest.get('/numbers', (req, res, ctx) => {
  return res(ctx.delay(), ctx.json([1, 2, 3])
})
```

**Possible implementation**
Check if the `durationMs` argument is given to the `delay()` function:

https://github.com/mswjs/msw/blob/4a49e733ed10cb2b11c1941160d9a46ec21182af/src/context/delay.ts#L8-L13

And if not, generate a random response delay number instead. Something similar to:

```js
res.delay = durationMs || RANDOM_REALISTIC_RESPONSE_TIME
```

UsamaHameed:
fixes #192

### DIFF
--- a/src/context/delay.test.ts
+++ b/src/context/delay.test.ts
@@ -1,4 +1,8 @@
-import { delay } from './delay'
+import {
+  delay,
+  MIN_SERVER_RESPONSE_TIME,
+  MAX_SERVER_RESPONSE_TIME,
+} from './delay'
 import { response } from '../response'
 
 describe('delay', () => {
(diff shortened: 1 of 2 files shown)


src/context/delay.test.ts
```diff
@@ -8,4 +8,11 @@ describe('delay', () => {
       expect(endResponse).toHaveProperty('delay', 1200)
     })
   })
+
+  describe('if no delay provided ', () => {
+    it('should set random delay on the response', () => {
+      const endResponse = response(delay())
+      expect(endResponse.delay).toBeDefined()
```
kettanaito: What do you think if we check that:

```js
expect(endResponse.delay).toBeGreaterThan(MIN_SERVER_RESPONSE_TIME)
expect(endResponse.delay).toBeLessThan(MAX_SERVER_RESPONSE_TIME)
```

That way we ensure that the default delay value lies in the pre-defined boundaries.

kettanaito (changes requested):
Awesome work! I've left only one comment regarding what we should be asserting. Let me know what's your opinion on that.

src/context/delay.test.ts
UsamaHameed: done!

kettanaito (approved):
Looks great! Thank you for the changes and welcome to contributors! 🎉

PR #201, opened 2020-06-06
Remove request json parsing from the worker file

### LINKED ISSUE #200: Uncaught (in promise) SyntaxError: Unexpected end of input
## Describe the bug
After updating the msw package from `0.15.5` to `0.19.0` and update the `mockServiceWorker.js` file, the service worker started throwing the following error 
 
`Uncaught (in promise) SyntaxError: Unexpected end of input` at line 111

After some debugging I was able to fix the issue ( at least for me ) by replacing 
```js
 const body = request.headers.get('content-type')?.includes('json')
        ? await request.json()
        : await request.text()
```
with 
```js
const isJSONContentType = request.headers.get('content-type')?.includes('json')
let body;
if (isJSONContentType) {
  try {
    body = await request.json()
  } catch (error) {
    console.warn(error)
    body = await request.text()
    console.log({body,headers: Object.fromEntries(request.headers.entries()), url: request.url})
  }
} else  {
  body = await request.text()
}
```
with this fix I'm getting the correct mocked responses but what I don't understand is why the body is showing up empty

## Environment

- `msw: 0.19.0`
- `nodejs: 12.13.1`
- `npm: 6.14.5`
- `webpack: 4.42.1`
- `webpack-dev-server: 3.10.3`

The requests that are failing match with a Webpack DevServer proxy that I have in my environment.
The proxy takes every request url starting with `"/api"` to a live server 

Webpack devServer config:

```js
devServer: {
    contentBase: "./public",
    headers: {
      "Access-Control-Allow-Origin": "*",
      "Access-Control-Allow-Headers": "Origin, X-Requested-With, Content-Type, Accept"
    },
    host: !process.argv.join(" ").includes("--sw") && require("ip").address(),
    disableHostCheck: true,
    open: true,
    hot: false,
    port,
    proxy: {
      "/api": {
        target: "https://live.server.domain.cloud",
        secure: false,
        changeOrigin: true,
        withCredentials: true,
      }
    }
  },
```

Chrome 83.0.4103.61

## To Reproduce
Steps to reproduce the behaviour:
1. Run the webapp with msw enabled 
2. in the console you'll see the error `Uncaught (in promise) SyntaxError: Unexpected end of input`
3. if you refresh the browser you'll find that the request for the webpack index.html file is hanging and will not resolve

## Expected behavior

Shouldn't throw if the suspected response has the `Content-Type` header containing `"json"` and the body is empty

## Screenshots
Original error 

image

chrome console after I added some logs and error handling on `mockServiceWorker.js`

image

Network tab of the same session / request after the fix showing expected mocked data

image

mouhannad-sh:
Fixes #200 

Do we need to add any tests to cover the case of empty body responses ? if so, can you guide me to where I need to start ?

Thanks


kettanaito:
Thank you for your work, @mouhannad-sh! Looks great.

Yes, let's add at least one basic test that asserts an explicit empty string body (`""`) and `Content-Type: application/json`. There's an existing [`test/rest-api/body.test.ts`](https://github.com/mswjs/msw/blob/master/test/rest-api/body.test.ts) test suite that asserts different body scenarios, I'd suggest to add this one where you think it makes a logical sense.

> The test performs a request with a certain options, and has a request handler in the [mocks file](https://github.com/mswjs/msw/blob/5defbf2b4f4d7ae1503cb68a4e7be14e994b3f7d/test/rest-api/body.mocks.ts#L3-L7) that returns any request body as a response body.

Take a look at how test actions/assertions are written for more details, I'm sure it'd be straightforward. Let me know if you have any questions.

kettanaito:
I've rebased your branch and will provide the tests, if you don't mind.

kettanaito:
Hey, I've took this pull request as an opportunity to structure the `body.test.ts` test suite better. Feel free to take a look at how it's now, and at that new integration test I've added for a request with a `Content-Type: application/json` and no body. I've verified it fails without your change and passes with it.

kettanaito (approved):
Thank you for these changes, you make the library much better! You rock.

mouhannad-sh:
Thanks a lot @kettanaito you're a life saver 🔥

sumitj18s:
Hey guys, I like this library, can you please release the current Master as we are stuck at #200 . Thanks so much!

kettanaito:
@sumitj18s thanks for the feedback. The fix is released in `0.19.1`. Make sure to follow the worker file update instructions in your browser's console.

sumitj18s:
Thanks @kettanaito a lot for such prompt release & response!

PR #219, opened 2020-06-17
Add a friently error message when you call setupServer in the browser

### LINKED ISSUE #214: Requests not properly resolving with axios in the browser
## Describe the bug
One the latest version of axios (0.19.2), and with any version >= 0.18.0 of msw any calls to axios.get do not ever resolve the promise. There are no errors in the console, and MSW and networks logs indicate the request was handled correctly

I haven't had time to make an isolated reproducible case yet, but will try to do so. In my app the request in question is one where I'm not actually mocking anything, just letting it pass through to the server, but I'm not sure that it's relevant. 

## Environment

- `msw: 0.18.0`
- `npm: 6.13.4`

Replicated in Firefox 77.0.1  and Chrome 83.0.4103.61
## To Reproduce

Should be reproducible with axios code like 
```
axios.get("/foo").then(res=> console.log(res))
```

## Expected behavior

Should log the response out to the console, but instead we never seem to make it inside the then statement.

AndrewSwerlick:
setupServer causes problem when run in the browser, but without any obvious
error messages. This commit changes setupServer so it detects when its not
running in a node environment and prints an error message letting you know that
calling setupServer is not supported.

Resolves #214


AndrewSwerlick:
I saw one failing test locally even before I made any changes to setupServer, so I'm not sure what that was. I assume that's why the ci build is failing though. Let me know what I can do to resolve

src/node/setupServer.ts
```diff
@@ -9,12 +9,23 @@ import { getResponse } from '../utils/getResponse'
 import { parseRequestBody } from '../utils/parseRequestBody'
 import * as requestHandlerUtils from '../utils/requestHandlerUtils'
 
+const isNode =
```
kettanaito: I suggest we move this into a utility function (somewhere at `src/utils/isNodeProcess.ts`, since it appears this will be needed for [another functionality](https://github.com/mswjs/msw/pull/205) as well. It'd be nice to have it declared in one place, and use everywhere it's needed.

src/node/setupServer.ts
```diff
@@ -9,12 +9,23 @@ import { getResponse } from '../utils/getResponse'
...
  * Sets up a server-side requests interception with the given mock definition.
  */
 export const setupServer = (...requestHandlers: RequestHandlersList) => {
   const interceptor = new RequestInterceptor()
 
+  if (!isNode) {
+    console.error(
+      '[MSW] It appears you are trying to set up Node Integration in the browser. Please make sure to move any calls to setupServer so they only run in a node/jsdom based testing environment',
+    )
+    throw 'Cannot call setupServer in the browser'
```
kettanaito: Do you think we should throw in this case?
This will be an unhandled exception on the developer's side, very likely crashing their app. I think that we shouldn't throw at this case, but rather:

1. Produce an error (like you're doing with `console.error`.
1. Have an explicit `return` within this `if` statement. This way calling `setupServer` in a non-Node environment won't return its usual API (`.listen`, `.use`, etc.), and will produce an error instead.

test/msw-api/setup-server/scenarios/browser.mocks.ts
```diff
@@ -0,0 +1,7 @@
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
+const worker = setupServer(
```
kettanaito: Could you please add a comment on top of this `setupServer` call that this is intentionally wrong code?
We are trying to keep test scenarios as actual usage examples, so it would be useful to mention this is _not_ how people should be using it.

src/node/setupServer.ts
```diff
@@ -9,12 +9,23 @@ import { getResponse } from '../utils/getResponse'
...
+
 /**
  * Sets up a server-side requests interception with the given mock definition.
  */
 export const setupServer = (...requestHandlers: RequestHandlersList) => {
   const interceptor = new RequestInterceptor()
 
+  if (!isNode) {
+    console.error(
+      '[MSW] It appears you are trying to set up Node Integration in the browser. Please make sure to move any calls to setupServer so they only run in a node/jsdom based testing environment',
```
kettanaito: What do you think if we rephrase it to:

```
[MSW] Failed to execute `setupServer` in the environment that is not NodeJS (i.e. a browser). Consider using `setupWorker` instead.
```

kettanaito (changes requested):
Thank you for these changes! I've provided a few suggestions in the review. Let me know what you think.

kettanaito:
CI fails on `POST https://mswjs.io/books`. Since that hostname exists, `/books` leads to a 404 page. I suppose `POST` is not allowed to that page, so instead of `404` we get `405 Method Not Allowed` response status.

The `msw-api/setup-server/resetHandlers.test.ts` integration test needs to be adjusted not to reference that production resource (let's handle this separately).

kettanaito:
Could you please rebase on the current `master`? I've fixed the failing test in #222. Thanks.

src/node/setupServer.ts
andrewswerlick-hmn: My thinking was throw to fail fast and crash rather than weird unexpected behavior, but I was back and forth on that myself. It's not a strongly held opinion, so I'll change it to remove the throw if that's what you prefer.

kettanaito: It should be fine to produce an error + don’t execute any code intended for Node (due to explicit return).

I’m thinking if we should add a similar check and error when running `setupWorker` in Node?

andrewswerlick-hmn:
@kettanaito All changes should be addressed now. Tests passed locally after rebasing

kettanaito:
@andrewswerlick-hmn, thanks! Could you please double check if you've committed `utils/isNodeProcess`, it shows as missing in CI and PR files?

test/msw-api/setup-server/scenarios/browser.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    captureConsole(test.page, errors, (message) => {
+      return message.type() === 'error'
+    })
+
+    await test.reload()
+
+    const nodeMessage = errors.find((message) => {
+      return message.includes(
+        '[MSW] It appears you are trying to set up Node Integration',
```
kettanaito: Could you please also change this to reflect the updated error message?

```js
message.startsWith('[MSW] Failed to execute `setupServer` in the environment that is not NodeJS')
```

Thanks!

kettanaito (changes requested):

src/node/setupServer.ts
```diff
@@ -15,6 +16,11 @@ import * as requestHandlerUtils from '../utils/requestHandlerUtils'
 export const setupServer = (...requestHandlers: RequestHandlersList) => {
   const interceptor = new RequestInterceptor()
 
+  if (!isNode) {
+    console.error(
+      '[MSW] Failed to execute `setupServer` in the environment that is not NodeJS (i.e. a browser). Consider using `setupWorker` instead.',
+    )
```
kettanaito, line 23: As a final touch, could you please add a `return` statement within this `if` condition?
This way we will short circuit the execution logic, preventing any Node-related code to run in a browser. Right now it will print the error, but continue to run, which would result into all sorts of issues. Thanks!

src/node/setupServer.ts
```diff
@@ -7,6 +7,7 @@ import { RequestHandlersList } from '../setupWorker/glossary'
 import { MockedRequest } from '../handlers/requestHandler'
 import { getResponse } from '../utils/getResponse'
 import { parseRequestBody } from '../utils/parseRequestBody'
+import isNode from '../utils/isNodeProcess'
```
kettanaito: If it doesn't burden you, could you please use named export here?

```js
import { isNode } from '...'

// isNodeProcess.ts
export function isNode(...) {...}
```

It's a nitpick, but keeping up the code style is beneficial in a long run. Thanks for understanding!

kettanaito (changes requested):
Thanks for updating the branch and providing the changes! I've got only two suggestions left, and then we'll be ready to ship this.

src/utils/isNodeProcess.ts
```diff
@@ -0,0 +1,3 @@
+export const isNode =
+  Object.prototype.toString.call(typeof process !== 'undefind') ===
```
kettanaito: Typo
```diff
- undefind
+ undefined
```

kettanaito: Please, could you make a function out of it? That way figuring out if it's Node is an _action_, and it's bound to the context where you call it, as opposed to where you declare it. Thanks!

kettanaito (changes requested):

andrewswerlick-hmn:
I think this should address the latest. No worries about the changes, it's a good thing anyways because on closer review, it turns out there was a bug with the isNode function causing failing tests once I made the if statement return and exit out. Let me know if you see anything else

kettanaito:
Hey, I've made a few changes:

- I think you were right, and I reworked those errors to be thrown exceptions. That's more evident to the user, and easier to test as well.
- I've added a similar exception when running `setupWorker` in a NodeJS environment (that's also an unintended behavior).

kettanaito (approved):

kettanaito:
Thank you once more for this improvement, @AndrewSwerlick! Welcome to contributors! 🎉

PR #224, opened 2020-06-17
Supports binary response types

### LINKED ISSUE #221: Support for binary response types
I tried to mock a response which contains a wav file. My initial attempt looked like this:

```js
res(
    ctx.status(200, 'OK'),
    ctx.set('Content-Length', arrayBuffer.byteLength.toString()),
    ctx.set('Content-Type', 'audio/wav'),
    ctx.body(arrayBuffer)
);
```

That didn't work. The response did not contain the contents of the `ArrayBuffer`. It was `[object Object]` instead.

I wonder if it would be possible to remove the JSON serialization and deserialization. Removing it from [here](https://github.com/mswjs/msw/blob/master/src/utils/createBroadcastChannel.ts#L27) and [here](https://github.com/mswjs/msw/blob/master/src/mockServiceWorker.js#L132) makes the example from above work.

I also tried converting the `ArrayBuffer` to a string but that didn't work out and specifying an `ArrayBuffer` directly is of course way more convenient.

chrisguttandin:
This pull request is meant to add support for binary response types by removing the serialization of messages send over the MessageChannel. It's meant to resolve #221.

As far as I can tell none of the existing functionality will be affected by this change.

Please let me know if there is some that needs to be changed.


test/rest-api/binary.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  })
+
+  afterAll(() => {
+    return test.cleanup()
+  })
+
+  describe('given a binary response', () => {
+    it('should return a binary response', async () => {
+      const res = await test.request({
+        url: 'https://test.msw.io/binary',
```
kettanaito: Could you please use `test.mswjs.io` (`msw` -> `mswjs`)? That way it points to the domain we own, not a third-party domain. 

> This is to ensure that tests have as less external dependencies as possible. Any link is potentially a dependency, as if it gets updated, it may affect the response and, thus, the test.

src/utils/createBroadcastChannel.ts
```diff
@@ -24,7 +24,7 @@ export const createBroadcastChannel = (event: MessageEvent) => {
       payload?: Record<string, any> | string
     }) {
       if (port) {
-        port.postMessage(JSON.stringify(message))
+        port.postMessage(message)
```
kettanaito, line 27: Okay, so when we sent a message from the worker, it's stringified and gets parsed in `handleRequestWith`, but when we communicate the response back to the worker, it's neither stringified nor parsed. I like that, wonder if the tests pass with this setup (will check).

test/rest-api/binary.test.ts
chrisguttandin: I updated the domain. I got it from copying it from another test. Should I create an issue for updating all the test domains to use `test.mswjs.io`?

kettanaito: Sorry, some tests were written at a different time, and may give a false example. 
No need, let's cover that as a separate change. Are you okay with that?

chrisguttandin: Yes, absolutely.

kettanaito:
I will test these changes locally (not sure why CI is not running, I've got forked builds enabled) and let you know! Meanwhile, could you please tell me if `yarn test` passes for you in this feature branch? Thanks.

chrisguttandin:
Thanks, I ran the tests locally as well and they kept passing for me.

src/utils/createBroadcastChannel.ts
chrisguttandin: I initially left it untouched as it seems to be not necessary to change this to support binary response types. However I tried to remove the serialization inside of `handleRequestWith()` now but I couldn't get it to work without it.

test/rest-api/binary.mocks.ts
```diff
@@ -0,0 +1,9 @@
+import { setupWorker, rest } from 'msw'
+
+const worker = setupWorker(
+  rest.get('https://test.mswjs.io/binary', (_, res, ctx) => {
+    return res(ctx.body(new Uint8Array([1, 2, 3, 4, 5, 6])))
```
kettanaito: I wonder, can we technically use an image as a mocked response body here?
That would be a more practical example.

test/rest-api/binary.mocks.ts
```diff
@@ -0,0 +1,9 @@
+import { setupWorker, rest } from 'msw'
```
kettanaito: Could you please rename this file to start with `body-`? We are getting more different body usage scenarios, it would be nice for them to be next to each other in the file tree.

```diff
- binary.mocks.ts
- binary.test.ts
+ body-binary.mocks.ts
+ body-binary.test.ts
```

src/utils/createBroadcastChannel.ts
kettanaito: For instance, that other serialization is used to convert `request.headers` to string, and then construct a `Headers` instance out of it during deserialization on the client.

test/rest-api/binary.mocks.ts
chrisguttandin: Yes, sure but in the end it will just be binary data. I thought the tests are simpler if they don't have to load an image first.

src/utils/createBroadcastChannel.ts
chrisguttandin: Ah, I see. I think it absolute makes sense to have a custom serialization for `Headers` as they will not be handled by the structural cloning algorithm that comes with `postMessage()`.

test/rest-api/binary.mocks.ts
chrisguttandin: Okay, I renamed the files.

test/rest-api/binary.mocks.ts
kettanaito: I see. I don't mind leaving it here for now and, perhaps, including a more realistic usage example in the docs' recipes.

kettanaito (dismissed):
Amazing changes! Thank you, @chrisguttandin. I'd like to release this as a minor release, as it's effectively a new feature. Let's merge it once #205 is done.

chrisguttandin:
Great. Once the release is done I'll try to put together a working example which shows how to use an image and add it to https://github.com/mswjs/mswjs.io/tree/master/docs/recipes.

kettanaito:
That would be lovely! Thank you so much for your contribution :)
Also, if you don't mind, I'd like to give you a credit for the great work you're doing. Do you happen to have a Twitter account I could reference when announcing this feature?

> This is completely up to you, I respect your privacy, so if you're against this I'm perfectly fine. Just let me know.

chrisguttandin:
Yes, sure. I go by the same name on Twitter. It's very unique. :-)

kettanaito:
I'm planning to release this in the next minor release (`0.20.0`). I'll give the previous release some time, in case any bugs would surface, and then start merging this feature and others to prepare them to the next release.

Thank you once more for the awesome work!

chrisguttandin:
Alright, thanks for letting me know.

kettanaito:
Updated the branch to include the latest CI improvements from the `master`. Now the PR build would verify that changes do not break usage examples.

kettanaito:
I've added the [Binary response type](https://github.com/mswjs/mswjs.io/pull/22) documentation page. @chrisguttandin, may I please ask you to take a look at that? It'd be awesome to get your opinion if you find that recipe useful.

I see that one of the examples fails after our changes (`test-smoke` CI job). Will try to look into it to find out what's wrong. Once the example is green we will merge this feature.

kettanaito:
I've discovered that the smoke test script in the Examples repository is misconfigured: it doesn't update the `mockServiceWorker.js` file, that results into any incremental changes to that file being ignored. The fix for that is issued.

kettanaito (approved):
Awesome! I'm super excited for this new feature. Thank you, @chrisguttandin 🎉 I'm glad to have you as a contributor.

PR #229, opened 2020-06-18
Use userEvent instead of fireEvent (as it is not imported / consistency)

markus-willems:
Hey, I found a small error in README.md.

The React example in the README makes use of `fireEvent`. However, it's not being imported from `@testing-library/react`. For consistency, use `userEvent` instead.


kettanaito (approved):
Hey, @markus-willems! Thanks for spotting this! Awesome fix.

PR #232, opened 2020-06-19
fix: add warning when mocking url with query params

### LINKED ISSUE #231: There is no warning when using query params in the handler URL
## Describe the bug
The following mock will never work since the query strings are removed from the original URL but not from the handler URL:

```js
rest.get("/foo?bar=10", (req, res, ctx) => {
  return res(
    ctx.json({
      foo: "bar",
    })
  );
}),
```

## Environment

- `msw: 0.19.0`
- `nodejs: 12.17.0`
- `npm: 6.14.4`

## To Reproduce
1. Add the handler mentioned above
2. In the app, do a request to `/foo?bar=10`
3. Error: the request is not mocked

## Expected behavior
There should be an warning explaining why the mock won't work and how to fix the problem

marcosvega91:
fix #231 

When you try to add a rest api where the `Mask` contains query parameters should give a warning.

I don't know if a warning message is useful because then you can't use the API, because if you try to call the API `getCleanUrl` will clean everything from URL.

Could be better to throw an error?

### DIFF
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -51,6 +51,12 @@ const createRestHandler = (method: RESTMethods) => {
     mask: Mask,
     resolver: ResponseResolver<MockedRequest, typeof restContext>,
   ): RequestHandler<MockedRequest, typeof restContext> => {
+    if (typeof mask === 'string' && new RegExp(/\?.+=.*/g).test(mask)) {
+      console.warn(
+        `It should not be good to add query parameters in the mask. 
+        It's better to use without them`,
+      )
+    }
     return {
       predicate(req) {
         // Ignore query parameters and hash when matching requests URI
--- a/src/setupWorker/setupWorker.test.ts
+++ b/src/setupWorker/setupWorker.test.ts
@@ -1,16 +1,15 @@
 import { setupWorker } from './setupWorker'
 import { rest, restContext } from '../rest'
 import { MockedRequest, ResponseResolver } from '../handlers/requestHandler'
+const simpleResolver: ResponseResolver<MockedRequest, typeof restContext> = (
+  req,
+  res,
+  { json },
+) => {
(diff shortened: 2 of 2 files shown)


kettanaito:
Hey, @marcosvega91! Thanks for shipping this. Sorry, I've coordinated it poorly, as I've been working on this for some time now. I'll close this in favor of #233. Please, I'd appreciate your code review on that pull request, if you don't mind.

marcosvega91:
I was looking your code, it's not a problem 😄

kotesh-arya:
But this warning should be **more targeted** because even If I use a Complete URL in my handler where I have to filter users using the query parameter itself, It's still showing up the `WARNING of redundant usage of query parameters`

As mentioned in this Issue #231, we should only show the warning when someone tries to use just query-parameter/s inside a handler, but not in the general case of a complete URL.

Being a beginner to Testing and using MSW for the first time, I believe a WARNING is something a beginner bothers with.
You can check this out : 
image

kettanaito:
Hi, @kotesh-arya. I believe that warning you're seeing is correct. You are including a `?state=IDLE` query parameter in your request handler URL. That won't have any effect because query parameters do not describe resource paths on the server (which is what you're trying to mock with MSW). Respectfully, the library warns you about that and suggests to drop that query parameter since it does nothing and may only confuse anybody reading the handler.

kotesh-arya:
> Hi, @kotesh-arya. I believe that warning you're seeing is correct. You are including a `?state=IDLE` query parameter in your request handler URL. That won't have any effect because query parameters do not describe resource paths on the server (which is what you're trying to mock with MSW). Respectfully, the library warns you about that and suggests to drop that query parameter since it does nothing and may only confuse anybody reading the handler.

Got it, thanks for the Heads-up!

Issue #228, opened 2020-06-18
Redirect response


mitchconquer:
**Is your feature request related to a problem? Please describe.**
Hi, I am creating a little app to handle OAuth and it needs to be able to mock redirect responses. 

**Describe the solution you'd like**
Is it possible implement redirect responses?

**Describe alternatives you've considered**
I tried overwriting the response object to include the Location header but it seems to get erased and I couldn't figure out why.

kettanaito:
Hey, @mitchconquer. Thanks for raising this. Could you please include the request handler code you've used to mock a redirect response?

I think you should be able to redirect by setting a proper response status code, and location header.

```js
rest.get('/auth', (req, res, ctx) => {
  return res(ctx.status(301), ctx.set('Location', 'https://google.com'))
})
```

I don't have any test currently to verify that. Let's use this issue as an opportunity to do that.

kettanaito:
I've added an integration test covering HTTP redirects:

https://github.com/mswjs/msw/blob/ebb85fb9f82c3e1bb0c117ebfc4b77e557d8cc61/test/rest-api/redirect.mocks.ts#L1-L17

Could you please take a look at tell me if that answers your question, @mitchconquer? I believe it behaves as the regular redirect at the moment.

Issue #227, opened 2020-06-18
undefined response when mocking axios get


lammypham:
## Describe the bug
I have a `server.get` set up to listen to a request url but whenever I am running the tests, I seem to end up with `undefined` in my response

## Environment

- `msw": "^0.19.3`
- `nodejs: 12.13.0`
- `npm: 6.12.0`

## To Reproduce
1. Run Test
2. Check to see if component is rendered correctly
3. See that loading component is still there.

## Expected behavior
I should be able to intercept the request and consume the payload I give `ctx`

## Screenshots
```const server = setupServer(
  rest.get('*', req => console.log(req.url.href)),
  rest.get(`/api/portfolioCalls`, (req, res, ctx) => {
    return res(ctx.status(200), ctx.json(propertiesData))
  }),
)
describe('PortfolioReport', () => {
  beforeAll(() => server.listen())
  afterAll(() => server.close())
  afterEach(() => server.resetHandlers())

  const checkForUserActivity = jest.fn()
  const switchToPropertyView = jest.fn()
  const setDateRange = jest.fn()
  beforeEach(() => {})

  it('should render component', async () => {
    act(() => {
      render(
        <UserContextProvider user={userData}>
          <PortfolioReport
            checkForUserActivity={checkForUserActivity}
            switchToPropertyView={switchToPropertyView}
            setDateRange={setDateRange}
          />
        </UserContextProvider>
      )
    })
    const { getByText } = screen
    expect(getByText('Portfolio')).toBeInTheDocument()
    await wait(() => {
      expect(getByText('Total Calls')).toBeInTheDocument()
    })

    screen.debug()
  })
})
```

this is the axios call i am trying to listen to
```
 try {
        const result = await axios.get(
          `${HOSTNAME}/api/portfolioCalls?startDate=${startDate}&endDate=${endDate}`
        )
        if (result?.status === 200 && result?.data) {
          const formattedData = formatData(result.data)
          dispatch({
            type: FETCH_PORTFOLIO_SUCCESS,
            payload: formattedData.properties,
            status: SUCCESS,
          })
          setData(formattedData)
        } else setData({})
      } catch (error) {
        dispatch({ type: FETCH_PORTFOLIO_ERROR, status: ERROR })
        console.log(error)
      }
    }
```

kettanaito:
Hey, @lammypham. Thanks for transferring this issue to the repo!

Could you please show me what is the value of `HOSTNAME` you use to compose a request URL for `axios`?

[...]

Issue #218, opened 2020-06-16
Question: GraphQL queries are not intercepted in Jest unit test


thawsitt:
## Issue
I want to use `msw` in Jest unit tests only. However, GraphQL queries are not being intercepted. Our stack is React + Jest + Apollo GraphQL.

I have already looked at https://github.com/mswjs/examples/tree/master/examples/graphql-react-apollo which is very helpful. Thanks for that.

## Environment

- `msw: 0.19.3`
- `nodejs: v12.14.0`
- `npm: 6.13.4`
- `Browser: Chrome 83.0.4103.97`

I have 2 questions.

### 1. Do I need any extra setup to make `msw` work in node (Jest unit tests)?

I believe this command is required to make `msw` work in browser for development. Do I need to do something similar for node?
```
$ npx msw init <PUBLIC_DIR>
```

### 2. Is there any config needed for different url end point?

Our end point is `/prism` instead of `/graphql`. Do I need to change any config to make it work besides [setting up apollo client](https://github.com/mswjs/examples/blob/master/examples/graphql-react-apollo/src/ApolloClient.js)?

Note: I have also tried changing our api url to `/graphql` but it still doesn't work.

thawsitt:
Just for debugging purposes, I added this in my Jest unit test file. (I only do this once. I comment this out in other runs)
```ts
const fn: any = (url: any, config: any) => {
    console.warn(url);
    console.warn(config);
};
jest.spyOn(window, 'fetch').mockImplementation(fn);
```

Here is what it prints out in terminal when I run the test.
```
http://localhost:3000/graphql

{
      method: 'POST',
      headers: { accept: '*/*', 'content-type': 'application/json' },
      credentials: undefined,
      signal: AbortSignal {},
      body: '{"operationName":"GetSourceDetail","variables":{"ids":["f3701660-1c06-4ff5-93e0-5ac772581553","0cab5f37-6050-4009-94ac-40509a40c330"]},"query":"query GetSourceDetail($ids: [String!]!) {\\n  getSourceDetailById(ids: $ids, type: LOGICAL_TABLE) {\\n    id\\n    name\\n    columns {\\n      name\\n      id\\n      dataType\\n      modified\\n      __typename\\n    }\\n    __typename\\n  }\\n}\\n"}'
    }
```

---

My `handlers.js` file
```js
// graphql-handlers.js
import { graphql } from 'msw';

export const handlers = [
    graphql.query('GetSourceDetail', (req, res, ctx) => {
[...]

Issue #217, opened 2020-06-16
Stopping and starting MSW will cause multiple "instances"


bjarkehs:
## Describe the bug
I have a use case where only I start MSW when user selects "development" data. First time it works fine, but if the user logs out and in again, i.e. stopping and starting MSW, then now it'll return twice requests twice. It will also write "Mocking enabled" twice. Basically every time I stop and start MSW it adds an extra instance.

## Environment

- `msw: 0.19.3`
- `nodejs: 13.11.0`
- `yarn: 1.22.4`

Microsoft Edge: Version 83.0.478.50

## To Reproduce
- Start a worker.
- Call .stop() on that worker.
- Start a new worker.

## Expected behavior
Expected mocking to only be enabled once

kettanaito:
Hey, @bjarkehs! Thank you for reporting this. That's definitely not the expected behavior.

I've also experienced a similar issue when doing a code change that produces a Fast Refresh update chunk, while preparing a [NextJS usage example](https://github.com/vercel/next.js/pull/13731). I'd be great if these two issues have the same origin, as their manifestations are almost identical.

marcosvega91:
I was looking into this and the problem is that when starting a new worker the instance remains the same( same clientId)  so the postMessage is handled by the stopped instance and the new one. 
I think that maybe we can remove handlers when we stop the insrance. What do you think?

kettanaito:
Looks interesting. Do I understand it correctly, that we are talking about this listener?

https://github.com/mswjs/msw/blob/219ca18b538957e6e44f4444f87ece528ceffbe3/src/setupWorker/start/utils/activateMocking.ts#L12-L32

It should also affect the listener responsible for response resolution, as the issue reports responses being sent twice as well. 

@marcosvega91, I think it's a good idea to remove all listeners from a worker instance once called `worker.stop()`. Would you be interested in making a proof of concept that it solves the issue?

marcosvega91:
Yes of course this listener but it will only print to console. So it's a problem but not the real problem. 
I think that the most important is the part below, because it is responsible for fetching :https://github.com/mswjs/msw/blob/219ca18b538957e6e44f4444f87ece528ceffbe3/src/setupWorker/start/createStart.ts#L37-L39

Maybe we could use `sendToClient` with a type `MOCK_DEACTIVATE` and intercept the message where we add listeners and remove them.

kettanaito:
[...]

PR #233, opened 2020-06-19
Warns on the redundant usage of query parameters

### LINKED ISSUE #231: There is no warning when using query params in the handler URL

kettanaito:
## Changes

- Resolves a `rest.*` mask to a `URL` instance, if possible. That instance is later used for both request matching in predicate and retrieving URL query params to produce a meaningful warning message on redundant query parameters.

## GitHub

- Fixes #231

### DIFF
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -88,7 +88,31 @@ const createRestHandler = (method: RESTMethods) => {
         return restContext
       },
 
-      log(req, res, handler) {
+      log(req, res, handler, parsed) {
+        // Warn on request handler URL containing query parameters.
+        if (resolvedMask instanceof URL && resolvedMask.search !== '') {
+          const queryParams: string[] = []
+          resolvedMask.searchParams.forEach((_, paramName) =>
+            queryParams.push(paramName),
+          )
+
+          console.warn(
+            `\
+[MSW] Found a redundant usage of query parameters in the request handler URL for "${method} ${mask}". Please match against a path instead, and access query parameters in the response resolver function:
+
+rest.${method.toLowerCase()}("${resolvedMask.pathname}", (req, res, ctx) => {
+  const query = req.url.searchParams
+${queryParams
+  .map(
+    (paramName) => `\
+  const ${paramName} = query.get("${paramName}")`,
+  )
+  .join('\n')}
+})\
+`,
+          )
+        }
+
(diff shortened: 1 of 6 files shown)


marcosvega91 (approved):
awesome work 🥳

src/rest.ts
```diff
@@ -51,12 +52,20 @@ const createRestHandler = (method: RESTMethods) => {
...
+        ? resolvedMask
+        : resolveRelativeUrl(resolvedMask)
+
     return {
       predicate(req) {
         // Ignore query parameters and hash when matching requests URI
         const cleanUrl = getCleanUrl(req.url)
         const hasSameMethod = isStringEqual(method, req.method)
-        const urlMatch = match(resolveRelativeUrl(mask), cleanUrl)
+        const urlMatch = match(cleanMask, cleanUrl)
```
kettanaito, line 68: I don't think it's necessary to pass a `cleanMask` here. It should be possible to revert back to `resolveRelativeUrl(mask)` usage, which would make `cleanMask` not needed (?).

marcosvega91: without the use of `cleanMask` I think that match will fail, because you'll compare the mask, with query parameters, and the cleanUrl

kettanaito: Nice catch! Thanks for highlighting that, I suppose I've overlooked that :)

kettanaito:
Thank you for the review, @marcosvega91!

Once more sorry for turning down the original pull request. Hey, if you're still interested in contributing to MSW, I can suggest #217, #218, #227, #228. There are quite a few issues I haven't got to yet, I'd appreciate a help. Thanks.

src/rest.ts
marcosvega91: It's good to contribute in someway :)

marcosvega91:
I'll take a look to one of them 🦹

PR #242, opened 2020-06-23
fix: improve error message on registration failed 

marcosvega91:
I have changed the error message that will be shown if  `navigator.serviceWorker.register` throw an error.

Now the error shows the complete message text from error object and a tip in the case the user didn't run `npx msw init <PUBLIC_DIR>` in his project.

An example of message is:

`[MSW] Failed to register a ServiceWorker for scope ('http://localhost:62036/') with script ('http://localhost:62036/invalidServiceWorker'): A bad HTTP response code (404) was received when fetching the script. 
            If the worker file has not been found maybe you didn't run "npx msw init <PUBLIC_DIR>"`

### DIFF
--- a/src/setupWorker/start/utils/activateMocking.ts
+++ b/src/setupWorker/start/utils/activateMocking.ts
@@ -18,7 +18,7 @@ export const activateMocking = (
             'color:orangered;font-weight:bold;',
           )
           console.log(
-            '%cDocumentation: %chttps://redd.gitbook.io/msw',
+            '%cDocumentation: %chttps://mswjs.io/docs',
             'font-weight:bold',
             'font-weight:normal',
           )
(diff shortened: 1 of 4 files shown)


marcosvega91:
I need to update the test, my stupid fault ✌️

marcosvega91:
now is ready, let me know :)

kettanaito:
Awesome changes, @marcosvega91! Thank you.

I've updated your feature branch, and slightly modified the error message when the worker is not found. I think it's better if we are definitely sure that the non-existing worker is the issue, and can provide a more targeted error message with instructions. All the other errors are handled using a generic error message printed to browser's console.

kettanaito (approved):
Thank you once more, @marcosvega91! Welcome to contributors 🎉

PR #243, opened 2020-06-23
setupServer: use timers module

kentcdodds:
This is all that needs to be done to avoid issues with this code:

```javascript
test('tolerates fake timers', async () => {
  server.use(
    rest.get('https://test.mswjs.io/pull', (req, res, ctx) => {
      return res(ctx.json({status: 'pulled'}))
    }),
  )
  jest.useFakeTimers()

  const res = await fetch('https://test.mswjs.io/pull')
  const body = await res.json()

  jest.useRealTimers()

  expect(body).toEqual({status: 'pulled'})
})
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,5 +20,6 @@ module.exports = {
       },
     ],
     '@typescript-eslint/no-namespace': 0,
+    '@typescript-eslint/no-var-requires': 0,
   },
 }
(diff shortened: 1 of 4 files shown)


kettanaito (approved):
Happy that you've figured out the right solution to this! Thank you and welcome to contributors! 🎉

PR #253, opened 2020-06-29
Added feature: res.networkError() mocks a network error

### LINKED ISSUE #280: Request handler should allow returning undefined or void
## Current behavior

Returning `undefined` or void raises a TypeScript error.

## Expected behavior

Request handler should allow returning undefined or void. This way request handlers support a fall-through functionality, as well as short-circuiting.

conifer64:
I added the feature so you can now use `res.networkError()` inside a request handler to mock a network request. This is useful because this will trigger the Fetch API to throw an error, and you can thus test whether your function handles this error correctly. 

## GitHub

- Fixes #280

### DIFF
--- a/test/msw-api/res/network-error.mocks.ts
+++ b/test/msw-api/res/network-error.mocks.ts
@@ -1,0 +1,10 @@
+import { setupWorker, rest } from 'msw'
+
+const worker = setupWorker(
+  rest.get('/user', (_, res) => {
+    return res.networkError()
+  }),
+)
+
+// @ts-ignore
+window.__MSW_START__ = worker.start
--- a/test/msw-api/res/network-error.mocks.ts
+++ b/test/msw-api/res/network-error.mocks.ts
@@ -1,8 +1,8 @@
 import { setupWorker, rest } from 'msw'
 
 const worker = setupWorker(
-  rest.get('http://test.io/user', (_, res) => {
-    return res.networkError('Network error')
+  rest.get('/user', (_, res) => {
+    return res.networkError('Custom network error message')
   }),
 )
 
--- a/test/msw-api/res/network-error.mocks.ts
+++ b/test/msw-api/res/network-error.mocks.ts
@@ -6,5 +6,4 @@ const worker = setupWorker(
   }),
 )
 
-// @ts-ignore
-window.__MSW_START__ = worker.start
+worker.start()
--- a/test/msw-api/res/network-error.mocks.ts
+++ b/test/msw-api/res/network-error.mocks.ts
@@ -1,7 +1,7 @@
 import { setupWorker, rest } from 'msw'
 
 const worker = setupWorker(
-  rest.get('/user', (_, res) => {
+  rest.get('http://test.io', (_, res) => {
     return res.networkError()
   }),
 )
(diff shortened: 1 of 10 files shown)


test/msw-api/res/network-error.test.ts
```diff
@@ -0,0 +1,16 @@
...
+beforeAll(async () => {
+  runtime = await runBrowserWith(
+    path.resolve(__dirname, 'network-error.mocks.ts'),
+  )
+})
+
+afterAll(() => runtime.cleanup())
+
+test('res.networkError causes Fetch API to throw error', async () => {
+  expect(runtime.request({ url: `${runtime.origin}/user` })).toThrow()
```
kettanaito: Does this behavior satisfies your network failure mocking scenario?

test/msw-api/res/network-error.mocks.ts
```diff
@@ -0,0 +1,10 @@
+import { setupWorker, rest } from 'msw'
+
+const worker = setupWorker(
+  rest.get('/user', (_, res) => {
+    return res.networkError()
+  }),
+)
+
+// @ts-ignore
+window.__MSW_START__ = worker.start
```
kettanaito: You don't have to delegate the start like this (I believe this is done only for specific tests). Please, call `worker.start()` directly.

test/msw-api/res/network-error.test.ts
jensmeindertsma: Currently the tests fail because the `networkError` function returns an empty object instead of throwing. I don't know where the empty object is coming from. Could you point me in the right direction?

test/msw-api/res/network-error.mocks.ts
jensmeindertsma: I changed this in the latest commit

kettanaito:
Hey. I've looked into your implementation and it looks good!

What happens now is that whenever you use `res.networkError()` it rejects a request promise and results into an unhandled error in the browser's console. However, that doesn't disrupt the request resolution, and since nothing has been returned from the handler, the request gets performed as-is. This _also_ results into a Network error (considering you're requesting a non-existing resource). 

I believe we need to find a way to cancel a pending request (or respond with a network error from the worker/node), so it doesn't get executed. `res.networkError()` is, effectively, a short-circuit for a request.

kettanaito:
I have some new information on how to properly handle an exception thrown within a request handler. Will update soon, it's a change to `node-request-interceptor`.

kettanaito:
Hey. I've rebased your feature branch against the latest `master` that should contain the changes necessary to treat exceptions in request handlers as network errors. Will revisit tests and CI.

jensmeindertsma:
Nice! I'm very busy this weekend but I'll try to take a look later.

kettanaito:
I've updated the branch, and added the logic that treats the exception thrown in `res.networkError()` as an _intended_ exception. Previously, we've treated any exception originating from a request handler as unintended (what if your function suddenly throws? that's most likely an issue with your function). With the introduction of network errors, we have a different group of intended exceptions now.

It's possible to emulate a network error on the Service Worker side by rejecting the pending request promise. Once done, it produces the very same errors as a regular network error. 

The only thing I couldn't get to is asserting the custom network error message in the browser test. Somehow both fetch rejection and captured console errors don't have it.

kettanaito (approved):
This looks awesome. I'm excited to ship `res.networkError` feature in the next release. 

Superb job, @jensmeindertsma! Thank you and welcome to contributors 🎉

Issue #203, opened 2020-06-07
React Native support


hsavit1:
**Is your feature request related to a problem? Please describe.**
Does React Native have the ability to support this?

**Describe the solution you'd like**
Some guide to do something similar with react native

**Describe alternatives you've considered**
Can something custom be set up with https://github.com/Blitz-Mobile-Apps/react-native-bg-thread#readme ?

**Additional context**

kettanaito:
I wonder if using `setupServer` in React Native would be enough. I'd appreciate if somebody could do a proof of concept with that.

hsavit1:
do you think, for react native, that it could work for grahphql requests (via apollo client) as well?

kettanaito:
@hsavit1, I believe so. It's a matter of understanding a request flow in React Native. I assume React Native app runs on Node, so that's why I've suggested to check if using `setupServer` doesn't intercept its requests already (that's the API that powers API mocking in Node). 

As of GraphQL, it's not much different than a regular REST request. In fact, it is a regular REST request, just compliant with the GraphQL specification (has a `query`, optional `variables`, and so on).

hsavit1:
I believe React Native has _some_ access to Node.js APIs via the node-js mobile library. See [this](https://code.janeasystems.com/nodejs-mobile/react-native/bridge) for more. 

Read more about the javascript core here: https://reactnative.dev/docs/0.60/javascript-environment

donovanhiland:
Stumbled across this issue while trying to set up a react native app to use msw for integration tests. I would love to help get this working for react-native outside of tests if someone could point me in the right direction.

kettanaito:
Hey, @donovanhiland! We'd really appreciate your help, if you have a spare minute. 

Could you please try using `setupServer` (regular integration with Node) in your React Native integration tests? Here's the [documentation](https://mswjs.io/docs/getting-started/integrate/node) on the process. Since tests run in Node environment, that integration should work. But due to the difference of the platform, it may have some unexpected behavior. 

Would love to hear how it goes from you.

donovanhiland:
Absolutely! @kettanaito I was able to get `setupServer` working in my react-native apollo jest integration test environment. There was a small caveat for react-native tests where I had to set a global fetch implementation so `msw` could intercept. Otherwise `fetch` could not be found.

[...]

PR #255, opened 2020-06-30
fix(isNodeProcess): avoid undefined error

jameslahm:
Hi, Thanks for the amazing tool. I try to use it in a browser, but `global is not defined` came out. I think it maybe caused by `isNodeProcess` function. Change it to `typeof process ==='object'` may solve this problem.

### DIFF
--- a/src/utils/isNodeProcess.react-native.test.ts
+++ b/src/utils/isNodeProcess.react-native.test.ts
@@ -1,47 +1,55 @@
-// @ts-nocheck
 /**
- * @jest-environment jsdom
+ * @jest-environment node
+ *
+ * This test suite asserts `isNodeProcess` behavior in the React Native environment.
+ * Its setup hooks emulate the aforementioned environment as much as it's possible.
  */
-/**
- * Test in React Native Environment
- * Please see https://github.com/mswjs/msw/pull/255
- */
-
 import { isNodeProcess } from './isNodeProcess'
 
+// Provide TypeScript overrides to operate with the `global` namespace
+declare global {
+  namespace NodeJS {
+    interface Global {
+      navigator: {
+        product: string
+      }
+    }
+  }
+}
+
 // need to mock process and navigator in React Native and restore them
-let originalNavigator
-let originalProcess
+let originalProcess: typeof global['process']
+let originalNavigator: typeof global['navigator']
 
 beforeAll(() => {
-  // back up
-  originalNavigator = global.navigator
+  // Store the original `global` properties.
   originalProcess = global.process
+  originalNavigator = global.navigator
 
   /**
-   * because it runs in node, Object.proyotype.toString(global.process) will equal '[object process]',
-   * we need to mock global.process and global.navigator. let them behave like in React Native
-   * PLease see https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpNavigator.js and
-   * https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpGlobals.js
+   * React Native has a unique property to determine that the process
+   * is running in a React Native environment.
+   * @see https://github.com/facebook/react-native/issues/1331
+   * @see https://github.com/facebook/react-native/blob/6d6c68c2c639b6473e049f7d916690b92e921c7e/Libraries/Core/setUpNavigator.js
+   * @see https://github.com/facebook/react-native/blob/6d6c68c2c639b6473e049f7d916690b92e921c7e/Libraries/Core/setUpGlobals.js
    */
-  global.process = {
-    env: {
+  Object.defineProperty(global.process, 'env', {
+    get: () => ({
       NODE_ENV: 'development',
-    },
-  }
-  Object.defineProperty(global.navigator, 'product', {
-    get: () => {
-      return 'ReactNative'
-    },
+    }),
   })
-})
 
-test('returns true when run in a react native environment', () => {
-  expect(isNodeProcess()).toBe(true)
+  global.navigator = {
+    product: 'ReactNative',
+  }
 })
 
 afterAll(() => {
-  // restore them
+  // Restore the stubbed `global` properties.
   global.process = originalProcess
   global.navigator = originalNavigator
+})
+
+test('treats React Native as a NodeJS environment', () => {
+  expect(isNodeProcess()).toBe(true)
 })
--- a/src/utils/isNodeProcess.react-native.test.ts
+++ b/src/utils/isNodeProcess.react-native.test.ts
@@ -1,25 +1,47 @@
 // @ts-nocheck
 /**
+ * @jest-environment jsdom
+ */
+/**
  * Test in React Native Environment
+ * Please see https://github.com/mswjs/msw/pull/255
  */
 
 import { isNodeProcess } from './isNodeProcess'
 
-test('returns true when run in a react native environment', () => {
-  const originalProcess = global.process
+// need to mock process and navigator in React Native and restore them
+let originalNavigator
+let originalProcess
+
+beforeAll(() => {
+  // back up
+  originalNavigator = global.navigator
+  originalProcess = global.process
+
+  /**
+   * because it runs in node, Object.proyotype.toString(global.process) will equal '[object process]',
+   * we need to mock global.process and global.navigator. let them behave like in React Native
+   * PLease see https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpNavigator.js and
+   * https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpGlobals.js
+   */
   global.process = {
     env: {
       NODE_ENV: 'development',
     },
   }
-
-  const originalNavigator = global.navigator
   Object.defineProperty(global.navigator, 'product', {
     get: () => {
       return 'ReactNative'
     },
   })
+})
+
+test('returns true when run in a react native environment', () => {
   expect(isNodeProcess()).toBe(true)
+})
+
+afterAll(() => {
+  // restore them
   global.process = originalProcess
   global.navigator = originalNavigator
 })
--- a/src/utils/isNodeProcess.react-native.test.ts
+++ b/src/utils/isNodeProcess.react-native.test.ts
@@ -1,0 +1,25 @@
+// @ts-nocheck
+/**
+ * Test in React Native Environment
+ */
+
+import { isNodeProcess } from './isNodeProcess'
+
+test('returns true when run in a react native environment', () => {
+  const originalProcess = global.process
+  global.process = {
+    env: {
+      NODE_ENV: 'development',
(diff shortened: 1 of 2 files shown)


src/utils/isNodeProcess.ts
```diff
@@ -2,5 +2,5 @@
  * Returns a boolean indicating if the current process is running in NodeJS environment.
  */
 export function isNodeProcess() {
-  return Object.prototype.toString.call(global.process) === '[object process]'
```
kettanaito, line 5: Can we use optional chaining here?

```js
global?.process
```

The reason why `global.process` is used and its prototype is stringified, is because not any object is necessarily a `process` object. That is to prevent false positives, when you have a global `process` object in your environment, but it's not Node.

kettanaito (changes requested):
Thank you for spotting this, @jameslahm! I've had one suggestion to the changes, could you please let me know what you think about it?

src/utils/isNodeProcess.ts
jameslahm: Thanks for the reply. But I think it may also cause `global is not defined error` ,because typescript will compile `global?.process` to `global === null || global === void 0 ? void 0 : global.process` .In browser, `global` is still not defined. So I think Change to `typeof global === 'object' && typeof global.process==='object'` may be better, what do you think about it?

kettanaito: That's a good point. We also need to take into account a React Native process, where, I believe, the `global` is set, but it doesn't have the `globlal.process` property (#203). 

```js
function isNodeProcess() {
  if (typeof global === 'undefined') {
    // Assuming that in React Native `global` exists
    return false
  }

  if (Object.prototype.toString.call(global.process) === '[object process]') {
    // Regular NodeJS process
    return true
  }

  if (SOME_REACT_NATIVE_CONDITION) {
    // We should treat a React Native environment similar to Node,
    // because you cannot setup a Service Worker in React Native,
    // just as you can't in Node.
    return true
  }
}
```

jameslahm: Thanks for the reply. I noticed it seems `global.process`  is also be set in React Native process. [Here](https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpGlobals.js) can see its definition. I agree that react native environment should be treated similarly to Node because the service worker is not available. Also, React Native implements XMLHttpRequest and fetch in its environment. Its fetch is implemented by `whatwg-fetch` polyfill, which is also based on XMLHttpRequest. I noticed `node-request-interceptor` intercepts requests using XMLHttpRequest. So it should also work in the React Native environment. But I think there is still one problem that `node-request-interceptor` has used `http` core module, [which is not supported by React Native](https://github.com/facebook/react-native/issues/15048). Although it can be patched by other packages, I think providing a function alone used for React Native may be better, such as setUpReactNativeServer? And may only need to intercept XMLHttpRequest.

kettanaito: That's a fair observation regarding the `http` module. I don't think that's a sufficient reason to ship a separate API, but what we may do is check whether `http`/`https` modules exist, and attach the interception only if they do (as a part of `node-request-interceptor` library).

jameslahm: Yes, I agree with you. So if check whether `http`/`https` module exist, should we distinguish React Native and node process in `isNodeProcess` function? I think if we check `http`/`https` module, then check `global` and `global.process` and treat React Native and node process similarly may be better. For example, like this
```
function isNodeProcess() {
    return typeof global === 'object' && typeof global.process==='object'
}
```
What do you think of it?

kettanaito: Yeah, looks good. I'd keep the prototype check, so an arbitrary object won't be treated as a `process`, even if it shares the namespace:

```js
function isNodeProcess() {
  return typeof global === 'object' && Object.prototype.toString.call(global.process) === '[object process]'
}
```

On the related note, I had some reports that `globals` is not defined in React Native (thanks for the source reference shared previously!). I'd like to approach that with caution. Do you have a chance to test a React Native project to see how this `isNodeProcess` function behaves there? It's expected to return `true` for React Native.

jameslahm: I try `console.log(global.process)` in React Native using `expo`, It shows 
```
Finished building JavaScript bundle in 40915ms.
Running application on sdk_gphone_x86.
Object {
  "env": Object {
    "NODE_ENV": "development",
  },
}
```
But `Object.prototype.toString.call(global.process)` shows `[object Object]`, thus Object.prototype.toString.call(global.process) === '[object process]' will be broken.
I noticed [here](https://github.com/facebook/react-native/issues/1331), navigator.product was set to 'ReactNative'. Also [here](https://github.com/facebook/react-native/blob/master/Libraries/Core/setUpNavigator.js) can see its definition. I also try `console.log(navigator.product) `, it shows `ReactNative`.
So I suggest this, just like you mentioned before
```
function isNodeProcess() {
  if (typeof global === 'undefined') {
    // Assuming that in React Native `global` exists
    return false
  }

  if (Object.prototype.toString.call(global.process) === '[object process]') {
    // Regular NodeJS process
    return true
  }

  if (navigator.product==='ReactNative') {
    // We should treat a React Native environment similar to Node,
    // because you cannot setup a Service Worker in React Native,
    // just as you can't in Node.
    return true
  }
}

```
What do you think about it?

kettanaito: That's yet again a great find. I like relying on `navigator.product` 👍 

I'd only suggest to unify last two conditions, as they both are treated the same:

```js
if (typeof global !== 'object') {
  return false
}

if (
  Object.prototype.toString.call(global.process) === '[object process]' ||
  navigator.product==='ReactNative'
) {
  return true
}
```

- checking if `typeof global` is not an `object`, to prevent global variable `global` from being treated as an env's global.
- unifying last two conditions.

Looks good to you?

jameslahm: Yes, Looks good to me. I have changed the code like that. Thanks for the reply!

kettanaito (dismissed):
Looks awesome!

I'd love to see a unit test in a React Native environment, but to reliably bootstrap one would be an overkill in our testing pipeline. We'd have to maintain that `navigator` manually. 

Let's wait for the CI to pass.

jameslahm:
Yes, there should be a unit test. I add a React Native Environment test just now. Copying environment settings from `React Native`, and because typescript will check `Process` type properties, which will break the mock of `global process`, I add `ts-nocheck`. What do you think about it?

src/utils/isNodeProcess.react-native.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+  }
+
+  const originalNavigator = global.navigator
+  Object.defineProperty(global.navigator, 'product', {
+    get: () => {
+      return 'ReactNative'
+    },
+  })
+  expect(isNodeProcess()).toBe(true)
```
kettanaito: If the assertion fails, we will never restore the original validator (the function will short-circuit).
I highly recommend to stub `global.navigator` in the `beforeAll` and restore it in `afterAll` hooks in Jest.

src/utils/isNodeProcess.react-native.test.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Test in React Native Environment
+ */
+
+import { isNodeProcess } from './isNodeProcess'
+
+test('returns true when run in a react native environment', () => {
+  const originalProcess = global.process
+  global.process = {
+    env: {
+      NODE_ENV: 'development',
```
kettanaito: Please, why do we have to simulate the development value here?

kettanaito (changes requested):

src/utils/isNodeProcess.react-native.test.ts
jameslahm: yes, I should stub global.navigator in the beforeAll and restore it in afterAll hooks, which will not break tests

src/utils/isNodeProcess.react-native.test.ts
jameslahm: here, my thought is that the value of `global.process` doesn't have influence, so I set it to the default value in `React Native`, which is `Object {
  "env": Object {
    "NODE_ENV": "development",
  },
}`.

src/utils/isNodeProcess.react-native.test.ts
kettanaito: I think not having `global.process` explicitly at all brings us closer to testing the React Native logic in isolation.

Let's say we've got an issue in our `isNodeProcess()` that produces false positives. Now, since we also stub `global.process`, it may fall into the first condition that checks the `global.process` and produce a false positive. 

I think we shouldn't touch `global` in this test. What do you think?

src/utils/isNodeProcess.react-native.test.ts
kettanaito: I think not having `global.process` explicitly at all brings us closer to testing the React Native logic in isolation.

Let's say we've got an issue in our `isNodeProcess()` that produces false positives. Now, since we also stub `global.process`, it may fall into the first condition that checks the `global.process` and produce a false positive. 

I think we shouldn't touch `global` in this test. What do you think?

jameslahm: But if we don't touch `global.process`. `Object.prototype.toString.call(global.process) === '[object process]'` will be true, which should be false in `React Native`, and `navigator.product` will also not be evaluated. `isNodeProcess` will always return true. So I think maybe we should mock `global.process` and let `Object.prototype.toString.call(global.process)` should not be '[object process]'?

kettanaito: Oh, you're totally right. Then this `global.process` assignment makes sense (still would move it to the `beforeAll` hook). Could you please also add a brief explanatory comment why we do that? (because the test itself runs in Node, we don't want it to return false positive). Thanks!

jameslahm: I have moved mock in `beforeAll` and `afterAll` hook, and add some comments. Looks good to you?

kettanaito:
Hey. I've rebased the branch against the latest `master` and cleaned up the React Native test suite a little. Will wait for the CI to pass and then merge.

kettanaito (approved):
Great changes. Thank you, @jameslahm, and welcome to contributors!

jameslahm:
The changes look good! Thanks!

jameslahm:
I just found a problem, that is because we mock `global.process` like this `Object.defineProperty(global.process, 'env', {
    get: () => ({
      NODE_ENV: 'development',
    }),
  })`, which leads `Object.prototype.toString.call(global.process)` still equals `[object process]`. It should be `[object Object]`? Would you mind to look at it again?

PR #257, opened 2020-07-01
Add an option to warn/error on unhandled requests

### LINKED ISSUE #191: Add an option to warn/error on unmocked requests
In my experience, when testing components, I've always mocked API responses. If I've forgotten to mock a response, it'll go something like: run test, assertion fails, wonder why, look at the html, realise I forgot to mock, add a mock. If MSW output a warning/error when a network request was made to an unmocked endpoint, it would be *immediately* obvious what's wrong.

There should be an option added to MSW to globally (and perhaps per-test) configure an option to warn/error on unmocked requests. Probably something that would go in an mswSetup.js file that goes in the jest config setup? I personally think this should be on by default, but I can imagine it possibly being a barrier to usage if it's on by default. If it isn't, it should be prominently displayed in the README and promoted, assuming you'd agree.

n/a

**Additional context**

conifer64:
## GitHub

- Resolves #191

### Specification
This specification will be kept up to date as we iterate over the functionality.

An _unhandled request_ is a request that has no corresponding request handler. It's possible to configure MSW to handle such requests differently. For example, MSW could notify the user with a warning or error message when a request has no corresponding request handler.

#### onUnhandledRequest
Both worker and server instances accept the `onUnhanledRequest` configuration option. This allows the user to choose a provided functionality or provide their own through a custom function:
```ts
type OnUnhandledRequest = 'bypass' | 'warn' | 'error' | (req: MockedRequest) => void
```
**Important:** all functionality is **opt-in**. This means that the default value for `onUnhanledRequest` is `'bypass'`.

#### Usage example
```ts
const worker = setupWorker(...)
const server = setupServer(...)

worker.start({ onUnhandledRequest: 'bypass' })
server.listen({ onUnhandledRequest: 'error' })
```

### Draft
This pull request is a _work in progress_. That means large portions of functionality have _not_ been implemented yet, and the functionality and specification details are still open for debate.

### DIFF
--- a/src/node/glossary.ts
+++ b/src/node/glossary.ts
@@ -1,3 +1,0 @@
-import { SharedOptions } from '../sharedOptions'
-
-export type ListenOptions = SharedOptions
--- a/src/node/glossary.ts
+++ b/src/node/glossary.ts
@@ -1,0 +1,5 @@
+import { OnUnhandledRequest } from '../onUnhandledRequest'
+
+export interface ListenOptions {
+  onUnhandledRequest?: OnUnhandledRequest
+}
--- a/src/node/glossary.ts
+++ b/src/node/glossary.ts
@@ -1,5 +1,3 @@
-import { OnUnhandledRequest } from '../onUnhandledRequest'
+import { SharedOptions } from '../sharedOptions'
 
-export interface ListenOptions {
-  onUnhandledRequest?: OnUnhandledRequest
-}
+export type ListenOptions = SharedOptions
--- a/src/node/setupServer.ts
+++ b/src/node/setupServer.ts
@@ -11,8 +11,9 @@ import { parseRequestBody } from '../utils/request/parseRequestBody'
 import { isNodeProcess } from '../utils/isNodeProcess'
 import * as requestHandlerUtils from '../utils/requestHandlerUtils'
 import { ListenOptions } from './glossary'
+import { onUnhandledRequest } from '../onUnhandledRequest'
 
-const DEFAULT_LISTEN_OPTIONS: DeepRequired<ListenOptions> = {
+const DEFAULT_LISTEN_OPTIONS: ListenOptions = {
   onUnhandledRequest: 'bypass',
 }
 
@@ -81,29 +82,7 @@ export const setupServer = (...requestHandlers: RequestHandlersList) => {
         const { response } = await getResponse(mockedRequest, currentHandlers)
 
         if (!response) {
-          if (resolvedOptions.onUnhandledRequest === 'warn') {
-            // Produce a developer-friendly warning
-            return console.warn(
-              `A request to ${mockedRequest.url} was detected but not mocked because no request handler matching the URL exists.`,
-            )
-          }
-
-          if (resolvedOptions.onUnhandledRequest === 'error') {
-            // Throw an exception
-
-            // throw new Error(`A request to ${req.url} was detected but not mocked because no request handler matching the URL exists.`)
-            return
-          }
-
-          if (typeof resolvedOptions.onUnhandledRequest === 'function') {
-            resolvedOptions.onUnhandledRequest(mockedRequest)
-            return
-          }
-
-          // resolvedOptions.onUnhandledRequest === 'bypass'
-
-          // Return nothing, if no mocked response associated with this request.
-          // That makes `node-request-interceptor` to perform the request as-is.
+          onUnhandledRequest(mockedRequest, resolvedOptions.onUnhandledRequest)
           return
         }
 
--- a/src/node/setupServer.ts
+++ b/src/node/setupServer.ts
@@ -10,8 +10,10 @@ import { getResponse } from '../utils/getResponse'
 import { parseRequestBody } from '../utils/request/parseRequestBody'
 import { isNodeProcess } from '../utils/isNodeProcess'
 import * as requestHandlerUtils from '../utils/requestHandlerUtils'
-import { ListenOptions } from './glossary'
+import { SharedOptions } from '../sharedOptions'
 import { onUnhandledRequest } from '../onUnhandledRequest'
+
+type ListenOptions = SharedOptions
 
 const DEFAULT_LISTEN_OPTIONS: ListenOptions = {
   onUnhandledRequest: 'bypass',
--- a/src/node/setupServer.ts
+++ b/src/node/setupServer.ts
@@ -10,6 +10,11 @@ import { getResponse } from '../utils/getResponse'
 import { parseRequestBody } from '../utils/request/parseRequestBody'
 import { isNodeProcess } from '../utils/isNodeProcess'
 import * as requestHandlerUtils from '../utils/requestHandlerUtils'
+import { ListenOptions } from './glossary'
+
+const DEFAULT_LISTEN_OPTIONS: DeepRequired<ListenOptions> = {
+  onUnhandledRequest: 'bypass',
+}
 
 /**
  * Sets up a server-side requests interception with the given mock definition.
@@ -32,7 +37,9 @@ export const setupServer = (...requestHandlers: RequestHandlersList) => {
     /**
      * Enables requests interception based on the previously provided mock definition.
      */
-    listen() {
+    listen(options?: ListenOptions) {
+      const resolvedOptions = Object.assign({}, DEFAULT_LISTEN_OPTIONS, options)
+
       interceptor.use(async (req) => {
         const requestHeaders = new Headers(
           flattenHeadersObject(req.headers || {}),
@@ -74,6 +81,22 @@ export const setupServer = (...requestHandlers: RequestHandlersList) => {
         const { response } = await getResponse(mockedRequest, currentHandlers)
 
         if (!response) {
+          if (resolvedOptions.onUnhandledRequest === 'warn') {
+            // Produce a developer-friendly warning
+            return
+          }
+
+          if (resolvedOptions.onUnhandledRequest === 'error') {
+            // Throw an exception
+            return
+          }
+
+          if (typeof resolvedOptions.onUnhandledRequest === 'function') {
+            return
+          }
+
+          // resolvedOptions.onUnhandledRequest === 'bypass'
+
           // Return nothing, if no mocked response associated with this request.
           // That makes `node-request-interceptor` to perform the request as-is.
           return
--- a/src/node/setupServer.ts
+++ b/src/node/setupServer.ts
@@ -96,7 +96,8 @@ export const setupServer = (...requestHandlers: RequestHandlersList) => {
           }
 
           if (typeof resolvedOptions.onUnhandledRequest === 'function') {
-            return resolvedOptions.onUnhandledRequest(mockedRequest)
+            resolvedOptions.onUnhandledRequest(mockedRequest)
+            return
           }
 
           // resolvedOptions.onUnhandledRequest === 'bypass'
--- a/src/node/setupServer.ts
+++ b/src/node/setupServer.ts
@@ -83,16 +83,20 @@ export const setupServer = (...requestHandlers: RequestHandlersList) => {
         if (!response) {
           if (resolvedOptions.onUnhandledRequest === 'warn') {
             // Produce a developer-friendly warning
-            return
+            return console.warn(
+              `A request to ${mockedRequest.url} was detected but not mocked because no request handler matching the URL exists.`,
+            )
           }
 
           if (resolvedOptions.onUnhandledRequest === 'error') {
             // Throw an exception
+
+            // throw new Error(`A request to ${req.url} was detected but not mocked because no request handler matching the URL exists.`)
             return
           }
 
           if (typeof resolvedOptions.onUnhandledRequest === 'function') {
-            return
+            return resolvedOptions.onUnhandledRequest(mockedRequest)
           }
 
           // resolvedOptions.onUnhandledRequest === 'bypass'
--- a/src/utils/handleRequestWith.ts
+++ b/src/utils/handleRequestWith.ts
@@ -75,6 +75,22 @@ export const handleRequestWith = (
       // Handle a scenario when there is no request handler
       // found for a given request.
       if (!handler) {
+        if (options.onUnhandledRequest === 'warn') {
+          // Produce a developer-friendly warning
+          return
+        }
+
+        if (options.onUnhandledRequest === 'error') {
+          // Throw an exception
+          return
+        }
+
+        if (typeof options.onUnhandledRequest === 'function') {
+          return
+        }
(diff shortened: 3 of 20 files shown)


jensmeindertsma:
Will throwing an error on these lines display an error message as intended or should I do `console.error`?
- https://github.com/jensmeindertsma/msw/blob/366a09992bfa7cb935ccb9ced6bf5c1761e31593/src/utils/handleRequestWith.ts#L84
- https://github.com/jensmeindertsma/msw/blob/366a09992bfa7cb935ccb9ced6bf5c1761e31593/src/node/setupServer.ts#L80

src/node/setupServer.ts
```diff
@@ -60,6 +67,27 @@ export const setupServer = (...requestHandlers: RequestHandlersList) => {
...
+            // throw new Error(`A request to ${req.url} was detected but not mocked because no request handler matching the URL exists.`)
+            return
+          }
+
+          if (typeof resolvedOptions.onUnhandledRequest === 'function') {
+            resolvedOptions.onUnhandledRequest(mockedRequest)
+            return
+          }
+
+          // resolvedOptions.onUnhandledRequest === 'bypass'
```
cloud-walker: I would personally avoid duplication of both the error / warn message and the `onUnhandledRequest` logic, by creating a util, like:

```typescript
export const onMissingRequestHandler = (
  request: MockedRequest,
  onUnhandledRequest: OnUnhandledRequest,
) => {
  if (typeof onUnhandledRequest === 'function') {
    onUnhandledRequest(request)
    return
  }

  if (['warn', 'error'].includes(onUnhandledRequest)) {
    const message = `[MSW] A request to ${request.url} was detected but not mocked because no request handler matching the URL exists.`

    if (onUnhandledRequest === 'warn') {
      console.warn(message)
    } else {
      throw new Error(message)
    }
    return
  }

  // if onUnhandledRequest is 'bypass' we have nothing to do.
}
```

I've added `[MSW]` tag at the beginning of the message and refactored a bit the logic.

src/utils/handleRequestWith.ts
```diff
@@ -70,6 +70,28 @@ export const handleRequestWith = (
...
+          // throw new Error(`A request to ${req.url} was detected but not mocked because no request handler matching the URL exists.`)
+          return channel.send({ type: 'MOCK_NOT_FOUND' })
+        }
+
+        if (typeof options.onUnhandledRequest === 'function') {
+          options.onUnhandledRequest(req)
+          return channel.send({ type: 'MOCK_NOT_FOUND' })
+        }
+
+        // options.onUnhandledRequest === 'bypass'
```
cloud-walker: We can avoid `channel.send` duplication and `onUnhandledRequest` logic by creating the util described in a previous comment.

cloud-walker: Not sure of how much is the effort, but we should test this logic on both `worker.listen` and `server.start`

cloud-walker (changes requested):

src/utils/handleRequestWith.ts
jensmeindertsma: Yes. Would you be willing to write the tests?

src/node/setupServer.ts
jensmeindertsma: Awesome!

src/utils/handleRequestWith.ts
kettanaito: Absolutely agree, there's no need to duplicate the `channel.send()` and any other possible logic in case of an unhandled request. You can think of the new feature as:

```ts
if (!response) {
  onUnhandledRequest(req, resolvedOptions) // analyzes "resolvedOptions.onUnhandledRequest"
  return channel.send({ type: 'MOCK_NOT_FOUND' })
}
```

src/node/glossary.ts
```diff
@@ -0,0 +1,5 @@
+import { OnUnhandledRequest } from '../onUnhandledRequest'
+
+export interface ListenOptions {
+  onUnhandledRequest?: OnUnhandledRequest
```
kettanaito: Considering we will be shipping the same behavior in terms of this option to both Browser and Node, do you find it useful to introduce a concept of "shared" configuration?

```ts
interface SharedOptions {
  onUnhandledRequest?: OnUnhandledRequest
}

type WorkerOptions = SharedOptions & {
  serviceWorker: WhatWeHaveNow
  // ...
}

type ServerOptions = SharedOptions
```

All those interfaces, apart from the `SharedOptions`, already exist in the respective `setupWorker`/`setupServer` declarations.

src/node/setupServer.ts
```diff
@@ -10,6 +10,11 @@ import { getResponse } from '../utils/getResponse'
 import { parseRequestBody } from '../utils/parseRequestBody'
 import { isNodeProcess } from '../utils/isNodeProcess'
 import * as requestHandlerUtils from '../utils/requestHandlerUtils'
+import { ListenOptions } from './glossary'
+
+const DEFAULT_LISTEN_OPTIONS: DeepRequired<ListenOptions> = {
```
kettanaito: No need to use `DeepRequired`, it's one level object (for now).

src/utils/handleRequestWith.ts
cloud-walker: I can try!

kettanaito: To clarify: I mean exactly what @cloud-walker suggested above: a custom function that encapsulates the logic (checks if the option value is a function, or one of the enums).

cloud-walker: > Yes. Would you be willing to write the tests?

How can contribute in this scenario? You give fork permissions to me? Or I should do another fork? 🤔

jensmeindertsma: I will give you permission to push to my fork

jensmeindertsma: I've merged the changes made by @kettanaito in the upstream repository so we're all up to date and I've given you access. Can you verify that it works? Haven't done this before.

src/node/glossary.ts
jensmeindertsma: Perhaps we could call it `IsomorphicOptions`? Doesn't really matter to me.

jensmeindertsma:
I added most of the changes that you requested, but there is still one problem: due to the use of `Object.assign`, `onUnhandledRequest` still has a possible type of `undefined`, after combining the passed options with the default options. This causes a type error when passing `options.onUnhandledRequest` to the function, because it doesn't accept `undefined`. To fix this, let's check whether  `options.onUnhandledRequest` is defined inside the function, and assign a default value of `'bypass'` here if it isn't. Are you OK with this idea @kettanaito ? It will move the default value for this options away from the `DEFAULT_OPTIONS` object.

cloud-walker:
@kettanaito I've no idea on how to run the integration tests, running `yarn test:integration` all test cases are failing for.. reasons.

You have any advice on how to start? Thanks!

jensmeindertsma:
They could be failing because of TypeScript errors?

jensmeindertsma:
This change fixes the TypeScript issues and probably also solves the issue with the integration tests for @cloud-walker , but it does make the code a little less DRY. Please give your opinion on this change and whether you have a better suggestion.

src/node/glossary.ts
kettanaito: "Shared" seems to describe that the options are shared between multiple units. I'm not sure what "isomorphic" means in relation to those settings.

kettanaito:
You've done an amazing work, fellows! I've rebased the branch and will add some tests to cover this feature. Looks solid!

jensmeindertsma:
Awesome!

kettanaito:
I'm having troubles testing the `"error"` scenario, because it's designed to throw an exception. The trick here is that the exception is not thrown as a part of `fetch` call, but within the process that runs Jest. I've tried various ways to catch that exception, but all had no effect:

- `expect(myFunc).toThrow()`
- `expect(myFunc()).rejects.toThrow()`
- `expect(myFunc()).rejects.toEqual()`
- `process.on('unhandledException')`
- `process.on('unhandledRejection')`

The culprit is that the `"error"` value is designed to fail the test, and it does, including the test on the MSW side. That _is_ the expected behavior, so we need to somehow convert that failed test into a "passing" test.

> Also having a little in-browser issues while testing, but should figure them out. Manual testing of the feature works great.

cloud-walker:
@kettanaito you can share the code of the tests?

kettanaito:
Sure. I've pushed the `setupServer` tests. Note that it would need https://github.com/mswjs/node-request-interceptor/pull/29 to be released in order to pass. 

> There's a trick towards throwing in the request promise: if not handled, that exception will terminate the current running process (i.e. a test). Although it seems expected, I couldn't find a reliable way to assert that in tests. Instead, I've introduced a feature to `node-request-interceptor` that, if an exception is thrown in a request promise, it imitates the request error (similar to network error). This results into error being thrown by _request issuing function_ (i.e. `fetch`).

kettanaito:
I've pushed browser integration tests. The following tests are failing, I cannot seem to figure why:

- `error.test.ts`
- `callback-throws.test.ts`

Manual testing produces expected behavior. There must be something on the Puppeteer's/test setup side that fails to assert that behavior properly. If somebody could take a look at those tests, I'd be much thankful!

cloud-walker:
I'm not an expert of puppeteer, but I can give an eye 👀 

@kettanaito  The command to run is `npm run test:integration`?

kettanaito:
@cloud-walker that'd be awesome.

Yeah, to run a specific integration test use:

```bash
$ yarn test:integration test/msw-api/setup-worker/start/on-unhandled-request/warn.test.ts
```

You can also run a specific test scenario in a browser to play around with by this command:

```bash
$ yarn test:focused test/msw-api/setup-worker/start/on-unhandled-request/warn.mocks.ts
```

For some reason both error and callback-throws scenarios fail (they don't produce an error in the console during runtime). Manual testing works as expected. Something must be off.

cloud-walker:
I get a console.error about the "msw" dep missing:

image

I should do something else? Like `yarn link`?

jensmeindertsma:
Have you run `yarn build` yet?

cloud-walker:
@kettanaito I get a `SIGABRT` error (by puppeteer I guess) or a `SIGSDEV` or `SIGBUS` at random on all the tests (bypass included). Seems to be in heisenbug..

My guess is that some cleanup of puppeteer is missing, but I'm not an expert of it T_T

kettanaito:
Yes, you need to `yarn build` the package, as the built version is used in the integration tests.
Those exceptions are rather strange. Please, @cloud-walker, which OS are you using?

cloud-walker:
I use MacOS mojave

kettanaito:
There's one nuance I've noticed: since **all** requests come through Service Worker and MSW, if you specify `onUnhandledRequest: 'error'` this, most likely, will break your entire page. Requests to third-party integrations, local assets, and even internal endpoints (like requests that webpack polls for HMR) will error, since they don't have (and shouldn't have?) a corresponding request handler.

I can't say I disagree with such behavior, since the `"error"` value is quite self-descriptive, but I'd be cautious about this. What do you think, folks?

jensmeindertsma:
Perhaps it would be better to include a required RegEx option with `onUnhandledRequest` where only routes that match the RegEx get "blocked" when using error. Also, the `error` option could mock a network error as well as throwing a error in the console. What do you think of this?

cloud-walker:
> There's one nuance I've noticed: since **all** requests come through Service Worker and MSW, if you specify `onUnhandledRequest: 'error'` this, most likely, will break your entire page. Requests to third-party integrations, local assets, and even internal endpoints (like requests that webpack polls for HMR) will error, since they don't have (and shouldn't have?) a corresponding request handler.
> 
> I can't say I disagree with such behavior, since the `"error"` value is quite self-descriptive, but I'd be cautious about this. What do you think, folks?

I personally think it's the expected behavior, at least for `setupServer`. During test development you want the code to fail as early as possible.

For `setupWorker` I agree, seems a bit odd. But I guess that the majority of users will choose `warn` here.

kettanaito:
@jensmeindertsma, I'm not against supporting a RegExp as the value of `onUnhandledRequest`, but I would start with shipping the current state of the feature first. That is to see if there are more people in need of regular expressions. My concern is to keep this feature small and predictable.

Regarding making the `error` value behave as a network error, I think this would be confusing. The next release will also include the actual network error emulation via `res.networkError()` (#253), which you can utilize on per-handled basis. You can achieve a similar behavior to unhandled requests and `res.networkError()` using a fall-through handler:

```js
// Handled request
rest.get('/user', (req, res) => res()),

// Fall-through handler matching any other request
res.get('*', (req, res) => res.networkError('Any other request raises a network error!')
```

The `error` option for unhandled requests is meant to raise an error in your test run, not in the mocked response.

jensmeindertsma:
Is there still work to be done to get the current version of this feature shipped?

kettanaito:
There are two test suits that fail:

- error.test.ts
- callback-throws.test.ts

We need to either construct them better, or make sure they pass in the current state before shipping this.

cloud-walker:
@kettanaito you think it's possible to reproduce the bugged scenarios on a dummy repo? Just to remove the real project noise?

kettanaito:
With the recent changes that an exception in a request handler can be intentional and handled as a network error (#253), throwing an error on the unhandled request started to pass as well. I'm verifying the remaining tests.

kettanaito:
I'm not sure, @cloud-walker. That noise may contain a valuable context. Once we bring a potentially solved test scenario in, it may start failing anyway. I'd say it's a good practice to approach tests exactly where they're being run.

This certainly takes more time, but I'm confident it's worth it. Sometimes Puppeteer is tricky, but most of the times it's just me now knowing how to use it. For example, I wonder why it doesn't get the console errors that are clearly there if you run Puppeteer in a headfull mode. The overall `page.on('console')` listener seems to be flaky at times, yet the error production must be consistent.

I'm a strong believer that no test is a thousand times better than a flaky one. I'll give this one more go, and if I cannot find a reliable way to test certain error scenarios they simply won't be added.

kettanaito:
After significant amount of effort I conclude `error` and `callback-throws` test scenarios as unreliable. The list of intercepted console errors sometimes is an empty array `[]` received from Puppeteer, and I can't find a reason why. 

> That being said, those scenarios do produce an expected behavior during manual testing.

I'm removing those tests and checking if the CI passes with the remaining browser and server tests. I've still added the `*.mocks.ts` files that represent the tested behavior, for manual testing mainly.

kettanaito (approved):
Thank you everybody for participating in developing this feature!
It looks great, let's publish it and give it a try. Approved 🎉

jensmeindertsma:
🎉

PR #262, opened 2020-07-04
test(isNodeProcess): mock global.process in react native

jameslahm:
This is following #255, I notice if we just mock global.process like this `Object.defineProperty(global.process, 'env', { get: () => ({ NODE_ENV: 'development', }), })`, which leads `Object.prototype.toString.call(global.process)` still equals `[object process]`. It should be `[object Object]`? Would you mind to look at it again?

### DIFF
--- a/src/utils/isNodeProcess.react-native.test.ts
+++ b/src/utils/isNodeProcess.react-native.test.ts
@@ -33,11 +33,12 @@ beforeAll(() => {
    * @see https://github.com/facebook/react-native/blob/6d6c68c2c639b6473e049f7d916690b92e921c7e/Libraries/Core/setUpNavigator.js
    * @see https://github.com/facebook/react-native/blob/6d6c68c2c639b6473e049f7d916690b92e921c7e/Libraries/Core/setUpGlobals.js
    */
-  Object.defineProperty(global.process, 'env', {
-    get: () => ({
+  // @ts-ignore
+  global.process = {
+    env: {
       NODE_ENV: 'development',
-    }),
-  })
+    },
+  }
 
   global.navigator = {
     product: 'ReactNative',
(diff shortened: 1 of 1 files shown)


src/utils/isNodeProcess.react-native.test.ts
```diff
@@ -33,11 +33,12 @@ beforeAll(() => {
    * @see https://github.com/facebook/react-native/blob/6d6c68c2c639b6473e049f7d916690b92e921c7e/Libraries/Core/setUpNavigator.js
    * @see https://github.com/facebook/react-native/blob/6d6c68c2c639b6473e049f7d916690b92e921c7e/Libraries/Core/setUpGlobals.js
    */
-  Object.defineProperty(global.process, 'env', {
-    get: () => ({
+  // @ts-ignore
```
kettanaito: Please, instead of `ts-ignore`, could you describe this in the `declare global` type override at the top of the file?
I understand that most likely you're using `ts-ignore` here to bypass `"process" is readonly", but I believe you can drop the `readonly` modifier form an interface key (we re-declare `interface Global` on top), allowing to write into `global.process`.

kettanaito (changes requested):
Hey, @jameslahm! Thank you once more for following on this. I completely agree that the assignment of `global.process` should change to preserve a proper `[object Object]` prototype.

Please, I've left one comment regarding TypeScript. I'd appreciate if you took a look.

src/utils/isNodeProcess.react-native.test.ts
jameslahm: Sorry, I don't quite understand. I think `re-declare interface Global` will merge `original interface Global declaration` and `global.process` will still be `Process` type? And that will refuse the assignment of `global.process`, which requires many other properties, such as `stdin, stdout` and so on. Please let me know if my thoughts are wrong. Thanks for the comment!

kettanaito: Oh, I see. Then I'd recommend to cast the override object `as any`. It's a less harmful override than ignoring the entire line:

```ts
global.proces = { ... } as any
```

What do you think?

jameslahm: Thanks for the comment! I have changed the code, `as any` is better.

kettanaito (dismissed):
Thanks for looking once more into this! Superb work.

kettanaito (approved):

PR #265, opened 2020-07-06
Ensures request handlers allow intended fall-through

### LINKED ISSUE #234: allow passthrough for undefined routes / graphql interferes with passthrough

I'd like to be able to allow certain matching urls to be passed through and handled normally instead of intercepting and giving 404.

Seems like a lot of work to manually define `ctx.fetch` for each request I would want to pass through on some wildcard domain / string match.

### LINKED ISSUE #258: Transparent handler ("*") doesn't allow fall-through
## Reproduction step 

```js
setupWorker(
  rest.get('*', () => console.log('caught')),
  rest.get('/user', (req, res, ctx) => res(ctx.json({ firstName: 'John' }))
)
```

## Expected behavior

1. Prints out `"caught"` in the console.
1. Returns `{ firstName: "John" }` as the mocked response.

## Current behavior

`GET /user` returns `404` (not handled).

marcosvega91:
- Fix #258 
- Fix #234

I have changed the logic inside `getResponse`. Now all handlers are looped until the first response is found.

In this case all handlers that match the request are executed until one of them return a response.


test/msw-api/setup-worker/fall-trought.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  )
+})
+
+afterAll(() => {
+  return runtime.cleanup()
+})
+
+it('should fall-trought until a response is found', async () => {
+  const messages = []
+  runtime.page.on('console', function (message) {
```
kettanaito: Could you please adjust this console spying with the recently updated `captureConsole` internal API?

```ts

const firstMessage = messages.log.filter(text => text.startsWith('[test] first caught'))
```

src/utils/getResponse.ts
```diff
@@ -23,14 +23,18 @@ export const getResponse = async <
...
   handlers: H,
 ): Promise<ResponsePayload> => {
-  const [relevantHandler, parsedRequest]: [
+  const [relevantHandler, parsedRequest, mockedResponse, publicRequest]: [
     RequestHandler<any, any>,
     R,
-  ] = handlers.reduce<any>((found, requestHandler) => {
+    MockedResponse | null,
+    any,
+  ] = (await handlers.reduce<any>(async (handlerPromise, requestHandler) => {
```
kettanaito: So the issue was that this reduce was looking for the relevant request handler, however, the first matched request handler may not have returned the response? Since the reduce halts once the first match was found, it never took into account that the found handled may return nothing, so there is no fall-through.

I don't think that putting the entire response resolution into this reduce is a good idea. We'd be recreating context and public request multiple times for each handler that doesn't return a mocked response. However, that data is the same and doesn't depend on the handler.

I think instead we should do something like:

```js
// Get the list of all handlers that match the current request.
// Instead of stopping after the first predicate match, continue
// to gather ALL the handlers.
const relevantHandlers = handlers.filter(...)

// Here, reduce the list of handlers into a mocked response.
// Create context and other relevant instances here.
const mockedResponse = relevantHandlers.reduce(...)
```

That would add readability and make sure that each method does exactly one thing. What do you think about it?

kettanaito (changes requested):

src/utils/getResponse.ts
marcosvega91: The approach used in my code is not good of course. 
If two requests return something what we'll do? 
What I have done there is to loop through requests and stop when one returning something was found.

Of course I was not considering the fact that I can have in `requestHandlers` first a request that returns something and then a request that returns nothing

kettanaito: If there is a request handler that returns a mocked response (or `null`), then it must stop the iteration, regardless if there are any other request handlers after it. If a request handler does not have a return statement, MSW should still execute its resolver.

So, it can be summarized as: MSW executes all request handlers which predicate matches the captured request _until_ any of them returns something.

kettanaito:
Hey, @marcosvega91 👋 
I think these changes look good. I've added a few commits:

- Added the respective `.node.test.ts` for the `setupServer` API.
- Flattened the test suites, so each `test()` call creates and clears its own teardown.
- Refactored the `getResponse` function, so that we parse request once, and also transformed the `Promise<[null, null, null, null]>` into `Promise<ReducedHandlersPayload>` (object) for easier reference.

I've also noticed removed one branch when there is `!mockedResponse` in the `getResponse` function. Setting `mockedResponse: null` on the reducer function eliminates the need to explicitly check if it's present. Instead, I'm checking if the `requestHandler` is present, which we were not doing before.

kettanaito (dismissed):

kettanaito (approved):
Thank you for making these changes happen, @marcosvega91! Approved.

PR #266, opened 2020-07-07
Node response patching

### LINKED ISSUE #264: Response patching in Node fails
When trying to use response patching as described in the documentation in a node process (`setupServer`), `ctx.fetch` fails with `Window is undefined` because it uses `window.fetch` directly. Am I missing something? Is there any way around that?

pmelab:
## Changes

- Uses `node-fetch` to enable response patching in Node.
- Adds integration tests for response patching in Node.

## GitHub

- Fixes #264

## Roadmap

- [x] Ensure call signature of `window.fetch` and `node-fetch` is compatible. Line out the corner-cases and limitations.
- [x] Ensure that `node-fetch` doesn't get bundled and executed on the client


package.json
```diff
@@ -64,6 +64,7 @@
     "@types/cookie": "^0.3.3",
     "chalk": "^4.0.0",
     "cookie": "^0.4.1",
+    "cross-fetch": "^3.0.5",
```
kettanaito: Can we try to reuse `node-fetch`, which we already use?

Conditionally deciding on the `fetch` would be enough (their call signatures should be identical):

```ts
// src/context/fetch.ts
import { isNodeProcess } from '../utils/isNodeProcess'

const useFetch = isNodeProcess() ? require('node-fetch') : fetch
```

The goal here is to conditionally require the installed dependency, only when used in a NodeJS environment. It may require to properly reference `fetch`, but we can handle that.

test/msw-api/setup-server/response-patching.test.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @jest-environment node
+ */
+import fetch from 'node-fetch'
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
+
+const server = setupServer(
+  rest.get('https://test.mswjs.io/user', async (req, res, ctx) => {
+    const originalResponse = await ctx.fetch(
+      'https://api.github.com/users/octocat',
```
kettanaito: GitHub API has rate limiting that may result into flaky CI. I suggest to use `https://httpbin.org/get`, which doesn't seem to have one (afaik).

kettanaito (changes requested):
Thank you for these changes, @pmelab! I've had a few suggestions, do you mind take a look?

pmelab:
@kettanaito Thank you for the review! I did the requested changes, and CI passed.

What's open are two `TODO`s I left in the code: The type definitions for `RequestInit` and `Response` in the browser and `node-fetch` are (in theory) not fully compatible and I did not find a way to make them so. In practice, I'm pretty sure they are. As far as I can tell, `cross-fetch` and `isomorphic-fetch` simply pretend compatibility in their type definitions.

src/context/fetch.ts
```diff
@@ -52,10 +52,8 @@ export const fetch = <ResponseType = any>(
   })
 
   return gracefully<ResponseType>(
-    isNodeProcess()
-      ? // TODO: Figure out cross typing of request and response.
-        /// @ts-ignore
-        nodeFetch(input.url.href, compliantReq)
-      : window.fetch(input.url.href, compliantReq),
+    // TODO: Figure out cross typing of request and response.
```
kettanaito, line 55: Please, what do you mean by `cross typing`? Type definitions?

kettanaito:
Thank you for the adjustments! Looks amazing. I share your caution regarding type definition (call signature compatibility). We need to research what are the exact limitations between in-browser `fetch` and `node-fetch`, apart from environment limitation (like absolute URLs in Node). My initial assumption is that everything `fetch` does should be able to reproduce with `http`, which I assume `node-fetch` uses under the hood.

kettanaito:
Hey. I've updated the feature branch against the latest `master`. Also added two more integration tests that asserts bypass/fall-through behavior (the difference between using `ctx.fetch` and native `fetch` directly). I've checked and confirm that `node-fetch` is not required when working in a browser, and I find the call signatures of `window.fetch` and `node-fetch` compatible.

kettanaito (approved):

kettanaito:
Thank you for improving the library! Welcome to contributors 🎉

PR #269, opened 2020-07-08
Add `native` runtime, alongside node

Federkun:
This is an attempt to support react native's runtime. https://github.com/mswjs/msw/issues/203

```
import 'react-native-url-polyfill/auto' // required polyfill
import { setupServer } from 'msw/native'

setupServer(...).listen()
```

After some thinking, I came to the conclusion that the `node` environment is just not the right place to target RN.
Currently, the main issue is that there's no practical way to conditionally import `timers`. This is not an option, as the metro packager of react native will still try to include it in the main bundle:

```
const aSetTimeout  = navigator.product !== 'ReactNative' ? require('timers').setTimeout : global.setTimeout
```
Another required change to this PR would be to import a `RequestInterceptor` from `node-request-interceptor` that doesn't attempt to import the `http` core module.

The solution of moving from `msw/node` to `msw/native` is not perfect either thought. When running your jest test suite, you may still want to use `msw/node`.

There may be some babel based alternatives worth exploring, but nothing I can think of at the moment.

Before creating a PR to `node-request-interceptor` with the change required to complete this PR, I may want to get some early feedback on the preferred direction.

### DIFF
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -1,4 +1,5 @@
 import * as path from 'path'
+import alias from '@rollup/plugin-alias'
 import resolve from '@rollup/plugin-node-resolve'
 import commonjs from '@rollup/plugin-commonjs'
 import typescript from 'rollup-plugin-typescript2'
@@ -97,4 +98,32 @@ const buildNode = {
   ],
 }
 
-export default [buildNode, buildEsm, buildUdm]
+const buildNative = {
+  input: 'src/native/index.ts',
+  external: ['events', 'node-request-interceptor'],
+  output: {
+    file: 'native/index.js',
+    format: 'cjs',
+  },
+  plugins: [
+    alias({
+      entries: [{ find: 'timers', replacement: '../utils/reactNativeTimers' }],
+    }),
+    json(),
+    resolve({
+      browser: false,
+      preferBuiltins: true,
+      extensions: ['.js', '.jsx', '.ts', '.tsx'],
+    }),
+    typescript({
+      useTsconfigDeclarationDir: true,
+      tsconfigOverride: {
+        outDir: './native',
+        declarationDir: './native',
+      },
+    }),
+    commonjs(),
+  ],
+}
+
+export default [buildNode, buildNative, buildEsm, buildUdm]
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -1,7 +1,7 @@
 import * as path from 'path'
-import alias from '@rollup/plugin-alias'
 import resolve from '@rollup/plugin-node-resolve'
 import commonjs from '@rollup/plugin-commonjs'
+import inject from '@rollup/plugin-inject'
 import typescript from 'rollup-plugin-typescript2'
 import json from '@rollup/plugin-json'
 import packageJson from './package.json'
@@ -94,6 +94,9 @@ const buildNode = {
         declarationDir: './node',
       },
     }),
+    inject({
+      setTimeout: ['timers', 'setTimeout'],
+    }),
     commonjs(),
   ],
 }
@@ -106,9 +109,6 @@ const buildNative = {
     format: 'cjs',
   },
   plugins: [
-    alias({
-      entries: [{ find: 'timers', replacement: '../utils/reactNativeTimers' }],
-    }),
     json(),
     resolve({
       browser: false,
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ __*
 .DS_*
 node_modules
 /lib
+/native/**/*.js
 /node/**/*.js
 tmp
 *-error.log
--- a/native/package.json
+++ b/native/package.json
@@ -1,0 +1,3 @@
+{
+  "types": "../lib/types/native"
+}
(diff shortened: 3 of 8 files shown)


hsavit1:
@kettanaito anything that I can do to help push this along?

hsavit1:
@Federkun maybe it is worth opening a dummy react native project just to prove that this can work?

kettanaito (changes requested):
I think the area of changes is too big to provide a conditional `setTimeout` function. I understand the intention, but I think at this point such abstraction brings more harm than solves a problem.

I believe the problem is that we cannot assign `setTimeout` conditionally? I wonder why the following code doesn't work:

```js
const useTimeout = isReactNativeProcess() ? setTimeout : require('timers').setTimeout
```

Federkun:
See https://github.com/mswjs/msw/issues/203#issuecomment-656046711 ; the gist is that the RN bundler needs to send all the javascript code to the device. It cannot require modules at runtime like that.
~~Another solution would be to create a babel plugin that should be installed along side msw, to make it work on native.~~

Probably just provide some documentation to modify the user's extra modules like this to the `metro.config.js` file, like:
```
const { nativeExtraModules } = require('msw/node')

module.exports = {
  resolver: {
    extraNodeModules: {
      ...nativeExtraModules
      // timers: ...
      // http: ..
    }
  }
};
```

Would that be better?
This would also not require any change in the `node-request-interceptor`, as we can provide our implementation of the core package.

kettanaito:
Thanks for explaining this to me once more, @Federkun!

I think a separate module like `msw/react-native` is a good idea, as there may be more corner cases to run the solution in React Native. What do you think if we focus on module replacement on the Rollup's side? Something like a provider plugin that replaces `setTimeout` with a different module.

Federkun:
I like that! It didn't occur to me that it was an option, but it seems the most sensible one ( both from a user and maintainer point of view ).
About `node-request-interceptor`, should we provide an equivalent api  (like `node-request-interceptor/native`) as well?

kettanaito:
With NRI the issue was, I believe, that it mustn't require http/https modules, correct? Because those are NodeJS pre-requisites not available in RN. If that’s the case, I wonder if we could incorporate a similar conditional logic as in XMLHttpRequest patching? Or would it also mean an unsupported a runtime require?

Federkun:
With `node-request-interceptor` we have the identical problem that we face here. But i just realised that the best solution is to just provide a `RequestIntereptor.native.js` file along side the other. The bundler will try to resolve all the imports by looking at `file.{platform}.js` first, then falling back to `file.js`.

I still think that creating a `msw/native` is necessary, as the public api may diverge from the node one, but now that I think about this problem better, it's not _strictly_ necessary. Should I close this PR, and open another one?

kettanaito:
> I still think that creating a `msw/native` is necessary, as the public api may diverge from the node one

I see the possibility of API deviation as well, but I think we can first build `src/node/index.ts` into `react-native/index.js`, applying the custom modules resolution through Rollup, retaining all the other API aspects. Once we have a reasonable amount of deviations we may consider a custom entry module. What do you think?

Feel free to proceed in this pull request (discard previous commits and force push the new ones), or create a separate pull request—it's up to you.

> The bundler will try to resolve all the imports by looking at `file.{platform}.js` first, then falling back to `file.js`.

Thanks, I didn't know about such module resolution. Does this mean that if I have:

```js
// src/foo.native.js
import foo from './foo'
```

```js
// src/foo.js
import foo from './bar'

foo()
```

The `foo` function in `foo.js` will be imported from `./foo`, not `./bar`?

Federkun:
I'm not entirely sure if I can improve the rollup config, and get rid of `reactNativeTimers.ts`. Suggestions are welcome.
This PR will remain as a draft, as it depends on https://github.com/mswjs/node-request-interceptor/pull/37/files

rollup.config.ts
```diff
@@ -97,4 +98,32 @@ const buildNode = {
...
+const buildNative = {
+  input: 'src/native/index.ts',
+  external: ['events', 'node-request-interceptor'],
+  output: {
+    file: 'native/index.js',
+    format: 'cjs',
+  },
+  plugins: [
+    alias({
+      entries: [{ find: 'timers', replacement: '../utils/reactNativeTimers' }],
```
kettanaito: Let's think of it this way: for React Native the `setTimeout` behavior should remain as usual, no specifics here. It is the Node build target that requires to swap `global.setTimeout` with `timers.setTimeout` in order to be used with fake timers in Jest.

I suggest we provide a module alias for Node build target, and leave React Native configuration without custom aliases. What do you think about this?

Federkun: Seems a good idea, I made the change

kettanaito (approved):
This looks perfect. Thank you so much for preparing this, @Federkun!
Welcome to contributors! 🎉

kettanaito:
Please, @Federkun, would you be interested in preparing a minimal usage example of React Native + MSW?
You can use [existing examples](https://github.com/mswjs/examples/tree/master/examples/rest-react) to get some inspiration. We'd love to have your expertise in this topic to have a superb example to reference and feature. Let me know.

Federkun:
Sure, I'll give it a shot this weekend

Issue #73, opened 2020-03-20
How to deal with requests made before msw has been activated?


kentcdodds:
If I have a situation like this:

```javascript
start('/msw.js')

fetch('/something-that-should-be-mocked')
```

MSW will not be registered as ready until *after* the request has already been forwarded along. What do you recommend for situations like this where the app is trying to make requests immediately?

kentcdodds:
For anyone coming along, here's the basic gist of my current workaround:

```javascript
window.serverReady = start('/msw.js')

async function bootstrapAppData() {
  await window.serverReady
  await fetch('/something-that-should-be-mocked')
  // etc...
}
```

In production I don't think this should be a problem because you'd basically just await undefined which is nbd.

kettanaito:
I'd recommend to start the Service Worker as early as possible in your app. I usually put my `mocks.js` in the root of my app:

```jsx
import React from 'react'
import ReactDOM from 'react-dom'
import App from './App'
import './mocks'

ReactDOM.render(<App />, rootElement)
```

> This implies I use `composeMocks()` and invoke the `start()` function within `mocks.js`.

However, since Service Worker activation is an async operation there is still a race condition in place of what happens first: your app's mount and potential data fetching, or MSW activation. To eliminate this race condition you can listen to the Promise returned from calling the `start()` function:

```js
// mocks.js
import { composeMocks } from 'msw'

return composeMocks(/* mock definition */)
```

```jsx
// index.js
import { start } from './mocks'

async function bootstrapApp() {
  await start()
  ReactDOM.render(<App />, rootElement)
}
bootstrapApp()
```

It is up to the user to define at which level awaiting the `start()` Promise is applicable, depending on their situation. One of the core principles of MSW is the absence of changes to the existing code, that's why I generally don't like the idea of making your app's logic dependent on the MSW activation status.

kentcdodds:
Yep, I think that's the best that can be done here. Thanks.

kentcdodds:
This should be re-opened. The promise returned by `start` can resolve before the service worker is actually ready.

You can observe this in my app which has MSW deployed:

1. https://5e7d73d0938fb965a65bbdf3--the-react-bookshelf.netlify.app
2. Register with username `abc` and password `123`
3. Do a "hard refresh" via <kbd>⌘</kbd> + <kbd>⇧</kbd> + <kbd>r</kbd>
[...]

PR #289, opened 2020-07-22
Fix: Return the correct mock worker when registering or updating

msutkowski:
This is in regard to the comments in https://github.com/mswjs/msw/issues/73 regarding CodeSandbox behavior. `getRegistration(url)`, whether providing an absolute or relative url, was just returning the CodeSandbox service worker being that the `scope` was on the root domain like 'https://kuwlp.csb.app/' during registration. Being that we already check for `mockRegistrations` above, we can just pull the existing one from there if it exists as it's already filtered.

Additional edit:

The behavior of `.register()` is a bit unexpected in the event that there is already a service worker that made the first claim. In this case, codesandbox's worker would alway so do that, so the `ServiceWorkerRegistration` would look like:
`{ active: CSBworker, waiting: mockWorker, ... normal registration fields }` - so what was happening is that when we called `getWorkerByRegistration`, even though the registration instance was correct, the returned Worker instance was not. I've enforced the filter for `scriptURL` === `asbsoluteWorkerUrl` as a default as this is always going to be the desired behavior and cuts out any future edge cases.

### DIFF
--- a/src/setupWorker/start/utils/getWorkerByRegistration.ts
+++ b/src/setupWorker/start/utils/getWorkerByRegistration.ts
@@ -4,6 +4,11 @@
  */
 export const getWorkerByRegistration = (
   registration: ServiceWorkerRegistration,
+  absoluteWorkerUrl: string,
 ): ServiceWorker | null => {
-  return registration.active || registration.installing || registration.waiting
+  return (
+    [registration.active, registration.installing, registration.waiting].find(
+      (worker) => worker?.scriptURL === absoluteWorkerUrl,
+    ) || null
+  )
 }
--- a/src/setupWorker/start/utils/getWorkerByRegistration.ts
+++ b/src/setupWorker/start/utils/getWorkerByRegistration.ts
@@ -1,14 +1,20 @@
 /**
- * Attempts to resolve a Service Worker instance from any of its states:
- * active, installing, or waiting.
+ * Attempts to resolve a Service Worker instance from a given registration,
+ * regardless of its state (active, installing, waiting).
  */
 export const getWorkerByRegistration = (
   registration: ServiceWorkerRegistration,
   absoluteWorkerUrl: string,
 ): ServiceWorker | null => {
-  return (
-    [registration.active, registration.installing, registration.waiting].find(
-      (worker) => worker?.scriptURL === absoluteWorkerUrl,
-    ) || null
-  )
+  const allStates = [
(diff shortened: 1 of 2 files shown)


msutkowski:
Actually, there may still be an issue here - investigating.

msutkowski:
Quick update: the other issue here is that the registration process will actually _sometimes_ return the wrong worker in the instance - we successfully register the mock worker if it doesn't exist, but the `getWorkerByRegistration` will potentially return the first `active` one, which in the CSB scenario will more often than not be the `sandbox-service-worker`.

msutkowski:
Alright, this is ready for review :tada:. Let me know what you think or if you have any concerns. I've tested this code locally as well as inside of a CodeSandbox _(by publishing a poorly named experimental package)_ and it works as expected. You can just call `worker.start()` now and it'll run. https://codesandbox.io/s/rtk-crud-msw-experimental-pkz3w?file=/src/index.tsx

**Note:** when using the CSB editor, the mocking doesn't always load before the network request, but any subsequent refreshes load properly. If you go to the direct URL, there is no issue at all. Ex: https://pkz3w.csb.app/. I'm not sure it'd be worth the effort to further down the path of trying to resolve that as it seems like a CSB issue at this point?

src/setupWorker/start/utils/getWorkerByRegistration.ts
```diff
@@ -1,9 +1,20 @@
...
   registration: ServiceWorkerRegistration,
+  absoluteWorkerUrl: string,
 ): ServiceWorker | null => {
-  return registration.active || registration.installing || registration.waiting
+  const allStates = [
+    registration.active,
+    registration.installing,
+    registration.waiting,
+  ]
+  const existingStates = allStates.filter(Boolean) as ServiceWorker[]
```
kettanaito, line 14: I'm filtering the possible states of the registration so when it comes to finding a relevant one, we're always operating on existing registrations.

kettanaito (approved):
@msutkowski, these changes are _amazing_. I'm thankful for you diving into this, and so excited to see MSW finally working with Codesandbox. This will simplify issue reproduction scenarios tremendously!

I've rebased the feature branch and provided some minor touches. Please use the rebase strategy for the future ;) Waiting for the CI to pass.

kettanaito:
@msutkowski, welcome to contributors! 🎉

PR #298, opened 2020-07-27
🐛Force usage of chokidar 3

### LINKED ISSUE #188: Integration tests fail locally with "function uv_close"

cowboyd:
Motivation
----------

fixes https://github.com/mswjs/msw/issues/188

The current version of webpack that we're using has [a bug][1] that causes the tests to crashes on MacOS with some regularity. That is because it depends on chokidar 2 which uses the `fsevents` api which has a [crictical bug][2] on MacOS.

Approach
--------

This manually forces all modules to use chokidar 3 which does not use the `fsevents` module at all. The v3 API is apparently similar enough that webpack is able to use it as a drop-in replacement for v2. Howeve, it is worth noting that because this is technically a semver violation, it will cause a console warning every time a `yarn install` is run.

This resolution can be removed once an upgrade is made to webpack 5 or to a different depedency bundler like [parcel][3] or [rollup][4].

Learning
--------

* [Selective dependency resolutions ][5]

[1]: https://github.com/webpack/webpack/issues/10037
[2]: https://github.com/fsevents/fsevents/issues/314
[3]: http://parceljs.org
[4]: https://rollupjs.org
[5]: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,5 +113,8 @@
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11",
     "webpack-dev-server": "^3.10.3"
+  },
+  "resolutions": {
+    "chokidar": "3.3.1"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1692,7 +1692,7 @@ anymatch@^2.0.0:
     micromatch "^3.1.4"
     normalize-path "^2.1.1"
 
-anymatch@^3.0.3:
+anymatch@^3.0.3, anymatch@~3.1.1:
   version "3.1.1"
   resolved "https://registry.yarnpkg.com/anymatch/-/anymatch-3.1.1.tgz#c55ecf02185e2469259399310c173ce31233b142"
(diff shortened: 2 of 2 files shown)


kettanaito:
Hey, @cowboyd. Thanks for preparing this.

> This resolution can be removed once an upgrade is made to webpack 5 or to a different depedency bundler like parcel or rollup.

This is a fair point. The library itself is bundled via Rollup. We use webpack to bundle usage examples (which are also test scenarios) and establish a dev server that's used in tests. I suppose something similar can be done with Rollup, either with native plugins, or manually (build + Express server, for example). I wouldn't tackle it here, but it's a noteworthy point.

kettanaito (dismissed):

kettanaito (approved):
Thank you for these changes! Approved.
Welcome to contributors 🎉

PR #300, opened 2020-07-28
Reworks "captureConsole" API for tests

kettanaito:
## Motivation

- Current `captureConsole` accepts the Array argument that it later mutates.
- Also refactors certain tests to be shallow and have each test scenario as a standalone reproducible test.

## Changes

The `captureConsole` function now returns the `messages`, where all captured console messages are sorted in groups ("info", "log", "warning", etc.).

```js
const neededMessage = messages.error.find(text => text.startsWith('Oops'))
```

### DIFF
--- a/src/utils/isNodeProcess.react-native.test.ts
+++ b/src/utils/isNodeProcess.react-native.test.ts
@@ -41,7 +41,7 @@ beforeAll(() => {
 
   global.navigator = {
     product: 'ReactNative',
-  }
+  } as any
 })
 
 afterAll(() => {
--- a/test/graphql-api/logging.test.ts
+++ b/test/graphql-api/logging.test.ts
@@ -18,9 +18,7 @@ describe('GraphQL: Logging', () => {
     let requestLog: string
 
     beforeAll(async () => {
-      const logs: string[] = []
-
-      captureConsole(test.page, logs, filterLibraryLogs)
+      const { messages } = captureConsole(test.page, filterLibraryLogs)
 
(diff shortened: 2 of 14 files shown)


kettanaito:
Hey, @marcosvega91 👋 Sorry for troubling, would you have a minute to review these changes?

marcosvega91:
Hey for me is ok 😄 . There is only a thing that I can't review because is not in the updated code, but I think that you can change in this PR.

There are some point in the code where `test.page.on('console')` is used. I think that you can change this part  
For example
https://github.com/mswjs/msw/blob/f41a1e2641417aa2286b49a2e2d6eaa8c1498978/test/msw-api/integrity-check.test.ts#L24-L28

marcosvega91 (changes requested):
Hey for me is ok 😄 . There is only a thing that I can't review because is not in the updated code, but I think that you can change in this PR.

There are some point in the code where `test.page.on('console')` is used. I think that you can change this part  
For example
https://github.com/mswjs/msw/blob/f41a1e2641417aa2286b49a2e2d6eaa8c1498978/test/msw-api/integrity-check.test.ts#L24-L28

kettanaito:
Thanks for the review!

I've remove the plain usage of `page.on('console')` from the tests in favor of `captureConsole`. Adjusted the `filterLibraryLogs` predicate to be shorthand passable to the `.filter()` Array method.

marcosvega91 (approved):
LGTM

PR #308, opened 2020-07-30
chore: node-fetch should be installed as dependency

marcosvega91:
The released version of `msw` can't be run on `node` because it requires `node-fetch` that is installed as dev dependency, but should be installed as dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,8 @@
     "node-match-path": "^0.4.4",
     "node-request-interceptor": "^0.3.3",
     "statuses": "^2.0.0",
-    "yargs": "^15.4.1"
+    "yargs": "^15.4.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -98,7 +99,6 @@
     "ignore-loader": "^0.1.2",
     "jest": "^26.1.0",
     "lint-staged": "^10.2.11",
-    "node-fetch": "^2.6.0",
```
kettanaito, line 101: Does this change produce any updated in `yarn.lock`?
Could you please run `rm -rf node_modules && yarn install` to verify?

marcosvega91: .lock file doesn't track the origin of the dependency

kettanaito (approved):
Thank you for yet another awesome change!

PR #309, opened 2020-07-31
chore: add native folder to files

marcosvega91:
Native folder was not added to `files` so when a new version is published the folder will not be in the lib

https://github.com/mswjs/msw/issues/203


kettanaito (approved):
Thank you for fixing this! One more valuable contribution 👍

PR #313, opened 2020-07-31
Do not set document.cookie on plain node

### LINKED ISSUE #312: `ctx.cookie` fails in a plain node environment
## Describe the bug
If I set a cookie in an MSW response in a Node.js environment without jsdom, a TypeError is raised. MSW should not attempt to set a cookie via `document.cookie` if it isn't available.

There is still value in supporting cookies in an environment without `document`, as the cookies can be retrieved by reading `Set-Cookie` headers.

## Environment

- `msw: 0.20.1`
- `nodejs: 12.18.1`
- `npm: 6.14.5`

## To Reproduce

Unit test: cookie.node.test.ts, based on cookie.test.ts

```typescript
/**
 * @jest-environment node
 */
import { cookie } from './cookie'
import { response } from '../response'

describe('cookie (Node.js without jsdom)', () => {
  describe('given I set a cookie', () => {
    let result: ReturnType<typeof response>

    beforeAll(() => {
      result = response(cookie('my-cookie', 'arbitrary-value'))
      expect(result).toBeDefined()
    })

    it('should be set on the response headers', () => {
      expect(result.headers.get('set-cookie')).toEqual(
        'my-cookie=arbitrary-value',
      )
    })
  })
})
```

## Expected behavior

The test above should pass.

A pull request will arrive shortly.

Enet4:
If `document` does not exist in the environment, `ctx.cookie` should not attempt to save cookies into `document.cookie`.

Fixes #312

### DIFF
--- a/src/context/cookie.node.test.ts
+++ b/src/context/cookie.node.test.ts
@@ -4,19 +4,9 @@
 import { cookie } from './cookie'
 import { response } from '../response'
 
-describe('cookie (Node.js without jsdom)', () => {
-  describe('given I set a cookie', () => {
-    let result: ReturnType<typeof response>
-
-    beforeAll(() => {
-      result = response(cookie('my-cookie', 'arbitrary-value'))
-      expect(result).toBeDefined()
-    })
-
-    it('should be set on the response headers', () => {
-      expect(result.headers.get('set-cookie')).toEqual(
-        'my-cookie=arbitrary-value',
-      )
-    })
-  })
+test('sets a cookie on the response headers, node environment', () => {
+  const result: ReturnType<typeof response> = response(
+    cookie('my-cookie', 'arbitrary-value'),
+  )
+  expect(result.headers.get('set-cookie')).toEqual('my-cookie=arbitrary-value')
 })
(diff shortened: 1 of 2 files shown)


src/context/cookie.node.test.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * @jest-environment node
+ */
+import { cookie } from './cookie'
+import { response } from '../response'
+
+describe('cookie (Node.js without jsdom)', () => {
+  describe('given I set a cookie', () => {
```
kettanaito: Do you mind adjusting this test so there is a single `test` scenario?

```js
test('does not set cookies on the document when run in NodeJS', () => {
  // ... merge two of your assertions here
})
```

The current state of test suites is a little misaligned, so you might have seen a test like this. We are trying to move towards shallow tests, here's [an example](https://github.com/mswjs/msw/blob/master/test/rest-api/body.test.ts).

kettanaito (changes requested):
Thank you for the fix!
Could you please address the test-related comment? Thank you.

src/context/cookie.node.test.ts
Enet4: Certainly, it's in. Hopefully the new test is OK.

kettanaito (approved):
Thank you for the fix! Welcome to contributors, @Enet4!

PR #316, opened 2020-08-01
Delay test flakiness resolved

### LINKED ISSUE #276: Flaky: "msw-api/context/delay.test.ts" integration test
## Scope

- Test: https://github.com/mswjs/msw/blob/3ca8d86aa4cee221b9965938b1f1833259d4c44b/test/msw-api/context/delay.test.ts

## Current behavior

- Failed job: https://app.circleci.com/pipelines/github/mswjs/msw/725/workflows/930f804b-a669-41a1-a2cd-dccf0020fcf5/jobs/996/steps

Current deviation value (`150` ms) is not sufficient to cover the random response time generated by `ctx.delay()`.

## Expected behavior

Integration test for implicit response delay passes. I recommend to setup a simple Shell script that would run that test numerous times (i.e. 100) and ensure all the runs pass.

## Suggested solution

- Increase the deviation amount to `200` ms.

JacobMGEvans:
This script used to stress test Delay in CLI
- `for i in {1..100}; do yarn test:integration delay --silent || (echo 'Failed after 80 attempts' && break); done`
- Delay test 30% fail rate

Fixes Applied:
- Increased deviation from `150 -> 200` as suggested.
- 0 bailouts on 100 consecutive runs.
Ran script with --bail to break out from script of even one test failed.
`for i in {1..100}; do yarn test:integration delay --silent --bail || (echo 'Failed after 80 attempts' && break); done`

should resolve #276

### DIFF
--- a/test/msw-api/context/delay.test.ts
+++ b/test/msw-api/context/delay.test.ts
@@ -78,7 +78,7 @@ test('uses realistic server response delay, when not provided', async () => {
   // Actual response time should lie within min/max boundaries
   // of the random realistic response time.
   const responseTime = endPerf - startPerf
-  expect(responseTime).toRoughlyEqual(250, 150)
+  expect(responseTime).toRoughlyEqual(250, 200)
 
   const status = res.status()
   const headers = res.headers()


kettanaito (approved):
Thank you for the analyzing and the fix, @JacobMGEvans! Approved.
Welcome to contributors 🎉

PR #318, opened 2020-08-01
Integration test for binary responses using setupServer

alessandrojcm:
This adds a (failing, for now) integration test for returning binary data using `setupServer`.

The mock:
* Reads an image from the file system using `fs.readFileSync`
* Sets the `Content-Length`, `Content-Type`, and returns the image; pretty much like the [recipe](https://mswjs.io/docs/recipes/binary-response-type) from the docs, but in node.

The test:
* Fetches the image using `fetch`
* Parses the response as a `blob`
* Tries to convert the response from blob to a base64 representation using `FileReader`'s `readAsDataURL` (like you would do on the browser)

The tests fail with a: `TypeError: Failed to execute 'readAsDataURL' on 'FileReader': parameter 1 is not of type 'Blob'`, probably because `FileReader` it's a browser API and the response from `node-fetch` does not conform to what's expected by the `readAsDataURL` method. Use cases for this kind of test could be code that is meant to fetch images browser-side (React custom hooks for example, which is my use case).

### DIFF
--- a/test/rest-api/body-binary.node.test.ts
+++ b/test/rest-api/body-binary.node.test.ts
@@ -1,0 +1,57 @@
+/**
+ * @jest environment-node
+ */
+import * as path from 'path'
+import * as fs from 'fs'
+import fetch from 'node-fetch'
+import { rest } from 'msw'
+import { setupServer } from 'msw/node'
+
+function getImageBuffer() {
+  return fs.readFileSync(path.resolve(__dirname, '../fixtures/image.jpg'))
+}
+
+const server = setupServer(
+  rest.get('http://test.mswjs.io/image', (_, res, ctx) => {
+    const imageBuffer = getImageBuffer()
+
+    return res(
+      ctx.set('Content-Length', imageBuffer.byteLength.toString()),
+      ctx.set('Content-Type', 'image/jpeg'),
+      ctx.body(imageBuffer),
+    )
+  }),
+)
+
+beforeAll(() => server.listen())
+afterAll(() => server.close())
+
+test('returns given buffer in the mocked response', async () => {
+  const response = await fetch('http://test.mswjs.io/image')
+  const { status, headers } = response
+  const actualImageBuffer = await response.buffer()
+  const expectedImageBuffer = getImageBuffer()
+
+  expect(status).toBe(200)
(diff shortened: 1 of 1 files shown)


kettanaito:
Hey, @alessandrojcm! Thanks for providing the integration test.

I've done a few changes to the test:

- Removed the cat image fixture (although extremely cute!), since we happened to have our own image fixture in `test/fixtures/image.png`.
- Moved the test into the `test/rest-api` folder, since there was an existing `body-binary.test` suite for the `setupWorker` API.
- Removed the `*.mock.ts` module, because it's designed to represent an in-browser usage. Wrote the necessary request handler in the test suite directly.
- Added a new test case for asserting the `response.array()` response type.
- Extended the `response.blob()` case to compare the blob size and the actual blob buffer with the expected image buffer we read from the file system.
- Removed the custom blob to base64 transformation function, because we don't need it.

Both test cases pass locally, let's see what CI brings.

kettanaito:
Please let me know if this usage example works in your application, so we could merge it. Thanks.

alessandrojcm:
Hey, thanks for the quick response! So my actual test case is still failing, but I'm not sure if it is my code or `msw`.

I'm using [ky](https://www.npmjs.com/package/ky), which has a `blob()` method (it works just like `fetch`'s method, a promise that resolves to a `Blob`); now, when that `blob()` resolves it is empty, thus the test fails with a `TypeError: Failed to execute 'readAsDataURL' on 'FileReader': parameter 1 is not of type 'Blob'` (that is why the blob to base64 transformation function was in the test, it is the actual code that is failing).

If I step in with the debugger and call `arrayBuffer()` in the response I can see that the image is indeed being sent. So I don't really know if the problem is with `msw` or my test code, I know the code itself works because when I run the app in the browser it does work ok.

alessandrojcm:
Ok, so I changed my test slightly (call `arrayBuffer()` on the response and then create a `Blob` from it) and now it passes and it continues to work correctly in the browser; so I think this can be merged. Curious why it doesn't work just by calling `blob()` directly tho.

kettanaito (approved):

kettanaito:
Thank you for this test and welcome to contributors 🎉

PR #329, opened 2020-08-08
fix(graphql): handle POST w/ no query in graphql

### LINKED ISSUE #328: Must Provide Source Error
## Describe the bug
when using msw w/ testcafe, testcafe sends a POST to `/messaging` but when MSW checks to see if this request matches a route, an error is thrown saying `Must provide source`.

## Environment

- `msw: 0.20.3`
- `nodejs: 12.18.0`
- `npm: 6.x`

## To Reproduce
see comments here: https://github.com/mswjs/msw/issues/259#issuecomment-670966154

## Expected behavior
request should _not_ throw any errors and be allowed to proceed to the original request.

benmonro:
fixes https://github.com/mswjs/msw/issues/328

### DIFF
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -127,7 +127,10 @@ const createGraphQLHandler = (operationType: OperationTypeNode, mask: Mask) => {
           }
 
           case 'POST': {
-            if (!req.body) {
+            if (
+              !req.body ||
+              !(req.body as GraphQLRequestPayload<VariablesType>).query
+            ) {
               return null
             }
 
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -127,10 +127,7 @@ const createGraphQLHandler = (operationType: OperationTypeNode, mask: Mask) => {
           }
 
           case 'POST': {
-            if (
-              !req.body ||
-              !(req.body as GraphQLRequestPayload<VariablesType>).query
-            ) {
+            if (!req.body?.query) {
               return null
             }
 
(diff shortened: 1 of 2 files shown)


benmonro:
@kettanaito feel free to take this PR over, wasn't sure what the right place was to put a test to cover this use case.

src/graphql.ts
```diff
@@ -127,7 +127,10 @@ const createGraphQLHandler = (operationType: OperationTypeNode, mask: Mask) => {
           }
 
           case 'POST': {
-            if (!req.body) {
+            if (
+              !req.body ||
+              !(req.body as GraphQLRequestPayload<VariablesType>).query
```
kettanaito: Let's use optional chaining here instead to keep a shorter notation:

```js
if (!req.body?.query) {
```

Once you do this, you'll get an error that `query` doesn't exist on type `string`. That is because the `parse` function uses a `req: MockedRequest` type, but instead it should respect the custom request type provided by the request handler. To fix this, go to `src/handlers/requestHandler.ts` and change:

```diff
-parse?: (req: MockedRequest) => ParsedRequest | null
+parse?: (req: RequestType) => ParsedRequest | null
```

Let me know if you have any questions.

kettanaito (changes requested):
Thanks for finding the cause of the issue, @benmonro! This is the perfect place to follow up, so you did the right thing. I've added one comment which can make this fix even better. Please, take a look. Thanks!

benmonro:
@kettanaito should be good to go now

kettanaito (approved):
Looks awesome!

benmonro:
@kettanaito i don't have permission to merge. 🙂

PR #333, opened 2020-08-10
Update ParsedRestRequest to be exported

### LINKED ISSUE #325: Export has or is using name 'ParsedRestRequest' from external module
## Describe the bug
When building a library that exports a function that generates a configured rest object, using typescript, a type error `Export has or is using name 'ParsedRestRequest' from external module ` occurs.  This is often due to an exposed method using a type that is not exported.  In this case, it appears that adding export to the ParsedRestRequest interface fixes the issue

## Environment

- `msw: 0.20.3`
- `nodejs: 14.7`
- `npm: 6.7`
- Typescript 3.9

## To Reproduce
1. Build a library that exports a function to generate a rest mock
2. Attempt to build
3. See error:
Exported variable '....' has or is using name 'ParsedRestRequest' from external module "..../node_modules/msw/lib/types/rest" but cannot be named.

## Expected behavior
Can build 

Updating the node_modules types to export the interface locally seems to make this work

mattcosta7:
When building with declaration: true in ts config for a library re-exporting a bunch of rest mocks, an error can arise due to the return type of rest[method] being a generic, where one of the types of the generic is not exported (and therefor not accessible to the compiler when building, directly). 

A fix is to export that interface `ParsedRestRequest` from src/rest.
This alone seems to allow the compiler to access this interface and work effectively, however, it seems as though a user might also want to explicitly type these options at some point, and doing so would require a re-export of that from the main package (to match the re-exports of MockedRequest, restContext)

fixes #325

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ export {
   AsyncResponseResolverReturnType,
   defaultContext,
 } from './handlers/requestHandler'
-export { rest, restContext, RESTMethods } from './rest'
+export { rest, restContext, RESTMethods, ParsedRestRequest } from './rest'
 export {
   graphql,
   graphqlContext,
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -45,7 +45,7 @@ export const restContext = {
   fetch,
 }
 
-interface ParsedRestRequest {
+export interface ParsedRestRequest {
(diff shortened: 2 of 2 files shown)


kettanaito (approved):
Thank you for fixing this, @mattcosta7! Welcome to contributors 🎉

Issue #335, opened 2020-08-11
Worker registered on different ports in Testcafe


benmonro:
## Describe the bug
i am now using msw in testcafe w/ no problems as of 0.20.4 (was thinking I should add a recipe for this to your docs, if you want let me know and I'll open a PR).

However, when i upgrade to 0.20.5, it appears that mocking is not working.

## Environment

- `msw: 0.20.4 & 5`
- `nodejs: 12`
- `yarn: latest`

> Please also provide your browser version.
chrome latest

## To Reproduce
1. clone this branch of my repo:  https://github.com/benmonro/msw-server/tree/msw-error
2. `yarn`
3. `yarn testcafe`
note, it works w/ this version as it's using 0.20.4
4. `yarn add -D msw@0.20.5`
5. see that tests fail, MSW doesn't register, network requests all fail.

## Expected behavior
mocks work. :)

msutkowski:
Hi - random person here. Just a few notes on this - I was trying to see if any of the changes I helped introduced into the 0.20.5 release impacted this. I had to remove `yarn.lock` from `./app`, as well as `yarn start` before running the tests. 

The current issue with this is that it was actually broken(maybe?) pre 0.20.5? The problem is in `getWorkerByRegistration` - testCafe is looking at port :1337, and the app itself is running in :3000. 

Script URL: https://localhost:1337/mockServiceWorker.js
absoluteWorkerUrl: http://localhost:3000/mockServiceWorker.js

I'm working on seeing if I can solve it quickly, but that's the starting info.

benmonro:
@msutkowski (aka Random person).  actually that's the correct port.  testcafe runs tests through a proxy.  you'll notice it uses the same port w/ no problems on 0.20.4.  that's what the file located at testcafe/registerServiceWorker.js is doing, it ensures that MSW is properly registered as a service worker w/ the correct URL w/out the proxy stuff that testcafe does.

msutkowski:
@benmonro Understood! What I was trying to say is that the absoluteWorkerUrl and worker.scriptURL have to match with 0.20.5. Previously, this wasn't actually enforced and would've called worker.start() on whatever serviceWorker was registered whether it was msw or not, and could've been unintended behavior for sure (ex: if you registered two workers at once for whatever reason, it never would've worked assuming msw was registered second). 

Here is a temporary solution: 

```js
        worker.start({
          serviceWorker: {
            url: "https://localhost:1337/mockServiceWorker.js"
          }
        })
```

[...]

PR #337, opened 2020-08-14
Add matchFilenameOnly start option

msutkowski:
- Addresses the conversation in #335 by introducing an escape hatch that allows a `serviceWorker` to be matched on the filename only, disregarding the strict `origin/filename` match.

### DIFF
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -54,6 +54,14 @@ export type StartOptions = SharedOptions & {
    * instance is ready. Defaults to `true`.
    */
   waitUntilReady?: boolean
+
+  /**
+   * A boolean that allows you to override the default strict matching behavior. If set to false, the first worker
+   * based on a matching filename only (ex: mockServiceWorker.js) will be returned. This is primarily useful in the scenario that
+   * you are using proxies and having difficulty to get the origin and `scriptURL` to match due to port differences.
+   * Defaults to `false`.
+   */
+  matchFilenameOnly?: boolean
 }
 
 export type RequestHandlersList = RequestHandler<any, any>[]
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -56,7 +56,7 @@ export type StartOptions = SharedOptions & {
   waitUntilReady?: boolean
 
   /**
-   * A boolean that allows you to override the default strict matching behavior. If set to false, the first worker
+   * A boolean that allows you to override the default strict matching behavior. If set to `true`, the first worker
    * based on a matching filename only (ex: mockServiceWorker.js) will be returned. This is primarily useful in the scenario that
    * you are using proxies and having difficulty to get the origin and `scriptURL` to match due to port differences.
    * Defaults to `false`.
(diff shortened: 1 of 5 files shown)


benmonro:
@msutkowski thank you! this looks great to me.

benmonro:
@kettanaito  any thoughts on this?  Would really make your testcafe user's experience with testcafe a lot better.  😎

src/setupWorker/glossary.ts
```diff
@@ -54,6 +54,14 @@ export type StartOptions = SharedOptions & {
...
    */
   waitUntilReady?: boolean
+
+  /**
+   * A boolean that allows you to override the default strict matching behavior. If set to `true`, the first worker
+   * based on a matching filename only (ex: mockServiceWorker.js) will be returned. This is primarily useful in the scenario that
+   * you are using proxies and having difficulty to get the origin and `scriptURL` to match due to port differences.
+   * Defaults to `false`.
+   */
+  matchFilenameOnly?: boolean
```
timdeschryver, line 64: I read the issue and I think this solves that use case (and probably others too).
Another option would be to give the user this responsibility, which would cover all use cases.
It would be more complicated for the user, but it would be more flexible.

I'm not sure it needs to be more flexible tho. Perhaps this is something to keep in mind for when another use case pops up, before adding other options.

e.g.

```ts
serviceWorkerFileMatcher?: (scriptUrl:string) => boolean

if (serviceWorkerFileMatcher) {
  return serviceWorkerFileMatcher(worker.scriptURL)
}
return worker.scriptURL === absoluteWorkerUrl
```

msutkowski: Hey, thanks for this idea! I'm a big fan of matcher funcs as well _(in this instance, as well as basically any other time it's relevant)_. I think the only real distinction is that what I'm proposing is the viewpoint: 'as a library, we've thought of what you're going to run into, and this is the easiest way to solve it', and the flag becomes obvious. The matcher func requires slightly more 🧠 power and documentation, but I'd agree that it is probably the better long-term solution. 

My only challenge to the matcher func is basically what you said: "not sure it needs to be more flexible". That being said, I'm more than happy to change this to your suggestion and add in the tests once we have a solid decision 🎉.

kettanaito: I agree that we should lean towards expanding API outwards (delegating more control to the user), not inwards (encapsulating specific behaviors under flags). Thanks for pointing this out, @timdeschryver!

Depending on when such matcher would be called, we can expose more information about the worker. Ideally, a `ServiceWorkerRegistration` reference, which contains about all you may need to create a custom registration predicate.

kettanaito:
Hey, @msutkowski 👋 Sorry for not getting a proper review on this one. Please, is there anything I can help with this pull request? The intention behind it definitely has a place to be, I don't see a custom worker registration predicate as an issue, especially if it helps to solve a problem.

msutkowski:
@kettanaito Sorry, I should've left a comment when closing. I'll be opening a new PR this afternoon unless you'd prefer the updates done in this branch/PR :)

PR #338, opened 2020-08-16
onUnhandledRequest: Adds request handler declaration suggestion

### LINKED ISSUE #336: onUnhandledRequest: Provide request handler declaration suggestion
Similar to how we handle a request with a query parameter in a request URL (by providing a suggestion of how to access those query parameters using `req.url.searchParams`), we should suggest request handler declaration on unhandled request option.

There is all necessary information about the request available to construct a request handler declaration suggestion that would be copy-pastable and ready to go.

## Current behavior

https://github.com/mswjs/msw/blob/a126de22c7604f84e1a904c3025282ad25ebb00d/src/onUnhandledRequest.ts#L20

## Expected behavior

Whenever the `onUnhandledRequest` option is set on `setupWorker`/`setupServer`, the warning/error message should contain a suggestion of a request handler declaration that would have handled currently unhandled request.

```
[MSW] captured a POST https://example.xyz request without a corresponding request handler. 

If you wish to intercept this request, consider creating a request handler for it:

rest.post('https://example.xyz', (req, res, ctx) => {
  return res(ctx.text('body'))
})
```

msutkowski:
Fixes #336 

- Updates the error message as requested
- Updates relevant tests to use `toMatchInlineSnapshot`

### DIFF
--- a/src/onUnhandledRequest.ts
+++ b/src/onUnhandledRequest.ts
@@ -1,4 +1,5 @@
 import { MockedRequest } from './handlers/requestHandler'
+import { getPublicUrlFromRequest } from './utils/getPublicUrlFromRequest'
 
 type UnhandledRequestCallback = (req: MockedRequest) => void
 
@@ -17,7 +18,17 @@ export function onUnhandledRequest(
     return
   }
 
-  const message = `captured a ${request.method} ${request.url} request without a corresponding request handler.`
+  const publicUrl = getPublicUrlFromRequest(request)
+
+  const message = `captured a ${request.method} ${
+    request.url
+  } request without a corresponding request handler.
+
+  If you wish to intercept this request, consider creating a request handler for it:
+
+  rest.${request.method.toLowerCase()}('${publicUrl}', (req, res, ctx) => {
+    return res(ctx.text('body'))
+  })`
 
   switch (onUnhandledRequest) {
     case 'error': {
--- a/test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
+++ b/test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
@@ -11,13 +11,23 @@ const server = setupServer(
   }),
 )
 
-beforeAll(() => server.listen({ onUnhandledRequest: 'error' }))
-afterAll(() => server.close())
+beforeAll(() => {
+  server.listen({ onUnhandledRequest: 'error' })
+})
+
+afterAll(() => {
+  server.close()
+})
 
 test('errors on unhandled request when using the "error" value', async () => {
   const getResponse = () => fetch('https://test.mswjs.io')
 
-  await expect(getResponse()).rejects.toThrow(
-    'request to https://test.mswjs.io/ failed, reason: [MSW] Error: captured a GET https://test.mswjs.io/ request without a corresponding request handler.',
-  )
+  await expect(getResponse()).rejects.toThrow(`\
+request to https://test.mswjs.io/ failed, reason: [MSW] Error: captured a GET https://test.mswjs.io/ request without a corresponding request handler.
+
+  If you wish to intercept this request, consider creating a request handler for it:
+
+  rest.get('https://test.mswjs.io/', (req, res, ctx) => {
+    return res(ctx.text('body'))
+  })`)
 })
--- a/test/msw-api/setup-server/scenarios/on-unhandled-request/warn.test.ts
+++ b/test/msw-api/setup-server/scenarios/on-unhandled-request/warn.test.ts
@@ -11,7 +11,10 @@ const server = setupServer(
   }),
 )
 
-beforeAll(() => server.listen({ onUnhandledRequest: 'warn' }))
+beforeAll(() => {
+  server.listen({ onUnhandledRequest: 'warn' })
+})
+
 afterAll(() => {
   server.close()
   jest.restoreAllMocks()
@@ -23,7 +26,12 @@ test('warns on unhandled request when using the "warn" value', async () => {
   const res = await fetch('https://test.mswjs.io')
 
   expect(res).toHaveProperty('status', 404)
-  expect(console.warn).toBeCalledWith(
-    '[MSW] Warning: captured a GET https://test.mswjs.io/ request without a corresponding request handler.',
-  )
+  expect(console.warn).toBeCalledWith(`\
+[MSW] Warning: captured a GET https://test.mswjs.io/ request without a corresponding request handler.
+
+  If you wish to intercept this request, consider creating a request handler for it:
+
+  rest.get('https://test.mswjs.io/', (req, res, ctx) => {
+    return res(ctx.text('body'))
+  })`)
 })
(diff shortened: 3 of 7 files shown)


msutkowski:
I'm not sure why this snapshot test is failing. `toMatchInlineSnapshot` is powered by prettier, so it should be uniform across environments? Would appreciate any advice here :)

test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
```diff
@@ -17,7 +17,13 @@ afterAll(() => server.close())
 test('errors on unhandled request when using the "error" value', async () => {
   const getResponse = () => fetch('https://test.mswjs.io')
 
-  await expect(getResponse()).rejects.toThrow(
-    'request to https://test.mswjs.io/ failed, reason: [MSW] Error: captured a GET https://test.mswjs.io/ request without a corresponding request handler.',
-  )
+  await expect(getResponse()).rejects.toThrowErrorMatchingInlineSnapshot(`
```
kettanaito: Can we replace snapshot with asserting an exact message? I don't think snapshot is needed here, as it adds an extra complexity to the test.

test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
```diff
@@ -17,7 +17,13 @@ afterAll(() => server.close())
...
+          "request to https://test.mswjs.io/ failed, reason: [MSW] Error: captured a GET https://test.mswjs.io/ request without a corresponding request handler.
+
+            If you wish to intercept this request, consider creating a request handler for it:
+
+            rest.get(\\"/\\", (req, res, ctx) => {
```
kettanaito: Please assert the entire suggestion, including request handler URL. We need to make sure we are suggesting a code that is copy-pastable for the exact warning/error scenario.

```diff
-rest.get(\\"/\\",
+rest.get('https://test.mswjs.io',
```

src/onUnhandledRequest.ts
```diff
@@ -17,7 +17,17 @@ export function onUnhandledRequest(
...
 
-  const message = `captured a ${request.method} ${request.url} request without a corresponding request handler.`
+  const message = `captured a ${request.method} ${
+    request.url
+  } request without a corresponding request handler.
+
+  If you wish to intercept this request, consider creating a request handler for it:
+
+  rest.${request.method.toLowerCase()}("${
+    request.url.pathname
```
kettanaito: As you may be drawing inspiration from here:

https://github.com/mswjs/msw/blob/a126de22c7604f84e1a904c3025282ad25ebb00d/src/rest.ts#L100

I notice that relying on `pathname` won't work in case of absolute URL given to a request handler.

- Given `https://test.mswjs.io/user`
- Suggestion: '/user'

That will not work during request capturing. I think we need to distinguish between absolute and relative URLs, providing the proper request handler URL here. Let's brainstorm on this, if you don't mind.

test/msw-api/setup-server/scenarios/on-unhandled-request/warn.test.ts
```diff
@@ -18,12 +18,18 @@ afterAll(() => {
 })
 
 test('warns on unhandled request when using the "warn" value', async () => {
-  jest.spyOn(global.console, 'warn')
+  const consoleSpy = jest.spyOn(global.console, 'warn')
```
kettanaito: Please remove the spy by calling `jest.clearAllMocks()`. `afterEach` may be a good place to do that.

test/msw-api/setup-worker/start/on-unhandled-request/warn.test.ts
```diff
@@ -26,10 +26,16 @@ test('warns on an unhandled request when using the "warn" option', async () => {
   expect(status).toBe(404)
 
   const libraryErrors = messages.error.filter(filterLibraryLogs)
-  const libraryWarnings = messages.warning.filter(filterLibraryLogs)
+  const [libraryWarning] = messages.warning.filter(filterLibraryLogs)
```
kettanaito: This assumes the first warning will be related to the unhandled request. That is not a safe assumption, as there is no guarantee of an exact warning index. 

Instead, access all warnings and find the one you are attempting to assert (how it was previously).

kettanaito (changes requested):

test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
msutkowski: I'm going to have to research how to do this, as `jest` appears to be automatically escaping things (or maybe that's happening elsewhere?).

test/msw-api/setup-worker/start/on-unhandled-request/warn.test.ts
msutkowski: Noted. Having the same issue I mentioned above regarding getting formatting to match.

src/onUnhandledRequest.ts
msutkowski: At a quick pass, the code that you have in `rest.ts` will solve this if I'm understanding the intended behavior correctly. Quick example:

```
const isRelativeRequest = request.referrer.startsWith(request.url.origin)
  const publicUrl = isRelativeRequest
    ? request.url.pathname
    : format({
        protocol: request.url.protocol,
        host: request.url.host,
        pathname: request.url.pathname,
      })

  const message = `captured a ${request.method} ${
    request.url
  } request without a corresponding request handler.

  If you wish to intercept this request, consider creating a request handler for it:

  rest.${request.method.toLowerCase()}('${publicUrl}', (req, res, ctx) => {
    return res(ctx.text('body'))
  })`
```

In usage: 

```
useEffect(() => {
fetch("/this-is-a-bad-request", {
      method: "POST",
      body: JSON.stringify({ data: "two" }),
    });

    fetch("https://superbatmandudeguy.com/this-is-a-bad-request");
}, [])
```

Console output:

image

test/msw-api/setup-worker/start/on-unhandled-request/warn.test.ts
msutkowski: Disregard, formatting isn't an issue when the output is right :)

src/onUnhandledRequest.ts
msutkowski: I pushed a quick update to show this, but we should probably add a couple more test cases to show both the absolute and relative behaviors, as well as add a test for the new util if you want to keep it. I think this achieves the goal, but I'm not sure if its exactly right. The previous behavior on the absolute URL on the `setupServer` test returned a trailing slash, and the updated test does this as well... is this correct behavior?

msutkowski:
Left some comments with thoughts. Thanks!

kettanaito:
I've cleaned up the commit history, but unfortunately messed up the authorship :( 
Otherwise your changes look great, if the CI passes they will be ready to go!

kettanaito (approved):
Thank you for this improvement, @msutkowski!
Sorry for messing up the commit authorship, I'll be more careful next time.

Issue #184, opened 2020-05-25
Improve GraphQL mocking


sibelius:
**Is your feature request related to a problem? Please describe.**
Right now you need to mock the whole GraphQL response like this https://github.com/mswjs/msw/blob/e31a344332f5951177bb17655a82c7cefe59445e/test/graphql-api/query.mocks.ts

**Describe the solution you'd like**
The idea is to only mock the types and fields that you need in your test

**Describe alternatives you've considered**
I've been using `relay-testing-utils` to mock only parts of my GraphQL response https://relay.dev/docs/en/testing-relay-components
We can also use graphql-tools mocking approach (https://www.apollographql.com/docs/graphql-tools/mocking/)

**Additional context**
This is based on a twitter discussion (https://twitter.com/sseraphini/status/1264877251587432448) about [graphql-mock-api](https://github.com/sibelius/graphql-mock-api)

brapifra:
Hey @kettanaito! I would love to help you with this. I was initially experimenting with the idea of building something on the top of msw, but I think it makes sense to add it as part of the library. Also, it should be quite straightforward to implement!
<details>
<summary><b>Experiment</b></summary>

```javascript
const { makeExecutableSchema, addMocksToSchema } = require('graphql-tools');
const { graphqlSync } = require('graphql');
const { graphql } = require('msw');
const { setupServer } = require('msw/node');
const { default: ApolloClient, gql } = require('apollo-boost');
const merge = require('lodash/merge');
const fetch = require('node-fetch');

// Creates a basic schema with mocks
function getSchemaWithMocks() {
  const schemaString = `
  type User { id: Int!, name: String! }
  type Query { user(id: Int): User! }
  `;

  const schema = makeExecutableSchema({ typeDefs: schemaString });

  const schemaWithMocks = addMocksToSchema({ schema });

  return schemaWithMocks;
}

const schemaWithMocks = getSchemaWithMocks();

// Setup server
const server = setupServer(
  graphql.query('getUser', (req, res, ctx) => {
    const { data, errors } = graphqlSync(schemaWithMocks, req.body.query);

    if (errors) {
      return res(ctx.errors(errors));
    }

    return res(
      ctx.data(merge(data, { user: { name: 'This is a custom name' } }))
    );
  })
);
server.listen();

// Create graphql client
const client = new ApolloClient({
  uri: 'https://fake_domain.com/graphql',
  fetch,
});

// Query user
client
  .query({
    query: gql`
      query getUser {
[...]

PR #343, opened 2020-08-19
Add graphql.operations to api

garhbod:
## Changes

- Adds `graphql.operation` API to match any valid GraphQL operation regardless of type (query/mutation).
- Adds respective integration test suite.

## GitHub

- Relates to #184 

---

Will it need refactoring? YES
Does it work? I think so

### DIFF
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -21,7 +21,7 @@ import { getStatusCodeColor } from './utils/logging/getStatusCodeColor'
 import { jsonParse } from './utils/internal/jsonParse'
 import { matchRequestUrl } from './utils/matching/matchRequestUrl'
 
-type ExpectedOperationTypeNode = OperationTypeNode | 'any'
+type ExpectedOperationTypeNode = OperationTypeNode | 'all'
 
 type GraphQLRequestHandlerSelector = RegExp | string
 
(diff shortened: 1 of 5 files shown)


src/graphql.ts
```diff
@@ -203,13 +229,137 @@ const createGraphQLHandler = (operationType: OperationTypeNode, mask: Mask) => {
...
+        })
+        console.log('Response:', loggedResponse)
+        console.groupEnd()
+      },
+    }
+  }
+}
+
 const graphqlStandardHandlers = {
+  operations: createGraphQLOperationsHandler('*'),
```
kettanaito: I think we should name it as a singular `operations`.

That way when you declare a request handler it reads: for any GraphQL _operation_ execute this resolver:

```js
graphql.operation((req, res, ctx) => {})
```

src/graphql.ts
```diff
@@ -203,13 +229,137 @@ const createGraphQLHandler = (operationType: OperationTypeNode, mask: Mask) => {
...
   query: createGraphQLHandler('query', '*'),
   mutation: createGraphQLHandler('mutation', '*'),
 }
 
 function createGraphQLLink(uri: Mask): typeof graphqlStandardHandlers {
   return {
+    operations: createGraphQLOperationsHandler(uri),
```
kettanaito: We need to find a way to abstract an existing `createGraphQLHandler` instead of creating an entire new request handler creator. You created a good request handler, but in the end it repeats too much logic from the `createGraphQLHandler` function: `resolver`, `parse`, `log`, and a few other request handler methods should be the same. Only the `predicate` (the part that decides if a request matches) will be different in case of `.operation()` handler.

Instead, we can:

- Introduce a new enum to `query` and `mutation`: `operation`, to which the `createGraphQLHandler` will react accordingly.

After some thought I think this is the best way to go. Abstracting the operation matching logic itself may get too tedious to maintain. At least, let's try it that way, and see what it brings.

test/graphql-api/operations.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      query: LOGIN_MUTATION,
+      variables: {
+        username: 'john',
+        password: 'super-secret',
+      },
+    },
+  })
+})
+
+test('.operations() does not match non-GraphQL requests', async () => {
```
kettanaito: This test still performs a GraphQL request (`executeOperation`). I think you meant "does not match invalid GraphQL requests", since the `query` you provide here is not a valid GraphQL query.

src/graphql.ts
```diff
@@ -73,6 +73,17 @@ interface ParsedQueryPayload {
   operationName: string | undefined
 }
 
+interface ParsedQueryAndTypePayload {
```
kettanaito: Can we use `GraphQLRequestParsedResult` instead? An arbitrary GraphQL operation can also have variables. Once we add those, two interfaces look identical.

kettanaito (changes requested):
Thank you for starting with this, @garhbod! This is no doubt a great beginning of a useful feature. 

I've left a few suggestions, please take a look and let me know what you think about them. Thanks.

src/graphql.ts
garhbod: The way it was reading for me is: for all GraphQL operations execute this resolver

src/graphql.ts
garhbod: Should be resolved in new commit

test/graphql-api/operations.test.ts
garhbod: Should be resolved in new commit

src/graphql.ts
garhbod: Should be resolved in new commit

kettanaito (dismissed):
This looks great!
I've added but a minor polishing and will rebase the branch. Thank you for the effort and welcome to contributors! 🎉

PR #350, opened 2020-08-24
Add support for more types passed to `context.json` helper

### LINKED ISSUE #349: Support more types passed to `context.json` helper
## Describe the bug
The [json](https://github.com/mswjs/msw/blob/847ec554646d67eec561b64c0843d340c6ef485c/src/context/json.ts) method on `context` object should support more types than `Record<string, any>`. Anything that results in a valid `JSON.stringify` should be supported.

## Environment

- `msw: 0.20.5`
- `nodejs: 12.16.1`
- `npm: 6.13.4`

> Please also provide your browser version. (N/A)

## To Reproduce
1. Create a `rest` request handler.
2. Use the `context.json` method.
3. Try passing anything other than an object or array.
4. See the type error.

## Expected behavior
I would expect no type error to occur when passing anything that would result in a valid `JSON.stringify` call.

## Screenshots
image

wyze:
Close #349

Expand the types the `context.json` helper accepts for the `body` argument so it is not just `Record<string, any>`. It should accept any type that would result in a successful `JSON.stringify` call.

### DIFF
--- a/src/context/json.test.ts
+++ b/src/context/json.test.ts
@@ -10,21 +10,31 @@ describe('json', () => {
     })
 
     it('should have body set to the given JSON', () => {
-      const object = response(json({ firstName: 'John' }))
+      const object = json({ firstName: 'John' })
 
-      expect(object).toHaveProperty('body', `{"firstName":"John"}`)
+      expect(response(object)).toHaveProperty('body', `{"firstName":"John"}`)
 
-      const array = response(json([1, '2', true, { ok: true }, '']))
+      const array = json([1, '2', true, { ok: true }, ''])
 
-      expect(array).toHaveProperty('body', '[1,"2",true,{"ok":true},""]')
+      expect(response(array)).toHaveProperty(
+        'body',
+        '[1,"2",true,{"ok":true},""]',
+      )
 
-      const string = response(json('Some string'))
+      const string = json('Some string')
 
-      expect(string).toHaveProperty('body', '"Some string"')
+      expect(response(string)).toHaveProperty('body', '"Some string"')
 
-      const bool = response(json(true))
+      const bool = json(true)
 
-      expect(bool).toHaveProperty('body', 'true')
+      expect(response(bool)).toHaveProperty('body', 'true')
+
+      const date = json(new Date(Date.UTC(2020, 0, 1)))
+
+      expect(response(date)).toHaveProperty(
+        'body',
+        '"2020-01-01T00:00:00.000Z"',
+      )
     })
   })
 })
--- a/src/context/json.test.ts
+++ b/src/context/json.test.ts
@@ -3,18 +3,28 @@ import { response } from '../response'
 
 describe('json', () => {
   describe('given a JSON body', () => {
-    let result: ReturnType<typeof response>
-
-    beforeAll(() => {
-      result = response(json({ firstName: 'John' }))
-    })
-
     it('should have "Content-Type" as "application/json"', () => {
+      const result = response(json({ firstName: 'John' }))
+
       expect(result.headers.get('content-type')).toEqual('application/json')
     })
 
     it('should have body set to the given JSON', () => {
-      expect(result).toHaveProperty('body', `{"firstName":"John"}`)
+      const object = response(json({ firstName: 'John' }))
+
+      expect(object).toHaveProperty('body', `{"firstName":"John"}`)
+
+      const array = response(json([1, '2', true, { ok: true }, '']))
+
+      expect(array).toHaveProperty('body', '[1,"2",true,{"ok":true},""]')
+
+      const string = response(json('Some string'))
+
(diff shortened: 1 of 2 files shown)


src/context/json.ts
```diff
@@ -1,11 +1,17 @@
 import { ResponseTransformer } from '../response'
 
+type JsonArray = (JsonValue | JsonRecord)[]
+type JsonRecord = Record<string, unknown>
+type JsonValue = string | boolean | number | JsonRecord | JsonArray | null
+
 /**
  * Sets the given Object as the JSON body of the response.
```
msutkowski:
```suggestion
 * Sets the given value as the JSON body of the response.
```

src/context/json.ts
```diff
@@ -1,11 +1,17 @@
 import { ResponseTransformer } from '../response'
 
+type JsonArray = (JsonValue | JsonRecord)[]
+type JsonRecord = Record<string, unknown>
+type JsonValue = string | boolean | number | JsonRecord | JsonArray | null
```
msutkowski: This looks pretty good! I like the idea of being this strict, but it seems like we should consider matching `JSON.stringify`'s type of `any` here. 

The only gotcha I can think of here with these types as-is is that this would potentially force a TypeScript user to transform `Date`s before being passed in. I don't personally think that's really a problem but wanted to bring it up for @kettanaito's review.

msutkowski (commented):
Awesome work! Just adding a few comments for further review.

src/context/json.ts
wyze: Oops! Definitely should support Date since the method converts it to a string.

marcosvega91: I agree with this because is what `JSON.stringify` does 

```js
/**
* Converts a JavaScript value to a JavaScript Object Notation (JSON) string.
* @param value A JavaScript value, usually an object or array, to be converted.
* @param replacer An array of strings and numbers that acts as a approved list for selecting the object properties that will be stringified.
* @param space Adds indentation, white space, and line break characters to the return-value JSON text to make it easier to read.
*/
stringify(value: any, replacer?: (number | string)[] | null, space?: string | number): string;
```

kettanaito: I agree with @msutkowski on typing the value as `any`, just as `JSON.stringify` has. That leaves the type annotation aligned with the underlying stringification, and reduces the cost of maintaining it in the future. If an invalid value is passed, an error will be raised when stringifying it. It's the end developer's responsibility to pass a valid JSON to `ctx.json()`.

src/context/json.test.ts
```diff
@@ -3,18 +3,28 @@ import { response } from '../response'
...
+
+      expect(array).toHaveProperty('body', '[1,"2",true,{"ok":true},""]')
+
+      const string = response(json('Some string'))
+
+      expect(string).toHaveProperty('body', '"Some string"')
+
+      const bool = response(json(true))
+
+      expect(bool).toHaveProperty('body', 'true')
```
kettanaito: Could you please add a `Date` usage scenario here?

src/context/json.test.ts
```diff
@@ -3,18 +3,28 @@ import { response } from '../response'
...
+      const result = response(json({ firstName: 'John' }))
+
       expect(result.headers.get('content-type')).toEqual('application/json')
     })
 
     it('should have body set to the given JSON', () => {
-      expect(result).toHaveProperty('body', `{"firstName":"John"}`)
+      const object = response(json({ firstName: 'John' }))
+
+      expect(object).toHaveProperty('body', `{"firstName":"John"}`)
```
kettanaito: I would truncate these assertions to call `response()` within the `expect`. It doesn't read right, as calling the `response` function always returns a response, not object/array/etc.

```js
expect(response(json({ firstName: 'John' }))).toHaveProperty('body', `{"firstName":"John"}`)
```

src/context/json.ts
```diff
@@ -1,11 +1,17 @@
...
+
 /**
  * Sets the given Object as the JSON body of the response.
  * @example
  * res(json({ foo: 'bar' }))
+ * res(json('Some string'))
+ * res(json([1, '2', false, { ok: true }]))
  */
-export const json = (body: Record<string, any>): ResponseTransformer => {
+export const json = (body: JsonValue): ResponseTransformer => {
```
kettanaito: Based on our discussion, let's annotate it as `body: any`, similar to `JSON.stringify` argument.

kettanaito (changes requested):
Thank you for preparing these improvements, @wyze! 

It looks great, I've had a few comments on how we can shorten our test scenarios, and a point about type annotating the JSON value. Could you please take a look and address them? Thanks.

wyze:
Thank you everyone for the feedback! I have pushed a commit that addresses all comments and suggestions.

kettanaito (approved):
Looks great! Thank you for your involvement, @wyze. Welcome to contributors! 🎉

PR #351, opened 2020-08-24
Add findWorker worker.start() option and tests

### LINKED ISSUE #335: Worker registered on different ports in Testcafe

msutkowski:
Fixes #335 by implementing feedback from #337.

- Adds tests for good and bad configurations
- Adds a log for DX when this option is used and no worker instance is returned

Most likely typical usage that would solve @benmonro 's case

```js
worker
  .start({
    findWorker: (scriptUrl, _mockServiceWorkerUrl) =>
      scriptUrl.includes('mockServiceWorker'),
  })
```

### DIFF
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -37,9 +37,9 @@ export type ServiceWorkerInstanceTuple = [
   ServiceWorkerRegistration,
 ]
 
-export type ServiceWorkerMatcher = (
+export type FindWorker = (
   scriptUrl: string,
-  absoluteWorkerUrl: string,
+  mockServiceWorkerUrl: string,
 ) => boolean
 
 export type StartOptions = SharedOptions & {
@@ -61,12 +61,10 @@ export type StartOptions = SharedOptions & {
   waitUntilReady?: boolean
 
   /**
-   * A predicate function that allows you override the default strict matching behavior.
-   * This is primarily useful in the scenario that you are using proxies you are using proxies
-   * and having difficulty getting the origin and scriptURL to match due to port differences.
-   * Default: (scriptURL, absoluteWorkerUrl) => worker.scriptURL ==== absoluteWorkerUrl
+   * A custom lookup function to find a Mock Service Worker in the list
+   * of all registered Service Workers on the page.
    */
-  serviceWorkerMatcher?: ServiceWorkerMatcher
+  findWorker?: FindWorker
 }
 
 export type RequestHandlersList = RequestHandler<any, any>[]
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -37,6 +37,11 @@ export type ServiceWorkerInstanceTuple = [
   ServiceWorkerRegistration,
 ]
 
+export type ServiceWorkerMatcher = (
+  scriptUrl: string,
+  absoluteWorkerUrl: string,
+) => boolean
+
 export type StartOptions = SharedOptions & {
   serviceWorker?: {
     url?: string
@@ -54,6 +59,14 @@ export type StartOptions = SharedOptions & {
    * instance is ready. Defaults to `true`.
    */
   waitUntilReady?: boolean
+
+  /**
+   * A predicate function that allows you override the default strict matching behavior.
+   * This is primarily useful in the scenario that you are using proxies you are using proxies
+   * and having difficulty getting the origin and scriptURL to match due to port differences.
+   * Default: (scriptURL, absoluteWorkerUrl) => worker.scriptURL ==== absoluteWorkerUrl
+   */
+  serviceWorkerMatcher?: ServiceWorkerMatcher
 }
 
 export type RequestHandlersList = RequestHandler<any, any>[]
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -19,6 +19,8 @@ const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
   quiet: false,
   waitUntilReady: true,
   onUnhandledRequest: 'bypass',
+  serviceWorkerMatcher: (scriptURL, absoluteWorkerUrl) =>
+    scriptURL === absoluteWorkerUrl,
 }
 
 export const createStart = (context: SetupWorkerInternalContext) => {
@@ -49,6 +51,7 @@ export const createStart = (context: SetupWorkerInternalContext) => {
 
       const [, instance] = await until<ServiceWorkerInstanceTuple | null>(() =>
         getWorkerInstance(
+          resolvedOptions.serviceWorkerMatcher,
           resolvedOptions.serviceWorker.url,
           resolvedOptions.serviceWorker.options,
         ),
@@ -61,6 +64,12 @@ export const createStart = (context: SetupWorkerInternalContext) => {
       const [worker, registration] = instance
 
       if (!worker) {
+        if (options?.serviceWorkerMatcher) {
+          console.warn(
+            `[MSW] worker.start() was provided a serviceWorkerMatcher predicate that failed to return a worker instance. Please verify your configuration.`,
+          )
+        }
+
         return null
       }
 
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -19,8 +19,8 @@ const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
   quiet: false,
   waitUntilReady: true,
   onUnhandledRequest: 'bypass',
-  serviceWorkerMatcher: (scriptURL, absoluteWorkerUrl) =>
-    scriptURL === absoluteWorkerUrl,
+  findWorker: (scriptURL, mockServiceWorkerUrl) =>
+    scriptURL === mockServiceWorkerUrl,
 }
 
 export const createStart = (context: SetupWorkerInternalContext) => {
@@ -51,9 +51,9 @@ export const createStart = (context: SetupWorkerInternalContext) => {
 
       const [, instance] = await until<ServiceWorkerInstanceTuple | null>(() =>
         getWorkerInstance(
-          resolvedOptions.serviceWorkerMatcher,
           resolvedOptions.serviceWorker.url,
           resolvedOptions.serviceWorker.options,
+          resolvedOptions.findWorker,
         ),
       )
 
@@ -64,9 +64,9 @@ export const createStart = (context: SetupWorkerInternalContext) => {
       const [worker, registration] = instance
 
       if (!worker) {
-        if (options?.serviceWorkerMatcher) {
+        if (options?.findWorker) {
           console.warn(
-            `[MSW] worker.start() was provided a serviceWorkerMatcher predicate that failed to return a worker instance. Please verify your configuration.`,
+            `[MSW] worker.start() registered the service worker, but was provided a findWorker predicate that failed to find a worker instance. Please verify your configuration.`,
           )
         }
 
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -65,9 +65,19 @@ export const createStart = (context: SetupWorkerInternalContext) => {
 
       if (!worker) {
         if (options?.findWorker) {
-          console.warn(
-            `[MSW] worker.start() registered the service worker, but was provided a findWorker predicate that failed to find a worker instance. Please verify your configuration.`,
-          )
+          console.error(`\
+[MSW] Failed to locate the Service Worker registration using a custom "findWorker" predicate.
+
+Please ensure that the custom predicate properly locates the Service Worker registration at "${resolvedOptions.serviceWorker.url}".
+More details: https://mswjs.io/docs/api/setup-worker/start#findworker
+`)
+        } else {
+          console.error(`\
+[MSW] Failed to locate the Service Worker registration.
+
+This most likely means that the worker script URL "${resolvedOptions.serviceWorker.url}" cannot resolve against the actual public hostname (${location.host}). This may happen if your application runs behind a proxy, or has a dynamic hostname.
+
+Please consider using a custom "serviceWorker.url" option to point to the actual worker script location, or a custom "findWorker" option to resolve the Service Worker registration manually. More details: https://mswjs.io/docs/api/setup-worker/start`)
         }
 
         return null
--- a/src/setupWorker/start/utils/getWorkerByRegistration.ts
+++ b/src/setupWorker/start/utils/getWorkerByRegistration.ts
@@ -1,4 +1,4 @@
-import { ServiceWorkerMatcher } from '../../glossary'
+import { FindWorker } from '../../glossary'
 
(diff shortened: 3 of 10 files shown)


src/setupWorker/start/utils/getWorkerByRegistration.ts
```diff
@@ -13,6 +16,9 @@ export const getWorkerByRegistration = (
   ]
   const existingStates = allStates.filter(Boolean) as ServiceWorker[]
   const mockWorker = existingStates.find((worker) => {
+    if (serviceWorkerFileMatcher) {
+      return serviceWorkerFileMatcher(worker.scriptURL)
+    }
     return worker.scriptURL === absoluteWorkerUrl
```
benmonro: should absoluteWorkerUrl be passed to the predicate?

then you can instead have serviceWorkerFileMatcher default to `(a,b) => a === b` (with better names)

msutkowski: Yea, I like this idea. I opened a PR into this one as an example for review here: https://github.com/msutkowski/msw/pull/1/files

msutkowski: I think we could safely rename `serviceWorkerFileMatcher` -> `serviceWorkerMatcher`. Not sure on this one, as they both make sense to me.

timdeschryver: 👍 I like both changes

marcosvega91: 👍

msutkowski: Thanks for the input everyone - I've updated this PR with these changes.

src/setupWorker/glossary.ts
```diff
@@ -54,6 +59,14 @@ export type StartOptions = SharedOptions & {
    * instance is ready. Defaults to `true`.
    */
   waitUntilReady?: boolean
+
+  /**
+   * A predicate function that allows you override the default strict matching behavior.
+   * This is primarily useful in the scenario that you are using proxies you are using proxies
+   * and having difficulty getting the origin and scriptURL to match due to port differences.
+   * Default: (scriptURL, absoluteWorkerUrl) => worker.scriptURL ==== absoluteWorkerUrl
```
kettanaito: Please do not include the default value in the JSDocs, that couples it too much to the implementation. Thanks.

src/setupWorker/start/utils/getWorkerInstance.ts
```diff
@@ -1,13 +1,17 @@
...
+  ServiceWorkerMatcher,
+} from '../../glossary'
 import { getAbsoluteWorkerUrl } from '../../../utils/url/getAbsoluteWorkerUrl'
 
 /**
  * Returns an active Service Worker instance.
  * When not found, registers a new Service Worker.
  */
 export const getWorkerInstance = async (
+  serviceWorkerMatcher: ServiceWorkerMatcher,
```
kettanaito: I suggest we add this new argument to the end of the arguments list. Would also look more aligned with other similar function, like `getWorkerByRegistration`.

src/setupWorker/glossary.ts
```diff
@@ -37,6 +37,11 @@ export type ServiceWorkerInstanceTuple = [
   ServiceWorkerRegistration,
 ]
 
+export type ServiceWorkerMatcher = (
+  scriptUrl: string,
+  absoluteWorkerUrl: string,
```
kettanaito: It's a little confusing to have two arguments different in URL relativity to be called differently: script and worker.
I vote for naming them both `script`, as that is the official name of the worker script.

```diff
scriptUrl: string,
-absoluteWorkerUrl: string,
+absoluteScriptUrl: string
```

src/setupWorker/start/createStart.ts
```diff
@@ -61,6 +64,12 @@ export const createStart = (context: SetupWorkerInternalContext) => {
       const [worker, registration] = instance
 
       if (!worker) {
+        if (options?.serviceWorkerMatcher) {
+          console.warn(
+            `[MSW] worker.start() was provided a serviceWorkerMatcher predicate that failed to return a worker instance. Please verify your configuration.`,
```
kettanaito: This warning message is misleading: the predicate doesn't return a worker instance, but finds it in the list of currently active service workers.

src/setupWorker/glossary.ts
```diff
@@ -54,6 +59,14 @@ export type StartOptions = SharedOptions & {
...
    */
   waitUntilReady?: boolean
+
+  /**
+   * A predicate function that allows you override the default strict matching behavior.
+   * This is primarily useful in the scenario that you are using proxies you are using proxies
+   * and having difficulty getting the origin and scriptURL to match due to port differences.
+   * Default: (scriptURL, absoluteWorkerUrl) => worker.scriptURL ==== absoluteWorkerUrl
+   */
+  serviceWorkerMatcher?: ServiceWorkerMatcher
```
kettanaito: What do you think if we call this method `findWorker`?

What it does is supplies the function to the `.find()`, and looks up a given worker script URL in the list of all registered workers (incl. third-party workers).

src/setupWorker/glossary.ts
kettanaito: Diving more into the implementation, I think the call signature of this function is confusing as well. `scriptUrl` is a worker script URL of each registered worker. `absoluteWorkerUrl` is the URL of the MSW's service worker (absolute version of the URL you provided as `start({ serviceWorker: { url } })`, given to you for convenience. 

I'm thinking of something like this:

```ts
export type FindWorker = (scriptUrl: string, mockServiceWorkerUrl: string) => boolean
```

src/setupWorker/glossary.ts
```diff
@@ -54,6 +59,14 @@ export type StartOptions = SharedOptions & {
    * instance is ready. Defaults to `true`.
    */
   waitUntilReady?: boolean
+
+  /**
+   * A predicate function that allows you override the default strict matching behavior.
```
kettanaito: People are most likely not familiar with what a "strict matching behavior" is, as it references to some internal logic.

I suggest we give a short description over what this method does:

```
A custom lookup function to find a Mock Service Worker in the list of all registered Service Workers on the page.
```

I'd also drop a usage suggestions and rather include them in the documentation for this option. What do you think?

src/setupWorker/start/createStart.ts
kettanaito: Maybe let's drop this warning for now? I'm not sure what value it brings, as when the code fails to locate the worker you will get a registration error anyway, wouldn't you?

kettanaito (changes requested):
Thank you for these improvements, @msutkowski! I've left a few comments, hope they make sense. Please let me know if you have any questions.

src/setupWorker/glossary.ts
msutkowski: Noted.

src/setupWorker/glossary.ts
msutkowski: I agree, this makes sense to me.

src/setupWorker/glossary.ts
msutkowski: Agreed. Will open another PR in the docs repo shortly.

src/setupWorker/start/createStart.ts
msutkowski: In this case, if someone only supplied a `findWorker` that failed to find a worker, it'd return `null` and there would be no error because the worker technically registers. 

Commented relevant code: 

```js
const [error, instance] = await until<ServiceWorkerInstanceTuple>(
    async () => {
      const registration = await navigator.serviceWorker.register(url, options) // 1. registers and won't throw
      return [
        // Compare existing worker registration by its worker URL,
        // to prevent irrelevant workers to resolve here (such as Codesandbox worker).
        getWorkerByRegistration(registration, absoluteWorkerUrl, findWorker), // 2. will be null with a bad findWorker option
        registration,
      ]
    },
  )
```

**Other options:**
We could throw an error if we fail to find a worker here after registration instead of handling it like I am proposing, but I don't think I like that idea because the worker would technically be registered already, and we'd then want to consider unregistering it? From a DX perspective, I'd prefer to keep the mockServiceWorker on the window if possible so it reactivates faster when the `findWorker` method is fixed.

Let me know what you think

src/setupWorker/glossary.ts
kettanaito: Thanks! Don't rush with the docs, let's merge this first :)

kettanaito: The docs need to be ready when we release this, not merge. It's perfectly fine to write them after, once we underwent the review process.

src/setupWorker/glossary.ts
msutkowski: Agreed and updated

src/setupWorker/start/createStart.ts
timdeschryver: I think the `if` predicate or the message is wrong. 
`options?.findWorker` doesn't execute the predicate method, it just verifies that the predicate is provided.

msutkowski: Well, in this case we're just checking to see if it was provided, so that is the intended behavior. `getWorkerInstance` calls `findWorker`. This does actually bring up another unhandled case (ben's original one).

Maybe stepping away from the code is a good idea. Here are the currently possible scenarios:

1. Worker is registered and then found
    - Mocking enabled 🎉 
2. Worker is provided a nonexistent file
    - `throw`s error, is caught and logged with DX message
3. Worker is registered, but then not found due to strict `scriptURL === mockServiceWorkerUrl`. 
    - unhandled, console would be empty
4. Worker is registered, but not found when provided `findWorker`
    - `console.warn`s to check configuration

I'm very much open to suggestions on how to improve this and appreciate the input. I can't currently think of any better language to improve the message(s).

msutkowski: Per 👆 , this could be an idea:

```js
if (!worker) {
        if (options?.findWorker) {
          console.warn(
            `[MSW] worker.start() registered the service worker, but was provided a findWorker predicate that failed to find a worker instance. Please verify your configuration.`,
          )
        } else {
          console.warn(`\
[MSW] worker.start() registered the service worker, but was unable to find a worker instance. 

You can most likely resolve this by calling worker.start() with the findWorker option. 

[link to docs].\
          `)
        }

        return null
      }
```

timdeschryver: Thanks for the clarification 🙏

kettanaito: @msutkowski I liked that different message based on the `findWorker` option. Added that in the latest commit. Also, converted the warning to an error, because in either of those cases the operability of the library is compromised.

src/setupWorker/start/createStart.ts
```diff
@@ -61,6 +64,22 @@ export const createStart = (context: SetupWorkerInternalContext) => {
...
+[MSW] Failed to locate the Service Worker registration using a custom "findWorker" predicate.
+
+Please ensure that the custom predicate properly locates the Service Worker registration at "${resolvedOptions.serviceWorker.url})".
+More details: https://mswjs.io/docs/api/setup-worker/start#findworker
+`)
+        } else {
+          console.error(`\
+[MSW] Failed to locate the Service Worker registration.
+
+This most likely means that the worker script URL ("${resolvedOptions.serviceWorker.url}") cannot resolve against the actual public hostname (${location.host}). This may happen if your application runs behind a proxy, or has a dynamic hostname.
```
kettanaito: @msutkowski may I please ask you to check if this error message contains proper hostname information from the #335 (if you have the reproduction repository)? I expect this error to display the worker script URL and the current proxied hostname, suggesting the difference.

I know the worker script URL is not absolute at this point, but it should be sufficient.

msutkowski: Alright, looks like we're good to go here. If someone else cares to validate (@benmonro ?), please feel free.

Without specifying any options and just calling `worker.start()` on the example project: 

image

Confirmed works as expected when passing in:

```js
 worker.start({
    findWorker: (scriptUrl, _mockServiceWorkerUrl) => scriptURL.includes('mockServiceWorker')
  })
```

image

msutkowski: And here's the full error with specifying a worker serviceWorker.url that doesn't match due to proxy:

image

msutkowski:
I updated the test to check for the new copy and we're passing now 🎉 .

src/setupWorker/start/createStart.ts
kettanaito: Thank you for verifying this, @msutkowski! Happy to see that error message being as expected.

src/setupWorker/start/createStart.ts
```diff
@@ -61,6 +64,22 @@ export const createStart = (context: SetupWorkerInternalContext) => {
       const [worker, registration] = instance
 
       if (!worker) {
+        if (options?.findWorker) {
+          console.error(`\
+[MSW] Failed to locate the Service Worker registration using a custom "findWorker" predicate.
+
+Please ensure that the custom predicate properly locates the Service Worker registration at ("${resolvedOptions.serviceWorker.url}").
```
kettanaito: Sorry for a nitpick, may I please ask to remove these parenthesis? It's a leftover from my side, should just say `registration at "${resolvedOptions.serviceWorker.url}"`. Otherwise it breaks the sentence.

test/msw-api/setup-worker/start/find-worker.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  const errorMessageIndex = messages.error.findIndex((text) => {
+    return text.includes('Error - no worker instance after starting')
+  })
+
+  const libraryErrors = messages.error.filter(filterLibraryLogs)
+
+  expect(libraryErrors).toContain(`\
+[MSW] Failed to locate the Service Worker registration using a custom "findWorker" predicate.
+
+Please ensure that the custom predicate properly locates the Service Worker registration at ("/mockServiceWorker.js").
```
kettanaito: If we change that error message, this string should also be adjusted to drop parenthesis around Service Worker registration URL. Thank you!

kettanaito (changes requested):
Please, may I ask for one final thing: a little change to that error message (dropping parenthesis)? Thank you 🙏

kettanaito (approved):
Thank you for the tremendous work here, @msutkowski! Approved.

PR #352, opened 2020-08-25
rest: Supports RequestBodyType, ResponseBodyType generics

### LINKED ISSUE #345: Generics for rest handles

The TS example on the homepage shows thats you could use generics for your REST handlers.
But, this isn't supported (yet).

Rest handlers should be able to accept generics and use it to infer the request/response types.

/

**Additional context**

Example from the homepage:
```ts
interface LoginBody {
  username: string
}
interface LoginResponse {
  username: string
  firstName: string
}
const worker = setupWorker(
  //        👇 notice the generic here
  rest.post<LoginBody, LoginResponse>('/login', (req, res, ctx) => {
    const { username } = req.body
    return res(
      ctx.json({
        username,
        firstName: 'John'
      })
    )
  }),
)
worker.start()
```

kettanaito:
## Changes

- Extends the `rest` request handler to accept `RequestBodyType` and `ResponseBodyType` generics.
- Extends request handler, response composition, and context utilities to inherit the given `ResponseBodyType` to provide proper type checking during response composition.

## Usage

```ts
import { rest } from 'msw'

interface LoginRequestBodyType {
  username: string
}

interface LoginResponseBodyType {
  accessToken: string
}

rest.post<LoginRequestBodyType, LoginResponseBodyType>('/login', (req, res, ctx) => {
  const { username } = req.body // "username" is type of "string"
  // req.body.foo // ERROR, "foo" doesn't exist on LoginRequestBodyType

  return res(ctx.json({ accessToken: 'abc-123' })) // OK
  // return res(ctx.json({ foo: 'bar' })) // ERROR, unknown property "foo"
  // return res(ctx.text('foo')) // ERROR, text not assignable to LoginResponseBodyType
})
```

## GitHub

- Fixes #345

## Roadmap

- [x] ~~`rest.get` should not have the `RequestBodyType`, only the response body type.~~
- [x] Use `unknown` instead of `any`

### DIFF
--- a/src/context/body.ts
+++ b/src/context/body.ts
@@ -1,11 +1,15 @@
-import { ResponseTransformer } from '../response'
+import { ResponseTransformer, MockedResponse } from '../response'
 
 /**
  * Sets the body of the response without any `Content-Type` header.
  * @example
- * res(body('foo'))
+ * res(body('message'))
  */
-export const body = <T>(value: T): ResponseTransformer => {
+export const body = <
+  BodyType extends string | Blob | BufferSource | ReadableStream | FormData
+>(
+  value: BodyType,
+): ResponseTransformer<BodyType> => {
   return (res) => {
     res.body = value
     return res
--- a/src/context/body.ts
+++ b/src/context/body.ts
@@ -1,4 +1,4 @@
-import { ResponseTransformer, MockedResponse } from '../response'
+import { ResponseTransformer } from '../response'
 
 /**
  * Sets the body of the response without any `Content-Type` header.
--- a/src/context/json.ts
+++ b/src/context/json.ts
@@ -1,4 +1,4 @@
-import { ResponseTransformer, MockedResponse } from '../response'
+import { ResponseTransformer } from '../response'
 
 /**
  * Sets the given value as the JSON body of the response.
--- a/src/context/json.ts
+++ b/src/context/json.ts
@@ -1,16 +1,18 @@
-import { ResponseTransformer } from '../response'
+import { ResponseTransformer, MockedResponse } from '../response'
 
 /**
  * Sets the given value as the JSON body of the response.
  * @example
- * res(json({ foo: 'bar' }))
+ * res(json({ key: 'value' }))
  * res(json('Some string'))
  * res(json([1, '2', false, { ok: true }]))
  */
-export const json = (body: any): ResponseTransformer => {
+export const json = <BodyType>(
+  body: BodyType,
+): ResponseTransformer<BodyType> => {
   return (res) => {
     res.headers.set('Content-Type', 'application/json')
-    res.body = JSON.stringify(body)
+    res.body = JSON.stringify(body) as any
     return res
   }
 }
--- a/src/context/errors.ts
+++ b/src/context/errors.ts
@@ -7,6 +7,6 @@ import { json } from './json'
  */
 export const errors = (
   errorsList: Partial<GraphQLError>[],
-): ResponseTransformer => {
+): ResponseTransformer<{ errors: typeof errorsList }> => {
   return json({ errors: errorsList })
 }
--- a/src/context/text.ts
+++ b/src/context/text.ts
@@ -1,11 +1,13 @@
-import { ResponseTransformer } from '../response'
+import { ResponseTransformer, MockedResponse } from '../response'
 
 /**
  * Sets a given text as a "Cotent-Type: text/plain" body of the response.
  * @example
- * res(text('Message'))
+ * res(text('message'))
  */
-export const text = (body: string): ResponseTransformer => {
+export const text = <BodyType extends string>(
+  body: BodyType,
+): ResponseTransformer<BodyType> => {
   return (res) => {
     res.headers.set('Content-Type', 'text/plain')
     res.body = body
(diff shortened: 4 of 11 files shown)


src/context/body.ts
```diff
@@ -1,11 +1,13 @@
...
+import { ResponseTransformer, MockedResponse } from '../response'
 
 /**
  * Sets the body of the response without any `Content-Type` header.
  * @example
- * res(body('foo'))
+ * res(body('message'))
  */
-export const body = <T>(value: T): ResponseTransformer => {
+export const body = <BodyType extends string>(
```
marcosvega91: I don't understand this,  `body` should be more general because already `text` does it, I'm missing something?

kettanaito: You are right! I've mistyped this generic and have just pushed the fix. `body` will extend text, but also `BufferSource`, `FormData`, and a few other types according to [this spec](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response).

marcosvega91: Yes that's right 🚀

src/context/json.ts
```diff
@@ -1,14 +1,16 @@
...
+import { ResponseTransformer, MockedResponse } from '../response'
 
 /**
  * Sets the given Object as the JSON body of the response.
  * @example
- * res(json({ foo: 'bar' }))
+ * res(json({ key: 'value' }))
  */
-export const json = (body: Record<string, any>): ResponseTransformer => {
+export const json = <BodyType>(
```
kettanaito, line 10: The way this works is that it inherits the type of the given argument (`body`) and returns a `ResponseTransformer` with the response body annotated with that inferred type.

Combine this with the `res` response composition function that infers the type from the request handler itself, and if the response transformer type from the response composition doesn't match a response transformer type inferred from the context utility's value, you get a TypeScript error.

marcosvega91: very good 🥇

marcosvega91: very good 🥇

timdeschryver (dismissed):
👍 thanks for using generic names, instead of just `T`, `V`, ....

One nit, I wonder if we should use `unknown` instead of `any`

msutkowski:
> 👍 thanks for using generic names, instead of just `T`, `V`, ....
> 
> One nit, I wonder if we should use `unknown` instead of `any`

Agreed on the `unknown`. I was going to propose type tests for this, but I feel like they're not really complex and don't warrant it. Any thoughts there?

kettanaito:
@timdeschryver, @msutkowski, good point, I'll use `unknown` instead of `any`.

@msutkowski what kind of type test are you speaking about? Could you please post some example? Thanks!

kettanaito:
By the way, I couldn't find a way to make the number of generics conditional to include only `ResponseBodyType` in case of `rest.get`. I'm afraid GET requests will support request body type as well.

msutkowski:
I was talking about something like: https://github.com/mmkal/ts/tree/master/packages/expect-type. I looked into it briefly after a conversation regarding migrating [redux-toolkit's type tests](https://github.com/reduxjs/redux-toolkit/blob/master/type-tests/files/createAction.typetest.ts). The only other commentary I have regarding such in-depth type tests like that is that they were implemented primarily due to the number of mapped types in the library as well as the possible generic abstractions people want to do _(which isn't really applicable here)_.

I think in the case of `msw`, it might make sense to just add a few simple use cases for the generic interfaces on response bodies.

kettanaito:
Yes, I believe the best test is an example. It should be sufficient to add one/two TypeScript examples that use `rest.*` and `graphql.*` handlers with generics to type annotate request/response body.

timdeschryver:
Great minds think alike 😅
I was also thinking about type tests here, but it's not like we're doing advanced typings.

It seems like there are a lot of libraries that provide type tests.
The ones I've used in the past are:
- dtslint ([example](https://github.com/ReactiveX/rxjs/blob/master/spec-dtslint/operators/buffer-spec.ts), but not a big fan of it now)
- expecter ([example](https://github.com/ngrx/platform/blob/master/modules/store/spec/types/store.spec.ts))
- tsd ([example](https://github.com/timdeschryver/rx-query/blob/master/rx-query/query.test-d.ts))

marcosvega91 (approved):
LGTM 👍 good changes 😄

kimroen:
Nice work on this, it feels very good to be able to provide these types!

Some thoughts as I was trying to upgrade:

1. Having everything fail unless you provide the response (because of `unknown`) was a little jarring, but it was easy enough to go in and add `<DefaultRequestBodyType, any>` to get things back to the way they were
2. Having the request type go before the response type means you need to specify it even when in a majority of cases (get-requests), the default type of `DefaultRequestBodyType` is actually what I want. Having to specify it in order to provide a response type is a little annoying
3. How do you set the status code with this change? Doing this doesn't work any more:
   ```ts
   type BlogPostResponse = {
     title: string;
     body: string;
   }

   ...

   rest.get<DefaultRequestBodyType, BlogPostResponse>('/post', async (req, res, ctx) => {
     const blogPost = await getBlogPostSomehow();
     return res(
       ctx.status(200),
       ctx.json(blogPost),
     );
   })
   ```

   This fails with this error:

   ```
   Argument of type 'ResponseTransformer<unknown>' is not assignable to parameter of type 'ResponseTransformer<{ title: string; body: string; }>'.
   ```

PR #358, opened 2020-08-28
getUrlByMask: Returns URL-like path string as-is

### LINKED ISSUE #357: Path Matching: Path with parameters, inconsistent behavior between browser and nodejs
I have a handler which works in nodejs but doesn't work in browser. 

## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.20.5   |
| browser | 84.0.4147.136 64bit     |
| OS      | Debian 9 + ChromeOS 84.0.4147.136     |

## Request handlers

```js
import { rest} from 'msw';

export const handlers = [
    // Handles GET http://<hash>.server.net/resource
    // This works nodejs but doesnt work for browser
    rest.get('http://*.server.net/resource', (req, res, ctx) => {
        return res(ctx.text('Hello world'));
    }),

    // This works for browser and nodejs
    //rest.get('*.server.net/resource', (req, res, ctx) => {
    //    return res(ctx.text('Hello world'));
    //}),
];
```

## Actual request

```js
fetch('http://randomhashrighthere.server.net/resource')
        .then(response => response.text())
```

## Current behavior

When wildcard is used somewhere between the URL it's not matched with browser's requests. It is however matched in nodejs.

## Expected behavior

Handlers should have exactly the same functionality between browser and nodejs. Path matching should work  identically.

### Reproduce issue
I have set a minimal repo for reproducing the issue: https://github.com/AriPerkkio/msw-path-pattern-issue-repro
```sh
yarn install

# Run tests to validate handler works on nodejs
yarn test

# Start development mode to validate browser doesnt work
yarn start
# Open http://localhost:1234/index.html

# Modify handlers.js and re-check browser
```

If you think this is a valid issue I'm ready to contribute for a fix.

kettanaito:
## Changes

- In `getUrlByMask` short-circuit when given a path string (i.e. a string with wildcard) being a valid URL. This prevents such string to be converted into a `URL` instance, escaping the wildcard character (`*` -> `%2A`) and resulting into URL mismatch.
- Adds a new unit test for `getUrlByMask` to test URL-like path strings.

## GitHub

- Fixes #357

### DIFF
--- a/src/utils/url/getUrlByMask.test.ts
+++ b/src/utils/url/getUrlByMask.test.ts
@@ -22,6 +22,14 @@ test('creates a URL instance given a path with parameters', () => {
   expect(url.toString()).toBe('http://localhost/user/:userId')
 })
 
+test('returns a URL-like path string as-is', () => {
+  const url = getUrlByMask('http://*.mswjs.io/:resourceName')
+  expect(typeof url).toBe('string')
+  expect(url).toBe('http://*.mswjs.io/:resourceName')
+})
+
 test('returns a path string as-is', () => {
-  expect(getUrlByMask('*')).toBe('*')
+  const url = getUrlByMask('*')
+  expect(typeof url).toBe('string')
+  expect(url).toBe('*')
 })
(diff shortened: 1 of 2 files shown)


kettanaito:
I considered but turned down an idea of making request URL matching encoding-insensitive. There is no evident reason to add such change as of now, and from the URL matching perspective such path strings, no matter if can be partially represented as URL instances, must _not_ be converted into URL instances.

MichaelDeBoey (approved):

timdeschryver (approved):

kettanaito:
Thanks for the review, folks, @MichaelDeBoey, @timdeschryver!

PR #360, opened 2020-08-30
Updates to "node-request-interceptor@0.5.0"

kettanaito:
## Changes

- Updates `node-request-interceptor` to version `0.5.0` to support granular interceptors for #203.
- Excludes `node-request-interceptor/lib/interceptors/**` from the CJS build of both NodeJS and React Native targets (see https://github.com/mswjs/node-request-interceptor/issues/52).
- Isolates a `setupServer` API into a high-order function to be able to create the same API for different context (with different interceptors for NodeJS and React Native).

## GitHub

- Necessary for React Native support (#203)


kettanaito:
# `SocketPolyfill` constructor (TypeScript build issue)

- See the [issue details in this gist](https://gist.github.com/kettanaito/60d3cf9bd9303dd05be2d1f2c45f425c)
- PR that introduces the change: https://github.com/mswjs/node-request-interceptor/pull/48

## Technical insights

The issue is not exactly with how `node-request-interceptor` is built (CommonJS), but that `msw` now requires one of its _standalone_ modules directly:

```js
import { interceptClientRequest } from 'node-request-interceptor/lib/interceptors/ClientRequest'
```

As I understand, although that module is compiled, it's rather a part of the whole CJS bundle, so importing it standalone makes little sense. 

I was able to partially resolve this by listing:

```js
// rollup.config.js
const buildNode = {
  externals: ['node-request-interceptor', 'node-request-interceptor/lib/interceptors/ClientRequest']
}
```

## Solution

I believe the right solution is to build `node-request-interceptor` so that its `inteceptors/*` modules are import-able standalone (from ESM or CJS context).

marcosvega91:
I was trying to update node-request-interceptor last week and I encouraged the same problem. I though a similar solution of yours and it worked

kettanaito:
@marcosvega91, thanks for letting me know. I'm glad I'm not the only once experiencing this. Ended up with explicitly excluding interceptors in `rollup.config.ts` of `msw`. I've tried creating a custom build configuration for NRI, but even when each interceptor is compiled standalone into CJS, it's still not usable from within ESM context (the same issues).

I've raised https://github.com/mswjs/node-request-interceptor/issues/52 to tackle this in the future.

kettanaito:
I've fixed the interceptors dependency issue. The tests are green.

@Tommos0, hey 👋 May I please ask you to check if this feature build of `msw` works as expected in React Native? It should now include granular request interceptors, so that in RN it only patches `XMLHttpRequest` class. I'd be most thankful.

marcosvega91 (dismissed):
Awesome 🥳

rollup.config.ts
```diff
@@ -103,7 +109,19 @@ const buildNode = {
 
 const buildNative = {
   input: 'src/native/index.ts',
-  external: ['events', 'node-request-interceptor'],
+  external: [
+    'tty',
```
kettanaito, line 113: Not sure if it's safe to rely on these modules as external built-in modules in React Native.

Issue #363, opened 2020-08-30
Windows environment development support


msutkowski:
**Is your feature request related to a problem? Please describe.**
The current test configuration does not run in a Windows environment, either through `pwsh` or `wsl`.

**Describe the solution you'd like**
Ideally, a user in any environment should be able to fork the repo and be able to run `yarn build` and `yarn test` and have the same end result.

**Describe alternatives you've considered**
N/A

**Additional context**
I don't think there is a need to run `test:smoke` here, but I could be wrong.

PR #364, opened 2020-08-30
Windows development environment support

msutkowski:
- Addresses #363

- Update failing `test/msw-api/cli/init.test.ts` test on windows
- Update package config for cross env compatibility
   - Windows can't use the bin files, so instead, we use the js file it would resolve

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,11 @@
   "scripts": {
     "start": "cross-env NODE_ENV=development rollup -c rollup.config.ts -w",
     "clean": "rimraf lib node/**/*.js",
-    "lint": "eslint '{cli,config,src,test}/**/*.ts'",
+    "lint": "eslint \"{cli,config,src,test}/**/*.ts\"",
     "build": "yarn clean && cross-env NODE_ENV=production rollup -c rollup.config.ts",
     "test": "yarn test:unit && yarn test:integration",
     "test:unit": "cross-env BABEL_ENV=test jest --runInBand",
-    "test:integration": "node --max_old_space_size=8000 node_modules/.bin/jest --config=test/jest.config.js --runInBand",
+    "test:integration": "node --max_old_space_size=8000 node_modules/jest/bin/jest.js --config=test/jest.config.js --runInBand",
     "test:smoke": "config/scripts/smoke.sh",
     "test:focused": "node_modules/.bin/ts-node --project=test/tsconfig.json test/focusedTest.ts",
     "prepublishOnly": "yarn lint && yarn test:unit && yarn build && yarn test:integration"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,13 +11,13 @@
...
-    "lint": "eslint '{cli,config,src,test}/**/*.ts'",
+    "lint": "eslint \"{cli,config,src,test}/**/*.ts\"",
     "build": "yarn clean && cross-env NODE_ENV=production rollup -c rollup.config.ts",
     "test": "yarn test:unit && yarn test:integration",
     "test:unit": "cross-env BABEL_ENV=test jest --runInBand",
-    "test:integration": "node --max_old_space_size=8000 node_modules/.bin/jest --config=test/jest.config.js --runInBand",
+    "test:integration": "node --max_old_space_size=8000 node_modules/jest/bin/jest.js --config=test/jest.config.js --runInBand",
     "test:smoke": "config/scripts/smoke.sh",
-    "test:focused": "node_modules/.bin/ts-node --project=test/tsconfig.json test/focusedTest.ts",
+    "test:focused": "node node_modules/ts-node/dist/bin.js --project=test/tsconfig.json test/focusedTest.ts",
```
msutkowski, line 20: this doesn't seem right does it? should probably be `node -r ts-node/register`

timdeschryver: I'm on a Windows environment 👀
Do you want me to try it out?

msutkowski: > I'm on a Windows environment 👀
> Do you want me to try it out?

That'd be great. I've tested on both and it works, but I think it's supposed to use register? I'm not 100% sure tho. Have to read up further on it

msutkowski: Read up a little further on this. Being that this `test:focused` command is designed to only run one `ts` file, this should be fine as is.

You should be able to run the following with no issues on windows (I've tested with `PowerShell` _(only runs from an elevated prompt)_ and `PowerShell 7`):

1. `yarn test:focused test/rest-api/basic.mocks.ts`
2. `yarn test`

Regarding `wsl` usage: that seems to be a can of worms and the instructions for getting puppeteer to work properly would be different depending on the distro someone was using. 😞

kettanaito:
I've tested this feature branch on MacOS and verify backwards-compatibility 👍 
Now what's left is for Windows users to confirm these changes allow them to contribute.

kettanaito (approved):
The changes are good. Please let's get some more feedback on how this runs on Windows, and then merge. Superb work, @msutkowski!

themindoverall:
Tested on Windows 10 Pro v2004. Looks good

kettanaito:
Thank you, @themindoverall! I think this pull request is ready to go.

PR #365, opened 2020-08-30
Adds integration test for deferring network traffic when registration fails

### LINKED ISSUE #362: App traffic is frozen if the service worker registration rejects (deferNetworkRequestsUntil)
## Describe the bug

To deal with a race condition between worker registration and application's requests, MSW performs network deferring. That deferring never restores if the worker registration Promise rejects, it only works if the registration Promise resolves.

## Environment

- `msw: 0.20.5`
- `nodejs: irrelevant`
- `npm: 6`

## To Reproduce

1. Call `worker.start()` that rejects (i.e. referencing non-existing worker script).
1. Perform a fetch call.
1. See nothing.

## Expected behavior
Deferring the network traffic must always restore the deferred requests, reacting to the `worker.start()` Promise as `.finally`, not `.then`:

https://github.com/mswjs/msw/blob/2943c080bbdaaf74ec9e23147592fe76a0e6a147/src/utils/deferNetworkRequestsUntil.ts#L15

kettanaito:
## GitHub

- Closes #362, but doesn't fix anything

### DIFF
--- a/test/support/spawnServer.ts
+++ b/test/support/spawnServer.ts
@@ -47,6 +47,9 @@ Using Service Worker build:
     mode: 'development',
     target: 'web',
     entry: [path.resolve(__dirname, 'utils'), absoluteMockPath],
+    output: {
+      publicPath: '/',
+    },
     module: {
       rules: [
         {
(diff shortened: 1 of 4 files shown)


test/support/spawnServer.ts
```diff
@@ -93,14 +96,12 @@ Using Service Worker build:
       // to be registered at the website's root.
       'Service-Worker-Allowed': '/',
     },
-    after(app, server, compiler) {
+    before(app, server, compiler) {
+      options?.withRoutes?.(app, server, compiler)
```
kettanaito, line 100: Per [WDS documentation](https://webpack.js.org/configuration/dev-server/#devserverbefore), `before` is the right option to attach custom route handlers.

Issue #361, opened 2020-08-30
 server-side behavior in NextJS


hauptrolle:
## Describe the bug
When adding response patching to an api endpoint which is called on the server-side of NextJS it throws an error.

## Environment

- `msw: ^0.20.4`
- `nodejs: v10.15.2`
- `npm: 6.4.1`

Chrome: 84.0.4147.135

> Please also provide your browser version.

## To Reproduce
1. Checkout this repository and follow the steps in the readme: https://github.com/hauptrolle/nextjs-msw-response-patching-bug

## Expected behavior
The response should be patched without an error.

marcosvega91:
Hi @hauptrolle thanks for raising this 😄 

The issue happens when `fetch` is called using the mocked request. When the `request` is of type `GET` or `HEAD` the body should be set to `null` when `useFetch` is called.

To do this the following diff should be enough 
```diff
diff --git a/src/context/fetch.ts b/src/context/fetch.ts
index 7265c94..58ee629 100644
--- a/src/context/fetch.ts
+++ b/src/context/fetch.ts
@@ -43,10 +43,18 @@ export const fetch = <ResponseType = any>(
   }
 
   const { body } = input
-  const compliantReq: RequestInit = augmentRequestInit({
+
+  const requestParameters: RequestInit = {
     ...input,
-    body: typeof body === 'object' ? JSON.stringify(body) : body,
-  })
+    body: null,
+  }
+
+  if (input.method !== 'GET' && input.method !== 'HEAD') {
+    requestParameters.body =
+      typeof body === 'object' ? JSON.stringify(body) : body
+  }
+
+  const compliantReq: RequestInit = augmentRequestInit(requestParameters)
 
   return gracefully<ResponseType>(useFetch(input.url.href, compliantReq))
 }
```

To you want to work on a PR 😄  ? We nee also a test for this. You contribution will be awesome 💯

hauptrolle:
@marcosvega91 Cool! Thank you for your fast feedback. Sure, I can work on a MR :) Will finish it probably around tomorrow in the evening. 👋

marcosvega91:
It's not a problem, thanks again for the support ❤️

kettanaito:
This issue was resolved in [`msw@0.21.0`](https://github.com/mswjs/msw/releases/tag/v0.21.0). Could you please update and let us know if it's fixed for you? Thanks.

PR #366, opened 2020-08-31
Set body to null for GET & HEAD requests

hauptrolle:
Hey,

this should fix the bug #361.

I added a check for `GET` & `HEAD` requests and set the body to`null` in this case.

### DIFF
--- a/src/context/fetch.test.ts
+++ b/src/context/fetch.test.ts
@@ -1,6 +1,4 @@
-import { augmentRequestInit, createFetchRequestParameters } from './fetch'
-import { MockedRequest } from '../utils/handlers/requestHandler'
-import { Headers } from 'headers-utils/lib'
+import { augmentRequestInit } from './fetch'
 
 describe('augmentRequestInit', () => {
   describe('given provided custom headers', () => {
@@ -70,60 +68,6 @@ describe('augmentRequestInit', () => {
       it('should preserve custom headers', () => {
         expect(headers.get('authorization')).toEqual('token')
       })
-    })
-  })
-})
-
-describe('createFetchRequestParameters', () => {
-  let mockedRequest: MockedRequest
-
-  beforeAll(() => {
-    mockedRequest = {
-      url: new URL('http://www.mswjs.io'),
-      method: 'GET',
-      headers: new Headers(),
-      cookies: {},
-      mode: 'same-origin',
-      keepalive: true,
-      cache: 'default',
-      destination: '',
-      integrity: '',
-      credentials: 'same-origin',
-      redirect: 'follow',
-      referrer: '',
-      referrerPolicy: 'origin',
-      body: {},
-      bodyUsed: true,
-      params: {},
-    }
-  })
-
-  describe('given a GET request', () => {
-    it('should set the body to null', () => {
-      const result = createFetchRequestParameters(mockedRequest)
-      expect(result.body).toEqual(null)
-    })
-  })
-
-  describe('given a HEAD request', () => {
-    it('should set the body to null', () => {
-      mockedRequest.method = 'HEAD'
-
-      const result = createFetchRequestParameters(mockedRequest)
-      expect(result.body).toEqual(null)
-    })
-  })
-
-  describe('given a POST request', () => {
-    it('should keep the body', () => {
-      const body = {
-        foo: 'bar',
-      }
-      mockedRequest.method = 'POST'
-      mockedRequest.body = body
-
-      const result = createFetchRequestParameters(mockedRequest)
-      expect(result.body).toEqual(JSON.stringify(body))
     })
   })
 })
--- a/src/context/fetch.test.ts
+++ b/src/context/fetch.test.ts
@@ -1,4 +1,6 @@
-import { augmentRequestInit } from './fetch'
+import { augmentRequestInit, createFetchRequestParameters } from './fetch'
+import { MockedRequest } from '../utils/handlers/requestHandler'
+import { Headers } from 'headers-utils/lib'
 
 describe('augmentRequestInit', () => {
   describe('given provided custom headers', () => {
@@ -68,6 +70,60 @@ describe('augmentRequestInit', () => {
       it('should preserve custom headers', () => {
         expect(headers.get('authorization')).toEqual('token')
       })
+    })
+  })
+})
+
+describe('createFetchRequestParameters', () => {
+  let mockedRequest: MockedRequest
+
+  beforeAll(() => {
+    mockedRequest = {
+      url: new URL('http://www.mswjs.io'),
+      method: 'GET',
+      headers: new Headers(),
+      cookies: {},
+      mode: 'same-origin',
+      keepalive: true,
+      cache: 'default',
+      destination: '',
+      integrity: '',
+      credentials: 'same-origin',
+      redirect: 'follow',
+      referrer: '',
+      referrerPolicy: 'origin',
+      body: {},
+      bodyUsed: true,
+      params: {},
+    }
+  })
+
+  describe('given a GET request', () => {
+    it('should set the body to null', () => {
+      const result = createFetchRequestParameters(mockedRequest)
+      expect(result.body).toEqual(null)
+    })
+  })
+
+  describe('given a HEAD request', () => {
+    it('should set the body to null', () => {
+      mockedRequest.method = 'HEAD'
+
+      const result = createFetchRequestParameters(mockedRequest)
+      expect(result.body).toEqual(null)
+    })
+  })
+
+  describe('given a POST request', () => {
+    it('should keep the body', () => {
+      const body = {
+        foo: 'bar',
+      }
+      mockedRequest.method = 'POST'
+      mockedRequest.body = body
+
+      const result = createFetchRequestParameters(mockedRequest)
+      expect(result.body).toEqual(JSON.stringify(body))
     })
   })
 })
--- a/src/context/fetch.ts
+++ b/src/context/fetch.ts
@@ -26,7 +26,7 @@ export const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
   }
 }
 
-export const createFetchRequestParameters = (input: MockedRequest) => {
+const createFetchRequestParameters = (input: MockedRequest) => {
   const { body } = input
   const requestParameters: RequestInit = {
     ...input,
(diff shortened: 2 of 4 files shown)


src/context/fetch.test.ts
```diff
@@ -71,3 +73,57 @@ describe('augmentRequestInit', () => {
...
+        foo: 'bar',
+      }
+      mockedRequest.method = 'POST'
+      mockedRequest.body = body
+
+      const result = createFetchRequestParameters(mockedRequest)
+      expect(result.body).toEqual(JSON.stringify(body))
+    })
+  })
+})
```
marcosvega91: I think that is better to add an integration test in the file `response-patching.test.ts`

marcosvega91 (changes requested):
Hi @hauptrolle thanks for your help, it's awesome 😄 . About your test I think is better to add an integration test as for the comment

src/context/fetch.test.ts
hauptrolle: Okay I will add an integration test for this. Should I delete the unit tests then? Or should we keep it?

hauptrolle: I tried my best to add an integration test, but when try to run these tests I always get: `ERROR in ./test/rest-api/response-patching.mocks.ts
    Module not found: Error: Can't resolve 'msw' in '/Users/achim.rolle/workspace/github/msw/test/rest-api'`

marcosvega91: before running the integration test you should build the library with `yarn build` script.
We can remove the unit for this case. Thanks a lot

test/rest-api/response-patching.test.ts
```diff
@@ -161,34 +157,7 @@ describe('REST: Response patching', () => {
...
-          )
-        },
-      })
-      const status = res.status()
-      const headers = res.headers()
-      const body = await res.json()
-
-      expect(status).toBe(200)
-      expect(headers).toHaveProperty('x-powered-by', 'msw')
-      expect(body).toEqual(null)
```
marcosvega91: we can keep the test with HEAD method

marcosvega91 (changes requested):
We have almost done 🥳

test/rest-api/response-patching.test.ts
hauptrolle: I tried to create a test for the `HEAD` method. Unfortunately I always get a `TimeoutError: Timeout exceeded while waiting for event`. Am I doing something wrong here? Would appreciate some help 😅

msutkowski: It looks like the `responsePredicate` is returning `false`, which is causing the `TimeoutError` you see. If you just `return true` there, you'll run into: 

```
  ● REST: Response patching › given a HEAD request to be patched › should be able to properly request and patch the response

    Protocol error (Network.getResponseBody): No data found for resource with given identifier
```

You'll want to make sure that you add handlers for the route. 

Try this out. in `response-patching.mocks.ts`, add:
```js
  rest.head('/headtest', async (req, res, ctx) => {
    return res(
      ctx.json(null),
    )
  }),
```

In `response-patching.test.ts`:

```js
withRoutes(app) {
          app.post('/headers-proxy', (req, res) => {
            const { authorization } = req.headers

            if (!authorization) {
              return res.status(403).json({ message: 'error' }).end()
            }

            return res.status(200).json({ message: 'success' }).end()
          })

          app.post('/posts', (req, res) => {
            res.status(200).json({ id: 101 }).end()
          })

          app.head('/posts', (req, res) => {
            res.status(200).end()
          })

          app.get('/posts', (req, res) => {
            res.status(200).json({ id: 101 }).end()
          })

          app.head('/headtest', (req, res) => {
            res.status(200).end()
          })
        },
```

And add the test:

```js
  describe('given a HEAD request to be patched', () => {
    it('should be able to properly request and patch the response', async () => {
      const res = await test.request({
        url: '/headtest',
        fetchOptions: {
          method: 'HEAD'
        },
        responsePredicate(res, url) {
          return (
            match(test.origin + url, res.url()).matches &&
            res.headers()['x-powered-by'] === 'msw'
          )
        },
      })
      const status = res.status()
      const headers = res.headers()
      const body = await res.json()

      expect(status).toBe(200)
      expect(headers).toHaveProperty('x-powered-by', 'msw')
      expect(body).toEqual(null)
    })
  })
```

msutkowski: disclaimer: even though this works, @marcosvega91 or @kettanaito probably have a better recommendation here.

hauptrolle: @msutkowski Thank you for your help! Added the test. Lets wait for the feedback from @kettanaito or @marcosvega91  😀

test/rest-api/response-patching.mocks.ts
```diff
@@ -48,6 +48,21 @@ const worker = setupWorker(
...
+
+    return res(
+      ctx.json({
+        ...originalResponse,
+        mocked: true,
+      }),
+    )
+  }),
+
+  rest.head('/headtest', async (req, res, ctx) => {
```
msutkowski:
```suggestion
  rest.head('/posts', async (req, res, ctx) => {
```

test/rest-api/response-patching.test.ts
```diff
@@ -23,6 +23,14 @@ describe('REST: Response patching', () => {
...
             res.status(200).json({ id: 101 }).end()
           })
+
+          app.get('/posts', (req, res) => {
+            res.status(200).json({ id: 101 }).end()
+          })
+
+          app.head('/headtest', (req, res) => {
+            res.status(200).end()
+          })
```
msutkowski:
```suggestion
```

msutkowski (changes requested):
Alright, I did some further testing, and these changes should clean this up and continue to work as expected. 

The reason that I used `/headtest` was that there was some weird behavior happening on my end due to me having a `return` in the handlers, which didn't belong. Give these changes a try and let us know :)

test/rest-api/response-patching.test.ts
hauptrolle: Thank you so much for your support. Updated the test according to your changes :)

test/rest-api/response-patching.test.ts
marcosvega91: HEAD request should make the original request using `ctx.fetch` to make the test valid. 
```js
rest.head('/posts', async (req, res, ctx) => {
   const originalResponse = await ctx.fetch(req) <----- this line 
   return res(ctx.json(null))
}),
```

The problem is that the head request is not resolving throwing the timeout error. @kettanaito have you any ideas of what the problem is ?

marcosvega91: I have found the problem. After the original fetch is called, a function `gracefully` will return the `json` or the `text` depending on the `header`. The problem is that when using method `HEAD` the header is set but the body will be always null. I don't know if is truly correct to use `gracefully` function because then you cannot access headers in your mocked function 🤔

test/rest-api/response-patching.mocks.ts
```diff
@@ -48,6 +48,21 @@ const worker = setupWorker(
...
+    return res(
+      ctx.json({
+        ...originalResponse,
+        mocked: true,
+      }),
+    )
+  }),
+
+  rest.head('/posts', async (req, res, ctx) => {
+    return res(ctx.json(null))
```
kettanaito: I would be careful with using `null` as the response body, as it may produce a false positive test result given there is unintentionally no body. Please, could you set a certain text/JSON body in this response resolver? So we are sure we are asserting the mocked response body in the test. Thanks.

src/context/fetch.ts
```diff
@@ -26,6 +26,21 @@ export const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
   }
 }
 
+const createFetchRequestParameters = (input: MockedRequest) => {
+  const { body } = input
+  const requestParameters: RequestInit = {
+    ...input,
+    body: null,
```
kettanaito: I believe that the default value of the request `body` is `undefined` when it's not set.

This is also how we are handling a GET request body received from the Service Worker:

https://github.com/mswjs/msw/blob/c254c99c9a658e5619ace45cbf03ddeb1fbd027b/src/utils/handleRequestWith.ts#L39-L50

src/context/fetch.ts
```diff
@@ -26,6 +26,21 @@ export const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
...
+  }
+
+  if (input.method !== 'GET' && input.method !== 'HEAD') {
```
kettanaito: Nitpick: I suggest you also get the `method` property from the `input` destructuring at the top of this function, the same as you do with the `body` property.

kettanaito: Reversed logical operators are generally much harder to read (`!a` than `a`), so I recommend to flip this condition like so:

```js
// If it's a GET/HEAD request, return the request parameters as-is.
if (['GET', 'HEAD'].includes(method)) {
  return requestParameters
}

// Otherwise handle the request body.
requestParameters.body = typeof body === 'object' ? JSON.stringify(body) : body

return requestParameters
```

src/context/fetch.ts
```diff
@@ -42,11 +57,9 @@ export const fetch = <ResponseType = any>(
...
   }
 
-  const { body } = input
-  const compliantReq: RequestInit = augmentRequestInit({
-    ...input,
-    body: typeof body === 'object' ? JSON.stringify(body) : body,
-  })
+  const requestParameters: RequestInit = createFetchRequestParameters(input)
+
+  const compliantReq: RequestInit = augmentRequestInit(requestParameters)
```
kettanaito: We should pick one: either `request` or `req`. Since we are already using the longer version in the codebase, I suggest to name this `compliantRequest`. Nothing wrong with being a little more specific :)

kettanaito (changes requested):
These are some awesome changes, @hauptrolle! Thank you for preparing them.
I've left a few comments, mainly nitpicks and tiny improvements. Could you please look at them once you have a spare minute? Thank you once more.

kettanaito (approved):
Looks awesome! Thank you for preparing this, @hauptrolle! Approved. Welcome to contributors 🎉

marcosvega91:
We need only to fix a test because is not making the original request. I can do in the PR opened for the `fetch`
```javascript
rest.head('/posts', async (req, res, ctx) => {
    return res(
      ctx.json({
        mocked: true,
      }),
    )
  }),
```

kettanaito:
@marcosvega91, that would be awesome! Thank you 🎉

PR #371, opened 2020-09-05
Implement keep alive interval

### LINKED ISSUE #367: Chrome removes idle service worker after 30-40 seconds
## Describe the bug
Because of the way user agents handle idle service worker (https://www.w3.org/TR/service-workers-1/#service-worker-lifetime) they can, and will, be disabled after a certain time has passed.

This is usually not an issue. However, we have an e2e-test that is running for +40 seconds, filling out a form in various ways, before making the request. During this time the service worker will have been disabled by the user agent. The request that should be intercepted will instead fall through and in this case, fail.

Basically. If the msw is idle long enough it will be disabled until the page is reloaded. If the first request to the msw is made after +40 seconds of other application interactions it will not be enabled once the request is made.

## Environment
See codesandbox package.json for details

- `msw: 0.20.5`
- `chrome: 84`

## To Reproduce
1. Go to https://codesandbox.io/s/hidden-currying-u0ec0?file=/src/App.tsx
2. When the app has loaded it's possible to make a fetch by pressing the button, this will get intercepted by MSW.
4. Reload the codesandbox preview browser and wait about 42 seconds (From my experiments I've found that having devtools open will prevent the service worker from timing out. It have to be closed to reproduce the error)
5. Press the fetch button
6. Notice the error message getting rendered. This message is the error from trying to JSON-parse the html that is returned by the SPA as the request to the `/test`-endpoint is not intercepted by the msw any longer.

## Expected behavior
It does not seem to be possible to prevent the user-agent from disabling the service worker if it's idle for too long. I see two alternatives: 
1. Keep the msw alive somehow, i.e. constantly make keep-alive requests (not ideal).
2. Somehow make the msw handle the request even though it was temporarily disabled by the user-agent from not having any events sent to it for too long.

msutkowski:
Fixes #367 

- Implements a simple interval to play ping pong with the worker once it is activated.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -30,7 +30,6 @@ self.addEventListener('message', async function (event) {
     case 'KEEPALIVE_REQUEST': {
       sendToClient(client, {
         type: 'KEEPALIVE_RESPONSE',
-        payload: new Date(),
       })
       break
     }
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -27,6 +27,14 @@ self.addEventListener('message', async function (event) {
   const allClientIds = allClients.map((client) => client.id)
 
   switch (event.data) {
+    case 'PING': {
+      sendToClient(client, {
+        type: 'PONG',
+        payload: new Date(),
+      })
+      break
+    }
+
     case 'INTEGRITY_CHECK_REQUEST': {
       sendToClient(client, {
         type: 'INTEGRITY_CHECK_RESPONSE',
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -27,9 +27,9 @@ self.addEventListener('message', async function (event) {
   const allClientIds = allClients.map((client) => client.id)
 
   switch (event.data) {
-    case 'PING': {
+    case 'KEEPALIVE_REQUEST': {
       sendToClient(client, {
-        type: 'PONG',
+        type: 'KEEPALIVE_RESPONSE',
         payload: new Date(),
       })
       break
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -11,6 +11,7 @@ export interface SetupWorkerInternalContext {
   worker: ServiceWorker | null
   registration: ServiceWorkerRegistration | null
   requestHandlers: RequestHandler<any, any>[]
+  keepAliveInterval?: ReturnType<typeof setInterval> | number
   events: {
     /**
      * Adds an event listener on the given target.
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -11,7 +11,7 @@ export interface SetupWorkerInternalContext {
   worker: ServiceWorker | null
   registration: ServiceWorkerRegistration | null
   requestHandlers: RequestHandler<any, any>[]
-  keepAliveInterval?: ReturnType<typeof setInterval> | number
+  keepAliveInterval?: number
   events: {
     /**
      * Adds an event listener on the given target.
@@ -66,6 +66,12 @@ export type StartOptions = SharedOptions & {
    * of all registered Service Workers on the page.
    */
   findWorker?: FindWorker
+
+  /**
+   * Disable the keep alive or use a custom interval.
+   * Defaults to 10000 (ms)
+   */
+  keepAlive?: false | number
 }
 
 export type RequestHandlersList = RequestHandler<any, any>[]
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -66,12 +66,6 @@ export type StartOptions = SharedOptions & {
    * of all registered Service Workers on the page.
    */
   findWorker?: FindWorker
-
-  /**
-   * Disable the keep alive or use a custom interval.
-   * Defaults to 10000 (ms)
-   */
-  keepAlive?: false | number
 }
 
 export type RequestHandlersList = RequestHandler<any, any>[]
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -124,6 +124,11 @@ If this message still persists after updating, please report an issue: https://g
         return null
       }
 
+      context.keepAliveInterval = setInterval(
+        () => worker.postMessage('PING'),
+        5000,
+      )
+
       return registration
     }
 
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -17,6 +17,7 @@ const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
     options: null as any,
   },
   quiet: false,
+  keepAlive: 10000,
   waitUntilReady: true,
   onUnhandledRequest: 'bypass',
   findWorker: (scriptURL, mockServiceWorkerUrl) =>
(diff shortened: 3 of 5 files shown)


msutkowski:
I could use some ideas on how to test this if it even needs tests. I don't think there is any real benefit in creating a test that would keep the browser open for an extended period of time as a part of the regular test suite, but perhaps it could be run separately for verifying this behavior? We could also consider adding PING/PONG logs to the console to check for like:

```ts
// handleRequestWith.ts

 // Ignore irrelevant worker message types
      if (type !== 'REQUEST') {
        
        if (type === 'PONG' && !options.quiet) {
          console.info('[MSW] Recieved PONG from mockServiceWorker')
        }

        return null
      }
```

Also, It might be more clear to use KEEP_ALIVE_REQUEST/RESPONSE instead of PING/PONG as well

chrisguttandin:
Hi @msutkowski, I came up with the concept of expectation tests for that use case.

I do for example have a project called [standardized-audio-context](https://github.com/chrisguttandin/standardized-audio-context) which tries to provide an implementation of the Web Audio API which behaves consistent across all browsers. Therefore the codebase is full of ugly workarounds which don't really make sense at first glance. In order to remember why I added a workaround and also to know if it is still necessary I added an [expectation test](https://github.com/chrisguttandin/standardized-audio-context/tree/master/test/expectation) for each workaround. These tests verify that my expectation that there is a browser bug which I need to circumvent is still valid. Once these tests fail I know I can safely remove the workaround.

Maybe the concept comes in handy here as well. You could add a test which checks if the Service Worker actually gets killed by the browser after a certain amount of time.

msutkowski:
That's a good thought and goes along with what I was hinting at by questioning the necessity. I think the only way we can achieve that is by making the `keepAlive` functionality an option and just defaulting it to true. We could then have 3 total test cases covered: 

1. PING/PONG events are logged when `quiet: false`
2. msw fails to handle a request after 2 minutes because the worker is stopped by the user agent
3. msw still handles a request after 2 minutes because the keepAlive works.

src/setupWorker/start/createStart.ts
```diff
@@ -124,6 +124,11 @@ If this message still persists after updating, please report an issue: https://g
         return null
       }
 
+      context.keepAliveInterval = setInterval(
+        () => worker.postMessage('PING'),
+        5000,
```
kettanaito, line 132: If the kill timeout is rather large for the Service Worker, what are the repercussions of raising this interval?

src/setupWorker/start/createStart.ts
```diff
@@ -124,6 +124,11 @@ If this message still persists after updating, please report an issue: https://g
         return null
       }
 
+      context.keepAliveInterval = setInterval(
+        () => worker.postMessage('PING'),
```
kettanaito: What would happen if the worker has terminated/been killed and we keep posting messages to it?
Should we account for that scenario in any way?

src/setupWorker/glossary.ts
```diff
@@ -11,6 +11,7 @@ export interface SetupWorkerInternalContext {
   worker: ServiceWorker | null
   registration: ServiceWorkerRegistration | null
   requestHandlers: RequestHandler<any, any>[]
+  keepAliveInterval?: ReturnType<typeof setInterval> | number
```
kettanaito: Can we use `number` only here? So the `keepAliveInterval` type is compatible with the `clearInterval` argument types?

We can also type it as `number | undefined`, because the interval reference is indeed undefined at some point of time, before the interval gets created in the `start()` call.

src/mockServiceWorker.js
```diff
@@ -27,6 +27,14 @@ self.addEventListener('message', async function (event) {
   const allClientIds = allClients.map((client) => client.id)
 
   switch (event.data) {
+    case 'PING': {
```
kettanaito: I agree that these events could be renamed into something like `KEEPALIVE_REQUEST` and `KEEPALIVE_RESPONSE` for better readability.

src/mockServiceWorker.js
```diff
@@ -27,6 +27,14 @@ self.addEventListener('message', async function (event) {
...
+      sendToClient(client, {
+        type: 'PONG',
+        payload: new Date(),
```
kettanaito: Is there any intricate reason to pass this `payload`? I wonder if this affects something 🤔

kettanaito:
Thank you for preparing this, @msutkowski! I think it's a solid foundation of dealing with service worker termination. I've left a few comments, would be glad to review again in the nearest time.

src/setupWorker/glossary.ts
msutkowski: Yea - the issue is apparently that I needed to use `window.setInterval` - just using `setInterval` is polluted by node types 🤦

src/mockServiceWorker.js
msutkowski: No - it was just an idea to look like there was some type of valid communication happening. It doesn't hurt to have it, but also doesn't seem to matter if it's removed.

src/setupWorker/start/createStart.ts
msutkowski: No repercussions - what I'm going to do is make this configurable with sensible defaults so someone could change the timeout to be lower or higher. 30s is supposed to be the time before the worker is stopped, but that behavior seems very random to me and changes by available resources on a machine.

src/setupWorker/start/createStart.ts
msutkowski: If we keep posting messages and the user agent stops the service worker, it will reactivate it, which is most likely the behavior that we'd want (assuming the user never calls `worker.stop`. There is no way for us to check if the worker is stopped or not, so the user would have to call `worker.stop()` to stop posting messages. I can't imagine a scenario where anyone would want to keep the worker active but stop the keep alive, as it'd just lead to weird behavior like some of these issues report.

src/utils/handleRequestWith.ts
```diff
@@ -57,6 +57,12 @@ export const handleRequestWith = (
 
       // Ignore irrelevant worker message types
       if (type !== 'REQUEST') {
+        if (type === 'KEEPALIVE_RESPONSE' && !options.quiet) {
+          console.info(
+            `[MSW] Received Keep Alive response from the mockServiceWorker`,
```
kettanaito: This is going to considerably clutter the developer's console. Maybe let's leave out the logging of internal communications?

src/setupWorker/glossary.ts
```diff
@@ -65,6 +66,12 @@ export type StartOptions = SharedOptions & {
    * of all registered Service Workers on the page.
    */
   findWorker?: FindWorker
+
+  /**
+   * Disable the keep alive or use a custom interval.
+   * Defaults to 10000 (ms)
+   */
+  keepAlive?: false | number
```
kettanaito: I take a general rule of thumb to parametrize public stuff. Since this is rather internal interval, I think we should skip the ability to change it. That way we shift the responsibility of it having optimal value to ourselves, rather than the end user.

src/mockServiceWorker.js
kettanaito: Seems trivial, but we need to think about this as changes to the `mockServiceWorker` file should be minimal across library updates (they require to re-initialize the worker file). I'd remove the `payload` property unless it's absolutely necessary. If ping-ponging empty events still keeps the worker alive, I'd leave it without any additional properties.

kettanaito (changes requested):

src/utils/handleRequestWith.ts
msutkowski: Sounds good. removed.

src/setupWorker/glossary.ts
msutkowski: In this case, I've just removed the option altogether. It's safe just to keep it a 5s and hide the behavior, as it's what users expect.

kettanaito (approved):
Looks great! Thank you for the improvement, @msutkowski! 🎉

PR #373, opened 2020-09-07
feat: add gracefullyFetch method and make fetch return original response

### LINKED ISSUE #372: ctx.fetch should return request and not the body

The actual behaviour of `ctx.fetch` is to `gracefully` return the body of the response. This will be a problem when the user wants to check `headers` and not only the `body`.
For example for `HEAD` request the `body` is not defined because only `headers` are populated.

I think that `ctx.fetch` should return the response and give the user the responsibility to handle it.
We can consider to add a parameter for example `gracefullyResponse` that by default is `true`.
In this way the actual behaviour will be maintained.

This will be a breaking change if someone is already using it and we decide to return always the original response.

marcosvega91:
fix #372

the pr will change the behaviour of `fetch`. Now the method will return the `Response` without any transformation.

This will be a breaking change for ones using the `fetch` method.


src/context/fetch.ts
```diff
@@ -26,20 +26,25 @@ export const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
   }
 }
 
+export const gracefullyFetch = <ResponseType = any>(
```
kettanaito: What do you think if we drop this graceful handling entirely? This encapsulated body resolution behavior doesn't help much, and using explicit `.text`, `.json`, etc. is much better for both compliance with an actual fetch and versatility.

src/context/fetch.ts
```diff
@@ -26,20 +26,25 @@ export const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
...
  * Wrapper around the native `window.fetch()` function that performs
  * a request bypassing MSW. Requests performed using
  * this function will never be mocked.
  */
-export const fetch = <ResponseType = any>(
+export const fetch = (
   input: string | MockedRequest,
   requestInit: RequestInit = {},
-) => {
+): Promise<Response> => {
```
kettanaito: Wonder if we can still use the given `ResponseType` generic with a regular `fetch` return statement 🤔

kettanaito (changes requested):

src/context/fetch.ts
marcosvega91: yes we can, is something that can be dropped

src/context/fetch.ts
marcosvega91: I don't know if it makes sense. Do you have an example of usage ?

kettanaito: My though was to let the end developer type-annotate the type of the response body they would receive. I suppose since there's no such possibility with the regular `fetch`, the `ctx.fetch` may drop it as well. One can use type casting for that purpose. What do you think?

marcosvega91: Yes it's right, it makes sense

kettanaito (dismissed):
Looks awesome! Thank you, @marcosvega91! 🎉

kettanaito:
Could you please update the branch against the latest `master`? Then we'll merge :) Thank you.

marcosvega91:
Yes I'm rebasing the branch and fix the test. It'll be ready in a minute

kettanaito (approved):

Issue #374, opened 2020-09-08
Question: Cypress MSW started, but not intercepted


deshiknaves:
I'm not sure where the best place to ask questions, so I'm asking them here. This is a simple example and I'll abstract it more when I can get the simple example working. What I'm trying to do is load MSW as part of the Cypress tests and not part of the application itself. This will allow me use use MSW to mock any part of the application — just like `fixtures` do in Cypress. The benefit being that MSW allows me to inspect the request before responding, something Cypress doesn't allow at the moment.

## Environment

| Name    | Version |
| -------- | ----------------- |
| msw       | ^0.20.5                  |
| browser |  85.0.4183.83       |
| OS          | macOS 10.15.6     |

## Request handlers

```js
// App.js in CRA
function App() {
  const [message, setMessage] = useState()

  useEffect(() => {
    setTimeout(() => {
      fetch('https://jsonplaceholder.typicode.com/todos/1')
        .then(response => response.json())
        .then(json => console.log(json))
    }, 1000)
  }, [])

  return (
    <div className="App">
      <h1>{message}</h1>
      <div>
        <button type="button" onClick={() => setMessage('First')}>
          First
        </button>
        <button type="button" onClick={() => setMessage('Second')}>
          Second
        </button>
        <button type="button" onClick={() => setMessage('Third')}>
          Third
        </button>
      </div>
    </div>
  )
}
```

```js
// In Cypress test
import { setupWorker, rest } from 'msw'

describe('First', () => {
  let worker

  beforeEach(() => {
    cy.visit('/', {
      onBeforeLoad(win) {
        worker = setupWorker(
          rest.get(
            'https://jsonplaceholder.typicode.com/todos/1',
            (req, res, ctx) => {
              return res(
                ctx.json({
                  userId: 1,
                  id: 1,
                  title: 'Lord of the rings',
                  completed: false,
                }),
              )
            },
          ),
        )
        worker.start()
        win.msw = { worker, rest }
      },
    })
  })
  it('should be able to set the first message', () => {
    cy.window().then(win => {
      console.log(win.msw)
    })
    cy.findByRole('button', { name: /first/i }).click()
    cy.findByRole('heading', { name: /first/i }).should('be.visible')
  })
})

```

## Actual request

```js
// Example of making a request. Provide your code here.
fetch('https://jsonplaceholder.typicode.com/todos/1')
        .then(response => response.json())
        .then(json => console.log(json))
```

## Current behavior

[...]

PR #383, opened 2020-09-12
Start option to share a client across all other client requests

deshiknaves:
This is in relation to #374 . The terminology might not be correct, there might be a better way to describe this other than `shared`.

The use case is that some one would be able to load `msw` as part of a Cypress test suite rather than as part of the application itself. This can be done by:

```javascript
// cypress/support/index.js

let worker

before(() => {
  worker = setupWorker()
  cy.wrap(worker.start())
})

Cypress.on('window:before:load', win => {
  if (!worker) return

  worker.resetHandlers()

  win.msw = { worker, rest }
})

Cypress.Commands.add('mock', (method, route, fn) => {
  worker.use(rest[method.toLowerCase()](route, fn))
})
```

Then in a test, the user can:

```javascript
  cy.mock(
      'GET',
      (req, res, ctx) => {
        return res(
          ctx.json({
            userId: 1,
            id: 1,
            title: 'Lord of the rings',
            completed: false,
          }),
        )
      },
    )
```

This currently doesn't work because the application that Cypress is testing has a different `clientId` than the one that started the worker and so the mocks aren't intercepted.

What I thought could be an option is to set an option of `{ serviceWorker: { shared: true } }` to `worker.start` so that the `mockServiceWorker` can capture that `sharedClientId` and use that when intercepting the requests.

~The other thing I'm not quite sure of is, how I can create a request from another `clientId` in the integration test.~ I have added tests for this.

I'd love to hear your thoughts.

**Note:** Some info on how this will be used: https://github.com/deshiknaves/cypress-msw-interceptor

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -37,7 +37,7 @@ self.addEventListener('message', async function (event) {
   const allClients = await self.clients.matchAll()
   const allClientIds = allClients.map((client) => client.id)
 
-  switch (event.data) {
+  switch (event.data.type) {
     case 'KEEPALIVE_REQUEST': {
       sendToClient(client, {
         type: 'KEEPALIVE_RESPONSE',
@@ -57,20 +57,12 @@ self.addEventListener('message', async function (event) {
       clients = ensureKeys(allClientIds, clients)
       clients[clientId] = true
 
+      if (event.data.shared) {
+        sharedClientId = clientId
+      }
+
       sendToClient(client, {
         type: 'MOCKING_ENABLED',
-        payload: true,
-      })
-      break
-    }
-
-    case 'SHARED_MOCK_ACTIVATE': {
-      clients = ensureKeys(allClientIds, clients)
-      clients[clientId] = true
-      sharedClientId = clientId
-
-      sendToClient(client, {
-        type: 'SHARED_MOCKING_ENABLED',
         payload: true,
       })
       break
@@ -101,7 +93,7 @@ self.addEventListener('message', async function (event) {
   }
 })
 
-self.addEventListener('fetch', async function (event) {
+self.addEventListener('fetch', function (event) {
   const { request } = event
   const requestId = uuidv4()
   const requestClone = request.clone()
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -87,6 +87,10 @@ self.addEventListener('message', async function (event) {
         return client.id !== clientId
       })
 
+      if (sharedClientId === clientId) {
+        sharedClientId = null
+      }
+
       // Unregister itself when there are no more clients
       if (remainingClients.length === 0) {
         self.registration.unregister()
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -11,6 +11,7 @@ const INTEGRITY_CHECKSUM = '<INTEGRITY_CHECKSUM>'
 const bypassHeaderName = 'x-msw-bypass'
 
 let clients = {}
+let sharedClientId
 
 self.addEventListener('install', function () {
   return self.skipWaiting()
@@ -63,6 +64,18 @@ self.addEventListener('message', async function (event) {
       break
     }
 
+    case 'SHARED_MOCK_ACTIVATE': {
+      clients = ensureKeys(allClientIds, clients)
+      clients[clientId] = true
+      sharedClientId = clientId
+
+      sendToClient(client, {
+        type: 'SHARED_MOCKING_ENABLED',
+        payload: true,
+      })
+      break
+    }
+
     case 'MOCK_DEACTIVATE': {
       clients = ensureKeys(allClientIds, clients)
       clients[clientId] = false
@@ -84,8 +97,8 @@ self.addEventListener('message', async function (event) {
   }
 })
 
-self.addEventListener('fetch', function (event) {
-  const { clientId, request } = event
+self.addEventListener('fetch', async function (event) {
+  const { request } = event
   const requestId = uuidv4()
   const requestClone = request.clone()
   const getOriginalResponse = () => fetch(requestClone)
@@ -97,7 +110,7 @@ self.addEventListener('fetch', function (event) {
 
   // Bypass mocking if the current client isn't present in the internal clients map
   // (i.e. has the mocking disabled).
-  if (!clients[clientId]) {
+  if (!clients[event.clientId]) {
     return
   }
 
@@ -109,6 +122,7 @@ self.addEventListener('fetch', function (event) {
 
   event.respondWith(
     new Promise(async (resolve, reject) => {
+      const clientId = sharedClientId || event.clientId
       const client = await self.clients.get(clientId)
 
       // Bypass mocking when the request client is not active.
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -15,9 +15,9 @@ import { createResponseListener } from '../../utils/worker/createResponseListene
 const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
   serviceWorker: {
     url: '/mockServiceWorker.js',
-    shared: false,
     options: null as any,
   },
+  shared: false,
   quiet: false,
   waitUntilReady: true,
   onUnhandledRequest: 'bypass',
@@ -127,7 +127,7 @@ If this message still persists after updating, please report an issue: https://g
 
       // Signal the Service Worker to enable requests interception
       const [activationError] = await until(() =>
-        activateMocking(context, options, resolvedOptions.serviceWorker.shared),
+        activateMocking(context, options),
       )
 
       if (activationError) {
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -15,6 +15,7 @@ import { createResponseListener } from '../../utils/worker/createResponseListene
 const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
   serviceWorker: {
     url: '/mockServiceWorker.js',
+    shared: false,
     options: null as any,
   },
   quiet: false,
@@ -126,7 +127,7 @@ If this message still persists after updating, please report an issue: https://g
 
       // Signal the Service Worker to enable requests interception
       const [activationError] = await until(() =>
-        activateMocking(context, options),
+        activateMocking(context, options, resolvedOptions.serviceWorker.shared),
       )
 
       if (activationError) {
--- a/src/setupWorker/start/utils/activateMocking.ts
+++ b/src/setupWorker/start/utils/activateMocking.ts
@@ -3,19 +3,16 @@ import { StartOptions, SetupWorkerInternalContext } from '../../glossary'
 export const activateMocking = async (
   context: SetupWorkerInternalContext,
   options?: StartOptions,
-  shared = false,
 ) => {
-  const message = shared ? 'SHARED_MOCK_ACTIVATE' : 'MOCK_ACTIVATE'
-  const event = shared ? 'SHARED_MOCKING_ENABLED' : 'MOCKING_ENABLED'
-  context.worker?.postMessage(message)
+  context.workerChannel.send('MOCK_ACTIVATE', { shared: options?.shared })
 
-  return context.events.once(event).then(() => {
+  return context.events.once('MOCKING_ENABLED').then(() => {
     if (!options?.quiet) {
       console.groupCollapsed(
-        `%c[MSW] ${shared ? 'Shared ' : ''}Mocking enabled.`,
+        `%c[MSW] ${options?.shared ? 'Shared ' : ''}Mocking enabled.`,
         'color:orangered;font-weight:bold;',
       )
-      if (shared) {
+      if (options?.shared) {
         console.log('All connected clients will be served these mocks')
       }
       console.log(
--- a/src/setupWorker/start/utils/activateMocking.ts
+++ b/src/setupWorker/start/utils/activateMocking.ts
@@ -3,15 +3,21 @@ import { StartOptions, SetupWorkerInternalContext } from '../../glossary'
 export const activateMocking = async (
   context: SetupWorkerInternalContext,
   options?: StartOptions,
+  shared = false,
 ) => {
-  context.workerChannel.send('MOCK_ACTIVATE')
+  const message = shared ? 'SHARED_MOCK_ACTIVATE' : 'MOCK_ACTIVATE'
+  const event = shared ? 'SHARED_MOCKING_ENABLED' : 'MOCKING_ENABLED'
+  context.worker?.postMessage(message)
 
-  return context.events.once('MOCKING_ENABLED').then(() => {
+  return context.events.once(event).then(() => {
     if (!options?.quiet) {
       console.groupCollapsed(
-        '%c[MSW] Mocking enabled.',
+        `%c[MSW] ${shared ? 'Shared ' : ''}Mocking enabled.`,
         'color:orangered;font-weight:bold;',
       )
+      if (shared) {
+        console.log('All connected clients will be served these mocks')
+      }
       console.log(
         '%cDocumentation: %chttps://mswjs.io/docs',
         'font-weight:bold',
--- a/test/msw-api/setup-worker/start/shared.test.ts
+++ b/test/msw-api/setup-worker/start/shared.test.ts
@@ -14,35 +14,41 @@ afterAll(() => {
 
 test('shares the client registration for all other clients when "shared" option is set to "true"', async () => {
   const { messages } = captureConsole(runtime.page)
-
-  await runtime.page.evaluate(() => {
-    // @ts-ignore
-    return window.__MSW_REGISTRATION__
-  })
+  const iframeUrl = `${runtime.origin}/test/support/template/iframe.html`
 
   await runtime.reload()
+
+  await runtime.page.evaluate((url) => {
+    const iframe = document.createElement('iframe')
+    iframe.src = url
+    document.body.appendChild(iframe)
+
+    return iframe
+  }, iframeUrl)
+
+  await runtime.page.waitForSelector('iframe')
+  const element = await runtime.page.$(`iframe[src="${iframeUrl}"]`)
+  const frame = await element.contentFrame()
+
+  const response = await frame.evaluate(() => {
+    return fetch(`${window.location.origin}/user`).then((response) =>
+      response.json(),
+    )
+  })
+
+  expect(response).toEqual({
+    firstName: 'John',
+    age: 32,
+  })
 
   const activationMessage = messages.startGroupCollapsed.find((text) => {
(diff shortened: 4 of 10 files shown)


deshiknaves:
So I dug into this further and replicated it in a test.

This happens when there is an iframe on the page and makes requests. They are not intercepted by parent mocks as the `clientId`s are different. This provides a way for child "clients" to use the mocks provided in the parent context.

deshiknaves:
The example above is irrelevant if you're not thinking in the context of Cypress, you can ignore that. The integration test will better illustrate the use case.

The iframe makes a request that should be intercepted by the mocks but isn't without this modification.

deshiknaves:
Hmm this failed in CI, but all the tests ran fine on multiple runs using `yarn test`. I'm not noticing anything different with the CI and my run.

kettanaito:
Hey, @deshiknaves. I'm a little overburdened at the moment, so I will look at these changes in a week or so. Sorry it takes so long.

marcosvega91:
Tomorrow I have enough time to take a look at this PR. 🚀

deshiknaves:
It's alright, we all have the pressures of work and life 😊. Keep me posted and I'll do what I need to.

src/mockServiceWorker.js
```diff
@@ -53,6 +54,18 @@ self.addEventListener('message', async function (event) {
       break
     }
 
+    case 'SHARED_MOCK_ACTIVATE': {
+      clients = ensureKeys(allClientIds, clients)
+      clients[clientId] = true
+      sharedClientId = clientId
```
marcosvega91: We should also clear the sharedClientId  onClose or deactivation

test/msw-api/setup-worker/start/shared.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    return text.includes('[MSW] Shared Mocking enabled.')
+  })
+  expect(activationMessage).toBeTruthy()
+
+  const requestsLog = messages.startGroupCollapsed.find((text) => {
+    return text.includes('[MSW]') && text.includes('GET /user')
+  })
+
+  expect(requestsLog).toMatch(/\[MSW\] \d{2}:\d{2}:\d{2} GET \/user 200/)
+})
```
marcosvega91:
```suggestion



test('shares the client registration for all other clients when "shared" option is set to "true"', async () => {
  const runtime = await runBrowserWith(path.resolve(__dirname, 'shared.mocks.ts'))
  const { messages } = captureConsole(runtime.page)
  const iframeUrl = `${runtime.origin}/test/support/template/iframe.html`

  await runtime.reload()

  await runtime.page.evaluate((url) => {
    const iframe = document.createElement('iframe')
    iframe.src = url
    document.body.appendChild(iframe)

    return iframe
  }, iframeUrl)

  await runtime.page.waitForSelector('iframe')
  const element = await runtime.page.$(`iframe[src="${iframeUrl}"]`)
  const frame = await element.contentFrame()

  const response = await frame.evaluate(() => {
    return fetch(`${window.location.origin}/user`).then((response) =>
      response.json(),
    )
  })

  expect(response).toEqual({
    firstName: 'John',
    age: 32,
  })

  const activationMessage = messages.startGroupCollapsed.find((text) => {
    return text.includes('[MSW] Shared Mocking enabled.')
  })
  expect(activationMessage).toBeTruthy()

  const requestsLog = messages.startGroupCollapsed.find((text) => {
    return text.includes('[MSW]') && text.includes('GET /user')
  })

  expect(requestsLog).toMatch(/\[MSW\] \d{2}:\d{2}:\d{2} GET \/user 200/)
  
  return runtime.cleanup()
})
```

marcosvega91 (changes requested):
Hi guys, 

I think that @deshiknaves has made a good work and, as for the other PR, I think that this is a good way to make Cypress work better with MSW. Of course as for the other PR I want an opinion also from @kettanaito and @msutkowski 

Do @deshiknaves have looked at the CI error? 

thanks all guys

src/mockServiceWorker.js
deshiknaves: Yes, that sounds like a good idea.

@marcosvega91 I found the CI error interesting. I've run them all locally and it all passes. I will update both branches tomorrow and see what's going on.

deshiknaves: This has been added

deshiknaves:
These have been open for a while. Please let me know what I can do to progress them 😊

deshiknaves:
Looks like something in the recent changes broke something. I'll dig into this.

Update: Hmm looks like master isn't passing the tests either.

filipemir:
Hey folks. Stumbled upon this PR while struggling with the exact same problem. Nice work, @deshiknaves. Being able to use MSW for request handling in Cypress would be a huge improvement to their very limited request handling. I'd love to see this merged if it gets the job done. Let me know I can help too.

marcosvega91:
I have thought about this for a different time, trying to find a way to not update service worker. I'll try to explain my idea about this.

### Goal 

We want to start `MSW` on cypress side and not on `app` side. In this way we can add more handlers on cypress like we do with jest for example.

### Problem

`Cypress` runs our application in an iframe, so when we try to start service worker on cypress side it will be registered by cypress itself and not by the app. When the app will try to make some requests they will not be intercepted because the app is not registered with the service worker.

### Possible solution

A possibile solution is to pass to `MSW` the current window instance. `MSW` uses the navigator object to register to the service worker. On `window:before:load` event we have the `window` instance of the application that is different than `global.window`. My idea is to pass the `window` instance to `MSW`.

our `setupWorker.ts` could be 

```ts
import { SetupWorkerInternalContext, RequestHandlersList } from './glossary'
import { createStart } from './start/createStart'
import { createStop } from './stop/createStop'
import * as requestHandlerUtils from '../utils/handlers/requestHandlerUtils'
import { isNodeProcess } from '../utils/internal/isNodeProcess'
import { ServiceWorkerMessage } from '../utils/createBroadcastChannel'

export interface SetupWorkerApi {
  start: ReturnType<typeof createStart>
  stop: ReturnType<typeof createStop>

  /**
   * Prepends given request handlers to the list of existing handlers.
   */
  use: (...handlers: RequestHandlersList) => void

  /**
   * Marks all request handlers that respond using `res.once()` as unused.
   */
  restoreHandlers: () => void

  /**
   * Resets request handlers to the initial list given to the `setupWorker` call, or to the explicit next request handlers list, if given.
   */
  resetHandlers: (...nextHandlers: RequestHandlersList) => void
}

/**
 * Configures a Service Worker with the given request handler functions.
 */

interface Listener {
  target: EventTarget
  event: string
  callback: EventListener
}

// Declare the list of event handlers on the module's scope
// so it persists between Fash refreshes of the application's code.
let listeners: Listener[] = []

function within(window: Window & typeof globalThis) {
  return function (...requestHandlers: RequestHandlersList): SetupWorkerApi {
    requestHandlers.forEach((handler) => {
      if (Array.isArray(handler))
        throw new Error(
          `[MSW] Failed to call "setupWorker" given an Array of request handlers (setupWorker([a, b])), expected to receive each handler individually: setupWorker(a, b).`,
        )
    })

    const context: SetupWorkerInternalContext = {
      worker: null,
      window: window,
      registration: null,
      requestHandlers: [...requestHandlers],
      events: {
        addListener(
          target: EventTarget,
          event: string,
          callback: EventListener,
        ) {
          target.addEventListener(event, callback)
          listeners.push({ event, target, callback })

          return () => {
            target.removeEventListener(event, callback)
          }
        },
        removeAllListeners() {
          for (const { target, event, callback } of listeners) {
            target.removeEventListener(event, callback)
          }
          listeners = []
        },
        once<T>(type: string) {
          const bindings: Array<() => void> = []

          return new Promise<ServiceWorkerMessage<T>>((resolve, reject) => {
            const handleIncomingMessage = (event: MessageEvent) => {
              try {
                const message = JSON.parse(event.data)

                if (message.type === type) {
                  resolve(message)
                }
              } catch (error) {
                reject(error)
              }
            }

            bindings.push(
              context.events.addListener(
                window.navigator.serviceWorker,
                'message',
                handleIncomingMessage,
              ),
              context.events.addListener(
                window.navigator.serviceWorker,
                'messageerror',
                reject,
              ),
            )
          }).finally(() => {
            bindings.forEach((unbind) => unbind())
          })
        },
      },
    }

    // Error when attempting to run this function in a NodeJS environment.
    if (isNodeProcess()) {
      throw new Error(
        '[MSW] Failed to execute `setupWorker` in a non-browser environment. Consider using `setupServer` for NodeJS environment instead.',
      )
    }

    return {
      start: createStart(context),
      stop: createStop(context),

      use(...handlers) {
        requestHandlerUtils.use(context.requestHandlers, ...handlers)
      },

      restoreHandlers() {
        requestHandlerUtils.restoreHandlers(context.requestHandlers)
      },

      resetHandlers(...nextHandlers) {
        context.requestHandlers = requestHandlerUtils.resetHandlers(
          requestHandlers,
          ...nextHandlers,
        )
      },
    }
  }
}

const setupWorker = within(window)

export { setupWorker, within }
```

simonsmith:
For those using Cypress they've recently added `intercept` which will allow control of all network requests - https://www.cypress.io/blog/2020/11/24/introducing-cy-intercept-next-generation-network-stubbing-in-cypress-6-0/

However I'd still love to see progress on this as it would be great to mock libraries like Stripe which run in their own iframe.

deshiknaves:
@marcosvega91 now that lifecycle hooks have been released. I'll have a look at this this week.

kettanaito:
Let us know if you need any help with this, @deshiknaves. We would love to have better Cypress support and this improvement would contribute to it greatly.

deshiknaves:
@kettanaito @marcosvega91 I played around with the within idea and it really doesn't work here. The service worker is not loaded inside the application, but is loaded in the Cypress app. If you pass the `window` as suggested, then the context can't communicate with with the service worker because the `message` event is not triggered in the same context.

This is an issue not only for Cypress, but also for ifames in an application. I think there are more than one use case for wanting to share the context across all frames.

The addition to the service worker is tiny and pretty safe, but it provides a viable way of sharing a context.

All the integration tests seem to be failing on master at the moment. I'll investigate soon.

src/setupWorker/start/createStart.ts
```diff
@@ -15,6 +15,7 @@ import { createResponseListener } from '../../utils/worker/createResponseListene
 const DEFAULT_START_OPTIONS: DeepRequired<StartOptions> = {
   serviceWorker: {
     url: '/mockServiceWorker.js',
+    shared: false,
```
kettanaito: Can we have MSW working in a "shared" mode by default? What implications would there be in that case?

kettanaito: I wouldn't put library-specific options under the `serviceWorker` key in the `StartOptions`. The intention for this key is to represent standard Service Worker options. We should probably move it to the root-level in that case.

src/mockServiceWorker.js
```diff
@@ -53,6 +54,18 @@ self.addEventListener('message', async function (event) {
       break
     }
 
+    case 'SHARED_MOCK_ACTIVATE': {
```
kettanaito: This case behaves the same as `MOCK_ACTIVATE` with the exception of setting the `sharedClientId` global variable. Once more I'm wondering if we can do that within `MOCK_ACTIVATE` and skip the introduction of a new event for that only purpose. We can still transfer additional data in `event.data`, so even `MOCK_ACTIVATE` event can contain something like `event.data.sharedClientId` to determine the need of setting `sharedClientId`.

src/mockServiceWorker.js
```diff
@@ -99,6 +116,7 @@ self.addEventListener('fetch', function (event) {
 
   event.respondWith(
     new Promise(async (resolve, reject) => {
+      const clientId = sharedClientId || event.clientId
```
kettanaito, line 121: This value can be computed outside of the `event.respondWith` scope, at the upper scope of the `fetch` event handler. Would you mind moving it up?

src/setupWorker/start/createStart.ts
```diff
@@ -122,7 +123,7 @@ If this message still persists after updating, please report an issue: https://g
 
       // Signal the Service Worker to enable requests interception
       const [activationError] = await until(() =>
-        activateMocking(context, options),
+        activateMocking(context, options, resolvedOptions.serviceWorker.shared),
```
kettanaito: I think `activeMocking` should accept `resolvedOptions` as the second argument instead of `options`. Then you can access the `shared` property within it, without having to pass it explicitly.

src/setupWorker/start/utils/activateMocking.ts
```diff
@@ -3,15 +3,21 @@ import { StartOptions, SetupWorkerInternalContext } from '../../glossary'
 export const activateMocking = async (
   context: SetupWorkerInternalContext,
   options?: StartOptions,
+  shared = false,
```
kettanaito: This option should be accessed from the second `options` argument already passed here.

kettanaito (changes requested):
Hey, @deshiknaves. I went through the changes and left a few comments. Meanwhile, I'm still trying to understand the technical side of how exactly this enables context sharing between parent/iframe windows.

deshiknaves:
> Hey, @deshiknaves. I went through the changes and left a few comments. Meanwhile, I'm still trying to understand the technical side of how exactly this enables context sharing between parent/iframe windows.

@kettanaito thanks, I'll update this soon. Yep, let me explain. What this option allows for is one client being responsible for providing the mocks to all other clients. When an iframe makes a request, it has a unique `clientId` which will never match the parent's `clientId`. However, the mocks exist in the context of parent. What this allows is for the parent to start up a server saying that it is responsible for providing the mocks to all other clients.

So when the registration happens, the service worker keep the `clientId` of the parent as the `sharedClientId`. When the iframe makes a request, if there is a `sharedClientId` then that is used instead of it's `clientId` allowing the parent to intercept the iframe request.

src/mockServiceWorker.js
deshiknaves: Yep, I'll do this

src/setupWorker/start/createStart.ts
deshiknaves: @kettanaito I actually have question to be honest. I've been trying to understand why a mock server can't just mock requests for all clients that connect to it. Seems like it was an intentional design decision. Does anyone have any more context on this?

deshiknaves:
@kettanaito @marcosvega91 do the tests pass for you currently on the `master` branch? Sorry, I haven't update this in a while, got busy with a few things. A lot of the integration tests are failing for me on the `master` branch.

src/mockServiceWorker.js
```diff
@@ -84,8 +101,8 @@ self.addEventListener('message', async function (event) {
   }
 })
 
-self.addEventListener('fetch', function (event) {
-  const { clientId, request } = event
+self.addEventListener('fetch', async function (event) {
```
kettanaito: If I remember correctly, the `fetch` event handler cannot be asynchronous. If it's async it loses the context of the request and may resolve sooner than the `event.respondWith` is called. I'd recommend keeping this handler synchronous (note that `event.respondWith` can still return a Promise, which can be an async function).

kettanaito:
I'd like to refresh my understanding of this: the issue is that the worker checks if the client is known by looking it up in the `clints` object by client ID. In the case of Cypress, that client ID (in an iframe) is not directly controlled by the worker, thus the lookup for existence fails, and requests originating from Cypress are not intercepted. Is this correct, @deshiknaves? 

I wonder how much information a `client` reference exposes. Maybe instead of manually keeping some shared client ID we could check if the client that originates a request has a parent frame that is controlled by the worker. 

Edit: I can see that the `client` instance has a `frameType` property. In the case of the immediate page it's `"top-level"`.  All possible frame types (per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Client/frameType)):

- auxiliary
- top-level
- nested
- none

 I suspect that in the case of an iframe `client.frameType` will equal `"nested"`. 

Edit: I can confirm that a request originating from an iframe has a `client.frameType` of `"nested"`. Of course, that's an entirely different client than the top frame: they have different IDs.

src/mockServiceWorker.js
kettanaito: More specifically, the following exception will be thrown:

```
Uncaught (in promise) DOMException: Failed to execute 'respondWith' on 'FetchEvent': The event handler is already finished.
```

src/setupWorker/start/createStart.ts
kettanaito: The main reason we've added the `clients` object is to give developers control over the mocking on a per-client basis. By calling `worker.stop()` you can disable mocking for the current client.

deshiknaves:
> I'd like to refresh my understanding of this: the issue is that the worker checks if the client is known by looking it up in the `clints` object by client ID. In the case of Cypress, that client ID (in an iframe) is not directly controlled by the worker, thus the lookup for existence fails, and requests originating from Cypress are not intercepted. Is this correct, @deshiknaves?
> 
> I wonder how much information a `client` reference exposes. Maybe instead of manually keeping some shared client ID we could check if the client that originates a request has a parent frame that is controlled by the worker.
> 
> Edit: I can see that the `client` instance has a `frameType` property. In the case of the immediate page it's `"top-level"`. All possible frame types (per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Client/frameType)):
> 
> * auxiliary
> * top-level
> * nested
> * none
> 
> I suspect that in the case of an iframe `client.frameType` will equal `"nested"`.
> 
> Edit: I can confirm that a request originating from an iframe has a `client.frameType` of `"nested"`. Of course, that's an entirely different client than the top frame: they have different IDs.

@kettanaito yes, exactly as your investigation went. The ids aren't the same. I'll just outline a bit more on that:

This is not only an issue for Cypress. This is also an issue if your application has an iframe in it. So though I want this for Cypress, there is also many use cases where this would not work. If you have a look at the unit test for this one, I think it illustrates the problem a bit.

The `clientId` of a child iframe doesn't match that of the parent. So when it looks for the client in the client map `clients[clientId]` it never finds the client and so it doesn't handle the requests.

So the solution here was to be able to say when the server starts up that every `clientId` that connects to it should be intercepted by these mocks.

kettanaito:
I've prepared a test that performs a request from an iframe, and I think I've fixed the issue on the worker's side. Will open a pull request today for you to review and try out, @deshiknaves, if you don't mind.

deshiknaves:
> I've prepared a test that performs a request from an iframe, and I think I've fixed the issue on the worker's side. Will open a pull request today for you to review and try out, @deshiknaves, if you don't mind.

Yep, I'm about to go to sleep. But I'll have a look in 7 hours or so 😊

kettanaito:
Please, always prioritize rest! You've been a great help. Thank you, @deshiknaves!

deshiknaves:
> @kettanaito @marcosvega91 do the tests pass for you currently on the `master` branch? Sorry, I haven't update this in a while, got busy with a few things. A lot of the integration tests are failing for me on the `master` branch.

This was my mistake. I keep forgetting that a `build` needs to be run before the integration test.

Issue #369, opened 2020-09-04
CRA now enforce public url to be used in dev server and this breaks MSW integration.


slowselfip:
## Describe the bug
Recently the react-dev-utils that is part of the react-scripts package used by create react app got an update that makes it so that the public path is used in the dev-server. The mockedServiceWorker will need to be registered with this public url but now it will not intercept any requests.

## Environment

- `msw: 0.20.5`
- `create-react-app: 3.4.1`

> Please also provide your browser version.
chrome 84

## To Reproduce
1. Make a create-react-app project using at least create-react-app 3.4.1.
2. Add msw as dependency and run init to put the mockServiceWorker.js in the public folder.
3. Add a "homepage" : "/foo/bar" property in the package.json file to make it the public path. The app is now hosted on localhost:3000/foo/bar.
3. Add browser.js and handler.js as usual and integrate the msw in the app. 
```js
//src/mocks/handlers.ts
import { rest } from "msw";

const handlers = [
  rest.get(`${process.env.PUBLIC_URL}/test`, (req, res, ctx) =>
    res(ctx.json({ message: "Handled by mock service worker" }))
  ),
];

export { handlers };
```

```js
//src/mocks/browser.ts
import { setupWorker } from 'msw'
import { handlers } from './handlers'

export const worker = setupWorker(...handlers);
```

Remember to add the process.env.PUBLIC_URL to the path of the service worker
```js
//index.tsx
import("./mocks/browser").then(({ worker }) => {
  worker
    ?.start({
      serviceWorker: {
        url: `${process.env.PUBLIC_URL}/mockServiceWorker.js`,
      },
    })
    .then(() => {
      ReactDOM.render(
        <React.StrictMode>
             <button
              onClick={() =>
                fetch(`${process.env.PUBLIC_URL}/test`)
                  .then((response) => response.json())
                  .then((json) => console.log("Got reponse:", json.message))
                  .catch(error => console.error(error))
              }
            >
              FETCH
            </button>
        </React.StrictMode>,
        document.getElementById("root")
      );
    });
});
```
4. Start the app, open developer console and notice the [MSW] Mocking enabled message. 
5. Press the FETCH-button
6. Check the console. There will be an error index.js:1 SyntaxError: Unexpected token < in JSON at position 0 because the mockedServiceWorker did not intercept the request.

## Expected behavior
I would expect to be able to run the application with a PUBLIC_URL and still use MSW

msutkowski:
[...]

PR #384, opened 2020-09-14
Support process.env.PUBLIC_URL in create-react-app

GioSensation:
When CRA is set to render in a subfolder, the worker was not found. This fixes the matter for CRA apps, not sure how other frameworks handle the same use case.

### DIFF
--- a/src/utils/url/getAbsoluteWorkerUrl.ts
+++ b/src/utils/url/getAbsoluteWorkerUrl.ts
@@ -3,5 +3,9 @@
  * relative URL (known during the registration).
  */
 export function getAbsoluteWorkerUrl(relativeUrl: string): string {
-  return new URL(relativeUrl, location.origin).href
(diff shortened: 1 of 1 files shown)


GioSensation:
[This issue](https://github.com/mswjs/msw/issues/369) helped solve the problem. The changes in this PR were not needed.

kettanaito:
Hey, @GioSensation. Thanks for opening this anyway!

Yeap, you can handle custom location of the worker script using the [`serviceWorker.url`](https://mswjs.io/docs/api/setup-worker/start#serviceworker) option of the `worker.start()` function.

We understand that the multiplicity of devstacks is enormous, so we try to come up with an API that would be flexible enough for various use cases. Plus, the ability to provide a custom worker URL and scope is available natively via Service Worker API, there's no need for us to drop it.

Glad you've resolved it!

PR #388, opened 2020-09-17
Request completion notification

deshiknaves:
I've been working on some integration with MSW and Cypress (https://github.com/deshiknaves/cypress-msw-interceptor). In order to be able to notify `Cypress` (or other tools) that a request has been completed (mocked or unmocked), there needs to be a mechanism to inform the consumer that a particular request has been completed. To this end, I've:

- Added a unique id to each request on the `REQUEST` message
- Added a new `messaged` called `REQUEST_COMPLETE` that responds with the `request` and the `response`. This will happen for `MOCK_SUCCESS` and `MOCK_NOT_FOUND`.

### DIFF
--- a/test/msw-api/setup-worker/notifications.mocks.ts
+++ b/test/msw-api/setup-worker/notifications.mocks.ts
@@ -1,0 +1,17 @@
+import { setupWorker, rest } from 'msw'
+
+const worker = setupWorker(
+  rest.get('/user', (req, res, ctx) => {
+    return res(
+      ctx.json({
+        firstName: 'John',
+        age: 32,
+      }),
+    )
+  }),
+)
+
+// @ts-ignore
+window.__MSW_REGISTRATION__ = worker.start({
+  serviceWorker: {},
+})
--- a/test/msw-api/setup-worker/notifications.mocks.ts
+++ b/test/msw-api/setup-worker/notifications.mocks.ts
@@ -12,6 +12,4 @@ const worker = setupWorker(
 )
 
 // @ts-ignore
-window.__MSW_REGISTRATION__ = worker.start({
-  serviceWorker: {},
-})
+worker.start()
--- a/test/msw-api/setup-worker/notifications.test.ts
+++ b/test/msw-api/setup-worker/notifications.test.ts
@@ -43,7 +43,6 @@ afterAll(() => {
 test('notifies when a fetch request initiates/completes with a mocked response', async () => {
   const { messages } = captureConsole(runtime.page)
 
-
   await runtime.reload()
 
   const res = await runtime.request({
--- a/test/msw-api/setup-worker/notifications.test.ts
+++ b/test/msw-api/setup-worker/notifications.test.ts
@@ -1,0 +1,186 @@
+import * as path from 'path'
+import { TestAPI, runBrowserWith } from '../../support/runBrowserWith'
+import { captureConsole } from '../../support/captureConsole'
+
+let runtime: TestAPI
+let postMessages = []
+
+beforeAll(async () => {
+  runtime = await runBrowserWith(
+    path.resolve(__dirname, 'notifications.mocks.ts'),
+  )
+
+  await runtime.page.exposeFunction('onPostMessage', function (evt) {
+    switch (evt.type) {
+      case 'REQUEST':
+      case 'REQUEST_COMPLETE': {
+        // Ignoring static files
+        if (evt.payload.url.match(/.js$/)) return
+        postMessages.push(evt)
+      }
+      default:
+      // Ignore the rest
+    }
+  })
+
+  await runtime.page.evaluateOnNewDocument(() => {
+    navigator.serviceWorker.addEventListener('message', function (message) {
+      const evt = JSON.parse(message.data)
+      // @ts-ignore
+      window.onPostMessage(evt)
+    })
+  })
+})
+
+beforeEach(() => {
+  postMessages = []
+})
+
+afterAll(() => {
+  return runtime.cleanup()
+})
+
+test('notifies when a fetch request initiates/completes with a mocked response', async () => {
+  const { messages } = captureConsole(runtime.page)
+
+  await runtime.page.evaluate(() => {
+    // @ts-ignore
+    return window.__MSW_REGISTRATION__
+  })
+
+  await runtime.reload()
+
+  const res = await runtime.request({
+    url: `${runtime.origin}/user`,
+  })
+
+  const headers = res.headers()
+  const body = await res.json()
+
+  expect(headers).toHaveProperty('x-powered-by', 'msw')
+  expect(body).toEqual({
+    firstName: 'John',
+    age: 32,
+  })
+
+  const requetsLog = messages.startGroupCollapsed.find((text) => {
+    return text.includes('[MSW]') && text.includes('GET /user')
+  })
+
+  expect(requetsLog).toMatch(/\[MSW\] \d{2}:\d{2}:\d{2} GET \/user 200/)
+
+  expect(postMessages).toHaveLength(2)
+  const request = postMessages.find((i) => i.type === 'REQUEST').payload
+  const response = postMessages.find((i) => i.type === 'REQUEST_COMPLETE')
+    .payload
+  const expectedRequest = {
+    url: `${runtime.origin}/user`,
+    method: 'GET',
+    headers: {
+      accept: '*/*',
+      'sec-ch-ua': '',
+      'sec-ch-ua-mobile': '?0',
+    },
+    cache: 'default',
+    mode: 'cors',
+    credentials: 'same-origin',
+    destination: 'empty',
+    integrity: '',
+    redirect: 'follow',
+    referrer: `${runtime.origin}/`,
+    referrerPolicy: 'no-referrer-when-downgrade',
+    body: '',
+    bodyUsed: false,
+    keepalive: false,
+  }
+  expect(request.id.toString()).toMatch(/^\d{3,}$/)
+  expect(request).toMatchObject(expectedRequest)
+  expect(response.request).toMatchObject(expectedRequest)
+  expect(response.response).toMatchObject({
+    status: 200,
+    statusText: 'OK',
+    body: '{"firstName":"John","age":32}',
+    delay: 0,
+    once: false,
+    headers: [
+      ['x-powered-by', 'msw'],
+      ['content-type', 'application/json'],
+    ],
+  })
+})
+
+test('notifies when a fetch request initiates/completes with no mocked response', async () => {
+  const { messages } = captureConsole(runtime.page)
+
+  await runtime.page.evaluate(() => {
+    // @ts-ignore
+    return window.__MSW_REGISTRATION__
+  })
+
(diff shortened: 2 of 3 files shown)


marcosvega91:
Hi @deshiknaves the idea of your integration project is cool. Great work !!

What is your use case for this feature?

deshiknaves:
@marcosvega91 once a request is mocked, it is often useful for the runner to wait for that response to complete before asserting that something has happened in the UI. Instead of waiting for some arbitrary time, it can wait for the request to complete.

Also, if the user has access to the `response`, then they can make sure that the UI is showing what was returned from the request and not hardcoded to some values. This enables a test to be able to run both as an integration and an e2e test.

I have a few examples in the readme of that repository, but for brevity, here is an example that outlines all the different scenarios:

```javascript
// When `cy.interceptRequest` doesn't receive a response resolver, it will make the request to the real endpoint
// but will still be able to wait on that request
function shouldMockResponse(fn) {
  return process.env.CYPRESS_E2E === 'true' ? fn : undefined
}

it('should be able to get the response and check that the correct text is displayed', () => {
  cy.visit('/')
  // This creates a mock with a response resolver
  cy.interceptRequest(
    'GET',
    shouldMockResponse((req, res, ctx) => {
      return res(
        ctx.delay(1000),
        ctx.json({
          userId: 1,
          id: 1,
          title: 'Lord of the rings',
          completed: false,
        }),
      )
    }),
  ).as('todos')

  // This waits for the completion and returns the response
  cy.waitForRequest('@todos').then(({ request, response }) => {
    cy.findByText(new RegExp(response.body.title, 'i')).should('be.visible')
  })
})
```

marcosvega91:
Hi @deshiknaves It makes sense but I don't know if is really useful. Because of response are mocked you already know what ii will be, so you can make assertion on it. I want to hear an opinion by @kettanaito. What do you think?

deshiknaves:
You don't for requests you haven't mocked. And even ones you have mocked being able to have the same API is quite useful. You also might have different response from the same resolver depending on what's in the request. So without knowing when something has completed and with what response, it's impossible to do a lot of things that is normally done in Cypress.

deshiknaves:
Along with that, there is currently no way to know when a request has complete and what the request is was.

marcosvega91:
Now is more clear, I didn't thought about integration tests. For me is ok. I'll take a look at your code. Thanks again 👍

test/msw-api/setup-worker/notifications.mocks.ts
```diff
@@ -0,0 +1,17 @@
...
+        age: 32,
+      }),
+    )
+  }),
+)
+
+// @ts-ignore
+window.__MSW_REGISTRATION__ = worker.start({
+  serviceWorker: {},
+})
```
marcosvega91: I think that i can be 

```suggestion
worker.start()
```

test/msw-api/setup-worker/notifications.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  return runtime.cleanup()
+})
+
+test('notifies when a fetch request initiates/completes with a mocked response', async () => {
+  const { messages } = captureConsole(runtime.page)
+
+  await runtime.page.evaluate(() => {
+    // @ts-ignore
+    return window.__MSW_REGISTRATION__
+  })
```
marcosvega91: this can be removed 

```suggestion
```

marcosvega91 (changes requested):
For me looks good, I want to hear other opinion on it. I leave only little suggestions

deshiknaves:
Just wondering if you had any thoughts on this?

deshiknaves:
These have been open for a while. Please let me know what I can do to progress them 😊

deshiknaves:
Looks like something in the recent changes broke something. I'll dig into this.

Update: Hmm looks like master isn't passing the tests either.

kettanaito (commented):
Thanks for preparing this, @deshiknaves.

I've got a few questions:

1. What is the use case to react to a request completion on such low level? Consider that most requests are reacted to as a part of the app's logic (i.e. UI update, side effect). Exposing such low-level listener mechanism is too bound to implementation and also lies beyond MSW responsibilities.

2. Have you considered built-in network interception tools of Cypress? I know it only supports `XMLHttpRequests` as of now, but its lack of support shouldn't be compensated by other tools. You can use MSW instead that API, if it's limiting, and react to requests by UI assertions, instead of listening to internal worker events.

I have a feeling that while this implementation is technically possible, it complicates MSW by expanding its responsibilities. It almost feels like it's a workaround for an issue in on entirely different surface.

kettanaito:
The master wasn't passing tests due to Cypress binary download timeouting. In case of this pull request you have a failing integration test:

```
Summary of all failing tests
 FAIL  msw-api/setup-worker/stop.test.ts (7.901 s)
  ● disables the mocking when the worker is stopped

    Protocol error (Network.getResponseBody): Target closed.

      at ../node_modules/puppeteer/lib/cjs/puppeteer/common/Connection.js:208:63
      at CDPSession.send (../node_modules/puppeteer/lib/cjs/puppeteer/common/Connection.js:207:16)
      at ../node_modules/puppeteer/lib/cjs/puppeteer/common/HTTPResponse.js:99:53
          at runMicrotasks (<anonymous>)
      at HTTPResponse.text (../node_modules/puppeteer/lib/cjs/puppeteer/common/HTTPResponse.js:111:25)
      at HTTPResponse.json (../node_modules/puppeteer/lib/cjs/puppeteer/common/HTTPResponse.js:124:25)
```

Could you please take a look into it? Run that test locally:

```bash
$ yarn test:integration msw-api/setup-worker/stop.test.ts
```

You can also run that test usage example in a browser for debugging:

```bash
$ yarn test:focused msw-api/setup-worker/stop.test.ts
```

deshiknaves:
@kettanaito did you by any chance have a look at the GitHub repo in the description? The readme has an extensive explanation as to why.

kettanaito:
Sorry, I've overlooked that. Will take a look and update the thread, thanks.

kettanaito:
If I understand correctly, the idea is to react to `MOCK_SUCCESS` and `MOCK_NOT_FOUND` internal events. With this intention we can expose a channel we use internally to send those events to the worker. Something like this:

```js
worker.start().then((registration, channel) => {
  channel.addEventListener('MOCK_SUCCESS', () => {
    // handle the "MOCK_SUCCESS" event
  })
}))
```

I'm speaking about this `channel`:

https://github.com/mswjs/msw/blob/327b384938b1fdf44e77c5bd79e91b7fab0aaad4/src/utils/createBroadcastChannel.ts#L16-L32

Used here to send events to the worker:

https://github.com/mswjs/msw/blob/327b384938b1fdf44e77c5bd79e91b7fab0aaad4/src/utils/handleRequestWith.ts#L24

Would this work for your use case? What caveats do you see with this approach?

kettanaito:
What I can spot is that using the listeners on `channel` won't let you access the original response in case of `MOCK_NOT_FOUND`. That's not necessarily a bad thing, but something to keep in mind.

deshiknaves:
@kettanaito thanks for pointing me to that. That will work to some extent, but in order for the e2e tests to be able to get access to every request it needs to have access to MOCK_NOT_FOUND and its response. I will check if there is some way to enable this using similar mechanism.

I'll also sort out the broken test.

kettanaito:
My main concern is that it's not a common use case, which goes against putting it into the worker script by default. If we could achieve the expected behavior via client-side API then I'm totally for it, as then we'd have control of it being opt-in. When it comes to the worker script the philosophy is to keep it as concise and straightforward as possible, which may imply omitting certain suggestions. 

> Note that it's also perfectly fine to modify the worker script for whichever purpose. The price for that is to ensure compatibility between worker script updates issued by the library. You can also [use multiple workers](https://github.com/kettanaito/service-worker-multiple) at once, but you can only handle the "fetch" event by one of them, as far as I know. 

I can suggest to dig into Cypress and how it uses the underlying browser. Puppeteer, for example, provides request/response interception in its API, and I can only assume they do that in some low-level fashion. Perhaps if Cypress has an ability to do something similar on the browser level, you could react to responses without touching the worker.

deshiknaves:
Yep, I can totally see that. The reason why I'm going down this path is that Cypress's own layer isn't as flexible as the MSW API.

I can see 2 options:
1) pass in an option which will notify using the channel mentioned above for MOCK_NOT_FOUND. If the option isn't passed in, then it will continue as it always was. The only addition becomes the clone and notification.
2) maintain a copy of the service worker. I can do that, but it's not particularly ideal as this is being actively developed here. However, if a start up option for returning the completion notification isn't acceptable, then that's the only viable option.

I'll be working on this in a few hours. I'll see how much the API changes can be reduced. Thought it is only doing 2 things:
- add a unique identifier to each request
- clone the response and notify. This can easily be converted into a start up option to be opt in.

deshiknaves:
> The master wasn't passing tests due to Cypress binary download timeouting. In case of this pull request you have a failing integration test:
> 
> ```
> Summary of all failing tests
>  FAIL  msw-api/setup-worker/stop.test.ts (7.901 s)
>   ● disables the mocking when the worker is stopped
> 
>     Protocol error (Network.getResponseBody): Target closed.
> 
>       at ../node_modules/puppeteer/lib/cjs/puppeteer/common/Connection.js:208:63
>       at CDPSession.send (../node_modules/puppeteer/lib/cjs/puppeteer/common/Connection.js:207:16)
>       at ../node_modules/puppeteer/lib/cjs/puppeteer/common/HTTPResponse.js:99:53
>           at runMicrotasks (<anonymous>)
>       at HTTPResponse.text (../node_modules/puppeteer/lib/cjs/puppeteer/common/HTTPResponse.js:111:25)
>       at HTTPResponse.json (../node_modules/puppeteer/lib/cjs/puppeteer/common/HTTPResponse.js:124:25)
> ```
> 
> Could you please take a look into it? Run that test locally:
> 
> ```shell
> $ yarn test:integration msw-api/setup-worker/stop.test.ts
> ```
> 
> You can also run that test usage example in a browser for debugging:
> 
> ```shell
> $ yarn test:focused msw-api/setup-worker/stop.test.ts
> ```

I just reinstalled `node_modules` the master branch is up to date with upstream. A lot of tests are failing in that branch. And that test is failing there too. I'll investigate further and get back.

c957b28bfd760dba56f5fedb0d846ec67ce27d84 — That's the last commit that passes `test:integration`

kettanaito:
Hey, @deshiknaves. May I please ask for your opinion on the lifecycle events we are currently working on (#482)?

The aim is to cover third-party scenarios that depend on the library's internal events flow (i.e. request started, request completed). Currently, that pull request introduces a `response:mocked` and `response:bypass` events, as well as `request:end` event. 

I think it can cover this effort as well. I'd like to close this pull request in favor of lifecycle events. I'd appreciate your feedback in doing so.

deshiknaves:
@kettanaito I'm closing this because #482 replaces this

kettanaito:
You can now use [Life-cycle events API](https://mswjs.io/docs/extensions/life-cycle-events) to know when a certain request is finished. Use the `request:end` event for that.

PR #402, opened 2020-09-27
fix: check setupWorker and setupServers args

### LINKED ISSUE #400: Always receiving TypeError: Network request failed
## Describe the bug
In my jest tests, if I call `server.listen()` all of my network requests inside of the test fail with the generic fetch error `TypeError: Network request failed`. These requests succeed if I don't attempt to run msw at all.

The failure occurs for EVERY request, regardless of whether or not msw has a handler that matches the request or not.

My test looks something like this:
```
import { rest } from "msw";

const handlers = [
  // behavior is the same regardless of whether or not I have any handlers defined
];

const server = setupServer(handlers);
describe("Some Thing", () => {
  beforeAll(() => server.listen());
  afterEach(() => server.resetHandlers());
  afterAll(() => server.close());

  test('example test', () => {
    await fetch(
      "https://api.fake.com/something/i/know/works",
      {
        method: "POST",
        headers: {
          "Content-Type": "application/json",
        },
        body: JSON.stringify({ foo: "bar" }),
      }
    );
  })
}
```

I am adding react-testing-library tests (using msw on the recommendation of KCD) to an older react application. The app was created with CRA ~3.5 years ago. I've updated react-scripts to a more recent version just in case this was the cause, but I've had no luck.

I think it's likely that either my configuration is wrong, or that I have a dependency problem, but the error from fetch is so opaque that I'm finding myself stuck. I don't know if there are any debug tools available to inspect or trace the path of this these failing requests to find out where they're getting jammed up. 

## Environment
My react application sits in a yarn workspace monorepo, but when I isolate the app from its "siblings" and just do `npm i && npm test` I see the same behavior.

- `msw: 0.21.2`
- `nodejs: 10.18.1` & `nodejs: 12.16.3`
- `npm: 6.14.4`
- `yarn: 1.22.4`
- `react-scripts: 3.2.0` which installs `jest: 24.9.0`
- I run my tests with `jest-environment-jsdom-sixteen: 1.0.3`

## To Reproduce
I'm not yet sure how to reproduce. I'm not even sure if this is a real bug. I think it's most likely a misconfiguration on my part or a weird dependency issue.

## Expected behavior
I expect that network requests are either handled by msw handlers or are passed through to be handled by the network.

marcosvega91:
Fix #400 

This PR will throw an error if the arg is an array. Args of the two functions make confusion sometime. Is better to inform the user which is the right way.

### DIFF
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -32,7 +32,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
     requestHandlers.forEach((handler) => {
       if (Array.isArray(handler))
         throw new Error(
-          `[MSW] Failed to call "setupServer": received a list of request handlers (setupServer([a, b])), expected to receive each handler individually: setupServer(a, b).`,
+          `[MSW] Failed to call "setupServer" given an Array of request handlers (setupServer([a, b])), expected to receive each handler individually: setupServer(a, b).`,
         )
     })
     const interceptor = new RequestInterceptor(interceptors)
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -29,6 +29,12 @@ export function createSetupServer(...interceptors: Interceptor[]) {
    * Sets up a server-side requests interception with the given mock definition.
    */
   return function setupServer(...requestHandlers: RequestHandlersList) {
+    requestHandlers.forEach((handler) => {
+      if (Array.isArray(handler))
+        throw new Error(
+          `[MSW] setupServer function receive every handler as an arg. You should call it as setupServer(...requestHandlers) with requestHandlers the array of handlers.`,
+        )
+    })
     const interceptor = new RequestInterceptor(interceptors)
 
     // Error when attempting to run this function in a browser environment.
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -32,7 +32,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
     requestHandlers.forEach((handler) => {
       if (Array.isArray(handler))
         throw new Error(
-          `[MSW] setupServer function receive every handler as an arg. You should call it as setupServer(...requestHandlers) with requestHandlers the array of handlers.`,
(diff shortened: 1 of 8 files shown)


test/msw-api/setup-worker/setupWorker.mocks.ts
```diff
@@ -0,0 +1,10 @@
+import { setupWorker, rest } from 'msw'
+
+// @ts-ignore
```
kettanaito: Can you please add a comment that this is an intentionally invalid usage?

test/msw-api/setup-worker/setupWorker.test.ts
```diff
@@ -0,0 +1,25 @@
+import * as path from 'path'
```
kettanaito: Lets rename this test suite so it’s clear what it does: `input-validation.test.ts`.

src/node/createSetupServer.ts
```diff
@@ -29,6 +29,12 @@ export function createSetupServer(...interceptors: Interceptor[]) {
    * Sets up a server-side requests interception with the given mock definition.
    */
   return function setupServer(...requestHandlers: RequestHandlersList) {
+    requestHandlers.forEach((handler) => {
+      if (Array.isArray(handler))
+        throw new Error(
+          `[MSW] setupServer function receive every handler as an arg. You should call it as setupServer(...requestHandlers) with requestHandlers the array of handlers.`,
```
kettanaito: - please don’t use contractions (args)
- we should include the wrong usage example, presumably how the user tried to call this function.
- we shouldn’t assume the user knows about the `requestHandlers` list (the name is rather internal)

```
[MSW] Failed to call “setupServer”: received a list of request handlers (setupServer([a, b])), expected to receive each handler individually: setupServer(a, b).

kettanaito (changes requested):

marcosvega91:
I have updated the repo, sorry for this stupid mistakes 😅

kettanaito (approved):
Thank you for adding this, @marcosvega91! I've retouched the `setupServer` test a little and change the terminology slightly in the error message. Let's wait for the CI.

marcosvega91:
thank you @kettanaito 😃

Issue #401, opened 2020-09-26
Support for GraphQL requests with partial results


tdeekens:
Thanks for this great library! It's a pleasure to use. While integrating it I ran into an issue which might be caused by lack of understanding on my side.

**Is your feature request related to a problem? Please describe.**

When using GraphQL requests which for instance load different parts of data such as

```gql
query MyQuery {
   myOrders {
      total
   }
   myProfile {
      name
   }
}
```

then this query can partially fail. In the example above the `myOrders` could be resolved while the `myProfile` can fail and be `null`. The resulting response could be something like

```js
{
   data: {
     myOrders: {
        total: 42
     },
     myProfile: null
   },
   errors: [{
      path: 'MyQuery.myProfile',
      extensions: { code: 'foo' }
   }]
}
```

I don't see a way to support this case via `msw` and the `graphql` mocking system. The `errors` context to how I read it's code terminates the request. While the `data` context seems to be a shorthand for `json.data()`. As a result I can't use both as in `ctx.data` and `ctx.errors`. While `ctx.json` is not exposed when using `graphql`. Mocking the request via `rest` feels cumbersome as matching the query is hard.

**Describe the solution you'd like**

Either the `errors` on `ctx` does not terminate the request and allows for combination with `data` or `json` is exposed on the context when using `graphql`.

emmenko:
Or we could introduce a new context method like `partial({ data, errors })`.

kettanaito:
Hey!
`ctx.errors` does not terminate the request. The thing is that both `ctx.data` and `ctx.errors` use `ctx.json` internally, which strictly overrides the mocked response body. So the next utility overrides the previous one, making it impossible to combine data and errors.

This is something we need to cover. Perhaps, `ctx.json` could accept an `extend` argument internally to extend the JSON body. Needs discussion.

tdeekens:
> ctx.errors does not terminate the request.

Thanks for the clarification. My bad I must've read that wrong somewhere. First time reading the code base. Which is very nice structured by the way.

I assume we have three options at the moment:

1. Exposing `json` in a `graphql` context so that both `data` and `errors` can be set
2. Allowing `json` to have a `merge` option so that `errors` doesn't overwrite `data` and vice-versa
3. Have a `partial` context which feels a bit like 2.

Really eager to hear what others think :)

emmenko:
I'm not sure exposing the `json` method is what we want here.

[...]

PR #403, opened 2020-09-27
feat: add support for composing data and errors in graphql

tdeekens:
This pull request implements one suggestion (without recommending it specifically discussed here https://github.com/mswjs/msw/issues/401.

It thereby implements  allowing `json` to use a `json({}, { merge: true })` option which `data` and `errors` of the `graphql` context can use. Internally it parses the prior `res.body` and shallowly merges the new `BodyType` in.

This internally is used by `data` and `errors` to support composing both with each other.

### DIFF
--- a/src/context/data.test.ts
+++ b/src/context/data.test.ts
@@ -3,19 +3,39 @@ import { errors } from './errors'
 import { response } from '../response'
 
 describe('data', () => {
-  describe('given a JSON data', () => {
-    let result: ReturnType<typeof response>
+  describe('given JSON data', () => {
+    describe('with specifying data once', () => {
+      let result: ReturnType<typeof response>
 
-    beforeAll(() => {
-      result = response(data({ name: 'msw' }))
+      beforeAll(() => {
+        result = response(data({ name: 'msw' }))
+      })
+
+      it('should have "Content-Type" as "application/json"', () => {
+        expect(result.headers.get('content-type')).toEqual('application/json')
+      })
+
+      it('should have body set to the given JSON nested in the "data" property', () => {
+        expect(result).toHaveProperty('body', `{"data":{"name":"msw"}}`)
+      })
     })
+    describe('with specifying data multiple times', () => {
+      let result: ReturnType<typeof response>
 
-    it('should have "Content-Type" as "application/json"', () => {
-      expect(result.headers.get('content-type')).toEqual('application/json')
-    })
+      beforeAll(() => {
+        result = response(data({ name: 'msw' }), data({ is: 'great' }))
+      })
 
-    it('should have body set to the given JSON nested in the "data" property', () => {
-      expect(result).toHaveProperty('body', `{"data":{"name":"msw"}}`)
+      it('should have "Content-Type" as "application/json"', () => {
+        expect(result.headers.get('content-type')).toEqual('application/json')
+      })
+
+      it('should have body set to the given JSON nested in the "data" property merging each "data" call', () => {
+        expect(result).toHaveProperty(
+          'body',
+          `{\"data\":{\"name\":\"msw\",\"is\":\"great\"}}`,
+        )
+      })
     })
   })
   describe('given composed with error', () => {
--- a/src/context/data.test.ts
+++ b/src/context/data.test.ts
@@ -2,57 +2,60 @@ import { data } from './data'
 import { errors } from './errors'
 import { response } from '../response'
 
-describe('data', () => {
-  describe('given JSON data', () => {
-    describe('with specifying data once', () => {
-      let result: ReturnType<typeof response>
+test('sets a single data on the response JSON body', () => {
+  const result = response(data({ name: 'msw' }))
 
-      beforeAll(() => {
-        result = response(data({ name: 'msw' }))
-      })
+  expect(result.headers.get('content-type')).toBe('application/json')
+  expect(result).toHaveProperty(
+    'body',
+    JSON.stringify({
+      data: {
+        name: 'msw',
+      },
+    }),
+  )
+})
 
-      it('should have "Content-Type" as "application/json"', () => {
-        expect(result.headers.get('content-type')).toEqual('application/json')
-      })
+test('sets multiple data on the response JSON body', () => {
+  const result = response(
+    data({ name: 'msw' }),
+    data({ description: 'API mocking library' }),
+  )
 
-      it('should have body set to the given JSON nested in the "data" property', () => {
-        expect(result).toHaveProperty('body', `{"data":{"name":"msw"}}`)
-      })
-    })
-    describe('with specifying data multiple times', () => {
-      let result: ReturnType<typeof response>
+  expect(result.headers.get('content-type')).toBe('application/json')
+  expect(result).toHaveProperty(
+    'body',
+    JSON.stringify({
+      data: {
+        description: 'API mocking library',
+        name: 'msw',
+      },
+    }),
+  )
+})
 
-      beforeAll(() => {
-        result = response(data({ name: 'msw' }), data({ is: 'great' }))
-      })
+test('combines with error in the response JSON body', () => {
+  const result = response(
+    data({ name: 'msw' }),
+    errors([
+      {
+        message: 'exceeds the limit of awesomeness',
+      },
+    ]),
+  )
 
-      it('should have "Content-Type" as "application/json"', () => {
-        expect(result.headers.get('content-type')).toEqual('application/json')
-      })
-
-      it('should have body set to the given JSON nested in the "data" property merging each "data" call', () => {
-        expect(result).toHaveProperty(
-          'body',
-          `{\"data\":{\"name\":\"msw\",\"is\":\"great\"}}`,
-        )
-      })
-    })
-  })
-  describe('given composed with error', () => {
-    let result: ReturnType<typeof response>
-
-    beforeAll(() => {
-      result = response(
-        data({ name: 'msw' }),
-        errors([{ message: 'is great' }]),
-      )
-    })
-
-    it('should have body set to the given JSON nested in the "data" property', () => {
-      expect(result).toHaveProperty(
-        'body',
-        `{\"data\":{\"name\":\"msw\"},\"errors\":[{\"message\":\"is great\"}]}`,
-      )
-    })
-  })
+  expect(result.headers.get('content-type')).toBe('application/json')
+  expect(result).toHaveProperty(
+    'body',
+    JSON.stringify({
+      errors: [
+        {
+          message: 'exceeds the limit of awesomeness',
+        },
+      ],
+      data: {
+        name: 'msw',
+      },
+    }),
+  )
 })
--- a/src/context/data.test.ts
+++ b/src/context/data.test.ts
@@ -1,4 +1,5 @@
 import { data } from './data'
+import { errors } from './errors'
 import { response } from '../response'
 
 describe('data', () => {
@@ -15,6 +16,23 @@ describe('data', () => {
 
     it('should have body set to the given JSON nested in the "data" property', () => {
       expect(result).toHaveProperty('body', `{"data":{"name":"msw"}}`)
+    })
+  })
+  describe('given composed with error', () => {
+    let result: ReturnType<typeof response>
+
+    beforeAll(() => {
+      result = response(
+        data({ name: 'msw' }),
+        errors([{ message: 'is great' }]),
+      )
+    })
+
+    it('should have body set to the given JSON nested in the "data" property', () => {
+      expect(result).toHaveProperty(
+        'body',
+        `{\"errors\":[{\"message\":\"is great\"}],\"data\":{\"name\":\"msw\"}}`,
+      )
     })
   })
 })
--- a/src/context/data.test.ts
+++ b/src/context/data.test.ts
@@ -31,7 +31,7 @@ describe('data', () => {
     it('should have body set to the given JSON nested in the "data" property', () => {
       expect(result).toHaveProperty(
         'body',
-        `{\"errors\":[{\"message\":\"is great\"}],\"data\":{\"name\":\"msw\"}}`,
+        `{\"data\":{\"name\":\"msw\"},\"errors\":[{\"message\":\"is great\"}]}`,
       )
     })
   })
--- a/src/context/data.ts
+++ b/src/context/data.ts
@@ -1,4 +1,5 @@
 import { ResponseTransformer } from '../response'
+import { json } from './json'
 
 export type DataContext<T> = (payload: T) => ResponseTransformer
 
@@ -6,9 +7,5 @@ export type DataContext<T> = (payload: T) => ResponseTransformer
  * Returns a GraphQL body payload.
  */
 export const data: DataContext<Record<string, any>> = (payload) => {
-  return (res) => {
-    res.headers.set('Content-Type', 'application/json')
-    res.body = JSON.stringify({ data: payload })
-    return res
-  }
+  return json({ data: payload }, { merge: true })
 }
--- a/src/context/errors.test.ts
+++ b/src/context/errors.test.ts
@@ -2,76 +2,61 @@ import { errors } from './errors'
 import { data } from './data'
 import { response } from '../response'
 
-describe('errors', () => {
-  describe('given at list of errors', () => {
-    let result: ReturnType<typeof response>
+test('sets a given error on the response JSON body', () => {
+  const result = response(errors([{ message: 'Error message' }]))
 
-    beforeAll(() => {
-      result = response(errors([{ message: 'Error message' }]))
-    })
+  expect(result.headers.get('content-type')).toEqual('application/json')
+  expect(result).toHaveProperty(
+    'body',
+    JSON.stringify({
+      errors: [
+        {
+          message: 'Error message',
+        },
+      ],
+    }),
+  )
+})
 
-    it('should have "Content-Type" as "application/json"', () => {
-      expect(result.headers.get('content-type')).toEqual('application/json')
-    })
+test('sets given errors on the response JSON body', () => {
+  const result = response(
+    errors([{ message: 'Error message' }, { message: 'Second error' }]),
+  )
 
-    it('should have body set to the given JSON nested in the "errors" property', () => {
-      expect(result).toHaveProperty(
-        'body',
-        JSON.stringify({
-          errors: [
-            {
-              message: 'Error message',
-            },
-          ],
-        }),
-      )
-    })
-  })
-  describe('given specifying errors multiple times', () => {
-    let result: ReturnType<typeof response>
+  expect(result.headers.get('content-type')).toEqual('application/json')
+  expect(result).toHaveProperty(
+    'body',
+    JSON.stringify({
+      errors: [
+        {
+          message: 'Error message',
+        },
+        {
+          message: 'Second error',
+        },
+      ],
+    }),
+  )
+})
 
-    beforeAll(() => {
-      result = response(
-        errors([{ message: 'First error message' }]),
-        errors([{ message: 'Second error message' }]),
-      )
-    })
+test('combines with data in the response JSON body', () => {
+  const result = response(
+    data({ name: 'msw' }),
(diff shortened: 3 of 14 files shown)


src/context/json.test.ts
```diff
@@ -36,5 +36,28 @@ describe('json', () => {
...
+
+    it('should allow merging with a prior body', () => {
+      const firstNonNestedObject = json({ firstName: 'John' }, { merge: true })
+      const secondNonNestedObject = json({ lastName: 'Doe' }, { merge: true })
+
+      expect(
+        response(firstNonNestedObject, secondNonNestedObject),
+      ).toHaveProperty('body', '{"lastName":"Doe","firstName":"John"}')
+
+      const firstNestedObject = json(
```
tdeekens: Just a test to verify that we don't deep merge things.

src/context/errors.test.ts
```diff
@@ -26,4 +27,21 @@ describe('errors', () => {
...
+      result = response(
+        data({ name: 'msw' }),
+        errors([{ message: 'is great' }]),
+      )
+    })
+
+    it('should have body set to the given JSON nested in the "data" property', () => {
+      expect(result).toHaveProperty(
+        'body',
+        `{\"errors\":[{\"message\":\"is great\"}],\"data\":{\"name\":\"msw\"}}`,
```
tdeekens: This is what we're after in the end. I tested it both in the `data` and `errors` context as both support it and do strictly are not concerned how it is achieved underneath them.

src/context/json.ts
```diff
@@ -1,5 +1,13 @@
 import { ResponseTransformer } from '../response'
 
+type JSONContextOptions = {
+  merge?: boolean
+}
+
+const defaultJSONContextOptions: JSONContextOptions = {
```
tdeekens: I just find this easier to read when having defaults on top a function.

src/context/json.ts
```diff
@@ -9,10 +17,23 @@ import { ResponseTransformer } from '../response'
  */
 export const json = <BodyType>(
   body: BodyType,
+  {
+    merge = defaultJSONContextOptions.merge,
```
tdeekens: So that if we allow more options in the future and one is `false` we don't overwrite it with others.

src/context/json.ts
```diff
@@ -9,10 +17,23 @@ import { ResponseTransformer } from '../response'
...
+  }: JSONContextOptions = defaultJSONContextOptions,
 ): ResponseTransformer<BodyType> => {
   return (res) => {
     res.headers.set('Content-Type', 'application/json')
-    res.body = JSON.stringify(body) as any
+
+    if (merge) {
+      try {
+        const previousBody = JSON.parse(res.body as string)
```
tdeekens: I think `BodyType` is typed wrong internally in the library from what I can tell. The `BodyType` is the generic of what is passed in as an e.g. object. It's then put as a JSON string on the `response`. The response typing however states that `res.body` is a `BodyType` while it is a stringified version of it. That's why there was the `as any` and we now need the `as string` to my understanding.

It is also why CI fails. TypeScript would like us to convert to an `unknown` first. Should be, if the assumption is correct, typing issue too?

src/context/data.ts
```diff
@@ -1,14 +1,11 @@
...
 /**
  * Returns a GraphQL body payload.
  */
 export const data: DataContext<Record<string, any>> = (payload) => {
-  return (res) => {
-    res.headers.set('Content-Type', 'application/json')
-    res.body = JSON.stringify({ data: payload })
-    return res
-  }
+  return json({ data: payload }, { merge: true })
```
tdeekens, line 10: Not sure why this wasn't using `json` just as `errors`. A likely worthwhile change regardless.

src/context/data.test.ts
```diff
@@ -17,4 +18,21 @@ describe('data', () => {
       expect(result).toHaveProperty('body', `{"data":{"name":"msw"}}`)
     })
   })
+  describe('given composed with error', () => {
+    let result: ReturnType<typeof response>
+
+    beforeAll(() => {
+      result = response(
+        data({ name: 'msw' }),
```
tdeekens: What won't work with shallow merging in `json` here is

```js
res(
  data({ foo: 'bar'}),
  data({ bar: 'bar'})
)
```

as it arrives at `json` as `data` which is one level nested meaning here we would only get `{ bar: 'bar }` has that's how `json` merges. Just for awareness.

tdeekens:
Any updates here. Should I be doing something different? How can I help to get this out there if we need it :)

src/context/data.ts
kettanaito: Most likely for historical reasons. I agree with your refactor here: it should reuse the `json` utility.

src/context/data.test.ts
kettanaito: That is incorrect behavior. We should merge _deeply_, with the latest `data` call taking priority. 

```js
res(data({ a: 1 }), data({ b: 2, c: { d: 3 } }))
// { data: { a: 1, b: 2, c: { d: 3 } } }
```

Using the existing `mergeRight` utility already does the deep merging. Let's reflect that in test suites.

src/context/data.test.ts
```diff
@@ -17,4 +18,21 @@ describe('data', () => {
       expect(result).toHaveProperty('body', `{"data":{"name":"msw"}}`)
     })
   })
+  describe('given composed with error', () => {
```
kettanaito: Need to add a test when using multiple `data()` utilities and assert that they merge the given payloads into one.

src/context/errors.test.ts
```diff
@@ -26,4 +27,21 @@ describe('errors', () => {
       )
     })
   })
+  describe('given composed with data', () => {
```
kettanaito: Need to also add a test where we use multiple `errors()` utilities and assert that they produce a list of merged errors.

src/context/json.ts
kettanaito: This is perhaps an optimization done too early. Let's utilize the default argument values for now, and if we get to supporting more options we'd adopt the approach you suggested.

```js
const json = (body, options = { merge: false }) => {}
```

src/context/json.ts
```diff
@@ -7,12 +16,27 @@ import { ResponseTransformer } from '../response'
...
+): ResponseTransformer<BodyTypeString> => {
   return (res) => {
     res.headers.set('Content-Type', 'application/json')
-    res.body = JSON.stringify(body) as any
+
+    if (merge) {
+      try {
+        const nextBody = JSON.parse(res.body)
+
+        res.body = JSON.stringify(mergeRight(body, nextBody)) as BodyTypeString
```
kettanaito: Seeing how tedious it becomes to operate with a JSON body I think we should drop its stringification within the `json` utility and instead do so on the `res()` chain itself. Then we won't have to parse and stringify the body back and forth, making the code simpler.

```js
const json = (body, options) => {
  return (res) => {
    res.headers.set('Content-Type', 'application/json')
    // No need to parse or stringify
    res.body = options.merge ? mergeRight(res.body, body) : body
    return res
  }
}
```

kettanaito (changes requested):
Hey, @tdeekens! Thank you for preparing these changes. I think they look awesome.

I've left a few comments, could you please take a look and address those? Everything is open to discussion, as always. Looking forward to seeing this feature land :)

src/context/data.test.ts
tdeekens: Thanks. Will add a spec.

src/context/errors.test.ts
tdeekens: Thanks. Will add a spec.

src/context/json.ts
tdeekens: That sounds nice @kettanaito can you maybe point me where the `res() chain itself` resides and give me a hint please. I struggle to find my way around.

src/context/json.ts
kettanaito: I think you're right, the type may be wrong. Does it work with `res.body` typed as `string`?

src/context/json.ts
kettanaito: Of course. Here's where the `res` is defined:

https://github.com/mswjs/msw/blob/2b133436cd405a9ebf6d6882d5b8e47997b943f6/src/response.ts#L40-L56

The `resolvedResponse` is the final instance of response composed by it and sent to the worker. I think we can stringify the response body before the return statement. Just pay attention if this is a safe thing to do for all `Content-Type` of responses (i.e. should we stringify body for `text/plain` response?).

tdeekens: Got ya, cheers. I think we should start with just `JSON.stringifying` the JSON Content-type. This will also solve the issue of `json` returning a `string` which caused some typing annoyances (e.g. the old `any` type). Will look into it soon.

kettanaito: Sounds reasonable. Keep in mind there are multiple JSON content-types, we may check them as something like `res.headers.get('content-type')?.endsWith('json')`.

tdeekens: Here would be my first propsal: 7bf236d. I will comment also inline.

src/context/json.ts
```diff
@@ -12,23 +12,17 @@ type JSONContextOptions = {
  * res(json('Some string'))
  * res(json([1, '2', false, { ok: true }]))
  */
-export const json = <BodyTypeJSON, BodyTypeString extends string>(
+export const json = <BodyTypeJSON>(
   body: BodyTypeJSON,
   { merge = false }: JSONContextOptions = {},
-): ResponseTransformer<BodyTypeString> => {
+): ResponseTransformer<BodyTypeJSON> => {
```
tdeekens, line 18: This removes any typing quirk I'd hope.

src/context/json.ts
```diff
@@ -12,23 +12,17 @@ type JSONContextOptions = {
...
 
     if (merge) {
-      try {
-        const nextBody = JSON.parse(res.body)
-
-        res.body = JSON.stringify(mergeRight(body, nextBody)) as BodyTypeString
-      } catch (e) {
-        res.body = JSON.stringify(body) as BodyTypeString
-      }
+      res.body = mergeRight(body, res.body || {})
```
tdeekens: The `|| {}` is only for any first merge to avoid an `if`.

src/response.ts
```diff
@@ -34,6 +34,14 @@ export const defaultResponse: Omit<MockedResponse, 'headers'> = {
   once: false,
 }
 
+const postProcessJSONBody = <T = Partial<MockedResponse>>(response: T): T => {
```
tdeekens: Would also be a tranform I guess we just compose in. Let me know what you think.

src/context/errors.ts
```diff
@@ -5,8 +5,6 @@ import { json } from './json'
 /**
  * Returns a list of GraphQL errors.
  */
-export const errors = (
-  errorsList: Partial<GraphQLError>[],
-): ResponseTransformer<{ errors: typeof errorsList }> => {
-  return json({ errors: errorsList })
+export const errors = (errorsList: Partial<GraphQLError>[]) => {
```
kettanaito: We need to be careful in dropping the `ResponseTransformer` return type here. It may cause some unexpected type issues when using `ctx.errors` in combination with other utils.

src/context/json.test.ts
```diff
@@ -36,5 +36,33 @@ describe('json', () => {
         '"2020-01-01T00:00:00.000Z"',
       )
     })
+
+    it('should allow merging with a prior body', () => {
+      const firstNonNestedObject = json({ firstName: 'John' }, { merge: true })
+      const secondNonNestedObject = json({ lastName: 'Santa' }, { merge: true })
+      const thirdNonNestedObject = json({ lastName: 'Doe' }, { merge: true })
```
kettanaito: I think we can limit this to 2 bodies in total. The fact that given JSON gets merged deeply is asserted in the unit tests for the `mergeRight` utility, there's no need to verify that once more. In `json` utility test we just test that given `merge: true` it merges 2 bodies, if not—it doesn't.

src/context/json.ts
kettanaito: Can you please explain more on how this works? I see that you've renamed 

```diff
- BodyType
+ BodyTypeJSON
```

But I'm certainly missing something here...

src/context/json.ts
```diff
@@ -7,12 +12,19 @@ import { ResponseTransformer } from '../response'
...
-): ResponseTransformer<BodyType> => {
+export const json = <BodyTypeJSON>(
+  body: BodyTypeJSON,
+  { merge = false }: JSONContextOptions = {},
+): ResponseTransformer<BodyTypeJSON> => {
   return (res) => {
     res.headers.set('Content-Type', 'application/json')
-    res.body = JSON.stringify(body) as any
+
+    if (merge) {
```
kettanaito: This `if` is unnecessary: you can assign a ternary expression to the `res.body`

```js
res.body = merge ? mergeRight(...) : body
```

That produces the same output as the if/else statement.

src/context/errors.ts
tdeekens: Ok. Thanks.

src/context/json.ts
tdeekens: True.

src/response.ts
```diff
@@ -34,6 +34,16 @@ export const defaultResponse: Omit<MockedResponse, 'headers'> = {
...
 
+const JSONBodyResponseTransformer = (res: MockedResponse): MockedResponse => {
+  if (res.body && res.headers.get('content-type')?.endsWith('json')) {
+    res.body = JSON.stringify(res.body)
+  }
+
+  return res
+}
+
+const postProcessTransformers = [JSONBodyResponseTransformer]
```
kettanaito: Let's drop the encapsulation of a single transformer into an array :) I know the intention here, but on practice this will remain an array with a single member forever. We can apply `JSONBodyResponseTransformer` directly.

src/response.ts
```diff
@@ -50,7 +60,9 @@ function createResponseComposition(
     )
 
     if (transformers.length > 0) {
-      return compose(...transformers)(resolvedResponse)
+      return compose(...postProcessTransformers.concat(transformers))(
```
kettanaito: That's an interesting design decision we are to make: should we stringify JSON body as a part of the response composition chain (potentially allowing to de-stringify it after you called `res()`), or should we do that right before sending the mocked response to the worker? 🤔

kettanaito (changes requested):
I've left a few comments, yet it looks awesome! We are close to merging this.

kettanaito:
Could you please check your local build? There seem to be a type violation (per CI):

```
[!] (plugin rpt2) Error: /home/circleci/release/src/context/json.ts(23,7): semantic error TS2322: Type 'Record<string, any>' is not assignable to type 'BodyTypeJSON'.
  'BodyTypeJSON' could be instantiated with an arbitrary type which could be unrelated to 'Record<string, any>'.
src/context/json.ts
Error: /home/circleci/release/src/context/json.ts(23,7): semantic error TS2322: Type 'Record<string, any>' is not assignable to type 'BodyTypeJSON'.
  'BodyTypeJSON' could be instantiated with an arbitrary type which could be unrelated to 'Record<string, any>'.
    at error (/home/circleci/release/node_modules/rollup/dist/shared/rollup.js:5211:30)
    at throwPluginError (/home/circleci/release/node_modules/rollup/dist/shared/rollup.js:17799:12)
    at Object.error (/home/circleci/release/node_modules/rollup/dist/shared/rollup.js:18418:24)
    at Object.error (/home/circleci/release/node_modules/rollup/dist/shared/rollup.js:17972:38)
    at RollupContext.error (/home/circleci/release/node_modules/rollup-plugin-typescript2/src/rollupcontext.ts:37:18)
    at /home/circleci/release/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:41:11
    at arrayEach (/home/circleci/release/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:516:11)
    at Function._.each [as forEach] (/home/circleci/release/node_modules/rollup-plugin-typescript2/node_modules/lodash/lodash.js:9368:14)
    at printDiagnostics (/home/circleci/release/node_modules/rollup-plugin-typescript2/src/print-diagnostics.ts:9:2)
    at Object.transform (/home/circleci/release/node_modules/rollup-plugin-typescript2/src/index.ts:242:5)
```

kettanaito:
If you run `yarn test:unit` there is a failing test:

```
Summary of all failing tests
 FAIL  src/context/errors.test.ts
  ● Test suite failed to run

    src/context/json.ts:23:7 - error TS2322: Type 'Record<string, any>' is not assignable to type 'BodyTypeJSON'.
      'BodyTypeJSON' could be instantiated with an arbitrary type which could be unrelated to 'Record<string, any>'.

    23       res.body = mergeRight(body, res.body || {})
```

You can also reproduce this by running that test in isolation locally: `yarn test:unit src/context/errors.test.ts`.

src/response.ts
tdeekens: Yeah, I was torn here knowing there is a premature abstraction lurking on the corner :). I will simplify it. We can always abstract it later.

src/response.ts
tdeekens: Yeah fair point. Initially I just wrapped the call with the transform than passed it down. Let me show what I had initally.

src/response.ts
```diff
@@ -60,8 +58,8 @@ function createResponseComposition(
     )
 
     if (transformers.length > 0) {
-      return compose(...postProcessTransformers.concat(transformers))(
-        resolvedResponse,
+      return JSONBodyResponseTransformer(
```
tdeekens, line 61: Is this what you had in mind? I find it confusing that the `createResponseComposition` returns a `ResponseFunction`. While `compose(...t)(response)` to my understanding doesn't return a `ResponseFunction` but just the `response` which I see also is being passed to the `Object.assign` below?

tdeekens:
@kettanaito thanks for the help. I am stuck on a last typing error and I fail to follow how it's meant to be.

The `resolvedResponse` [here](https://github.com/mswjs/msw/pull/403/commits/a25611cac7928a531a7e08f6ef544b63dba8f982#diff-635146f32166c4a4f9ab579dae7e2253L51) is a `Partial<MockedResponse>` which is combined with the `overrides` which are also a `Partial<MockedResponse>`. If this we have `transformers` we pipe the `resolvedResponse` through them if not we return it [here](https://github.com/mswjs/msw/pull/403/commits/a25611cac7928a531a7e08f6ef544b63dba8f982#diff-635146f32166c4a4f9ab579dae7e2253L68). The `createResponseComposition` itself returns a `ResponseFunction` which returns a `MockedResponse<BodyType>` **not** a `Partial` of it. I am confused how the typing ever worked. For one I assume the `compose` returned an `any` which is assignable to anything. Now the `JSONBodyResponseTransformer` returns a `Partial<MockedReponse>` which seems logical. However, then the return of `createResponseComposition` is not as expected. How do you see this? Can you shed some light here? To me it seems, if I am not mistaken, the last bit to get this through 😄.

tdeekens:
@kettanaito sorry to ping and bother but any feedback here?

kettanaito:
Hey, @tdeekens. Thanks once more for preparing these changes! I think they look fantastic. 

I took this opportunity to refine the pieces around request/response JSON transformation in general, as well as to fix a wrongly typed test suite. 

What concerns your typing question, I've decided to go with the type casting:

```ts
mergeRight(...) as BodyTypeJSON
```

Nevertheless, I'm open to suggestions here. Other than that let's wait for the CI and merge this, if it's green.

src/context/data.test.ts
```diff
@@ -2,57 +2,60 @@ import { data } from './data'
...
+    JSON.stringify({
+      errors: [
+        {
+          message: 'exceeds the limit of awesomeness',
+        },
+      ],
+      data: {
+        name: 'msw',
+      },
+    }),
```
tdeekens, line 59: That makes sense. Thought of that too but didn't dare to make that change.

kettanaito (approved):
The changes looks great. Thank you, @tdeekens, and welcome to contributors! 🎉

tdeekens:
@kettanaito any plans to release this? 🤔

kettanaito:
Hi, @tdeekens. The next version's release took some time, but I think it'll happen this week.

PR #417, opened 2020-10-07
mockServiceWorker: Short-circuits the fetch event on bypass state

### LINKED ISSUE #412: Chrome Dev tools XHR
## Describe the bug
Not sure if it is even a msw problem, but maybe someone knows the answer. For some reason, when msw is enabled, I can not set up a filter to see only XHR requests in Chrome network tab. See how it looks like below (note, that static files are displayed even though the XHR filter is on.)

## Environment

- `msw: 0.21.2`
- `nodejs: 12.16.1`
- `npm: 6.14.8`
- `Chrome Version 85.0.4183.121 (Official Build) (64-bit)`

## To Reproduce
1. Set up msw for browser as described in official docs
2. Open Chrome Dev Tools
3. Turn on filter XHR
4. Reload the page. 
5. You will see all requests.

## Expected behavior
Only XHR requests are displayed

## Screenshots
Screenshot 2020-10-02 at 15 33 02

kettanaito:
> This requires to update the worker script.

## GitHub

- Closes #412 

## Changes

~~Instead of requesting a bypasses request once more, short circuit the `fetch` handler in the worker for it to bypass that request. This will get rid of requests duplications and won't list static assets in the XHR tab of a browser's DevTools.~~

- Short circuits on the navigation requests outside of the `respondWith` scope.
- Short circuits on the `!clients[clientId]` requests outside of the `respondWith` scope.
- Fixes an issue that some requests are still marked as "from ServiceWorker" after the worker has been unregistered.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -74,10 +74,21 @@ self.addEventListener('message', async function (event) {
   }
 })
 
-self.addEventListener('fetch', async function (event) {
+self.addEventListener('fetch', function (event) {
   const { clientId, request } = event
   const requestClone = request.clone()
   const getOriginalResponse = () => fetch(requestClone)
+
+  // Bypass navigation requests.
+  if (request.mode === 'navigate') {
+    return
+  }
+
+  // Bypass mocking if the current client isn't present in the internal clients map
+  // (i.e. has the mocking disabled).
+  if (!clients[clientId]) {
+    return
+  }
 
   // Opening the DevTools triggers the "only-if-cached" request
   // that cannot be handled by the worker. Bypass such requests.
@@ -89,20 +100,15 @@ self.addEventListener('fetch', async function (event) {
     new Promise(async (resolve, reject) => {
       const client = await event.target.clients.get(clientId)
 
-      if (
-        // Bypass mocking when no clients active
-        !client ||
-        // Bypass mocking if the current client has mocking disabled
-        !clients[clientId] ||
-        // Bypass mocking for navigation requests
-        request.mode === 'navigate'
-      ) {
+      // Bypass mocking when the request client is not active.
+      if (!client) {
         return resolve(getOriginalResponse())
       }
 
       // Bypass requests with the explicit bypass header
       if (requestClone.headers.get(bypassHeaderName) === 'true') {
         const modifiedHeaders = serializeHeaders(requestClone.headers)
+
         // Remove the bypass header to comply with the CORS preflight check
         delete modifiedHeaders[bypassHeaderName]
 
--- a/test/msw-api/unregister.test.ts
+++ b/test/msw-api/unregister.test.ts
@@ -35,10 +35,7 @@ test('unregisters itself when not prompted to be activated again', async () => {
   })
   const secondBody = await secondResponse.json()
 
-  // Although the Service Worker unregisters itself upon refreshing the page,
-  // it still remains "active and running" with the "deleted" status.
-  // This results into requests go through the Service Worker until the next reload.
-  expect(secondResponse.fromServiceWorker()).toBe(true)
+  expect(secondResponse.fromServiceWorker()).toBe(false)
   expect(secondBody).not.toEqual({
     mocked: true,
   })


kettanaito:
We may also want to short circuit on `MOCK_NOT_FOUND` case, where a request was captured, but there was no respective request handler, so it got performed as-is.

marcosvega91:
I didn't know that trick, fine !!

src/mockServiceWorker.js
```diff
@@ -97,7 +97,7 @@ self.addEventListener('fetch', async function (event) {
         // Bypass mocking for navigation requests
         request.mode === 'navigate'
       ) {
-        return resolve(getOriginalResponse())
+        return
```
malykhinvi: Is it ok that the promise is never resolved in that case? Why we can not do an early return before the `event.respondWith`?

kettanaito: Early return before `event.respondWith` throws an exception that request cannot be handled because the event is closed. 

Good point about pending Promise. I'm not sure if it's an issue, though. We shouldn't resolve it, that will cause a fake response without any data.

marcosvega91: I'm not sure that it will work. Intercepting the `fetch` I think that we should provide a way to resolve the request, fetching from original one or from cache, or mocking as our case.

kettanaito: The fact that it performs the request may imply that it actually works as intended. Had the promise given to `respondWith` been pending forever, we'd receive a forever pending request in the network.

That being said, there are some failing tests on CI with a navigation timeout. This may be caused by multiple factors, including forever pending requests. I haven't encountered those while manually testing.

kettanaito: You are right: not resolving a Promise here causes an infinitely pending request. I haven't spotted this due to some race conditions between the worker registration and the request dispatch. Now I can see that.

src/mockServiceWorker.js
```diff
@@ -74,11 +74,22 @@ self.addEventListener('message', async function (event) {
   }
 })
 
-self.addEventListener('fetch', async function (event) {
```
kettanaito, line 77: Fetch event handler function cannot be async per specification. Marking it as `async` here is misleading: you cannot use `await` within its top scope, it will cause an invalid state exception (cannot handle fetch event that's already handled).

src/mockServiceWorker.js
```diff
@@ -74,11 +74,22 @@ self.addEventListener('message', async function (event) {
...
+self.addEventListener('fetch', function (event) {
   const { clientId, request } = event
   const requestClone = request.clone()
   const getOriginalResponse = () => fetch(requestClone)
 
+  // Bypass navigation requests.
+  if (request.mode === 'navigate') {
```
kettanaito, line 83: Navigation requests can be bypassed early, so I'm moving them outside of the `event.respondWith` promise. This way navigation requests are performed as-is, not within the worker's scope.

src/mockServiceWorker.js
```diff
@@ -74,11 +74,22 @@ self.addEventListener('message', async function (event) {
...
+    return
+  }
+
+  // Bypass mocking if the current client isn't present in the internal clients map
+  // (i.e. has the mocking disabled).
+  if (!clients[clientId]) {
+    return
```
kettanaito, line 90: When the client hasn't yet sent the `MOCK_ACTIVATE` event bypass early as well. Requests made between the worker registration and worker activation are performed as-is. This includes static assets as well, they won't be marked as XHR in this time window.

Note that the time window is subjected to a race condition, as worker activation and registration are both asynchronous actions and may happen with a different time distance.

test/msw-api/unregister.test.ts
```diff
@@ -35,10 +35,7 @@ test('unregisters itself when not prompted to be activated again', async () => {
   })
   const secondBody = await secondResponse.json()
 
-  // Although the Service Worker unregisters itself upon refreshing the page,
-  // it still remains "active and running" with the "deleted" status.
-  // This results into requests go through the Service Worker until the next reload.
-  expect(secondResponse.fromServiceWorker()).toBe(true)
+  expect(secondResponse.fromServiceWorker()).toBe(false)
```
kettanaito, line 38: Moving the navigation request bypassing out of the `respondWith` scope now fixes the issue that marked the first request as served from the worker, even after the worker has been unregistered. This test is now actually correct.

src/mockServiceWorker.js
marcosvega91: this is a great catch !

marcosvega91 (dismissed):
everything ok !! I didn't thought that solution, Great !

PR #421, opened 2020-10-12
TypeScript typing for REST path parameters

### LINKED ISSUE #393: TypeScript typing for REST query parameters

Request parameters have `any` types, which is easy to miss when you use MSW, and have untyped code.

https://github.com/mswjs/msw/blob/eec2a9ee3450af2ad3f33369157a1220117289c6/src/utils/handlers/requestHandler.ts#L41-L43

image

An ideal solution would be having a generic type for query parameters, similar to the body and response:

```tsx
rest.get<{}, {}, {bookingSessionId:? string}>(
```

This shouldn't be a breaking change if we make this type optional.

As a short term solution, we could replace `any` with `unknown` be this looks like a breaking change because the code that used to work (but untyped) would require new types.

WojciechMatuszewski:
I'm not 100% sure about this one in terms of design. It seems kinda off that the 3rd type parameter is for something related to the request while the second one is related to the response.

Maybe we can use overloading here?

The typings would work as proposed in the issue.

Closes #393

### DIFF
--- a/test/rest-api/params.mocks.ts
+++ b/test/rest-api/params.mocks.ts
@@ -1,7 +1,17 @@
 import { setupWorker, rest } from 'msw'
 
+interface ResponseType {
+  username: string
+  messageId: string
+}
+
+interface RequestParams {
+  username: string
+  messageId: string
+}
+
 const worker = setupWorker(
-  rest.get(
+  rest.get<any, ResponseType, RequestParams>(
     'https://api.github.com/users/:username/messages/:messageId',
     (req, res, ctx) => {
       const { username, messageId } = req.params
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -3,6 +3,7 @@ import {
   ResponseResolver,
   MockedRequest,
   DefaultRequestBodyType,
+  RequestParams,
 } from './utils/handlers/requestHandler'
 import { Mask } from './setupWorker/glossary'
 import { set } from './context/set'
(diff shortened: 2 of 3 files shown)


timdeschryver (commented):
I don't want to block the PR (which looks good to me), but I wonder if we could and want to use [TS 4.1 template literal types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/#template-literal-types) here in a follow-up PR. 

If I'm not mistaken, we should be able to infer the request params based on the URL.
See this [playground](https://www.typescriptlang.org/play?ts=4.1.0-pr-40336-88#code/C4TwDgpgBAogHsATgQwMbAEoHsCuwIAKyKAtgM4A8AKlBAhAHYAmZUZSAlgwOYB8UAXgBQUNpx616zVlRFQA-FAwRUWREwrtEXbgBox2nrzkAuKDTr5pUAAYASAN5cAZhERQAysGLAAviccXNygiUl8AekCGV3dldl8bOUUHAG0AayguEOJkEigAHyg0iBAsZ1gEFHRsPEIc8go44F4AXTMtHV9Tc0krFlsomM9vRD8Ap2jg0NyEpKhUjKzpkjaDTu6HXwBuISFQSBDBCqQ0TFx8ZcoAcnCwLHYycJM79gBJJiveHd2f-egAYSO8BO1XOdVI11u92Aj2e0PeT1UJBIjGA70+312QA) for an example (not mine).

test/rest-api/params.mocks.ts
```diff
@@ -1,7 +1,17 @@
...
+}
+
+interface RequestParams {
+  username: string
+  messageId: string
+}
+
 const worker = setupWorker(
-  rest.get(
+  rest.get<any, ResponseType, RequestParams>(
```
kettanaito, line 14: I've added this usage example for the reference.

kettanaito (approved):
A superb work done! Thank you, @WojciechMatuszewski.

kettanaito:
That's an awesome suggestion, @timdeschryver. Would you mind to create a follow-up issue so we can plan an implement that? I suppose it would require a TypeScript version bump, so we can roll this out in the next breaking version.

kettanaito:
Renamed the pull request since it annotates _path_ parameters, not query parameters. We'd have to approach query parameters type safety as a separate task.

test/rest-api/params.mocks.ts
SalahAdDin: This line gave me some ideas to fix **typescript** errors, thank you!

kid1412621:
How about query params? Currently only works with path params.

SalahAdDin:
> How about query params? Currently only works with path params.

It requires a new PR.

PR #424, opened 2020-10-13
feat: supports custom predicate as the "quiet" option value

### LINKED ISSUE #420: Predicate for Logs
First off thank you for the excellent library. It is amazingly helpful for mocking endpoints.

I have certain requests which are fired over and over again (for example an autosave endpoint, or polling the server for some information). I find logging very useful but I would like a way to filter out some of these more common requests. I saw the implementation of quiet in https://github.com/mswjs/msw/issues/111. Would it be possible to extend that concept to use a function which returns True for requests which should be logged and False for requests which should not be logged? The alternative I can think of is to override console.log which is probably less desirable.

balavishnuvj:
- Closes #420

### DIFF
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -55,7 +55,7 @@ export type StartOptions = SharedOptions & {
    */
   quiet?:
     | boolean
-    | ((req?: MockedRequest, res?: ResponseWithSerializedHeaders) => boolean)
+    | ((req: MockedRequest, res: ResponseWithSerializedHeaders) => boolean)
 
   /**
    * Defer any network requests until the Service Worker
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -1,5 +1,5 @@
 import { HeadersList } from 'headers-utils'
-import { RequestHandler } from '../utils/handlers/requestHandler'
+import { MockedRequest, RequestHandler } from '../utils/handlers/requestHandler'
 import { MockedResponse } from '../response'
 import { SharedOptions } from '../sharedOptions'
 import { ServiceWorkerMessage } from '../utils/createBroadcastChannel'
@@ -53,7 +53,9 @@ export type StartOptions = SharedOptions & {
    * Disable the logging of captured requests
    * into browser's console.
    */
-  quiet?: boolean
+  quiet?:
+    | boolean
+    | ((req?: MockedRequest, res?: ResponseWithSerializedHeaders) => boolean)
 
   /**
    * Defer any network requests until the Service Worker
--- a/src/setupWorker/start/utils/activateMocking.ts
+++ b/src/setupWorker/start/utils/activateMocking.ts
@@ -7,7 +7,9 @@ export const activateMocking = async (
   context.worker?.postMessage('MOCK_ACTIVATE')
 
   return context.events.once('MOCKING_ENABLED').then(() => {
-    if (!options?.quiet) {
+    const isQuiet =
+      typeof options?.quiet === 'function' ? options?.quiet() : options?.quiet
+    if (!isQuiet) {
       console.groupCollapsed(
         '%c[MSW] Mocking enabled.',
         'color:orangered;font-weight:bold;',
--- a/test/msw-api/setup-worker/start/quiet.fn.true.mocks.ts
+++ b/test/msw-api/setup-worker/start/quiet.fn.true.mocks.ts
@@ -22,8 +22,7 @@ const worker = setupWorker(
 // @ts-ignore
 window.__MSW_REGISTRATION__ = worker.start({
   // Disable logging of matched requests into browser's console
-  quiet: (req, res) => {
-    console.info('anything?', req.url.href, res.status)
+  quiet: (req) => {
     return req.url.href.includes('foo')
   },
 })
--- a/test/msw-api/setup-worker/start/quiet.fn.true.mocks.ts
+++ b/test/msw-api/setup-worker/start/quiet.fn.true.mocks.ts
@@ -1,0 +1,29 @@
+import { setupWorker, rest } from 'msw'
+
+const worker = setupWorker(
+  rest.get('/user', (req, res, ctx) => {
+    return res(
+      ctx.json({
+        firstName: 'John',
+        age: 32,
+      }),
+    )
+  }),
+  rest.get('/foo', (req, res, ctx) => {
+    return res(
+      ctx.json({
+        firstName: 'John',
+        age: 32,
+      }),
+    )
+  }),
+)
+
+// @ts-ignore
+window.__MSW_REGISTRATION__ = worker.start({
+  // Disable logging of matched requests into browser's console
+  quiet: (req, res) => {
+    console.info('anything?', req.url.href, res.status)
+    return req.url.href.includes('foo')
+  },
+})
--- a/test/msw-api/setup-worker/start/quiet.fn.true.test.ts
+++ b/test/msw-api/setup-worker/start/quiet.fn.true.test.ts
@@ -1,0 +1,84 @@
+import * as path from 'path'
+import { TestAPI, runBrowserWith } from '../../../support/runBrowserWith'
+import { captureConsole } from '../../../support/captureConsole'
+
+let runtime: TestAPI
+
+beforeAll(async () => {
+  runtime = await runBrowserWith(
+    path.resolve(__dirname, 'quiet.fn.true.mocks.ts'),
+  )
+})
+
+afterAll(() => {
+  return runtime.cleanup()
+})
+
+test('log the captured request when the "quiet" option is returns to "false"', async () => {
+  const { messages } = captureConsole(runtime.page)
+
+  await runtime.page.evaluate(() => {
+    // @ts-ignore
+    return window.__MSW_REGISTRATION__
+  })
+
+  await runtime.reload()
+
+  const activationMessage = messages.startGroupCollapsed.find((text) => {
+    return text.includes('[MSW] Mocking enabled.')
+  })
+  expect(activationMessage).toBeFalsy()
+
+  const res = await runtime.request({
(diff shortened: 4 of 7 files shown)


src/setupWorker/start/utils/activateMocking.ts
```diff
@@ -7,7 +7,9 @@ export const activateMocking = async (
   context.worker?.postMessage('MOCK_ACTIVATE')
 
   return context.events.once('MOCKING_ENABLED').then(() => {
-    if (!options?.quiet) {
+    const isQuiet =
+      typeof options?.quiet === 'function' ? options?.quiet() : options?.quiet
```
balavishnuvj, line 11: I think this part doesn't work. Can someone help here?

marcosvega91: this is working, but I think that the quit function should check if the req is provided

```js
window.__MSW_REGISTRATION__ = worker.start({
  // Disable logging of matched requests into browser's console
  quiet: (req) => {
    if (req) {
      return req.url.href.includes('foo')
    }
    return true
  },
})
```

test/msw-api/setup-worker/start/quiet.fn.true.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    firstName: 'John',
+    age: 32,
+  })
+
+  const requetsLog = messages.startGroupCollapsed.find((text) => {
+    return text.includes('[MSW]') && text.includes('GET /foo')
+  })
+
+  expect(requetsLog).toBeUndefined()
+})
```
marcosvega91: I think that we can integrate this test in `quiet.test.ts`.

we can update the `quiet.mock.ts` as 

```ts

const worker = setupWorker(
    return res(
      ctx.json({
        firstName: 'John',
        age: 32,
      }),
    )
  }),
  rest.get('/foo', (req, res, ctx) => {
    return res(
      ctx.json({
        firstName: 'John',
        age: 32,
      }),
    )
  }),
)

// @ts-ignore
window.__MSW_START__ = worker.start
```

the in the `quiet.test.ts` we can do as 

```diff
+  await runtime.reload()
+
   await runtime.page.evaluate(() => {
     // @ts-ignore
-    return window.__MSW_REGISTRATION__
+    return window.__MSW_START__({
+      // Disable logging of matched requests into browser's console
+      quiet: function (req) {
+        if (req) {
+          return req.url.href.includes('foo')
+        }
+        return true
+      },
+    })
   })
 
-  await runtime.reload()
```

In this way we can call the function `__MSW_START__` with the right options. What do you think?

marcosvega91 (changes requested):
Hey @balavishnuvj thanks for working on this great feature ! :) 

I have a suggestion about the project structure. Let me know that you think :)

test/msw-api/setup-worker/start/quiet.fn.true.test.ts
balavishnuvj: yeah this looks great, will update. What does `await runtime.reload()` do? Do I have to remove it?

marcosvega91: it will reload the page, because we have a single instance of puppeteer it is necessary to have everything clean between test. It should be called before the `evaluate`. 

So when the test runs it will reload the page and then start MSW.

marcosvega91: I think that we can add in a `beforeEach` instead
```js
beforeEach(() => {
  await runtime.reload()
})
```

src/setupWorker/glossary.ts
```diff
@@ -53,7 +53,9 @@ export type StartOptions = SharedOptions & {
    * Disable the logging of captured requests
    * into browser's console.
    */
-  quiet?: boolean
+  quiet?:
+    | boolean
+    | ((req?: MockedRequest, res?: ResponseWithSerializedHeaders) => boolean)
```
kettanaito: By the time we analyze the `quiet` option and output a successful mock to console both `req` and `res` exist. No need to annotate them as optional with `?`.

src/setupWorker/start/utils/activateMocking.ts
kettanaito, line 11: I don't think we should execute a `quiet` predicate function here. How about we only check the `options.quiet` for truthiness?

```js
// Basically, keeping the previous implementation.
if (!options?.quiet)
```

This is an internal log message not bound to any request. It should still abide by `quiet: true` in case you want to silence all logs, but it shouldn't support predicate as there is no request context to execute the predicate in. 

What are your thoughts on this?

test/msw-api/setup-worker/start/quiet.fn.true.mocks.ts
```diff
@@ -0,0 +1,28 @@
...
+const worker = setupWorker(
+  rest.get('/user', (req, res, ctx) => {
+    return res(
+      ctx.json({
+        firstName: 'John',
+        age: 32,
+      }),
+    )
+  }),
+  rest.get('/foo', (req, res, ctx) => {
```
kettanaito: I'd suggest to use realistic paths so that these usage examples resemble actual usage.

test/msw-api/setup-worker/start/quiet.fn.true.test.ts
kettanaito: We can also introduce a different test suite like `quiet-predicate` to scope the predicate functionality.

src/setupWorker/start/utils/activateMocking.ts
balavishnuvj: wouldn't the `options?.quiet` be a function?

src/setupWorker/glossary.ts
balavishnuvj: this was to handle `src/setupWorker/start/utils/activateMocking.ts`, there quiet can be a function without and req and res. Or am I missing anything?

test/msw-api/setup-worker/start/quiet.fn.true.test.ts
balavishnuvj: both works for me, I created new file because I thought that was convention that being followed.
eg. `wait-until-ready.false.mocks.ts` and `wait-until-ready.error.mocks.ts`
let me know which works better?

src/setupWorker/start/utils/activateMocking.ts
kettanaito: A good point, as a function value produces a truthy value, while the option describes a negative value (`true` for disable).

src/setupWorker/glossary.ts
kettanaito: Yeah, I understand the intention. As I've stated earlier, such internal events like `ACTIVATE_MOCKING` shouldn't be controlled by the predicate function of the `quiet` option.

balavishnuvj:
@kettanaito @marcosvega91 please do let me know your concerns are not addressed.

marcosvega91 (dismissed):
LGTM :) thanks :)

src/setupWorker/glossary.ts
balavishnuvj: Should we remove the quiet check from this log altogether?

src/setupWorker/start/utils/activateMocking.ts
kettanaito: This raises an entirely new concern: how can one have a custom logging predicate for requests and also disable the internal messages? The current `quiet` option value cannot handle this.

balavishnuvj: either we make `req` and `res` in callback optional or introduce a new option

src/setupWorker/glossary.ts
kettanaito: That's a sensible approach, yet it would be a breaking change then. Previously this log was also controlled by `options.quiet`. 

Alternatively, maybe we should distribute a function-as-value in a _different_ option?

balavishnuvj: yeah, it makes sense to introduce different option to handle this case.
but then usage seems a bit tricky,
```
{
quiet: true,
quietPredicate: (req) => {
      return req.url.href.includes('profile')
  },
}
```
what is expected might be a not sure.

we could ntroducte another option  for internal logs
```
quiet:  (req) => {
      return req.url.href.includes('profile')
  },
quietInternals: true
```
we can come up with better names for sure 😅

balavishnuvj:
@kettanaito what needs to be done here? bit confused 😅

lukesmurray:
looks like a simpler `quiet: boolean` was merged [here](https://github.com/mswjs/msw/pull/113). Not sure what the status of this is.

kettanaito:
@balavishnuvj, sorry, I've lost the context of these changes after a while. Worry not, I will check what's left to do, if anything, and update the review status. The issues/active collaborators ratio is rather high, so certain features although useful and awesome don't get as much attention as they deserve.

kettanaito:
@lukesmurray, hey 👋 You're right, you can already silence the logging by passing `quiet: false` to either `worker.start()` or `server.listen()`. This pull request aims for that `quiet` option to support a custom predicate, so you may choose which requests to log and which to ignore.

rpowis:
Any update on this? I was just looking for exactly this functionality (or at least something very similar).

Our use case is making sure we don't miss any unhandled requests to controlled paths (i.e. `/api/untested-route`) while filtering out the noise of paths/domains we never want to handle (i.e. Google Analytics).

Unless there's another way to achieve that without this being merged, I'd be happy to contribute to get this feature released if needed.

kettanaito:
Thanks once again for working on this, @balavishnuvj!

I think I will close this one. I can see this being a useful feature but it has a few cost points that don't quite justify the gain here. It's an additional config, and I'd like to be extremely strict when it comes to it. Moreover, this hasn't been requested as a missing feature for three years the pull request has been open. 

I'd be glad to circle back on this if more people find this missing. Until then, closing.

PR #430, opened 2020-10-20
GraphQL: support undefined errors with "ctx.errors"

aganglada:
This PR make graphql `errors` able to be undefined, as the proposed docs are the following:

```
graphql.operation(async (req, res, ctx) => {
  const payload = await graphqlRequest(
    schema,
    req.body.query,
    root,
    null,
    req.variables,
  )
  return res(ctx.data(payload.data), ctx.errors(payload.errors))
})
```

Where errors are passed straight away to the response.
If errors are undefined, there won't be any errors property added to the response.

Reference to the docs PR: https://github.com/mswjs/mswjs.io/pull/88

### DIFF
--- a/src/context/errors.ts
+++ b/src/context/errors.ts
@@ -3,10 +3,16 @@ import { ResponseTransformer } from '../response'
 import { json } from './json'
 
 /**
- * Returns a list of GraphQL errors.
+ * Sets a given list of GraphQL errors on the mocked response.
  */
-export const errors = (
-  errorsList: Partial<GraphQLError>[],
-): ResponseTransformer<{ errors: typeof errorsList }> => {
+export const errors = <
+  ErrorsType extends Partial<GraphQLError>[] | null | undefined
+>(
+  errorsList: ErrorsType,
+): ResponseTransformer<{ errors: ErrorsType }> => {
+  if (errorsList == null) {
+    return (res) => res
+  }
+
   return json({ errors: errorsList }, { merge: true })
 }
--- a/src/context/errors.test.ts
+++ b/src/context/errors.test.ts
@@ -60,3 +60,10 @@ test('combines with data in the response JSON body', () => {
     }),
   )
 })
+
+test('bypasses undefined errors', () => {
+  const result = response(errors(undefined), errors(null))
+
+  expect(result.headers.get('content-type')).not.toEqual('application/json')
+  expect(result).toHaveProperty('body', null)
+})


src/context/errors.ts
```diff
@@ -6,7 +6,9 @@ import { json } from './json'
  * Returns a list of GraphQL errors.
  */
 export const errors = (
-  errorsList: Partial<GraphQLError>[],
-): ResponseTransformer<{ errors: typeof errorsList }> => {
+  errorsList?: Partial<GraphQLError>[],
+): ResponseTransformer | ResponseTransformer<{ errors: typeof errorsList }> => {
```
kettanaito: I believe the correct annotation would be 

```ts
: ResponseTransformer<{ errors: typeof errorsList | undefined }>
```

Would this produce a proper type checking when given `undefined` as the `errors` value?

src/context/errors.ts
```diff
@@ -6,7 +6,9 @@ import { json } from './json'
...
+  if (!errorsList) return json({})
```
kettanaito: That will set the response body to an empty object and may not play nicely when combined with `ctx.data`, as it may override any preceding data. 

Can we return the response transformer without any transformations in this case?
```js
if (!errorsList) {
  return res => res
}
```

> We may later replace this with something like `identity()` function for a shorter notation.

kettanaito (changes requested):
Thank you for addressing this, @aganglada!
I've left a few suggestions in the review. Could you please take a look and let me know what you think of them?

src/context/errors.ts
aganglada: I think is exactly the same isn't it?

`typeof errorsList` references a value that can be undefined, therefore inherit its types right?

src/context/errors.ts
aganglada: make sense 👍

src/context/errors.ts
kettanaito: Not really. `errorsList` is `Partial<GraphQLError>[]` and doesn't contain a nullable type (always demands an array). 

So `undefined` is an incompatible type with `errorsList`. To handle this, we can use a union type `| undefined`.

kettanaito:
Hey, @aganglada. Thanks for preparing this! I've rebased the feature branch and rewrote the test to be flat, just as the rest of the tests in the suite (introduced recently).

kettanaito (approved):

kettanaito:
Welcome to contributors, @aganglada 🎉

aganglada:
👏 You rock! Sorry I didn't have time to finish it, I was running into a TS error that didn't make sense as well as work 🎉

PR #432, opened 2020-10-21
docs: add build note to integration tests

timdeschryver:
I tried to run the tests but the integration tests were failing.
I think this note can help future devs that encounter the same problem.


marcosvega91:
Hi @timdeschryver instead of `build` script I think is better to use the `start` script. In this way if you made some changes to codebase with `start` script it will rebuild on fly.

kettanaito (approved):
Thank you for adding this, @timdeschryver!

Issue #70, opened 2020-03-19
Status 400 (OK) displaying when it should display (Bad Request)


kentcdodds:
Here's what a request looks like when I set the status to 400 with msw vs a real server sending a 400:

![image](https://user-images.githubusercontent.com/1500684/77124278-08e92080-6a08-11ea-9063-25d0d29bd3ca.png)

I'm not sure what could be causing this.

kettanaito:
My guess is that since MSW constructs a new `Response` it has `statusText: 'OK'` by default. I presume you set the status code alone (i.e. via `ctx.status(400)`), so it doesn't map it to any status text associated with that status code. 

One way to solve it for now could be to provide an explicit status text alongside the status code:

```js
ctx.status(400, 'Bad request')
```

I value the validity of requests and responses, but I wonder whether it's the library's responsibility to map status codes to texts implicitly. I'd expect this to be done by `Request` constructor. Would have to look into this in more detail.

kentcdodds:
That's fine. Thank you for helping me know what to do there.

kentcdodds:
I think that having a default statusText that's the same as what the statusText would be for a given status code would be useful. I could always override it if I want to, but having a default to what the browser does would be sensible IMO.

kettanaito:
I was thinking of keeping a status code-status text map internally in MSW and mapping it by default, as you described. I guess there's little harm in that, but a big gain in developer experience.

kettanaito:
I went with the solution that reuses the [`statuses`](https://github.com/jshttp/statuses/) package thats's used in Express to map status codes to status texts. 

Starting from `0.9.3` MSW now implicitly maps given status codes to the associated status text, if such is present, and if no custom status text is provided. 

@kentcdodds, could you please update to `0.9.3` and let me know it produced correct status texts for you? Thanks.

kentcdodds:
Confirmed. This worked. Thanks!

pandrodo:
Sorry for bumping that old issue, but when I using setup server with such handler: 
`rest.post('http://127.0.0.1:3000/getUserPoints', (req, res, ctx) => res(ctx.status(401))),`
I receiving response with status text `OK`. Its not only for 401, but for other statuses too.
![](https://puu.sh/GHndZ/3cff3d9b20.png)

kettanaito:
Hey, @pandrodo. That's perfectly fine.

Can you clarify a few things:

1. Which version of `msw` are you using?
1. What version of NodeJS are you on?
[...]

PR #438, opened 2020-10-28
ctx.status: Fixes synthetic default import of the "statuses" package

kettanaito:
## GitHub

- Regression of #70

## Changes

- Sets `allowSyntheticDefaultImports: true` in `tsconfig.json` to enable imports of non-JS modules (i.e. `statuses`).
- Fixes the issue that `ctx.status` did not set the status text based on the status code implicitly.
- Flattens unit tests for `ctx.status`.
- Adds integration test that asserts in-browser usage (the issue surfaced here, unit tests passed in NodeJS).

### DIFF
--- a/src/context/status.ts
+++ b/src/context/status.ts
@@ -1,4 +1,4 @@
-import * as statuses from 'statuses/codes.json'
+import statuses from 'statuses/codes.json'
 import { ResponseTransformer } from '../response'
 
 export const status = (
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,7 +5,7 @@
     "target": "es6",
     "module": "ESNext",
     "moduleResolution": "node",
-    "allowSyntheticDefaultImports": false,
+    "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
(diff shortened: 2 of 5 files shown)


src/context/status.ts
```diff
@@ -1,4 +1,4 @@
-import * as statuses from 'statuses/codes.json'
+import statuses from 'statuses/codes.json'
```
kettanaito, line 1: Importing it as `* from X` resulted into the `statuses` being an object with the `{ default }` key. 

Getting the value as `statuses[statusCode]` failed, since it was in fact `statuses.default[statusCode]`.

tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "target": "es6",
     "module": "ESNext",
     "moduleResolution": "node",
-    "allowSyntheticDefaultImports": false,
+    "allowSyntheticDefaultImports": true,
```
kettanaito, line 8: I don't see any obvious implications of changing this configuration value. We should still be mindful about synthetic imports, as they may produce unexpected behavior in NodeJS. We must have tests that assert that.

PR #439, opened 2020-10-28
fix: .json() must be awaited

marcosvega91:
should fix the issue in CI. `res.json()` was not awaited 

image


kettanaito (approved):
Reading my mind, @marcosvega91! Thank you.

PR #441, opened 2020-10-29
feat: add custom transformers

### LINKED ISSUE #440: Support alternative JSON library

This is both a feature request and a bug report. Consider the following : 

```js
import { rest } from 'msw';
import axios from 'axios';
import JSONbig from 'json-bigint';

// This is required by Jest to work properly
BigInt.prototype.toJSON = function() { return this.toString(); };

const mockBaseUrl = "http://localhost/services/acme/api";

const customer = {
  "username": "Dude",
  "balance": BigInt(1597928668063727616) // yes, the Dude is rich
};

const server = setupServer(
  rest.get(`${mockBaseUrl}/customers/me`, (req, res, ctx) => {
    return res(
      ctx.status(200),
      ctx.json(customer),
    )
  }),
)

beforeAll(() => server.listen({ onUnhandledRequest: 'error' }))
afterEach(() => server.resetHandlers())
afterAll(() => server.close())

const apiAxios = axios.create({
  transformResponse: [data => data ? JSONbig.parse(data) : data],
});

describe("Mocks", () => {

  it("should mock a basic GET call", () => {
    const url = `${mockBaseUrl}/customers/me/`;

    return expect(apiAxios.get(url).then(r => r.data)).resolves.toEqual(customer);
  });
});
```

This test result in the following error : 

```
Error: expect(received).resolves.toEqual(expected) // deep equality

- Expected
+ Received

  Object {
-   "balance": 1597928668063727616n,
+   "balance": "1597928668063727616",
    "username": "Dude",
  }
```

One simple solution would be to be able to set the JSON library to use while transforming the response, as [it looks like](https://github.com/mswjs/msw/blob/master/src/response.ts#L43) you internally transform to json.
In my case the common library to handle `BigInt` is [json-bigint](https://www.npmjs.com/package/json-bigint).

I'm still looking into creating my own response transformer to handle `BigInt` but there is no documentation about it so it's quite time consuming.

marcosvega91:
fix #440

This PR will add support for `defaultTransformers`.  This could be useful if you want to apply your own body transformation as for the issue.

Now the function `createResponseComposition` is exported from the library. Users could create there own response function like 

```js
import fetch from 'node-fetch'
import { rest, createResponseComposition } from 'msw'
import * as JSONbig from 'json-bigint'

const customReponse = createResponseComposition({
  defaultTransformers: [
    (res) => {
      if (res.body && res.headers?.get('content-type')?.endsWith('json')) {
        res.body = JSONbig.stringify(res.body)
      }
      return res
    },
  ],
})

const server = setupServer(
  rest.get('http://test.mswjs.io/me', (req, res, ctx) => {
    const me = {
      username: 'Dude',
      balance: BigInt(1597928668063727616),
    }
    return customReponse(ctx.json(me))
  }),
)
```


src/response.ts
```diff
@@ -46,10 +38,26 @@ const stringifyJsonBody: ResponseTransformer = (res) => {
...
+export type ResponseCompositionOptions<BodyInit> = {
+  defaultTransformers?: ResponseTransformer<BodyInit>[]
+  mockedResponseOverrides?: Partial<MockedResponse>
+}
+
+export function createResponseComposition<BodyInit>(
+  options: ResponseCompositionOptions<BodyInit> = {
+    defaultTransformers: [stringifyJsonBody],
+    mockedResponseOverrides: {},
+  },
```
marcosvega91: I have added not in the `MockedResponse` type because it was something not problem regarding to the response

src/response.ts
```diff
@@ -58,20 +66,24 @@ function createResponseComposition(
           'x-powered-by': 'msw',
         }),
       },
-      overrides,
+      options.mockedResponseOverrides,
     )
+    const transformers = [
+      ...(options.defaultTransformers ?? [stringifyJsonBody]),
+      ...userTransformes,
+    ].filter(Boolean)
```
marcosvega91: We are applying `compose` from right to left, maybe we should change and apply from left to right?

package.json
```diff
@@ -104,6 +105,7 @@
     "husky": "^4.3.0",
     "ignore-loader": "^0.1.2",
     "jest": "^26.4.2",
+    "json-bigint": "^1.0.0",
```
marcosvega91, line 108: I have used the library mentioned in the issue to make the test more accurate

marcosvega91 (commented):
I have added some comments

JesusTheHun:
Works for me !

package.json
kettanaito: That’s okay. We can also leverage examples repo (smoke tests) for purposes like this. I didn’t decide yet where the border lies on this.

src/response.ts
kettanaito: In the functional domain compose works right-to-left. Changing this would be a breaking change for the users.

What are the reasons to change the order?

package.json
marcosvega91: ok, fine

src/response.ts
marcosvega91: > In the functional domain compose works right-to-left

yes right

> What are the reasons to change the order?

Thinking about FP it is not real problem

kettanaito (approved):
Great work on this, @marcosvega91!

I've flattened the `createResponseComposition` options so we don't have to deal with default + custom options merging when given partial options object. The library now publicly exposes certain response-related TypeScript definitions so it's possible to use our internal functions and have type safety. 

Looks great, I'm looking forward for the CI to pass.

timdeschryver:
I think it would be great to add this as a recipe to the [docs](https://mswjs.io/docs/recipes).
Since you implemented this feature @marcosvega91 , are you planning to add this to the recipes?
If not, I can add it (if we want to add this to the docs) 🙂

marcosvega91:
I'm really bad writing things. Also my English is not very good, so feel free to open a PR with your changes 😉. Thanks @timdeschryver :)

kettanaito:
@marcosvega91, thank you once more for adding these changes, Marco :) 
Please, know that I'm always happy to proof-read and help with the writing. You're not alone in this.

marcosvega91:
It's always a pleasure 🙂

PR #450, opened 2020-11-08
Supports listing current request handlers

### LINKED ISSUE #310: List of all current handlers
### Feature

It would be helpful to know the list of active handlers at a given point of time. For eg. when used for testing, I would like to know already present handlers. It is a bit easier to debug. It could easy to track when `server.use` is used to add more handlers.

### Solution
It would be quite simple, API could be something like `server.listHandlers()`, which would return all the active listeners.

### Current Alternative
The current alternative is to use `onUnhandledRequest` and kind of guess the typo in the request handler. 

PS: I would be happy to give PR if this is a feature that others would want to have too.

kettanaito:
## Changes

- Adds a new `.getMetaInfo()` method on a request handler object. Designed to return a publicly inspectable information about the handlers (i.e. mask, path, operation type, etc.)
- Adds a new `.list()` API to list currently active request handlers. Works for both browser and NodeJS API. 

## GitHub

- Fixes #310 

## Roadmap

- [x] Add browser integration tests
- [x] Come up with a more suitable API name than `.list()` (conflicts with auto-suggestions upon `server.listen`)
- [x] Consider if `getMetaInfo` should be a function, as its return data is static
- [x] Consider adding a matcher function for a given handler: pass actual URL string, see if it matches straight away
- [x] Exclude path matching REPL link for `graphql.*` request handlers (not supported on the web)

### DIFF
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -154,6 +154,7 @@ ${queryParams
 
       getMetaInfo() {
         return {
+          type: 'rest',
           header: `[rest] ${method} ${mask.toString()}`,
           mask,
           callFrame,
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -155,6 +155,8 @@ ${queryParams
       getMetaInfo() {
         return {
           header: `[rest] ${method} ${mask.toString()}`,
+          mask,
+          method,
           callFrame,
         }
       },
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -25,6 +25,7 @@ import { getStatusCodeColor } from './utils/logging/getStatusCodeColor'
 import { isStringEqual } from './utils/internal/isStringEqual'
 import { matchRequestUrl } from './utils/matching/matchRequestUrl'
 import { getUrlByMask } from './utils/url/getUrlByMask'
+import { getCallFrame } from './utils/internal/getCallFrame'
 
 export enum RESTMethods {
   HEAD = 'HEAD',
@@ -72,6 +73,7 @@ const createRestHandler = (method: RESTMethods) => {
     ResponseBodyType
   > => {
     const resolvedMask = getUrlByMask(mask)
+    const callFrame = getCallFrame()
 
     return {
       parse(req) {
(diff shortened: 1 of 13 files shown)


kettanaito:
> Consider if getMetaInfo should be a function, as its return data is static

I'm leaving `getMetaInfo` as a method of request handlers as outputting a handler for preview may require additional computations, such as composing a proper header, deriving properties, etc. When `getMetaInfo` is a function, those computations can be done within it. If it's a plain Object, that data would have to be computed on the root level of the handler, meaning it always computes, even if you never list the handlers at all, making a slightly more expensive impact on the users.

src/rest.ts
```diff
@@ -149,6 +151,15 @@ ${queryParams
         console.log('Response', loggedResponse)
         console.groupEnd()
       },
+
+      getMetaInfo() {
+        return {
+          header: `[rest] ${method} ${mask.toString()}`,
+          mask,
+          method,
```
timdeschryver: Should `method` be returned here?

timdeschryver (dismissed):

src/rest.ts
kettanaito: I suppose we can drop it. Originally I thought it could be used in the REPL, but then realized it's expected/actual paths comparison at the moment, so a request's method is not of much use.

timdeschryver: I was only asking it because the property isn't required by the `RequestHandlerMetaInfo ` type.

kettanaito: Oh, a great point. I wasn't sure how to approach the meta data type-wise as I can see a potential room for arbitrary properties in case of custom request handlers. I need to type that better 💯

marcosvega91 (dismissed):
🤯  , great ! I like it :)

src/setupWorker/setupWorker.ts
```diff
@@ -127,5 +132,19 @@ export function setupWorker(
...
+
+    printHandlers() {
+      context.requestHandlers.forEach((handler) => {
+        const meta = handler.getMetaInfo()
+
+        console.groupCollapsed(meta.header)
+        console.log(`Declaration: ${meta.callFrame}`)
+        console.log('Resolver: %s', handler.resolver)
+        if (['rest'].includes(meta.type)) {
+          console.log('Match:', `https://mswjs.io/repl?path=${meta.mask}`)
```
timdeschryver, line 119: 🤩

timdeschryver (approved):

kettanaito:
Let's release this after the bugs are fixed:

- #466 
- #458

timdeschryver (dismissed):

PR #454, opened 2020-11-16
Implement virtual cookie store

chrisguttandin:
This PR is not really meant to be merged as is. It's just my quick and dirty implementation of a virtual cookie store. I only implemented what I needed for my specific use case. It does:

- store all cookies set by the mocked responses in memory
- scope cookies by origin
- add the appropriate cookies to each request depending on its `credentials` property
- delete expired cookies
- save cookies to local storage to make sure they survive a page reload

Given that the correct handling of cookies is very complex I think it would make sense to move all of this into a separate module. It could provide a virtual cookie store that has two methods which could have the following signature.

```ts
interface VirtualCookieStore {
    // This would add the appropriate cookies to a request.
    appendCookies(request: Request): void;

    // This would read cookies from the response to store them internally.
    storeCookies(request: Request, response: Response): void;
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
   "dependencies": {
     "@open-draft/until": "^1.0.3",
     "@types/cookie": "^0.4.0",
+    "@types/set-cookie-parser": "^2.4.0",
     "chalk": "^4.1.0",
     "chokidar": "^3.4.2",
     "cookie": "^0.4.1",
@@ -71,6 +72,7 @@
     "node-fetch": "^2.6.1",
     "node-match-path": "^0.5.2",
     "node-request-interceptor": "^0.5.1",
+    "set-cookie-parser": "^2.4.6",
     "statuses": "^2.0.0",
     "yargs": "^16.0.3"
   },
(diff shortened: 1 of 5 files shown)


kettanaito:
Hey, @chrisguttandin 👋  Happy to see you around.
Thank you for this feature, I'll review it at my first free slot!

kettanaito:
Hey, @chrisguttandin. Thank you once more for providing this implementation.

I've created a [separate repository](https://github.com/mswjs/cookies) to manage this a a separate package. I've also added you as a maintainer, so we can iterate on the cookie store together. 

Please, if you have time, could you test that repo and let me know whether it behaves how you've originated designed? There are a few unit tests, but I didn't get much into implementation details. Thank you.

marcosvega91:
Maybe we can provide different way of persistence, the default one is using local storage on a compatible environment but we can add more ways for non dom environment.

kettanaito:
@marcosvega91 I was thinking of using a singleton cookie store that would ensure persistency for non-DOM environments, where JS doesn't re-execute as it does in a browser. Do you think that would work?

marcosvega91:
I have looked into the code and I think is fine as it is implemented :), cool 🥳

kettanaito:
I've published the draft as the [`virtual-cookies`](https://www.npmjs.com/package/virtual-cookies) package.

You can try installing it and replacing the logic from the pull request using the package's API. Please let me know how that goes.

PR #458, opened 2020-11-18
Declares and exports SetupServerApi interface

### LINKED ISSUE #455: [TypeScript] missing return types of the setupServer method.
## Describe the bug
Hi!
I'm creating a mocked server as described in the guides:

```jsx
import { rest } from 'msw';

server = setupServer( ... my mocks )
```

I looked at [the code](https://github.com/mswjs/msw/blob/a446e9b1d5/src/node/createSetupServer.ts#L45), and it seems that an object is returned that contains all of the listen, use, close, etc. methods. So I understand from this that TypeScript simply does type inference when using the result of this method.

It could be great to have a type definition for this (I sometimes pass this server object to methods and don't want to use the any type).

Since I'm going to declare these types anyway in my project, I thought it would be better to just add them to this project so everybody could enjoy it.

What do you think?
Thanks 👍

ofekdeitch-fiverr:
## Description

Add a return type to the `setupServer` method.

### Why do we need this?
- Fixes https://github.com/mswjs/msw/issues/455

### DIFF
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -11,10 +11,8 @@ import { getResponse } from '../utils/getResponse'
 import { parseBody } from '../utils/request/parseBody'
 import { isNodeProcess } from '../utils/internal/isNodeProcess'
 import * as requestHandlerUtils from '../utils/handlers/requestHandlerUtils'
-import { SharedOptions } from '../sharedOptions'
 import { onUnhandledRequest } from '../utils/request/onUnhandledRequest'
-
-type ListenOptions = SharedOptions
+import { ListenOptions, ServerInstance } from './glossary'
 
 const DEFAULT_LISTEN_OPTIONS: ListenOptions = {
   onUnhandledRequest: 'bypass',
@@ -28,7 +26,9 @@ export function createSetupServer(...interceptors: Interceptor[]) {
   /**
    * Sets up a server-side requests interception with the given mock definition.
    */
-  return function setupServer(...requestHandlers: RequestHandlersList) {
+  return function setupServer(
+    ...requestHandlers: RequestHandlersList
+  ): ServerInstance {
     requestHandlers.forEach((handler) => {
       if (Array.isArray(handler))
         throw new Error(
@@ -52,7 +52,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
       /**
        * Enables requests interception based on the previously provided mock definition.
        */
-      listen(options?: ListenOptions) {
+      listen(options?: ListenOptions): void {
         const resolvedOptions = Object.assign(
           {},
           DEFAULT_LISTEN_OPTIONS,
@@ -125,21 +125,21 @@ export function createSetupServer(...interceptors: Interceptor[]) {
       /**
        * Prepends given request handlers to the list of existing handlers.
        */
-      use(...handlers: RequestHandlersList) {
+      use(...handlers: RequestHandlersList): void {
         requestHandlerUtils.use(currentHandlers, ...handlers)
       },
 
       /**
        * Marks all request handlers that respond using `res.once()` as unused.
        */
-      restoreHandlers() {
+      restoreHandlers(): void {
         requestHandlerUtils.restoreHandlers(currentHandlers)
       },
 
       /**
        * Resets request handlers to the initial list given to the `setupServer` call, or to the explicit next request handlers list, if given.
        */
-      resetHandlers(...nextHandlers: RequestHandlersList) {
+      resetHandlers(...nextHandlers: RequestHandlersList): void {
         currentHandlers = requestHandlerUtils.resetHandlers(
           requestHandlers,
           ...nextHandlers,
@@ -149,7 +149,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
       /**
        * Stops requests interception by restoring all augmented modules.
        */
-      close() {
+      close(): void {
         interceptor.restore()
       },
     }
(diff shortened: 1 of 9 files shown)


src/node/createSetupServer.ts
```diff
@@ -149,7 +149,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
       /**
        * Stops requests interception by restoring all augmented modules.
        */
-      close() {
+      close(): void {
```
kettanaito: I believe that both argument and return types are inferred from the `ServerInstance` return type you assign to the `setupServer` function declaration. There's no need to explicitly annotate their return types as `void`.

src/node/glossary.ts
```diff
@@ -0,0 +1,12 @@
+import { SharedOptions } from '../sharedOptions'
+import { RequestHandlersList } from '../setupWorker/glossary'
+
+export type ListenOptions = SharedOptions
```
kettanaito: We can remove the `ListenOptions` type as it's effectively a copy of the `SharedOptions`. Let's use the latter directly, unless we have some extra types to union with the shared options.

```ts
listen(options?: SharedOptions): void
```

src/node/glossary.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface ServerInstance {
```
kettanaito: Let's name this interface `SetupServerApi`, to be aligned with how a similar `setupWorker` return type is called (`SetupWorkerApi`):


https://github.com/mswjs/msw/blob/a851536cc00b23ce7c3a7d71cb5a02ae865adec0/src/setupWorker/setupWorker.ts#L8

kettanaito (changes requested):
Hey, @ofekdeitch. Thank you for preparing these improvements!
I've left a few minor comments, could you please have a look at them?

src/node/glossary.ts
marcosvega91: we can move then `SetupWorkerApi` in his glossary file? In this way we have the same structure

kettanaito: That's a sound suggestion, @marcosvega91!

ofekdeitch-fiverr:
Hey guys ( @kettanaito  @marcosvega91 ),
Fixed the comments 👍 

Please notice that I also moved the comments from the implementation of `SetupServerApi` to its definition in the glossary file (the same has been done before with `SetupWorkerApi`).

ofekdeitch-fiverr:
One last comment - I guess I should also bump the patch\minor version - whatever you guys prefer.

kettanaito:
@ofekdeitch-fiverr, thank you for addressing the comments!
No need to bump the version, that will be done by one of the library's maintainers once your improvement lands.

kettanaito:
I've rebased the feature branch and added an additional commit that drops the inferred argument types from the `setupWorker` return object. Let's wait for the CI.

src/node/glossary.ts
```diff
@@ -0,0 +1,29 @@
+import { SharedOptions } from '../sharedOptions'
+import { RequestHandlersList } from '../setupWorker/glossary'
+
+export interface SetupServerApi {
```
marcosvega91, line 4: To make the interface accessibile we should export from `node/index.ts`

src/setupWorker/glossary.ts
```diff
@@ -76,3 +78,22 @@ export type ResponseWithSerializedHeaders<BodyType = any> = Omit<
 > & {
   headers: HeadersList
 }
+export interface SetupWorkerApi {
```
marcosvega91, line 82: Same as `setup-worker` but from `index.ts`

test/msw-api/setup-worker/resetHandlers.test.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from 'path'
 import { rest } from 'msw'
 import { runBrowserWith } from '../../support/runBrowserWith'
-import { SetupWorkerApi } from '../../../src/setupWorker/setupWorker'
+import { SetupWorkerApi } from '../../../src/setupWorker/glossary'
```
marcosvega91:
```suggestion
import { rest, SetupWorkerApi } from 'msw'
```

test/msw-api/setup-worker/restoreHandlers.test.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from 'path'
 import { rest } from 'msw'
 import { runBrowserWith } from '../../support/runBrowserWith'
-import { SetupWorkerApi } from '../../../src/setupWorker/setupWorker'
+import { SetupWorkerApi } from '../../../src/setupWorker/glossary'
```
marcosvega91:
```suggestion
import { rest, SetupWorkerApi } from 'msw'
```

test/msw-api/setup-worker/use.test.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from 'path'
 import { rest } from 'msw'
 import { runBrowserWith } from '../../support/runBrowserWith'
-import { SetupWorkerApi } from '../../../src/setupWorker/setupWorker'
+import { SetupWorkerApi } from '../../../src/setupWorker/glossary'
```
marcosvega91:
```suggestion
import { rest, SetupWorkerApi } from 'msw'
```

marcosvega91 (changes requested):
We should export the new interfaces to make them accessibile from the library ?

src/node/glossary.ts
kettanaito: Thank you for the suggestion. Addressed.

kettanaito:
Hey, @ofekdeitch-fiverr. I've rebased the branch against the latest `master` and provided some of the suggestions Marco has mentioned above. I hope it's okay.

kettanaito (dismissed):

kettanaito:
I trust we can merge this once the CI is green.

PR #463, opened 2020-11-19
Parses JSON req/res body if its content-type includes 'json'

### LINKED ISSUE #462: [v0.22.0] Not parsing JSON body
## Describe the bug
Hi, I am not sure if this is a bug or was a planned changed but after upgrading msw to the latest version, part of my tests started to fail. I fount this [commit](https://github.com/mswjs/msw/commit/e6b60c0fbc467850ca3a4cbd117ad900bf19dd66) that changes the condition to parse the body.

Now when we have the following header:
```
Content-Type: application/json; charset=UTF-8
```

The body is not parsed.

## Environment

- `msw: 0.22.0`
- `nodejs: 12.19.0`
- `npm: 6.14.8`

## To Reproduce
1. Create a REST request handler
2. Send a request with json body and add header `Content-Type: application/json; charset=UTF-8`
3. `req.body` inside the request handler will be a string

## Expected behavior
When the `content-type` header is `application/json`, the body is automatically parsed.

otaciliolacerda:
This PR fixes #462. 

Version 0.22.0 added this [commit](https://github.com/mswjs/msw/commit/e6b60c0fbc467850ca3a4cbd117ad900bf19dd66) which changes the condition to parse the body.

So given the header:
```
Content-Type: application/json; charset=UTF-8
```

The body was not parsed.

This change rollback to the previous behaviour where we use `includes` instead of `endsWith`

### DIFF
--- a/src/utils/request/parseBody.test.ts
+++ b/src/utils/request/parseBody.test.ts
@@ -22,6 +22,17 @@ test('parses a body if the "Content-Type*/json" header is set', () => {
   })
 })
 
+test('parses a body if the "Content-Type:application/json; charset=UTF-8" header is set', () => {
+  expect(
+    parseBody(
+      `{"property":2}`,
+      new Headers({ 'Content-Type': 'application/json; charset=UTF-8' }),
+    ),
+  ).toEqual({
+    property: 2,
+  })
+})
+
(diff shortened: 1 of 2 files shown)


kettanaito (approved):
Thank you for the fix and especially for the test suite, @otaciliolacerda!

kettanaito:
Welcome to contributors! 🎉

PR #467, opened 2020-11-20
Supports async response transformers

### LINKED ISSUE #465: Support asynchronous response transformers
## What

I suggest to consider adding asynchronous support to [response transformers](https://mswjs.io/docs/basics/response-transformer).

## Why

- Example in #457 

Creating a custom `jpeg` response transformer includes converting a Base64 string into an `ArrayBuffer` via `fetch`, which is an asynchronous operation. This makes it not exactly suitable in the response transformer.

However, if response transformers can be async function it becomes much easier.

```js
import { context } from 'msw'

export const jpeg = async (base64) => {
  const buffer = await fetch(base64).then(res => res.arrayBuffer())

  return (res) => {
    context.set('Content-Length', buffer.bufferLength.toString())(res)
    context.body(buffer)(res)
  }
}

res(jpeg(base64))
```

kettanaito:
## Changes

- Now it's possible to create async response transformers.
- Exports `compose` function publicly to compose multiple standard response transformers as a part of a custom response transformer.
- Adds browser integration tests for an async response transformer.
- Handles `response` calls as potentially async in unit tests.
- Aligns unit tests to be shallow (drops `describe` blocks).

## GitHub

- Closes #465 

## Roadmap

- [x] Annotate `compose` properly (infer function type, set return type)

### DIFF
--- a/test/msw-api/context/async-response-transformer.mocks.ts
+++ b/test/msw-api/context/async-response-transformer.mocks.ts
@@ -1,0 +1,20 @@
+import { ResponseTransformer, setupWorker, rest, context, compose } from 'msw'
+import base64Image from 'url-loader!../../fixtures/image.jpg'
+
+async function jpeg(base64: string): Promise<ResponseTransformer> {
+  const buffer = await fetch(base64).then((res) => res.arrayBuffer())
+
+  return compose(
+    context.set('Content-Length', buffer.byteLength.toString()),
+    context.set('Content-Type', 'image/jpeg'),
+    context.body(buffer),
+  )
+}
+
+const worker = setupWorker(
+  rest.get('/image', async (req, res, ctx) => {
+    return res(ctx.status(201), await jpeg(base64Image))
+  }),
+)
+
+worker.start()
--- a/test/msw-api/context/async-response-transformer.mocks.ts
+++ b/test/msw-api/context/async-response-transformer.mocks.ts
@@ -1,4 +1,11 @@
-import { ResponseTransformer, setupWorker, rest, context, compose } from 'msw'
+import {
+  ResponseTransformer,
+  setupWorker,
+  rest,
+  context,
+  compose,
+  createResponseComposition,
+} from 'msw'
 import base64Image from 'url-loader!../../fixtures/image.jpg'
 
 async function jpeg(base64: string): Promise<ResponseTransformer> {
@@ -11,9 +18,23 @@ async function jpeg(base64: string): Promise<ResponseTransformer> {
   )
 }
 
+const customResponse = createResponseComposition(null, [
+  async (res) => {
+    res.statusText = 'Custom Status Text'
+    return res
+  },
+  async (res) => {
+    res.headers.set('x-custom', 'yes')
+    return res
+  },
+])
+
 const worker = setupWorker(
   rest.get('/image', async (req, res, ctx) => {
     return res(ctx.status(201), await jpeg(base64Image))
+  }),
+  rest.post('/search', (req, res, ctx) => {
+    return customResponse(ctx.status(301))
   }),
 )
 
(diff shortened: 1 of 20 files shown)


test/msw-api/context/async-response-transformer.mocks.ts
```diff
@@ -0,0 +1,20 @@
...
+  return compose(
+    context.set('Content-Length', buffer.byteLength.toString()),
+    context.set('Content-Type', 'image/jpeg'),
+    context.body(buffer),
+  )
+}
+
+const worker = setupWorker(
+  rest.get('/image', async (req, res, ctx) => {
+    return res(ctx.status(201), await jpeg(base64Image))
```
kettanaito, line 34: I find it extremely readable to await the transformer upon invocation, as opposed to drilling asynchronicity all the way down to the `response` composition tree.

test/msw-api/context/async-response-transformer.mocks.ts
```diff
@@ -0,0 +1,20 @@
+import { ResponseTransformer, setupWorker, rest, context, compose } from 'msw'
+import base64Image from 'url-loader!../../fixtures/image.jpg'
+
+async function jpeg(base64: string): Promise<ResponseTransformer> {
```
kettanaito, line 11: This is an example of a custom `jpeg` response transformer that appends a given Base64 image to the mocked response. It utilizes newly exported `compose` function, as well as the ability to make the transformer async.

timdeschryver (commented):
Don't we have to `await` somewhere (I think in `compose`)?

My thoughts:
`ResponseTransformer` can now be a promise  and is a parameter (`defaultTransformers`) for `createResponseComposition`. This parameter is used to build the `resolvedTransformers`, which is passed to the `compose` method. So, when we're in `compose`, we still have the promise, and we're not `awaiting` its result.

(I could be wrong tho, I didn't test it)

kettanaito:
Thanks for the feedback, @timdeschryver.

> Don't we have to await somewhere (I think in compose)?

Not with the current implementation. We allow response transformer to be async, but it must be awaited within the response resolver (which is already supported as async). That way the list of transformers that we supply to `compose` consists of already resolved transformer functions and there's no need to await in the `compose` itself:

```js
async (req, res, ctx) => {
  // `res` will get the resolved value of `customTransformer`, not the Promise
  return res(await customTransformer())
}
```

> ResponseTransformer can now be a promise and is a parameter (defaultTransformers) for createResponseComposition. This parameter is used to build the resolvedTransformers, which is passed to the compose method.

That's a great point. I'll add a test that ensures that async transformer passed to the custom response composition function still executes properly 👍

kettanaito:
What do you think about this extension to the `compose` function?

https://github.com/mswjs/msw/blob/d526e8e6a575e0f6cec2b63ea57b62fca4d9facf/src/utils/internal/compose.ts#L5-L17

I didn't want to make it always return a Promise, which would ruin a synchronous execution. At the same time if any of the functions returns a Promise the `compose` returns a Promise accordingly. 

> I only wish we could still annotate `compose` so it understands that return signature. 

I've also added an integration test for an async response transformer as the value of `defaultTransformers` of a custom response.

timdeschryver (dismissed):
LGTM, nice @kettanaito!

kettanaito:
Experiencing some TS issues during the compilation of an Angular project as a part of smoke tests.

timdeschryver:
@kettanaito this could be due to miss matching typescript versions.
I'm going to upgrade the angular example to the new angular version, which also supports TS 4.x.

kettanaito:
This is amazing, @timdeschryver! Thank you for figuring that out. I'll review your Angular pull request.

timdeschryver:
No problem! It was an error that I've had seen before 😅

PR #468, opened 2020-11-20
fix: do not stringify res.body if content is string

### LINKED ISSUE #466: [v0.22.x] Regression on parsing response
## Describe the bug
Hi again. 😀
For what I got the new version now supports the definition of custom transformers to parse the response. From [here](https://github.com/mswjs/msw/pull/441/files#diff-8c103f2199265c6a0659504dac715a5bc176ab1f494fc64fb06a5f4a814094c0R64) I see we have a default transformer that will stringify the response body if the content-type header ends with "json". This is a regression from version 0.21.3 since now the `ctx.body` does not work as expected. From the docs:
```
ctx.body()
Sets a raw response body. Does not apply any Content-Type headers or other response transformations.
```

So giving:
```js
rest.get('api/resource/:id', (req, res, ctx) => {
    const status = 400;
    const problemJson = {
       type: 'app-error-code',
       title: 'Mock response provided by msw',
       status,
       detail: 'Mock response provided by msw',
       instance: 'string',
    };
    return res(
      ctx.status(status),
      ctx.body(JSON.stringify(problemJson)),
      ctx.set('Content-Type', 'application/problem+json')
    );
  }),
```
This will not work because now the default transformer sees that the content-type header ends with "json" and will stringify the body again. That will surround my string body with double quotes: `""MY_STRINGIFIED_BODY""`.

Changing the `ctx.body(JSON.stringify(problemJson))` to `ctx.body(problemJson)` works. Before version 0.21.3 it was working as the documentation specifies.

I can work in a PR to change this but it depends on which direction it should go:
- Just update the docs to make it clear
- Change the `stringifyJsonBody` function to not stringify again if the `res.body` is already a string.

## Environment

- `msw: 0.22.1`
- `nodejs: 12.19.0`
- `npm: 6.14.8`

## Expected behavior
Avoid stringify the `res.body` more than once.

otaciliolacerda:
This PR fixes #466.

### Summary

From the docs:
```
ctx.body()
Sets a raw response body. Does not apply any Content-Type headers or other response transformations.
```

This PR changes the default transformer to not stringify the response body if the content is already a string. This way we do not surround the existing content with quotes.

It covers the following use case:
```js
res(
  ctx.body('Lorem ipsum'), 
  ctx.set('Content-Type', 'application/problem+json')
)
```


marcosvega91 (dismissed):
Hi @otaciliolacerda  thanks for your time fixing this :) It is good for me :)

kettanaito:
The changes look good. If you don't mind, I'd like to merge https://github.com/mswjs/msw/pull/467 first, as it refactors unit tests for response transformers a little.

src/context/json.test.ts
```diff
@@ -14,7 +14,7 @@ test('sets given Array as the response JSOn body', () => {
 
 test('sets given string as the response JSON body', () => {
   const result = response(json('some string'))
-  expect(result).toHaveProperty('body', `"some string"`)
+  expect(result).toHaveProperty('body', 'some string')
```
kettanaito: Do we drop the extra quotes assuming the body is going to be stringified later on by the default response transformer?

otaciliolacerda: Hi @kettanaito, the `json` function does not stringify the content, the responsible to do that is the response transformer. In this case the default response transformer.

The difference is that the default transformer checks the content is already a string so it does not stringify it again (cause of the extra quotes).

kettanaito: Can we ensure that parsing of such JSON response body works as intended?

For example, `JSON.parse('some string')` would throw an exception as the given string indeed lacks those extra quotes. Since here we are asserting on the `response()` invocation result, that's what the consumer may get. 

There is no additional response transformation since the response composition moment: body gets transferred to the worker script and a `Response` instance is constructed with that body:

```js
new res = new Response('some string')
await res.json()
```

The example above will throw a JSON parsing exception as the body string misses extra quotes around it to be a valid JSON.

otaciliolacerda: Hi @kettanaito I've got you point and it makes sense but I am not sure how to make the change though. 

To be honest I am not that familiar with `msw` internals so I was reading the code and the past PRs and trying to figure that out. I saw that PR #403 changed the `ctx.json` which, as expected, was the responsible to stringify the body content. Now, it's responsible to set the header `content-type: application/json` and enable merges. 

To me the merge functionality seems like something we could move to a utils function so when calling `json`, `data` or `errors` functions we could have something like:
```js
ctx.json(merge(data1, data2, data3))
```
It seems the project already have a function like this (`mergeRight`) so it was just a matter of export the function and update the docs.

Also, correct me if I am wrong, but it seems the response transformers were added just to enable different json parsers in the body (#440). If the merge was a util/helper function we could solve the problem just defining a custom resolve transformer like:
```js
import { rest } from 'msw'
import * as JSONbig from 'json-bigint'

const jsonBigInt = body => {
  return (res) => {
    res.headers.set('Content-Type', 'application/json')
    res.body = JSONbig.stringify(body)
    return res
  }
}

const server = setupServer(
  rest.get('http://test.mswjs.io/me', (req, res, ctx) => {
    const me = {
      username: 'Dude',
      balance: BigInt(1597928668063727616),
    }
    return res(ctx.jsonBigInt(me))
  }),
)
```

This pattern can be use to extend the available context functions for different use cases. This is what I do in my projects.

I know is not that simple to change the API of a open source library so as an alternative solution we could create a flag in the default response:

```ts
export interface MockedResponse<BodyType = any> {
  body: BodyType
  status: number
  statusText: string
  headers: Headers
  once: boolean
  isJson: boolean // new flag to point to json responses
  delay?: number
}
```

The default would be **false** and we would set it to **true** in the `json` function. Then we could check for the flag in the default response transformer like:
```ts
const stringifyJsonBody: ResponseTransformer = (res) => {
  if (res.body && res.isJson) {
    res.body = JSON.stringify(res.body)
  }
  return res
}
```

Notice we don't need to check for the header anymore. We only guarantee the body to be stringified if the `ctx.json` function is used. In case the users set the content-type header with any json variation manually and use `ctx.text`  or `ctx.body` to set the body, the user would need to stringify the body themselves.


I personally try to avoid flags as much as I can because they add complexity.

Let me know what you think.

kettanaito: It's a little different than that. We are exposing an explicit set of context utilities depending on the request handler (rest/graphql). That way your mocks resemble a spec-compliant response composition respective to the handler used. For instance, in `graphql.*` there's no `ctx.json` utility, as you cannot set an arbitrary JSON as a GraphQL endpoint response, it must follow the spec. 

That was the reason #403 got introduced. Under the hood it meant an internal extension of the `ctx.json` utility that powers `ctx.data` and `ctx.errors` under the hood, to merge all the given objects of the response body. During the implementation we realized that in order to operate with those objects in a more comfortable fashion they should be stored as objects, not strings. Thus stringification of the JSON response body was moved to be the last step of the response composition function.

When it comes to this test, I believe this becomes a breaking change, as the response body is not a valid JSON. 

> If the merge was a util/helper function we could solve the problem just defining a custom resolve transformer like:

This is a good point. I like the idea of being able to incorporate certain fixed logic to the response composition function itself, but such custom response transformer also satisfies the use case. We should seriously consider this.

> I personally try to avoid flags as much as I can because they add complexity.

I couldn't agree more. I don't think that extending the mocked response object is the way to go. 

Do you think something like a custom default response transformer (like the current JSON BigInt) has a place to be? Or can any use case be substituted by a custom transformer?

kettanaito: Regarding the `isJson` property, I find it confusing, as this information should be extracted from the response's headers, treating them as the source of truth. To brainstorm, here are the factors we have:

- `ctx.body` should set a given body as-is, not bothering with its stringification.
- `ctx.json` must always result into a valid JSON body (stringified).
- `ctx.json` may potentially support cross-transformer merges. This item may be obsolete if we fully replace default response transformers with custom `ctx.*` functions developers may create.
- `typeof res.body === 'string` is insufficient to understand whether the response body should be stringified or not. Strings are also a valid JSON.

kettanaito: @msutkowski, hey, Matt 👋 What do you think about the points above?

kettanaito:
Hey, @otaciliolacerda. I've rebased your feature branch against the latest `master`. 

The current tests will fail, as I've retained the stringification of a JSON body according to our discussion above. 

```diff
-'some string'
+ '"some string"'
```

kettanaito:
I've pushed the rework of how we should handle merged JSON for `ctx.data` and `ctx.errors` utility, as well as unified the stringification of the JSON response body.

- `ctx.json` now always stringifies the given body.
- `ctx.data` and `ctx.errors` now attempt to parse the previously set body, then merge their bodies into the previous one to compose a single JSON body.
- Removes the default JSON body response transformer. The response body is not implicitly stringified, no matter the `Content-Type` header.
- Suggests to use a custom context utility than a default response transformer to handle the `BigInt` scenario (custom `JSONBig.stringify` function).

kettanaito:
@otaciliolacerda, hey 👋 Could you please use the Codesandbox PR build above to test that your usage scenario works as intended now? This PR also contains a failing test with your use case, which now should be passing. Thanks.

src/context/body.test.ts
```diff
@@ -7,3 +8,12 @@ test('sets a given body value without implicit "Content-Type" header', async ()
...
 })
+
+test('does not stringify raw body twice if content is string and "Content-Type" header is "json"', async () => {
+  const result = await response(
+    set('Content-Type', 'application/json'),
+    body(JSON.stringify('some text')),
+  )
+
+  expect(result).toHaveProperty('body', `"some text"`)
+})
```
marcosvega91, line 19: we need this test anymore?

src/context/data.ts
```diff
@@ -7,5 +9,10 @@ export type DataContext<T> = (payload: T) => ResponseTransformer
  * Returns a GraphQL body payload.
  */
 export const data: DataContext<Record<string, any>> = (payload) => {
-  return json({ data: payload }, { merge: true })
+  return (res) => {
+    const prevBody = jsonParse(res.body) || {}
+    const nextBody = mergeRight(prevBody, { data: payload })
+
+    return json(nextBody)(res)
```
marcosvega91, line 16: We should add somewhere a message if the user has passed an object into res.body?

The `jsonParse` function will return undefined in this way, maybe we can add a message into the catch

marcosvega91 (changes requested):
Makes sense to me, so here we are only dropping the `merge` functionality right?

src/context/body.test.ts
kettanaito: We still need to ensure that the response body doesn't get stringified twice, no matter the header. This is mainly to prevent regressions in the future, as this test was missing and we didn't catch this issue.

src/context/data.ts
kettanaito: > if the user has passed an object into res.body?

This would violate the rule that the response body must always be set as a string. This affects both standard context utilities and the ones you may have created by yourself. With these changes `req.body` cannot be anything other than string or `null`.

otaciliolacerda:
Hi @kettanaito sorry for the long delay. I went out on vacation and did not have a connection for a few days. I did not test yet but looking at the code changes it seems good now. Thanks.

kettanaito:
Hey, @otaciliolacerda. No worries. Thank you for the feedback on this. Our discussion has highlighted that there has been a mistake in handling JSON as objects internally, and I'm glad this surfaced. Let me know if the issue is fixed once you try it out. Cheers.

slaweet:
As an end-user of msw, I don't understand the Migration guidelines in v0.24.0. I have a simple use of ctx.json as in https://mswjs.io/docs/api/context/json#examples - in my case: `return res(ctx.json(mock.data));`

After upgrading to msw v0.24.0, typescript is complaining
```
94:18 Argument of type 'ResponseTransformer<string>' is not assignable to parameter of type 'ResponseTransformer<unknown>'.
  Types of parameters 'res' and 'res' are incompatible.
    Type 'MockedResponse<unknown>' is not assignable to type 'MockedResponse<string>'.
      Type 'unknown' is not assignable to type 'string'.
    92 |     const mock = getMatchingMock(method, req.url.pathname, params);
    93 |     if (mock) {
  > 94 |       return res(ctx.json(mock.data));
       |                  ^
    95 |     }
    96 |   }) as RequestHandler<any, any, unknown, any, unknown>
    97 | ));

```

Can you please update the release "Migration guidelines" in [release notes of v0.24.0](https://github.com/mswjs/msw/releases/tag/v0.24.0) to say what to do so that the type of return value of `ctx.json` matches the type of the argument of `res`?

kettanaito:
Thanks for reporting this, @slaweet. I believe this was a breaking change:

https://github.com/mswjs/msw/commit/1bd21c2b1f7292029b6da55e757dc313250170a7#diff-2d1882322615ff8e09b8a2e5ab407a7ea552f3d0f82b5457c196978a7952ddacR12

However, I don't think it's executed properly. The intention is that `res.body` is always a string, so the response transformer `ctx.json` returns annotates the body as a `ResponseTransformer<string>` generic, opposed to the previous value of `BodyTypeJSON` inferred from the given object. Looks like this causes type incompatibility with the existing code, but mist importantly doesn't annotate things correctly, as the code you have above should be 100% valid.

PR #469, opened 2020-11-22
feat: implement virtual cookie store

### LINKED ISSUE #435: add a virtual cookie store

Let's say there are two mocked routes which look like this:

```js
rest.get('https://example.org/set-cookie', (_, res, ctx) => {
    return res(ctx.cookie('key', 'value'));
}),
rest.get('https://example.org/require-cookie', (req, res, ctx) => {
    if (req.cookies.key === 'value') {
        return res(ctx.status(204));
    }

    return res(ctx.status(400));
}),
```

When invoking those routes with two chained request ...

```js
fetch('https://example.org/set-cookie').then(() => fetch('https://example.org/require-cookie'))
```

... the second request will get a 400 response since there is no cookie header send with the request.

As far as I understand, this currently doesn't work since [the cookie is set by using document.cookie](https://github.com/mswjs/msw/blob/master/src/context/cookie.ts#L19). This means that the cookie will always be associated with the domain of the current document which is not necessarily the domain the response came from. There seems to be no way to change this behavior.

I would therefore like to propose adding a virtual cookie store. It would keep track of the cookies set by calling `ctx.cookie()` and it would automatically set the appropriate header to each mocked request to the same domain.

As mentioned above I think there is no alternative right now since setting cookies via JavaScript is restricted for good reasons.

**Additional context**

I'm happy to start a PR for this if we conclude that this is something useful.

chrisguttandin:
- Closes #435 

This PR uses the virtual-cookies package to implement a virtual cookie store.

It does so far hydrate the store before setting the cookies on a request and it persists the cookies after reading them from a response. This could probably be done more cleverly.

I needed to enable the definitions for `ES2019.Object` to allow the usage of `Object.fromEntries()`.

I also needed to cast the MockedRequest and MockedResponse instances since the virtual-cookies package expects a Request and a Response instead. Maybe we could define two types with the minimal set of common properties which would be expected to be used by the  virtual-cookies package instead.

### DIFF
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -16,6 +16,9 @@ import { ServerLifecycleEventsMap, SetupServerApi } from './glossary'
 import { SharedOptions } from '../sharedOptions'
 import { uuidv4 } from '../utils/internal/uuidv4'
 import { MockedRequest, RequestHandler } from '../handlers/RequestHandler'
+import { setRequestCookies } from '../utils/request/setRequestCookies'
+import { readResponseCookies } from '../utils/request/readResponseCookies'
+import { MockedResponse } from '../response'
 
 const DEFAULT_LISTEN_OPTIONS: SharedOptions = {
   onUnhandledRequest: 'bypass',
@@ -101,7 +104,8 @@ export function createSetupServer(...interceptors: Interceptor[]) {
             credentials: 'same-origin',
           }
 
-          emitter.emit('request:start', mockedRequest)
+          // Attach all the cookies stored in the virtual cookie store.
+          setRequestCookies(mockedRequest)
 
           if (requestCookieString) {
             // Set mocked request cookies from the `cookie` header of the original request.
@@ -110,6 +114,8 @@ export function createSetupServer(...interceptors: Interceptor[]) {
             // Unlike browser, where interception is on the worker level, _before_ the request happens.
             mockedRequest.cookies = cookieUtils.parse(requestCookieString)
           }
+
+          emitter.emit('request:start', mockedRequest)
 
           if (mockedRequest.headers.get('x-msw-bypass')) {
             emitter.emit('request:end', mockedRequest)
@@ -145,6 +151,9 @@ export function createSetupServer(...interceptors: Interceptor[]) {
               headers: response.headers.getAllHeaders(),
               body: response.body as string,
             }
+
+            // Store all the received response cookies in the virtual cookie store.
+            readResponseCookies(mockedRequest, response)
 
             // the node build will use the timers module to ensure @sinon/fake-timers or jest fake timers
             // don't affect this timeout.
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "node-request-interceptor": "^0.6.3",
     "statuses": "^2.0.0",
     "strict-event-emitter": "^0.1.0",
+    "virtual-cookies": "^0.1.2",
     "yargs": "^16.2.0"
   },
   "devDependencies": {
--- a/src/utils/request/getRequestCookies.ts
+++ b/src/utils/request/getRequestCookies.ts
@@ -9,6 +9,13 @@ function getAllCookies() {
  * Returns relevant document cookies based on the request `credentials` option.
  */
 export function getRequestCookies(request: MockedRequest) {
+  /**
+   * @note No cookies persist on the document in NodeJS: no document.
+   */
+  if (typeof location === 'undefined') {
+    return {}
+  }
+
   switch (request.credentials) {
     case 'same-origin': {
       // Return document cookies only when requested a resource
--- a/src/utils/request/parseWorkerRequest.ts
+++ b/src/utils/request/parseWorkerRequest.ts
@@ -1,7 +1,7 @@
 import { Headers } from 'headers-utils'
 import { MockedRequest } from '../../handlers/RequestHandler'
 import { ServiceWorkerIncomingRequest } from '../../setupWorker/glossary'
-import { getRequestCookies } from './getRequestCookies'
+import { setRequestCookies } from './setRequestCookies'
 import { parseBody } from './parseBody'
 import { pruneGetRequestBody } from './pruneGetRequestBody'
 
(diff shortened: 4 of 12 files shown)


src/utils/handleRequestWith.ts
```diff
@@ -64,7 +64,18 @@ export const handleRequestWith = (
       req.body = parseBody(req.body, req.headers)
 
       // Set document cookies on the request.
-      req.cookies = getRequestCookies(req)
+      store.hydrate()
+      req.cookies = Object.fromEntries(
```
kettanaito: Can we move this logic into `getRequestCookies` function? 
`req.cookies` and `req.headers.set` blocks are quite verbose, and are inseparable, so I'd vote for a standalone function that we could unit test.

kettanaito: Alternatively, we could have a designated method on the cookie store to either:

A. Return a `{ cookies, headers }` object already derived from the internal store values.
B. Accept a request and set those properties on it implicitly.

src/utils/handleRequestWith.ts
```diff
@@ -92,6 +103,9 @@ export const handleRequestWith = (
         return channel.send({ type: 'MOCK_NOT_FOUND' })
       }
 
+      store.add(<Request>(<unknown>req), <Response>(<unknown>response))
+      store.persist()
```
kettanaito: With the cookie store persisting on each request handling, we need to seriously consider failure scenarios. 

- Local storage entry may become corrupted.
- Failures to sync cookies with the storage must not raise an exception and fail a request.

Perhaps we could hydrate/persist at a different step? What advantages do you see in doing so upon each request handling?

kettanaito:
> Maybe we could define two types with the minimal set of common properties which would be expected to be used by the virtual-cookies package instead.

💯 That is a great idea. We should allow compatible request/response instances in `virtual-cookie`.

kettanaito (changes requested):
Thank you for putting this down, @chrisguttandin. I've left a few comments and appreciate if you took a look.

chrisguttandin:
I created mswjs/virtual-cookies#3 to add the minimal required type definitions.

src/utils/handleRequestWith.ts
chrisguttandin: I created a `setRequestCookies()` and a `readResponseCookies()` function. I think that also resolves the comment below.

chrisguttandin:
I made another update to update `virtual-cookies` to v0.1.2.

kettanaito:
I've migrated from the `virtual-cookies` package to `@mswjs/cookies`. Will give this a final review and it will be ready for merge.

src/utils/handleRequestWith.ts
kettanaito: I suppose a corrupted persisted value is handled as a part of https://github.com/mswjs/cookies/pull/2.

kettanaito:
@chrisguttandin, should we support cookie storage for requests in NodeJS as well?

Looks like we count on running the package in NodeJS: 
https://github.com/mswjs/cookies/blob/7ab4213ce5c35c5791e0f7fd437b22f09fbb042a/src/CookieStore.ts#L24

Should we add the usage of the store to request handling in NodeJS?

https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/node/createSetupServer.ts#L62

https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/node/createSetupServer.ts#L89

chrisguttandin:
> @chrisguttandin, should we support cookie storage for requests in NodeJS as well?

I'm not sure what the semantics of the Node.js support are. Should it emulate the behavior of a browser? In that case we might need to add it to `setupServer()` as well.

@kettanaito  What do you think?

kettanaito:
Hey, @chrisguttandin. Yes, I believe it should emulate the browser's behavior to the possible extent. There are two cases when `setupWorker` may be used:

- In a DOM-like environment, which is still NodeJS, but with some DOM object polyfills.
- In a pure NodeJS environment, like when developing a NodeJS server.

In the first case, there is a `document` object that we can use as a cookie container. In fact, we're already. using it like such in the `ctx.cookie` utility: 

https://github.com/mswjs/msw/blob/91cefafafe178f233d30c50ccfb8d249cb2c7db9/src/context/cookie.ts#L17-L19

If it's a pure NodeJS environment, the `document` doesn't exist, and the cookies don't persist anywhere. In that case, we don't do anything during persistency/hydration. 

I think we can add the same logic you've added to the in-browser request handling to `setupServer`, with the exception that it'd check the existence of `window` to confirm we are in a DOM-like environment. 

What do you think about this approach?

chrisguttandin:
Hi @kettanaito, I think this is a good approach. However I think I need some guidance on how to do it.

I just rebased this branch and noticed that the tests are broken as well. I'm afraid I don't really know why they are failing.

I would appreciate your help. Thanks in advance.

chrisguttandin:
Should I add the virtual cookie store [here](https://github.com/mswjs/msw/blob/f418117669ce40cc3d454759b1ccf570b35df1d1/src/node/createSetupServer.ts#L109-L114) as well?

kettanaito:
I've rebased the feature branch to be in-sync with the latest updates to the Request Handler API. Not many changes were necessary (mainly imports to `RequestHandler.ts`).

kettanaito:
I've also added an integration test that verifies that the cookies are inherited on a request once they are set in a response to a different request (according to your usage scenario #435). The test is passing, which confirms your client-side addition is a great one!

kettanaito:
The build is failing due to some caching issue, builds and all tests passing locally. 

_Edit:_ the build failed due to the case-sensitive module names (`handlers/requestHandler` -> `handlers/RequestHandler`).

kettanaito:
The branch now should have an integration test for the virtual cookie store in NodeJS and its implementation. Turns out the current state of `virtual-cookie` can work in NodeJS quite nicely. The only thing I adjusted is the `getRequestCookies` function that now returns an empty object (`{}`) when run in NodeJS (no document to retrieve the cookies from).

src/node/createSetupServer.ts
```diff
@@ -111,6 +115,8 @@ export function createSetupServer(...interceptors: Interceptor[]) {
             mockedRequest.cookies = cookieUtils.parse(requestCookieString)
           }
 
+          emitter.emit('request:start', mockedRequest)
```
kettanaito, line 118: Moving the emit of the `request:start` even so that its listener would accept an already transformed request (all cookies attached, any other coercions and transformations finished).

kettanaito (approved):
This is a great addition! Thank you, @chrisguttandin, for your work on this!

PR #470, opened 2020-11-22
fix: duplicate parameter in function with default parameter values

psoch:
**Summary**

When running a msw service worker with storybook in `Safari` I get the following error: `SyntaxError: Duplicate parameter 'onUnhandledRequest' not allowed in function with default parameter values.` It seems to be related to usage of the same function and parameter name with a default value in this [file](https://github.com/mswjs/msw/blob/e1656555e925d672b576e974716ad8e5e0a8aa18/src/utils/request/onUnhandledRequest.ts#L14). 

(For reference I found a similar type of issue in safari in another library https://github.com/jupyterlab/jupyterlab/pull/7993)

### DIFF
--- a/src/utils/request/onUnhandledRequest.ts
+++ b/src/utils/request/onUnhandledRequest.ts
@@ -11,10 +11,10 @@ export type OnUnhandledRequest =
 
 export function onUnhandledRequest(
   request: MockedRequest,
-  onUnhandledRequest: OnUnhandledRequest = 'bypass',
+  handler: OnUnhandledRequest = 'bypass',
 ): void {
-  if (typeof onUnhandledRequest === 'function') {
-    onUnhandledRequest(request)
+  if (typeof handler === 'function') {
+    handler(request)
     return
   }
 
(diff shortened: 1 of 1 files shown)


src/utils/request/onUnhandledRequest.ts
```diff
@@ -11,10 +11,10 @@ export type OnUnhandledRequest =
 
 export function onUnhandledRequest(
   request: MockedRequest,
-  onUnhandledRequest: OnUnhandledRequest = 'bypass',
+  onUnhandledReq: OnUnhandledRequest = 'bypass',
```
kettanaito: Let's name this parameter `handler`.

That way we don't have to shorten it while still preserving its semantic meaning: a function that handles an `onUnhandled` event.

kettanaito (changes requested):
Hi, @psoch. Thank you for creating a pull request with this improvement!

I've left one comment, could you please take a look?

kettanaito (approved):
This is awesome! Thank you, @psoch, and welcome to contributors!

psoch:
Thank you @kettanaito for the feedback, quick response, and great library!

PR #396, opened 2020-09-23
Adds WebSocket support

kettanaito:
## GitHub

- Closes #156 

## Goal

Provide a designated API that would allow the mocking of a WebSocket server. Keep the API as close to the actual server's API as possible, to minimize the library-specific knowledge needed to operate with this API.

## Changes

- Adds basic classes for WebSocket support.
- Adds integration tests.

## Roadmap

- [ ] Support execution in Node.js.
  - [ ] Utilize `WebSocketStorage` class as a singleton to store all server references.
  - [ ] There is no `Event` in Node.js.
  - [x] There is no `BroadcastChannel` in Node.js (not needed, we can use singletons in Node.js).
- [ ] Integration tests
  - [ ] Server `error` event.
  - [x] Server `close` event.
  - [ ] Ensure client does not receive messages when it's being closed (gets removed from the internal `connections` list of the server).
  - [x] Test multi-tab event propagation
- [ ] Ensure there are no memory leaks when multiple clients (tabs) are opened.

PR #471, opened 2020-11-22
Add codesandbox ci config

msutkowski:
Just a test to see if we can get CSB CI to build PRs 🎉

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,0 +1,3 @@
+{
+  "sandboxes": ["vanilla", "vanilla-ts", "new"]
+}
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,3 @@
 {
-  "sandboxes": ["vanilla", "vanilla-ts", "goofy-darwin-xx1c8"]
+  "sandboxes": ["msw-react-xx1c8"]
 }
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,3 @@
 {
-  "sandboxes": ["vanilla", "vanilla-ts", "new"]
+  "sandboxes": ["vanilla", "vanilla-ts"]
 }
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,3 +1,3 @@
 {
-  "sandboxes": ["vanilla", "vanilla-ts"]
+  "sandboxes": ["vanilla", "vanilla-ts", "goofy-darwin-xx1c8"]
 }


msutkowski:
@kettanaito Hey Artem, would you be willing to add the CodesandboxCI app and allow it to build PRs for `msw`? You just need to [add the app as mentioned here](https://codesandbox.io/docs/ci). I'd like to get a slug for your `ws` PR over at #396 so I can setup an example for it for `rtk-query` using `ws` as a baseQuery. We use this _extensively_ with RTK... examples:

https://ci.codesandbox.io/status/rtk-incubator/rtk-query
https://ci.codesandbox.io/status/reduxjs/redux-toolkit

If you're okay with this, I can setup an example with `msw` to just build automatically in a CSB as well, like I do in the examples seen throughout these docs: https://deploy-preview-43--rtk-query-docs.netlify.app/concepts/prefetching

Thanks!!!!!!!!

.codesandbox/ci.json
```diff
@@ -0,0 +1,3 @@
+{
+  "sandboxes": ["vanilla", "vanilla-ts", "new"]
```
kettanaito: Can we specify a Codesandbox document to use as a template for incremental builds?

As we have a [React sandbox](https://codesandbox.io/s/goofy-darwin-xx1c8?file=/src/index.js), we may create a new sandbox dedicated for PR previews. Unlike the React one, it would use TypeScript, plain JavaScript and have a few tests. That way we could see browser and NodeJS integrations in a single sandbox.

kettanaito:
Hey, @msutkowski. That's an awesome suggestion!

I've added CSB App to this repository. Let me know what else is needed to have a CSB integration. 

It'd be a great idea to populate the docs with some Codesandbox examples indeed!

.codesandbox/ci.json
msutkowski: That's actually what I do in `rtk-query`... there is some config shenanigans we had to deal with to get the examples to run locally as well in CSB, but I could probably get both scenarios working for `msw` without too much an issue. If you look, the Svelte example uses the node template, while React uses CRA. https://github.com/rtk-incubator/rtk-query/tree/main/examples

msutkowski:
Alright, it builds the sandbox that you linked me to, and if you check the package.json in it, it points to the build of the latest commit :). You can safely merge this, and if you want to have the examples locally in the repo, we can try and set that up as well... but it [looks like this works as expected](https://codesandbox.io/s/msw-react-reproduction-5g45m?file=/package.json) :)

.codesandbox/ci.json
msutkowski: Just to elaborate on 'shenanigans': it was problematic getting the tests to run and to avoid errors regarding multiple versions of React. The examples would try to reference the React/jest/etc versions in the root and throw 'multiple versions of react' errors regarding hooks usage. It took a bit of tinkering to get everything configured so that things run everywhere without problems.

kettanaito:
This looks awesome, @msutkowski!
Before we merge I'd like for those sandboxes to contain a meaningful reproduction scenario (such as a usage and a test for that usage). Currently vanilla and TypeScript sandboxes are good, but one'd have to write usage/tests manually in order to try the library, which decreases the usefulness of this feature.

kettanaito:
I'll try to set up such sandboxes and link them here. You can also do this, if you have time, of course. 
Thank you for the amazing work!

kettanaito:
Hey, @msutkowski 👋 I've added a single template for now (MSW React) so we could merge this change straight away. I'm looking forward to other sandboxes you may create when you have time! 

Will rebase the branch and merge. Once more thank you for this awesome addition!

kettanaito (approved):

PR #473, opened 2020-11-23
feat: add new recorder api

marcosvega91:
Hi guys,

I want to add this new feature. It's in draft and under development. 

### Goal
To use MSW the first thing to do is create `handlers`, so why not create a function to `record` responses from real APIs?

### How 
I have added in the `setupWorker` a new property `recorder`. You can:

1) `start`, start recording by using `worker.recorder.start()`. All your handlers will be paused until you called `stop` on the `recorder`. All intercepted handlers will be saved in a `logs` property of the `recorder`. All handlers are saved as `functions` so then you can copy and paste the code.

2) `stop`, stop recording by using `worker. recorder.stop()`. After that all your previous handlers will be restored.

3) `getLogs`, will return the list of functions for the latest recording.

### Readmap

- [ ] Create the new Recorder Class
- [ ] Support REST 
- [ ] Support graphql
- [ ] Support binary ( then the user will have a function with all the binary inside, I don't know if it is really useful ) 
- [ ] Add tests

### DIFF
--- a/src/setupWorker/setupWorker.ts
+++ b/src/setupWorker/setupWorker.ts
@@ -4,7 +4,7 @@ import { createStop } from './stop/createStop'
 import * as requestHandlerUtils from '../utils/handlers/requestHandlerUtils'
 import { isNodeProcess } from '../utils/internal/isNodeProcess'
 import { ServiceWorkerMessage } from '../utils/createBroadcastChannel'
-import { Recorder, ApiRecorder } from '../utils/Recorder'
+import { Recorder } from '../utils/Recorder'
 export interface SetupWorkerApi {
   start: ReturnType<typeof createStart>
   stop: ReturnType<typeof createStop>
@@ -26,7 +26,7 @@ export interface SetupWorkerApi {
   /**
    * Record your APIS
    */
-  recorder: ApiRecorder
+  recorder: Recorder
 }
 
 interface Listener {
@@ -144,10 +144,6 @@ export function setupWorker(
       })
     },
 
-    recorder: {
-      record: () => recorder.record(),
-      getLogs: () => recorder.getLogs(),
-      stop: () => recorder.stop(),
-    },
+    recorder,
   }
 }
--- a/src/setupWorker/setupWorker.ts
+++ b/src/setupWorker/setupWorker.ts
@@ -1,33 +1,14 @@
-import { SetupWorkerInternalContext, RequestHandlersList } from './glossary'
+import {
+  SetupWorkerInternalContext,
+  RequestHandlersList,
+  SetupWorkerApi,
+} from './glossary'
 import { createStart } from './start/createStart'
 import { createStop } from './stop/createStop'
 import * as requestHandlerUtils from '../utils/handlers/requestHandlerUtils'
 import { isNodeProcess } from '../utils/internal/isNodeProcess'
 import { ServiceWorkerMessage } from '../utils/createBroadcastChannel'
 import { Recorder } from '../utils/Recorder'
-export interface SetupWorkerApi {
-  start: ReturnType<typeof createStart>
-  stop: ReturnType<typeof createStop>
-
-  /**
-   * Prepends given request handlers to the list of existing handlers.
-   */
-  use: (...handlers: RequestHandlersList) => void
-
-  /**
-   * Marks all request handlers that respond using `res.once()` as unused.
-   */
-  restoreHandlers: () => void
-
-  /**
-   * Resets request handlers to the initial list given to the `setupWorker` call, or to the explicit next request handlers list, if given.
-   */
-  resetHandlers: (...nextHandlers: RequestHandlersList) => void
-  /**
-   * Record your APIS
-   */
-  recorder: Recorder
-}
 
 interface Listener {
   target: EventTarget
@@ -42,7 +23,6 @@ let listeners: Listener[] = []
 export function setupWorker(
   ...requestHandlers: RequestHandlersList
 ): SetupWorkerApi {
-  const recorder = new Recorder()
   requestHandlers.forEach((handler) => {
     if (Array.isArray(handler))
       throw new Error(
@@ -53,7 +33,6 @@ export function setupWorker(
     worker: null,
     registration: null,
     requestHandlers: [...requestHandlers],
-    recorder,
     events: {
       addListener(target: EventTarget, event: string, callback: EventListener) {
         target.addEventListener(event, callback)
@@ -111,11 +90,11 @@ export function setupWorker(
     )
   }
 
-  return {
+  const instance: SetupWorkerApi = {
     start: createStart(context),
     stop: createStop(context),
 
-    use(...handlers) {
+    use(...handlers: RequestHandlersList) {
       requestHandlerUtils.use(context.requestHandlers, ...handlers)
     },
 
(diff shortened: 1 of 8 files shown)


src/utils/Recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const hasJsonContent = response.headers?.get('content-type')?.includes('json')
+  if (hasJsonContent) {
+    composeLines.push(`ctx.json(${body})`)
+  } else {
+    composeLines.push(`ctx.body(${body})`)
+  }
+
+  response.headers.forEach((value, key) => {
+    composeLines.push(`ctx.set('${key}', '${value}')`)
```
kettanaito: How would this handle a header with multiple values?

src/utils/Recorder.ts
```diff
@@ -0,0 +1,83 @@
...
+import { fetch } from '../context/fetch'
+
+async function createFunctionFromRequest(
+  request: MockedRequest,
+  response: Response,
+) {
+  const lines = []
+  const composeLines = []
+  lines.push(
+    `rest.${request.method.toLowerCase()}('${
```
kettanaito, line 21: We can also check if the request body/query parameters have `operation`/`variables` properties, treating such request as a GraphQL request.

src/setupWorker/setupWorker.ts
```diff
@@ -111,5 +133,10 @@ export function setupWorker(
         ...nextHandlers,
       )
     },
+    recorder: {
```
kettanaito: A conceptual question: can `recorder` be a request handler?

```js
setupWorker(
  recorder()
)
```

With it being a request handler we get a few things baked-in:

1. Request interception. Request handler will receive all unhandled requests prior to its positioning in the list of handlers. This position dependency may be used as a feature to filter out certain requests you don't wish to record.
1. Short-circuit. Since `recorder` always performs requests as-is, it can return them short-circuiting the request handlers lookup. There's no need for an extra modification of `handleRequestWith`.
1. Potentially more straightforward NodeJS support. 

Potential disadvantages:

1. Disrupted flow. `recorder` being a request handler makes it confusing where to access the recorded handlers.

src/utils/Recorder.ts
marcosvega91: `ctx.set` will append the value to the Header so if one value for the same key already exist the will be merged

src/utils/Recorder.ts
marcosvega91: yes, this is a draft. Of course I would support GraphQL

src/setupWorker/setupWorker.ts
marcosvega91: `recorder` as `handler` could be great. My first idea was to add in this way but I had to create all handlers for every method and I was not sure if it was the right thing. It could remain a property of the worker but we can, under the hood, add some handlers instead of changing `handleRequestWith`. 

Find a simple way to make it works with both node and browser, could be very awesome.

The main idea is to create a devtools then

kettanaito: This functionality can be embedded into the devtools directly, with MSW only exposing some hooks to listen to. However, it's a good incremental step to introduce the recorder feature gradually, first without the devtool, and then with it. 

If that doesn't look like a lot of trouble, could you try `recorder` as a request handler? I'm extremely curious what difference that would produce.

marcosvega91: I have updated the code with the hybrid approach. In this way the `recorder` is something that can't be confused with the handler, but at the same time we reuse the handler approach to make the recorder works

marcosvega91: Using only as `handler` like you have done above, how the user then could access to the recorder requests?

marcosvega91:
How we can handle binary entities ? To record the API , we should include the binary in the created function, but it will be not very clean for the user

src/utils/Recorder.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (body) {
+    const hasJsonContent = response.headers
+      ?.get('content-type')
+      ?.includes('json')
+    if (hasJsonContent) {
+      composeLines.push(`ctx.json(${body})`)
+    } else {
+      composeLines.push(`ctx.body(\`${body.replace(/`/g, '\\`')}\`)`)
+    }
+  }
```
marcosvega91, line 38: How we can handle binary entities ? To record the API , we should include the binary in the created function, but it will be not very clean for the user

kettanaito:
I had a though in #481 that we could establish an event emitter in MSW that would dispatch various events like `request-start` or `request-intercepted`, allowing third-party tools to build extensions and plugins around that. Recorder API could also be a separate package if we support emitters:

```js
// msw-recorder/index.js
export const logs = []

export function record(worker) {
  worker.events.on('request-end', (req, res) => {
    logs.push(req.url.href)
  })
}
```

I'm not saying we should do this as a part of this pull request, but that may be a good API to move extensions forward. 

What is your take on this, Marco?

marcosvega91:
Yes it makes sense and it is also better. Anyway, we should intercept all requests even if you don't have a defined an handler.
We can in this way do:

1. `recorder.start()` will remove all handlers and ad a lister as you suggested
2. `recorder.stop()` will restore all handlers before recording and return a list of recorded responses 

We should need a way to "pause" MSW, to avoid problem with the recorder, In my example I have added an API `removeHandlers` that removed and returned all the attached handlers. I think that we need something specific because you can eventually add the option `onunhandledrequest` set to `error`. I was solving this problem because my generic handlers was returning a response.

kettanaito:
Now when Life-cycle events are published (#482), we can utilize them for the recording capabilities. 

I'd vote for a separate package, like an MSW extension, to be responsible for the recording and request handler generation. We can publish it under `@mswjs/*` namespace, so it'll be an officially supported extension to generate handlers from actual traffic. What do you think about this, @marcosvega91?

marcosvega91:
Yes I agree with it, MSW now offers more possibility of integration. Awesome 🥳 . I'm closing the PR

tobiasbueschel:
Just stumbled upon this issue as I was researching the same functionality 👍 @marcosvega91 & @kettanaito have you already started / looked into creating this recorder package that one could use to e.g. quickly integrate and mock out all API calls by running the application against it once? Would be keen to contribute 🙏

marcosvega91:
Hello @tobiasbueschel, I have closed this PR to start a new one with the features implemented by @kettanaito. I didn't start it yet because of time. Your help will be very awesome ❤️ .

kettanaito:
Hey, @tobiasbueschel. I think this feature has a place to be and we would greatly appreciate your help with it!

If you don't mind, let's start by opening it as an issue in this repository and brainstorm about the API and how it would work. The better a feature is defined, the smoother its implementation is. Looking forward to having productive discussions with you!

tobiasbueschel:
Great @kettanaito & @marcosvega91 I've created it here: https://github.com/mswjs/msw/issues/628 By the way, how do you usually align on how to implement certain features? Does that mostly happen here on GitHub or is there also a Discord for example?

Hope you're having a good week :)

kettanaito:
Thank you, @tobiasbueschel.

> By the way, how do you usually align on how to implement certain features?

We usually create a GitHub issue, discuss and brainstorm in there, and once the implementation path is clear we make the thing happen. Ability for anybody to contribute to the discussion is the heart of open-source.

PR #482, opened 2020-11-30
Adds life-cycle events API

### LINKED ISSUE #481: Middleware for Requests
Middlewares would help in building dev tools/extend MSW without having to fork the repo.

Let's say we want to "monitor" all the requests that are responded by the server. So that we can assert the requests made.
Middleware/emitting responded event would help in building that.

Should be able to add "middleware" or event emitted.

```js
server.middlewares(middleware)
```

Right now to achieve this, we can add some logic to catch request in each handler.

```js
let loginRequests = [];

server.use(
    rest.post('/login', (req, res, ctx) => {
        loginRequests.push(req);
        return res(ctx.json({ data: 'data'}));
    }),
);
```
But this would have to duplicated and we can't reliably abstract this logic out.

kettanaito:
## GitHub

- Fixes #481 
- Relates to #388 

## Motivation

Lifecycle events would allow developers to hook into the internal operations of MSW and execute their arbitrary logic. This feature is primarily designed for dev tools, extensions, and plugin developers to get easier access to MSW internals. 

## Changes

- Adds a new lifecycle events API to allow to listen to the following events:
  - [x] `request:start`—any request is dispatched.
  - [x] `request:match`—a request is dispatched that has a corresponding request handler.
  - [x] `request:unhandled`—a request is dispatched that doesn't have a request handler.
  - [x] `request:end`—any request has ended (succeeded with the mocked or original response).
  - [x] `response:mocked`—a mocked response has been returned.
  - [x] `response:bypass`—an original response has been returned. 
- Type annotates sent and received messages from the worker.
- Refactors client-worker communication channel. 
- Updates `node-request-interceptor` to support the `response` event.
- Each request in a browser and NodeJS now has a UUID assigned to it. 

## Roadmap

- [x] Remove all listeners when the worker stops/unregisters.
- [x] Set a unique request ID on the request object to reference the same request across events.
- [x] Retrieve actual response from the worker in case of a bypassed request and provide it to respective listeners. 
- [x] Type annotate worker-client messaging. 
- [x] Review the lifecycle event names.
- [x] Support NodeJS.
  - [x] Extend `node-request-interceptor` to signal back the response event (`response:bypass`).
  - [x] Wait for https://github.com/mswjs/node-request-interceptor/issues/82
- [ ] Conduct performance testing.
- [x] Add integration tests for all lifecycle events.
- [x] Use [`strict-event-emitter`](https://github.com/open-draft/strict-event-emitter) instead of a custom emitter solution.


kettanaito:
@deshiknaves, hi. May I get your feedback on this feature and whether it would satisfy the requirements towards the Cypress integration you are developing? Thanks.

marcosvega91:
I think that the `response:bypass` should also return the response. It will be also util for the recorder.

kettanaito:
You're right, @marcosvega91. One of the steps of this feature should be a post message from the worker with the original response, so we could provide it as an argument to certain lifecycle events, like `response:bypass`.

marcosvega91:
Maybe we can have a single event and use a  parameter to distinguish between the two kind of requests

src/mockServiceWorker.js
```diff
@@ -239,3 +253,11 @@ function ensureKeys(keys, obj) {
     return acc
   }, {})
 }
+
+function uuidv4() {
```
timdeschryver, line 265: I hope you don't mind that I'm commenting on a draft, but I think this can be removed and replaced with `utils/uuidv4`?.

src/utils/handleRequestWith.ts
```diff
@@ -89,6 +95,9 @@ export const handleRequestWith = (
           response,
         )
 
+        context.emitter.emit('response:bypass', req)
```
timdeschryver: Would it make sense to add the response here, if that's possible?
Because for the mocked instances, we do add the response to the emitter.

src/mockServiceWorker.js
kettanaito: No, of course, I don't mind. 
The worker script is not transpiled, only copied as-is into the build directory. I suppose you suggest to reuse the same utility we have for internal usage, but that'd mean having to transpile the worker. I'm not strictly against that, but 1-1 representation of the script that will end up on developers' machines, as well as a guarantee of its integrity, is a reasonable advantage to consider against an extra transformation step.

src/utils/handleRequestWith.ts
kettanaito: Yes, this is one of my todo items. Once I refactor the worker-client signaling system, the library will receive a `RESPONSE_ORIGINAL` (or similar) event from the worker when an original response is retrieved. This `response:bypass` event should be executed at that time, not during request handling.

src/mockServiceWorker.js
timdeschryver: Thanks for the info, I wasn't aware of this.

src/utils/handleRequestWith.ts
kettanaito: Request end and response retrieval happen in two different moments of time and two different contexts. Therefore, they will be handled separately, with you being able to connect them post-factum by the persistent `req.id` and `response.requestId`.

src/mockServiceWorker.js
kettanaito: No worries. I agree that it may introduce certain code duplication, yet I'm unconvinced that the file should be built in any way. Plus, code reuse would still be dangerous, as the worker executes in a limited context, potentially making certain code un-reusable (i.e. if a code relies on `window` or `document`).

src/setupWorker/glossary.ts
```diff
@@ -56,6 +56,21 @@ export interface SetupWorkerInternalContext {
   requestHandlers: RequestHandler<any, any>[]
   emitter: StrictEventEmitter
   keepAliveInterval?: number
+  workerChannel: {
```
kettanaito, line 94: I'm scoping the way to send and receive events from the worker to these typed methods, so that unknown events cannot be sent/listened to.

kettanaito:
With the introduction of the `RESPONSE` message event, where the worker signals back _any_ response that happens (mocked or original), I wonder what is the right `res` type to expose?

At the moment I went with the unaltered `Response` instance because it's more familiar to developers and allows to transform the response in any way (i.e. `res.json()`, `res.arrayBuffer()`). The drawback to this decision is that any library-related information will be exposed separately. For example, a mocked response and a request ID are available separately:

```js
worker.on('response:bypass', (res, requestId) => {})
worker.on('response:mocked', (res, requestId) => {})
```

At the same time, the `req` instance exposed in other lifecycle events, like `request:start` or `request:match`, is always an internal request representation (`MockedRequest`). That is for the developer's convenience and for that request instance to resemble the same `req` they have in request handlers. 

What do you think about these choices?

src/mockServiceWorker.js
```diff
@@ -186,14 +188,34 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
...
-    }),
+    })
+      .then(async (response) => {
+        const client = await event.target.clients.get(clientId)
+        const clonedResponse = response.clone()
+
+        sendToClient(client, {
+          type: 'RESPONSE',
+          payload: {
+            requestId,
```
tigerabrodi, line 199: This will be also of key `requestId` or?

src/setupWorker/setupWorker.ts
```diff
@@ -38,6 +41,31 @@ export function setupWorker(
...
     registration: null,
     requestHandlers: [...requestHandlers],
+    emitter: new StrictEventEmitter(),
+    workerChannel: {
+      on(type, callback) {
+        context.events.addListener(
+          navigator.serviceWorker,
+          'message',
+          (event: MessageEvent) => {
+            const message = jsonParse<ServiceWorkerMessage<typeof type, any>>(
```
tigerabrodi: I had a hard time understanding right away that `type` comes from the parameter, my suggestion would be to rename it, perhaps `eventType`?

src/utils/internal/uuidv4.ts
```diff
@@ -0,0 +1,7 @@
+export function uuidv4() {
```
tigerabrodi, line 1: Why is this custom `uuid` function needed? Also can we not name it `uuid` just if we gonna use something custom?

src/utils/request/parseWorkerRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+    body: pruneGetRequestBody(rawRequest),
+    bodyUsed: rawRequest.bodyUsed,
+    headers: new Headers(rawRequest.headers),
+  }
+
+  // Set document cookies on the request.
+  request.cookies = getRequestCookies(request)
+
+  // Parse the request's body based on the "Content-Type" header.
+  request.body = parseBody(request.body, request.headers) as any
```
tigerabrodi, line 35: Here, is there a better way of casting it as `any`? Perhaps we could cast it to another type if casting is needed?

src/utils/worker/createResponseListener.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  ServiceWorkerIncomingEventsMap,
+  SetupWorkerInternalContext,
+} from '../../setupWorker/glossary'
+import { ServiceWorkerMessage } from '../createBroadcastChannel'
+
+export function createResponseListener(context: SetupWorkerInternalContext) {
+  return (
+    _: MessageEvent,
```
tigerabrodi, line 9: Is there a reason you named this parameter ` _`?

tigerabrodi (dismissed):

deshiknaves:
> @deshiknaves, hi. May I get your feedback on this feature and whether it would satisfy the requirements towards the Cypress integration you are developing? Thanks.

Hey @kettanaito this will work perfectly for that 👏🏽. This enables us to get start and end of every request which is what is needed.

kettanaito:
Glad to hear that, @deshiknaves! Could it also solve, or eliminate the need of sharing a client (https://github.com/mswjs/msw/pull/383)?

src/utils/worker/createResponseListener.ts
kettanaito: To satisfy the linting rule, denoting that the first argument here (`event`) is not used.

src/utils/request/parseWorkerRequest.ts
kettanaito: I'd say this logic is quite raw, and the return type of `parseBody` could be much more permissive.

src/utils/internal/uuidv4.ts
kettanaito: I don't wish to introduce an extra dependency for a few lines of code. This function generates spec-compliant V4 UUID, so there's no ambiguity in naming it accordingly.

src/mockServiceWorker.js
kettanaito: Sorry, what do you mean?

timdeschryver (commented):
I do think that this is the right decision @kettanaito 
I suppose having the real response object would help with #473 

It's a bit of a bummer that the API is a little different with the 2nd argument being the request id. What about wrapping the response in another object? If this is the API we prefer over having 2 arguments, perhaps we could also wrap the request object?

```ts
interface ResponseWrapper {
  requestId: string;
  response: Response;
}
```

kettanaito:
Thank you for the feedback, @timdeschryver. What is your argumentation for creating such response abstraction?
I'd like to learn more about how that would be beneficial. 

I worry that such `ResponseWrapper` is a completely new thing, unlike `req` which is of the `MockedRequest` type also exposed publicly in request handlers. As I've stated above, `MockedResponse` is the wrong structure to describe an actual or even a mocked response, as it's an internal representation of the response and contains certain extra properties that bear no use to the consumer (`delay`, `once`), as well as lacking other properties to be compatible with the native `Response` instance (`ok`, `bodyUsed`, `json`, etc.). 

The current benefit I see over such `ResponseWrapper` abstraction is for the listener function to accept one argument instead of two. However, that shouldn't be the issue, as we are in charge of what arguments are exposed.

timdeschryver:
You're right, the only argument I can make would be that the API will be the same for requests and responses.

deshiknaves:
> Glad to hear that, @deshiknaves! Could it also solve, or eliminate the need of sharing a client (#383)?

@kettanaito no, this does not solve that problem. Currently if it's in an iframe, it won't share the mocks from the parent. That one is needed to allow other clients to use the mocks served from the parent context.

src/setupWorker/setupWorker.ts
kettanaito: Renamed ✅

kettanaito:
Let's ensure that these changes don't negatively affect the performance of medium-large apps, and it will be ready for release in the next minor version.

src/mockServiceWorker.js
msutkowski: @tigerabrodi Yes - this is the same as `payload: { requestId: requestId }`

src/setupWorker/glossary.ts
```diff
@@ -2,26 +2,115 @@ import { HeadersList } from 'headers-utils'
...
+  | 'body'
+  | 'json'
+  | 'blob'
+  | 'arrayBuffer'
+  | 'formData'
+  | 'clone'
+  | 'signal'
+  | 'isHistoryNavigation'
+  | 'isReloadNavigation'
+>
```
msutkowski, line 25: For clarity, I would consider splitting this into two types, because `clone`, `signal`, `isHistoryNavigation`, `isReloadNavigation` are not 'methods'

src/mockServiceWorker.js
tigerabrodi: Yes, my bad for the late response! Sorry @kettanaito

test/msw-api/life-cycle-events/life-cycle-events.mocks.ts
```diff
@@ -0,0 +1,39 @@
...
+worker.on('request:unhandled', (req) => {
+  console.warn(`[request:unhandled] ${req.method} ${req.url.href} ${req.id}`)
+})
+
+worker.on('request:end', (req) => {
+  console.warn(`[request:end] ${req.method} ${req.url.href} ${req.id}`)
+})
+
+worker.on('response:mocked', async (res, reqId) => {
+  const body = await res.text()
```
msutkowski, line 29: Will a user always use `await res.text()`? I know this is produced by:

`const response = new Response(payload.body, payload)`

But what's the full use-case look like here? I can't really tell from the test what _should_ be going on. Ideally, I'd imagine we would recreate the original response with it's content-type, then provide an unwrap utility so someone can work with the object or text without trying to reconstruct it?

I'll poke around a bit in the run time... maybe I'm missing something :)

msutkowski (commented):
This looks good to me overall, but there are a few things I don't understand from a usability perspective based on just the code. If you could help me understand, that'd be awesome! Either way, I'll test this out and see if I can provide anything n addition. Great work @kettanaito!!

test/msw-api/life-cycle-events/life-cycle-events.mocks.ts
kettanaito: It's up to the consumer to handle the response in this listener. The intention is to return a cloned original/mocked responses here so that the consumer may operate on them in any way needed. That includes constructing a response body (`res.text`/`res.json`/etc.), accessing headers, etc. There is nothing that _should_ happen precisely, as it's up to the consumer to utilize these listeners. All these listeners are read-only and don't affect what happens or happened in the library. 

I find the native `Response` instance to be more suitable here than a custom abstraction, because:

1. It's more familiar (`fetch()` returns the same instance).
1. It gives more control and is quite convenient to handle.
1. It makes `response:mocked` and `response:bypass` behave homogenously, although the source of response differs for those two events.

timdeschryver (dismissed):
If it doesn't have a performance hit it LGTM 👍
Nice one!

src/setupWorker/glossary.ts
kettanaito: Perhaps renaming that interface would be enough? I agree that `RequestWithoutMethods` is a poor name.

I thought `clone` is a method to clone the request 🤔 At least it works like one.

kettanaito:
The final touch on the `response:bypass` support in NodeJS has been merged. Let's wait for the CI and merge the life-cycle events.

PR #493, opened 2020-12-05
Print console message when calling "worker.stop()"

### LINKED ISSUE #485: Add "[MSW] Mocking disabled." console message
Would currently get this console message when MSW is enabled: `[MSW] Mocking enabled.`

It would be great to add the `[MSW] Mocking disabled.` when it's disabled.

My use case:
I'm building a small web app to mesure the performance of a web library. I'm mocking backend calls with MSW. I added a checkbox that can be used to toggle activation, deactivation of the mocking and it would be great to have confirmation in the console with this message.

Thank you!

tigerabrodi:
This closes https://github.com/mswjs/msw/issues/485 :+1: :tada: :100:

### DIFF
--- a/src/setupWorker/stop/createStop.ts
+++ b/src/setupWorker/stop/createStop.ts
@@ -11,5 +11,12 @@ export const createStop = (context: SetupWorkerInternalContext) => {
     context.events.removeAllListeners()
     context.emitter.removeAllListeners()
     window.clearInterval(context.keepAliveInterval)
+
+    if (!context.startOptions?.quiet) {
+      console.log(
+        '%c[MSW] Mocking disabled.',
+        'color:orangered;font-weight:bold;',
+      )
+    }
   }
 }
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -86,6 +86,7 @@ export interface WorkerLifecycleEventsMap {
 }
 
 export interface SetupWorkerInternalContext {
+  startOptions: StartOptions | undefined
   worker: ServiceWorker | null
   registration: ServiceWorkerRegistration | null
   requestHandlers: RequestHandler<any, any>[]
--- a/src/setupWorker/setupWorker.ts
+++ b/src/setupWorker/setupWorker.ts
@@ -38,6 +38,7 @@ export function setupWorker(
       )
   })
   const context: SetupWorkerInternalContext = {
+    startOptions: undefined,
     worker: null,
     registration: null,
     requestHandlers: [...requestHandlers],
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -31,6 +31,10 @@ export const createStart = (context: SetupWorkerInternalContext) => {
   return function start(options?: StartOptions) {
     const resolvedOptions = mergeRight(DEFAULT_START_OPTIONS, options || {})
 
+    // Store the start options in the context so that other methods (like `stop`)
+    // could reference them.
+    context.startOptions = resolvedOptions
+
     const startWorkerInstance = async () => {
       if (!('serviceWorker' in navigator)) {
         console.error(
--- a/test/msw-api/setup-worker/stop/quiet.mocks.ts
+++ b/test/msw-api/setup-worker/stop/quiet.mocks.ts
@@ -1,0 +1,8 @@
+import { setupWorker } from 'msw'
+
+const worker = setupWorker()
+
+// @ts-ignore
+window.msw = {
+  worker,
+}
--- a/test/msw-api/setup-worker/stop/quiet.test.ts
+++ b/test/msw-api/setup-worker/stop/quiet.test.ts
@@ -1,0 +1,48 @@
+import * as path from 'path'
+import { SetupWorkerApi } from 'msw'
+import { captureConsole } from '../../../support/captureConsole'
+import { runBrowserWith } from '../../../support/runBrowserWith'
+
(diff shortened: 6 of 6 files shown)


src/setupWorker/stop/createStop.ts
```diff
@@ -10,5 +13,11 @@ export const createStop = (context: SetupWorkerInternalContext) => {
     context.worker?.postMessage('MOCK_DEACTIVATE')
     context.events.removeAllListeners()
     window.clearInterval(context.keepAliveInterval)
+    if (!options?.quiet) {
+      console.groupCollapsed(
```
kettanaito: Currently, this is an unterminated console group. Unless there's a `console.groupEnd()` following it, it will embrace all logs that are coming after this message and group them in `[MSW] Mocking disabled.` group.

Instead, let's use plain `console.log`, since we have no additional information to group in this case.

src/setupWorker/stop/createStop.ts
```diff
@@ -1,6 +1,9 @@
-import { SetupWorkerInternalContext } from '../glossary'
+import { SetupWorkerInternalContext, StartOptions } from '../glossary'
 
-export const createStop = (context: SetupWorkerInternalContext) => {
+export const createStop = (
+  context: SetupWorkerInternalContext,
+  options?: StartOptions,
```
kettanaito: It appears we don't have access to start options in the `createStop` function:

They are both created with the same internal `context` as an argument:

https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/setupWorker/setupWorker.ts#L99-L100

However, the `createStart` later resolves its options

https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/setupWorker/start/createStart.ts#L30-L31

and passed them to the function that attached the activation listener (and outputs the "Mocking enabled" message):

https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/setupWorker/start/createStart.ts#L120-L123

https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/setupWorker/start/utils/activateMocking.ts#L3-L10

In the case of `createStop`, we don't know anything about the start options. Wonder if we should persist them in context in any way 🤔

kettanaito (changes requested):

kettanaito:
Thank you for this improvement, @tigerabrodi! I've left a few comments, I'd appreciate if you had a look at them.

src/setupWorker/stop/createStop.ts
tigerabrodi: Alright :raised_hands:

src/setupWorker/stop/createStop.ts
tigerabrodi: @kettanaito Seems to be implemented in two different ways. Would we need a `disableMocking` util for this, and use that in `createStop`? :thinking:

kettanaito: We don't have to break it into separate functions, as `createStop` does only a few things. The challenge is to reference the start options in this `createStop` function. Currently, there's no way to do that. What comes to my mind is to store those options in the `context` object available for these internal methods, but I bet it needs some more thinking.

tigerabrodi: @kettanaito Do you mean storing them, and then using it like `context.options`, also with more thinking, do you mean as in decision making or it would take more time?

If storing options in `context`, it would require some refactoring as well I believe.

tigerabrodi:
This PR is set to Draft, how we gonna implement this is still under planning/thinking. :+1:

src/setupWorker/stop/createStop.ts
kettanaito: Yes, that's one of the options. We need to be careful with how we store the start options in the context because the context can be accessed while the worker hasn't been started yet. In other words, the info we're about to store is set asynchronously and that should be evident when we're trying to access it. 

My initial thought was to create a getter function that returns a start options object or `null` if it's not defined yet. Not sure if that's the best way to go. 

@timdeschryver, what would you recommend in this situation?

tigerabrodi: Alright! I also get notified for issues and PRs since I have enabled watch on MSW haha :wink: :raised_hands: :muscle: :smiling_face_with_three_hearts:

timdeschryver: I think there are two options here, and I think the solution @kettanaito suggested is the best one.

Since the options are initialized in `start`, we have no other choice than adding it to `context`.

Another option would be to move the options from `start` to `createServer`, that way the options can be passed down to `start` and `stop` but this is a breaking change. I also think the DX is better when the user uses `start` to provide the options (otherwise, the user will have to provide the options twice if they're using the browser and the node environment).

kettanaito:
I'll try to finalize this in the upcoming days. Once more thank you for the superb work.

kettanaito:
I've polished things a little and added an integration test. Looks good, let's wait for the CI.

kettanaito (dismissed):

kettanaito:
Welcome to the contributors, @tigerabrodi 🎉 Happy to see you joining.

kettanaito:
Strange CI issue of failing `test/msw-api/setup-worker/stop/quiet.test.ts` test. I've re-run it multiple times locally and everything passes.

kettanaito (approved):

PR #509, opened 2020-12-14
Export the library in the IIFE format

### LINKED ISSUE #507: "Uncaught ReferenceError: process is not defined" when running msw in a script tag
## Describe the bug
I tried running msw in a `script` tag. The error "Uncaught ReferenceError: process is not defined" is thrown and not handled. When looking at the minified umd file, the error occurs on the instruction: `process.env.NODE_ENV === "production"` which obviously will not work in a browser without further transformation.

My understanding is the following: when msw is run in a browser, it is assumed that a static bundler like webpack is used, in which case the esm bundle would be used and webpack would substitute process.env.NODE_ENV at build time. When msw is used in node where the umd version is used, then process.env.NODE_ENV is defined, and this is not a problem. However, to use msw in a browser without webpack, [like when using karmajs](https://github.com/mswjs/msw/issues/447) minus webpack, it might be useful to be able to run msw directly from a script tag.

I was able to work around this issue by creating a project that re-exports msw and is bundled using [rollup-plugin-replace](https://github.com/rollup/rollup-plugin-replace) to replace the process.env.NODE_ENV string in the output. Works for me but it removes the NODE_ENV variable use at runtime so the result is not suited for node applications.

If there is appetite to run msw directly from a script tag, I'm happy to spend a little more time and (tentatively) open a PR. The most universal solution would be to export msw as an iife and use [the rollup replace plugin](https://github.com/rollup/plugins/tree/master/packages/replace#usage) to remove the process.env.NODE_ENV instruction.

The result would be three formats, umd for node, esm for bundlers, and an iife for script tags (in which we would always be in a "development" environment).

## Environment

- `msw: 0.24.2`
- `nodejs: 12.18.3`
- `npm: 6.14.6`
- `Chrome 87.0.4280.88 on macOS`

## To Reproduce
```
$ mkdir msw-html
$ cd msw-html
$ npm init -y
$ npm install serve msw
$ touch index.html
```

Edit the index.html file and paste the following content:

```
<html>
  <head>
    <script src='./node_modules/msw/lib/umd/index.js'></script>
  </head>
</html>
```

```
$ serve .
```

Open the page in a browser (by default http://localhost:5000), open the console.

## Expected behavior
msw should be loaded and available

## Screenshots
Screen Shot 2020-12-11 at 4 35 35 PM

arnaudNYC:
Adds support for bundling the library into the IIFE format.

- Fixes #507

### DIFF
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -158,7 +158,6 @@ const buildIife = {
     esModule: false,
   },
   plugins: [
-    ,
     replace({
       'process.env.NODE_ENV': JSON.stringify('development'),
     }),
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -159,7 +159,7 @@ const buildIife = {
   },
   plugins: [
     replace({
-      'process.env.NODE_ENV': JSON.stringify('development'),
+      'process.env.NODE_ENV': JSON.stringify('production'),
     }),
     ...plugins,
     terser(),
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -4,6 +4,8 @@ import commonjs from '@rollup/plugin-commonjs'
 import inject from '@rollup/plugin-inject'
 import typescript from 'rollup-plugin-typescript2'
 import json from '@rollup/plugin-json'
+import replace from '@rollup/plugin-replace'
+import { terser } from 'rollup-plugin-terser'
 import packageJson from './package.json'
 
 const integrityCheck = require('./config/plugins/rollup-integrity-check-plugin')
@@ -53,7 +55,7 @@ const buildEsm = {
 /**
  * Configuration for the UMD build
  */
-const buildUdm = {
+const buildUmd = {
   input: 'src/index.ts',
   output: {
     file: packageJson.main,
@@ -144,4 +146,25 @@ const buildNative = {
   ],
 }
 
-export default [buildNode, buildNative, buildEsm, buildUdm]
+/**
+ * Configuration for the iife build
+ */
+const buildIife = {
+  input: 'src/index.ts',
+  output: {
+    file: 'lib/iife/index.js',
+    name: 'MockServiceWorker',
+    format: 'iife',
+    esModule: false,
+  },
+  plugins: [
+    ,
+    replace({
+      'process.env.NODE_ENV': JSON.stringify('development'),
+    }),
+    ...plugins,
+    terser(),
+  ],
+}
+
+export default [buildNode, buildNative, buildEsm, buildUmd, buildIife]
--- a/yarn.lock
+++ b/yarn.lock
@@ -6030,7 +6030,7 @@ jest-watcher@^26.6.2:
     jest-util "^26.6.2"
     string-length "^4.0.1"
 
-jest-worker@^26.6.2:
+jest-worker@^26.2.1, jest-worker@^26.6.2:
   version "26.6.2"
   resolved "https://registry.yarnpkg.com/jest-worker/-/jest-worker-26.6.2.tgz#7f72cbc4d643c365e27b9fd775f9d0eaa9c7a8ed"
   integrity sha512-KWYVV1c4i+jbMpaBC+U++4Va0cp8OisU185o73T1vo99hqi7w8tSJfUXYswwqqrjzwxa6KpRK54WhPvwf5w6PQ==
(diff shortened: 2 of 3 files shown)


rollup.config.ts
```diff
@@ -144,4 +146,25 @@ const buildNative = {
...
+const buildIife = {
+  input: 'src/index.ts',
+  output: {
+    file: 'lib/iife/index.js',
+    name: 'MockServiceWorker',
+    format: 'iife',
+    esModule: false,
+  },
+  plugins: [
+    ,
```
timdeschryver:
```suggestion
```

timdeschryver (commented):
This LGTM, have you tried using the built files locally to verify that it works?

rollup.config.ts
arnaudNYC: Fixed, sorry about that!

You can see the file I used to test here: https://gist.github.com/arnaudNYC/a1491c87c86ba38a948482494f04878b

```html
<html>

  <head>
    <script src="./mockServiceWorker.js"></script>
    <script src="./index.js"></script>
  </head>

  <body>
    <script type="text/javascript">
      var rest = MockServiceWorker.rest;
      var setupWorker = MockServiceWorker.setupWorker;

      var handlers = [
        rest.get('/user', function (req, res, ctx) {
          return res(
            ctx.status(200),
            ctx.json({
              username: 'admin',
            }),
          )
        }),
      ];

      var worker = setupWorker.apply(void 0, handlers);
      var p = worker.start();

      function doFetch() {
        return fetch('/user')
          .then(function (resp) {
            return resp.json();
          })
          .then(function (data) {
            console.log('Data:', data)
          });
      }

      p.then(doFetch);
    </script>
  </body>

</html>
```

Screen Shot 2020-12-15 at 9 30 46 AM

rollup.config.ts
```diff
@@ -53,7 +55,7 @@ const buildEsm = {
 /**
  * Configuration for the UMD build
  */
-const buildUdm = {
+const buildUmd = {
```
kettanaito, line 58: Oops. Thank you for patching this 😊

rollup.config.ts
```diff
@@ -144,4 +146,24 @@ const buildNative = {
...
+  input: 'src/index.ts',
+  output: {
+    file: 'lib/iife/index.js',
+    name: 'MockServiceWorker',
+    format: 'iife',
+    esModule: false,
+  },
+  plugins: [
+    replace({
+      'process.env.NODE_ENV': JSON.stringify('development'),
```
kettanaito: Could you please remind me why IIFE module should be bundled for the "development" environment?

rollup.config.ts
```diff
@@ -144,4 +146,24 @@ const buildNative = {
...
 }
 
-export default [buildNode, buildNative, buildEsm, buildUdm]
+/**
+ * Configuration for the iife build
+ */
+const buildIife = {
+  input: 'src/index.ts',
+  output: {
+    file: 'lib/iife/index.js',
```
kettanaito, line 155: Should we specify this entry in `package.json`? I'm afraid I don't know how IIFE modules are being resolved. Is it sufficient to explicitly provide the path to the module when importing?

yarn.lock
```diff
@@ -7485,7 +7485,7 @@ querystringify@^2.1.1:
   resolved "https://registry.yarnpkg.com/querystringify/-/querystringify-2.1.1.tgz#60e5a5fd64a7f8bfa4d2ab2ed6fdf4c85bad154e"
   integrity sha512-w7fLxIRCRT7U8Qu53jQnJyPkYZIaR4n5151KMfcJlO/A9397Wxb1amJvROTK6TOnp7PfoAmg/qXiNHI+08jRfA==
 
-randombytes@^2.0.0, randombytes@^2.0.1, randombytes@^2.0.5:
+randombytes@^2.0.0, randombytes@^2.0.1, randombytes@^2.0.5, randombytes@^2.1.0:
```
kettanaito, line 7495: Wow, we rely on 4 versions of the same lib, 3 of which are compatible. That looks like it could use some dependency alignment 👀 

Don't bother with this, but I'll raise an internal issue to improve this.

kettanaito: Raised in https://github.com/mswjs/msw/issues/516

kettanaito (commented):
Thank you for preparing this, @arnaudNYC!

I've left a few comments, mainly questions for me to take this opportunity to learn. I'd appreciate it if you could elaborate on those. Thanks!

rollup.config.ts
arnaudNYC: Out of an abundance of caution, which now that I am revisiting this does not seem necessary.

In development mode, there is an extra check resulting in an error warning "when multi-realm constructors are detected", you can see it in the compiled UMD version. So by the time we get to a production ready bundle, warnings should have been identified and taken care of. I'll change this to production.

To test this, I re-ran the test code above with a "development" and "production" build

rollup.config.ts
arnaudNYC: There is a `browser` field in package.json that can be used for just that, however that causes an issue with webpack. When webpack finds both a module and a browser field, it will choose the browser field, which in this case is the less desirable iife format which doesn't allow tree shaking. So in this case, since we want webpack to use the esm format under the module field, we should not specify a value for browser.

The way I saw this being used is by either manually copying the iife to whatever app the users are using, or by linking directly to the file using a cdn like is explained here: https://mswjs.io/docs/recipes/usage-with-cdn but on the iife. I am happy contributing that documentation should this change be accepted.

So not having the browser field specified in package.json makes it a little less obvious that an iife format is available, but hopefully good documentation can somewhat fill that gap.

arnaudNYC:
> Thank you for preparing this, @arnaudNYC!
> 
> I've left a few comments, mainly questions for me to take this opportunity to learn. I'd appreciate it if you could elaborate on those. Thanks!

My pleasure! Sorry I didn't get around to it sooner, I need to fix my github notifications, I only saw your comments today. I tried to address them the best I could.

Happy holidays,
Arnaud

rollup.config.ts
kettanaito: Thank you for such an insightful explanation. That documentation page was also something that came to my mind when I saw this pull request. I think starting with the documentation is always a good choice, and we'd be happy to welcome your pull request to extend that page once this lands.

kettanaito (approved):
This is a great addition. Thank you for this, @arnaudNYC, and welcome to contributors 🎉

PR #513, opened 2020-12-15
feat: support batched graphql operations

### LINKED ISSUE #510: Support batched GraphQL queries
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.24.2   |

## Request handlers

```js

export const handlers = [
    return res(
      ctx.data({
        user: {
          username: 'test',
          firstName: 'John',
        },
      })
    );
  }),

    return res(
      ctx.data({
        user: {
          username: 'test',
          firstName: 'John',
        },
      })
    );
  }),
];

```

## Actual request

ie.
```
```

## Current behavior

## Expected behavior

batch queries get intercepted

msutkowski:
Closes #510

### DIFF
--- a/src/utils/getResponse.ts
+++ b/src/utils/getResponse.ts
@@ -30,7 +30,7 @@ export const getResponse = async <
   req: R,
   handlers: H,
 ): Promise<ResponsePayload> => {
-  const relevantHandlers = handlers
+  let relevantHandlers = handlers
     .filter((requestHandler) => {
       // Skip a handler if it has been already used for a one-time response.
       return !requestHandler.shouldSkip
@@ -56,7 +56,14 @@ export const getResponse = async <
       handler: null,
       response: null,
     }
+  } else if (relevantHandlers.length > 1) {
+    // TODO: this is probably not the best approach... we need some way to limit the result set of relevantHandlers to 1 of each type.
+    // Is this really even possible though? Will one request ever _really_ match multiple handlers?
+    // We can match many graphql endpoints in a bulk request, but we should only use the graphql handler?
+    relevantHandlers = relevantHandlers[0]
   }
+
+  // TODO: What needs to happen here is that if we have an array in `parsedRequest`, we need to build one singular response with the aggregate of the operations.
 
   const {
     requestHandler,
@@ -106,7 +113,7 @@ export const getResponse = async <
   )
 
   // Although reducing a list of relevant request handlers, it's possible
-  // that in the end there will be no handler associted with the request
+  // that in the end there will be no handler associated with the request
   // (i.e. if relevant handlers are fall-through).
   if (!requestHandler) {
     return {
--- a/test/graphql-api/utils/executeOperation.ts
+++ b/test/graphql-api/utils/executeOperation.ts
@@ -1,4 +1,5 @@
 import { Page, Response } from 'puppeteer'
+import { Override } from '../../../src/tsHelpers'
 
 export const HOSTNAME = 'http://localhost:8080/graphql'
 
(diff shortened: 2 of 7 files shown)


msutkowski:
This actually needs more work. I finally set up an environment to actually test the batch link behavior and ran into issues. Will update later on :)

dkhuntrods:
@msutkowski 👋 

I've [done some work on this](https://github.com/BulbEnergy/msw/pull/1) (branched from an earlier version as we need to support older TS versions) as we needed the functionality ASAP.

Hopefully it's of help - we'd much rather to be able to use and support original package!

/cc @kettanaito

kettanaito:
Thank you for providing that implementation reference, @dkhuntrods! 

With the recent changes to the request handlers, this pull request needs a little adjustment. I've tried to structure the steps that a request handler performs, so the new logic we add will fall under one of the steps (most likely a parsing phase). I'd be happy to give a hand in aligning these changes with the latest state of the repository. 

A huge thank you, @msutkowski, for preparing this. Let me know if you need help with rebasing this branch.

kettanaito:
Hey, @msutkowski. During our discussion, I've mentioned a utility that can differentiate between REST and GraphQL requests. Here it is:

https://github.com/mswjs/msw/blob/655c93f83b7fcebe9a9093cacd7d891939028d29/src/utils/internal/parseGraphQLRequest.ts#L139-L172

We are already using it in the GraphQL request handler during the parsing phase:

https://github.com/mswjs/msw/blob/655c93f83b7fcebe9a9093cacd7d891939028d29/src/handlers/GraphQLHandler.ts#L107-L109

That parsing result is later taken into account in the `predicate` step that decides whether a request should be treated as a GraphQL request and be handled by the respective handler:

https://github.com/mswjs/msw/blob/655c93f83b7fcebe9a9093cacd7d891939028d29/src/handlers/GraphQLHandler.ts#L121-L124

It seems that no extra logic should be necessary as a part of the batched GraphQL queries support. As always, I may be wriong. Would appreciate your expertise on this. Thank you.

src/utils/getResponse.ts
```diff
@@ -56,8 +56,15 @@ export const getResponse = async <
       handler: null,
       response: null,
     }
+  } else if (relevantHandlers.length > 1) {
+    // TODO: this is probably not the best approach... we need some way to limit the result set of relevantHandlers to 1 of each type.
+    // Is this really even possible though? Will one request ever _really_ match multiple handlers?
```
kettanaito, line 61: If a request has a matching handler and that handler returns a mocked response, _only that handler_ will be used. Even if there are other handlers that match, they are resolved in the order they were attached. A single request cannot respect multiple matching handlers, this is a no-op. 

Why do we need to have this logic, though?

test/graphql-api/utils/executeOperation.ts
```diff
@@ -83,3 +66,59 @@ export const executeOperation = async (
...
+) => {
+  const { uri = HOSTNAME, method = 'POST' } = options || {}
+  const url = new URL(uri)
+  const urlString = url.toString()
+
+  const results = await Promise.all(
+    payload.map(({ query, variables }) =>
+      performRequest({
+        page,
+        urlString,
```
kettanaito, line 116: As these requests will be to the same URL, `performRequest` will resolve on the first match. This results in both results having the same JSON body.

kettanaito:
@msutkowski, hey 👋 Quite a bunch of things has changed around request handlers since this improvement was opened. I think you are right to consider opening a new pull request, it'll be less hassle than trying to migrate the current state. 

Thank you so much for working on this one!

otaciliolacerda:
Hi all, by the state of this branch can I assume `msw` currently does not support batch requests?

kettanaito:
## Released: v2.1.3 🎉

---

PR #531, opened 2021-01-08
fix: send null body and not empty string

### LINKED ISSUE #530: Cannot mock 204 response
## Describe the bug
It seems to be impossible to construct a 204 (No content) response. When I just invoke
`return res(ctx.status(204))`
I get the following error:
**_Failed to construct 'Response': Response with null body status cannot have body_**
It does not help trying to add `ctx.json({})` or `ctx.body(null)` - the body seems to end up as the empty string regardless. This fails when msw is trying to invoke `const response = new Response(payload.body, payload)` in the function `createResponseListener` as responses with status 204 are not allowed to have any body.

## Environment

- `msw: 0.25.0`
- `Chrome: 87.0.4280.141`

marcosvega91:
close #530

`response.text()` will return an empty string if the body is null. This will create some issues when for example the mocked response has a status 204 (no content). The event listener that will handle the response was creating a `Response` object with an empty string  instead of null.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -200,7 +200,8 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
             ok: clonedResponse.ok,
             status: clonedResponse.status,
             statusText: clonedResponse.statusText,
-            body: await clonedResponse.text(),
+            body:
+              clonedResponse.body === null ? null : await clonedResponse.text(),
             headers: serializeHeaders(clonedResponse.headers),
             redirected: clonedResponse.redirected,
           },
--- a/test/msw-api/life-cycle-events/regression/null-body.test.ts
+++ b/test/msw-api/life-cycle-events/regression/null-body.test.ts
@@ -1,0 +1,22 @@
+import * as path from 'path'
+import { runBrowserWith } from '../../../support/runBrowserWith'
+import { captureConsole } from '../../../support/captureConsole'
+import { sleep } from '../../../support/utils'
+
+function createRuntime() {
+  return runBrowserWith(path.resolve(__dirname, 'null-body.mocks.ts'))
+}
+
+test('lifecycle mothod should support null body', async () => {
+  const runtime = await createRuntime()
+  const { messages } = captureConsole(runtime.page)
+
+  await runtime.request({
+    url: 'https://test.mswjs.io/api/books',
+  })
+  await sleep(500)
+
+  expect(messages.warning).toEqual([`[response:mocked] null`])
+
+  return runtime.cleanup()
+})
--- a/test/msw-api/life-cycle-events/regression/null-body.test.ts
+++ b/test/msw-api/life-cycle-events/regression/null-body.test.ts
@@ -8,15 +8,19 @@ function createRuntime() {
 }
 
 test('lifecycle mothod should support null body', async () => {
+  let error
   const runtime = await createRuntime()
-  const { messages } = captureConsole(runtime.page)
+
+  runtime.page.on('pageerror', (err) => {
+    error = err
+  })
 
   await runtime.request({
     url: 'https://test.mswjs.io/api/books',
   })
   await sleep(500)
 
-  expect(messages.warning).toEqual([`[response:mocked] null`])
+  expect(error).not.toBeDefined()
 
   return runtime.cleanup()
 })
--- a/test/msw-api/life-cycle-events/regression/null-body.mocks.ts
+++ b/test/msw-api/life-cycle-events/regression/null-body.mocks.ts
@@ -6,8 +6,4 @@ const worker = setupWorker(
   }),
 )
 
-worker.on('response:mocked', async (res) => {
-  console.warn(`[response:mocked] ${res.body}`)
-})
-
 worker.start()
--- a/test/msw-api/life-cycle-events/regression/null-body.mocks.ts
+++ b/test/msw-api/life-cycle-events/regression/null-body.mocks.ts
@@ -1,0 +1,13 @@
+import { setupWorker, rest } from 'msw'
+
+const worker = setupWorker(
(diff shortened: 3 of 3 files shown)


src/mockServiceWorker.js
```diff
@@ -200,7 +200,8 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
             ok: clonedResponse.ok,
             status: clonedResponse.status,
             statusText: clonedResponse.statusText,
-            body: await clonedResponse.text(),
+            body:
+              clonedResponse.body === null ? null : await clonedResponse.text(),
```
kettanaito, line 204: And we don't need the same logic on the line 124 because `body` in the mocked response is `null` by default?

https://github.com/mswjs/msw/blob/9c894760fef38f84d8fd4841a837a5ab6e1c5ec7/src/response.ts#L37

test/msw-api/life-cycle-events/regression/null-body.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+function createRuntime() {
+  return runBrowserWith(path.resolve(__dirname, 'null-body.mocks.ts'))
+}
+
+test('lifecycle mothod should support null body', async () => {
+  const runtime = await createRuntime()
+  const { messages } = captureConsole(runtime.page)
+
+  await runtime.request({
```
kettanaito, line 18: And without the fix this test throws an exception upon trying to construct a `Response` instance internally?

marcosvega91: Yes right, If I add 

```js
runtime.page.on('pageerror', (err) => {
    console.log('pageerror: ', err)
  })
```

The console log is called with the error and no warning message is printed . I could add an additional check, what do you think?

src/mockServiceWorker.js
marcosvega91: ah yes, I didn't see it. I think that we can add the same code there and add a new test

test/msw-api/life-cycle-events/regression/null-body.test.ts
kettanaito: I think we can't go wrong with including this behavior into the assertions directly. If the error behavior can be observed by an exception thrown on the page, the test should assert that exception doesn't get thrown. Ideally, it doesn't even assert anything else.

src/mockServiceWorker.js
kettanaito: Not sure about adding the same code there. Line 124 is controlled by the end developer. If they provide a body or an empty string as a body, for a 204 response they are making a mistake. It would be good for the underlying logic to tell them about it, instead of coercing their mistakes. 

In case of the `response` event (to which this entire `.then` clause is dedicated), the mistake is on our side, because the end developer doesn't directly construct the `Response` instance, instead they act as a consumer of an already valid response instance.

test/msw-api/life-cycle-events/regression/null-body.test.ts
marcosvega91: I have updated the assertion, removing the warning log

src/mockServiceWorker.js
marcosvega91: > Not sure about adding the same code there. Line 124 is controlled by the end developer. If they provide a body or an empty string as a body, for a 204 response they are making a mistake. It would be good for the underlying logic to tell them about it, instead of coercing their mistakes.

Maeby we can create a real `Response` object somewhere, for example in `createResponseComposition`. In case of error we could inform the user of the issue.

kettanaito: We can, but it would be only for that purpose alone, as we cannot send the `Response` instance to the worker, it needs to be converted into JSON beforehand. I think it's not the issue that the error happens in the worker, if that's your concern.

kettanaito (approved):
Looks great to me, @marcosvega91. Thank you! Feel free to merge once it's ready.

src/mockServiceWorker.js
marcosvega91: yes yes I know it was only for a check. Surround `Response` with a try-catch. If an error is caught we could inform the user otherwise we will continue as is

marcosvega91: This is of course not regarding to this issue.

I mean something like the following with a better message than mine hahahaha 

```diff
diff --git a/src/response.ts b/src/response.ts
index 5fe12a4..6590907 100644
--- a/src/response.ts
+++ b/src/response.ts
@@ -67,11 +67,18 @@ export function createResponseComposition<BodyType>(
       ...transformers,
     ].filter(Boolean)
 
-    const resolvedResponse =
-      resolvedTransformers.length > 0
-        ? compose(...resolvedTransformers)(initialResponse)
-        : initialResponse
+    const resolvedResponse = await (resolvedTransformers.length > 0
+      ? compose(...resolvedTransformers)(initialResponse)
+      : initialResponse)
 
+    try {
+      new Response(resolvedResponse.body, { ...resolvedResponse })
+    } catch (e) {
+      throw new Error(`
+There is an issue with the response in your handler. This is the error thrown:
+"{e}"
+In most cases your body has a conflict with status. Make sure that your are following the principles of fetch. See https://fetch.spec.whatwg.org/#responses`)
+    }
     return resolvedResponse
   }
 }
```

kettanaito: However, the same (but less user-friendly) error would surface to the user anyway. I'm not sure this extra check is necessary. Everybody will pay for it, as the `Response` that's not used will be constructed for each `res()` invocation, but only those who make mistakes would get the benefit of this check. Moreover, this check would guard against a spec-incompliant usage more than an invalid usage of MSW. I'd say this check is not necessary as of now.

marcosvega91: for me is ok

PR #538, opened 2021-01-13
Support delay modes (ctx.delay)

kettanaito:
## Motivation

Using `ctx.delay(Infinity)` executes the response timer immediately instead of delaying the response indefinitely, as expected. That's due to how `setTimeout` (that powers response callback to the worker) operates in JavaScript. These changes aim to add support for infinite response delays and add some safeguards against using large integers as the `ctx.delay` inpt. 

## Changes

- `ctx.delay` now supports delay modes—enums that abstract specific delay behaviors:
  - `real`, uses a random realistic server response time.
  - `infinite`, uses a maximum allowed integer as a `setTimeout` duration to delay a response indefinitely (useful when emulating pending/loading states).
- `ctx.delay` now throws an exception when given `Infinity`, `Number.MAX_VALUE`, `Number.MAX_SAFE_INTEGER`, or any integer value larger than the maximum allowed `setTimeout` duration.
- `ctx.delay` now throws an exception when given an unknown delay mode.

## Usage

```js
ctx.delay() // realistic response time
ctx.delay(500) // explicit response time
ctx.delay('real') // the same as `ctx.delay()`
ctx.delay('infinite') // infinite response time
```

### DIFF
--- a/src/context/delay.ts
+++ b/src/context/delay.ts
@@ -17,7 +17,7 @@ const getRandomServerResponseTime = () => {
   )
 }
 
-type DelayMode = 'real' | 'infinite'
+export type DelayMode = 'real' | 'infinite'
 
 /**
  * Delays the response by the given duration (ms).
--- a/src/context/delay.ts
+++ b/src/context/delay.ts
@@ -1,6 +1,7 @@
 import { ResponseTransformer } from '../response'
 import { isNodeProcess } from '../utils/internal/isNodeProcess'
 
+export const SET_TIMEOUT_MAX_ALLOWED_INT = 2147483647
 export const MIN_SERVER_RESPONSE_TIME = 100
 export const MAX_SERVER_RESPONSE_TIME = 400
 export const NODE_SERVER_RESPONSE_TIME = 5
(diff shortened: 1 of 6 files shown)


src/context/delay.ts
```diff
@@ -16,16 +17,54 @@ const getRandomServerResponseTime = () => {
...
   return (res) => {
-    res.delay = durationMs ?? getRandomServerResponseTime()
+    let delayTime: number
+
+    if (typeof durationOrMode === 'string') {
+      switch (durationOrMode) {
+        case 'infinite': {
+          // Using `Infinity` as a delay value executes the response timeout immediately.
+          // Instead, use the maximum allowed integer for `setTimeout`.
+          delayTime = SET_TIMEOUT_MAX_ALLOWED_INT
```
timdeschryver, line 41: I think these `case` statements need a `break` in between, otherwise it cascades and end up by throwing an error.

timdeschryver (dismissed):

kettanaito:
Failing to test an infinite delay scenario in Puppeteer, because it throws an exception `Target is closed` when attempting to close a browser with a pending network request.

kettanaito:
I've added only the `"real"` integration test, others seem to be too brittle.

kettanaito:
Experienced a failing test. Passes locally.

```bash
 FAIL  msw-api/setup-server/scenarios/fall-through.node.test.ts (60.105 s)
  ● falls through all relevant handler even if none returns response

    : Timeout - Async callback was not invoked within the 60000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 60000 ms timeout specified by jest.setTimeout.Error:

      41 | })
      42 | 
    > 43 | test('falls through all relevant handler even if none returns response', async () => {
         | ^
      44 |   const res = await fetch('https://test.mswjs.io/blog/article', {
      45 |     method: 'POST',
      46 |   })

      at new Spec (../node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Object.<anonymous> (msw-api/setup-server/scenarios/fall-through.node.test.ts:43:1)
```

PR #540, opened 2021-01-15
Support an option for request body to be an object

shevelevs:
Since request body is parsed into an object if request is of type JSON, DefaultRequestBodyType needs to support this. Current workaround without having to explicitly cast to any is `rest.post<any, any>`, but should be supported by default type as well.

### DIFF
--- a/src/utils/handlers/requestHandler.ts
+++ b/src/utils/handlers/requestHandler.ts
@@ -14,7 +14,10 @@ export const defaultContext = {
 }
 
 export type DefaultRequestBodyType = 
-  | Record<string, any> | string | any | undefined
+  | Record<string, any>
+  | string
+  | any
+  | undefined
 
 export interface MockedRequest<
   BodyType = DefaultRequestBodyType,
--- a/src/utils/handlers/requestHandler.ts
+++ b/src/utils/handlers/requestHandler.ts
@@ -13,7 +13,8 @@ export const defaultContext = {
   fetch,
 }
 
-export type DefaultRequestBodyType = Record<string, any> | string | any | undefined
+export type DefaultRequestBodyType = 
+  | Record<string, any> | string | any | undefined
 
 export interface MockedRequest<
   BodyType = DefaultRequestBodyType,
--- a/src/utils/handlers/requestHandler.ts
+++ b/src/utils/handlers/requestHandler.ts
@@ -13,7 +13,7 @@ export const defaultContext = {
   fetch,
 }
 
-export type DefaultRequestBodyType = Record<string, any> | string | undefined
+export type DefaultRequestBodyType = Record<string, any> | string | any | undefined
 
 export interface MockedRequest<
   BodyType = DefaultRequestBodyType,
--- a/src/utils/handlers/requestHandler.ts
+++ b/src/utils/handlers/requestHandler.ts
@@ -13,7 +13,7 @@ export const defaultContext = {
   fetch,
 }
 
-export type DefaultRequestBodyType = 
+export type DefaultRequestBodyType =
   | Record<string, any>
   | string
   | any


src/utils/handlers/requestHandler.ts
```diff
@@ -13,7 +13,11 @@ export const defaultContext = {
   fetch,
 }
 
-export type DefaultRequestBodyType = Record<string, any> | string | undefined
+export type DefaultRequestBodyType =
+  | Record<string, any>
+  | string
+  | any
```
kettanaito, line 19: Adding `any` to the union type effectively make a type equal to just `any` type`:

```ts
type A = string | number | any
type B = any
```

Types `A` and `B` are the same when it comes to type-checking. I think we should annotate the missing body types, but not via `any`.

kettanaito:
Hey, @shevelevs. Thanks for opening this. Could you please elaborate more on what issue are you experiencing?

> Since request body is parsed into an object if request is of type JSON, DefaultRequestBodyType needs to support this.

I believe the `DefaultRequestBodyType` type already supports that by including `Record<any, any>` in its union type. 

I'd like to learn more about your use-case so we could find a suitable solution. Thanks.

shevelevs:
> Hey, @shevelevs. Thanks for opening this. Could you please elaborate more on what issue are you experiencing?
> 
> > Since request body is parsed into an object if request is of type JSON, DefaultRequestBodyType needs to support this.
> 
> I believe the `DefaultRequestBodyType` type already supports that by including `Record<any, any>` in its union type.
> 
> I'd like to learn more about your use-case so we could find a suitable solution. Thanks.

Maybe it's the matter of removing the options, not adding them. 

But basically, let's say I'm posting something _as json_, say an object `{foo: "bar"}`. What are my options as far as handler body in this case?

Option 1
```    
rest.post('/some/endpoint', (req, res, ctx) => {
        const foo = req.body["foo"];
...
}
```

Option 2
```    
rest.post<any, any>('/some/endpoint', (req, res, ctx) => {
        const foo = req.body.foo;
...
}
```

Option 3
``` 
interface Foo {
	foo: string;
}
rest.post<Foo, any>('/some/endpoint', (req, res, ctx) => {
        const foo = req.body.foo;
...
}
```

Note, that in order for me to access `.foo` I have to either use Record's interface (`req["foo"]`) or specify types for the generic method (`post<any, any>` or `post<Foo, any>`). So it kind of feels inconsistent that what seems to be the most frequently used use case (I assume) is not made easy by the "Default" request body type, as in 

(non-existent) Option 4
```    
rest.post('/some/endpoint', (req, res, ctx) => {
        const foo = req.body.foo;
...
}
```

But again, maybe it's the matter or removing everything but string and undefined, like
```DefaultRequestBodyType = string | undefined```. If one needs more specific type, they should instantiate the template.

**Comment by kettanaito** [comment]
Thanks for the detailed overview of the issue. 

I find Option 3 to be the right one to use if you wish to handle a request body as JSON. The request body is assumed as text by default, and I see no reason to change that. If you know that your request body is of a certain type/shape, you can take an explicit action of describing that shape and providing it to the request handler generic:

```ts
interface RegisterRequestBody {
  email: string
  password: string
}

rest.post<RegisterRequestBody>('/register', (req, res, ctx) => {
  const { email, password } = req.body
})
```

I wouldn't recommend Option 1 or 2 since you're not getting any type safety that way: the request body can indeed be anything. 

Do you see any particular issues with using Option 3?

kettanaito:
Thank you for this suggestion, but I don't think that there's an issue that needs fixing. 

Until you provide an explicit type, `req.body` is de-facto either a string or a parsed JSON. That's why you'd have to treat it that way in your handlers (described options 1 and 2 above). 

If you know your request body type, which in most cases you do, annotate your request handler using its first generic:

```ts
interface Body {
  username: string
}

rest.post<Body>('/user', (req, res, ctx) => {
  const { username } = req.body // OK
})
```

PR #542, opened 2021-01-16
Makes "ctx.data" and "ctx.errors" input types compatible with "ExecutionResult" of "graphql"

### LINKED ISSUE #522: ctx.data() and ctx.error() types are incompatible with `graphql` package `ExecutionResult` result type
## Describe the bug
Passing the result of an executed graphql query to `ctx.data()` or `ctx.errors()` results in a TypeScript error.

## Environment

- `msw: 0.24.2`
- `nodejs: 14.11.2`
- `graphql: 15.3.0`
- `typescript: 4.1.2`

## To Reproduce

```ts
import { graphql as mswGraphql } from "msw";
import { graphql } from "graphql";
import { buildSchema } from "graphql/utilities";

mswGraphql.query("Foo", async (req, res, ctx) => {
  const rsp = await graphql(
    buildSchema(""),
    req.body?.query || "",
    {},
    req.variables
  );

  // Typescript errors thrown on both ctx.data() and ctx.errors()
  return res(ctx.data(rsp.data), ctx.errors(rsp.errors));
});
```

[The return value of `await graphql()` is this `ExecutionResult` type](https://github.com/graphql/graphql-js/blob/607345275f60e07dba1b7156a23b9ddf8b086fc9/src/execution/execute.d.ts#L49-L57), and that is incompatible with the types of `ctx.data` and `ctx.error`

**`ctx.data()` error:**

```
Argument of type '{ [key: string]: any; } | null | undefined' is not assignable to parameter of type 'Record<string, any>'.
  Type 'undefined' is not assignable to type 'Record<string, any>'.
```

**`ctx.errors()` error:**

```
Argument of type 'readonly GraphQLError[] | undefined' is not assignable to parameter of type 'Partial<GraphQLError>[] | null | undefined'.
  The type 'readonly GraphQLError[]' is 'readonly' and cannot be assigned to the mutable type 'Partial<GraphQLError>[]'.
```

## Expected behavior

I would expect MSW's types to be compatible with `graphql()`.

While simply fixing the types on those two methods would be sufficient, perhaps it would be worth introducing an additional `ctx` method that takes the return value of `graphql()` directly, e.g.

```ts
res(ctx.graphql(rsp))
```

This would allow it to handle any data/type transformations that may be required for compatibility with `ctx.data/error()`, while also adding a convenient API for users.

kettanaito:
## GitHub

- Fixes #522

### DIFF
--- a/src/context/errors.ts
+++ b/src/context/errors.ts
@@ -10,7 +10,7 @@ import { json } from './json'
  * @see {@link https://mswjs.io/docs/api/context/errors}
  */
 export const errors = <
-  ErrorsType extends Partial<GraphQLError>[] | null | undefined
+  ErrorsType extends readonly Partial<GraphQLError>[] | null | undefined
 >(
   errorsList: ErrorsType,
 ): ResponseTransformer<string> => {
--- a/test/graphql-api/compatibility.test.ts
+++ b/test/graphql-api/compatibility.test.ts
@@ -1,0 +1,82 @@
+import { graphql as executeGraphql, GraphQLError } from 'graphql'
+import { buildSchema } from 'graphql/utilities'
+import { createApolloFetch } from 'apollo-fetch'
+import { graphql } from 'msw'
+import { setupServer } from 'msw/node'
+
+const schema = `
+type User {
+  firstName: String!
+}
+
+type Query {
+  user: User!
+}
+`
+
+const server = setupServer(
+  graphql.query('GetUser', async (req, res, ctx) => {
+    const executionResult = await executeGraphql(
+      buildSchema(schema),
+      req.body.query,
+      {
+        user: {
+          firstName: 'John',
+        },
+      },
+      req.variables,
+    )
+
+    return res(
+      ctx.data(executionResult.data),
+      ctx.errors(executionResult.errors),
+    )
+  }),
+)
+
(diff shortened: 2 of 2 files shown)


kettanaito:
Hey, @ericbiewener. This pull request should fix the issue you've described.

Could you please check if these changes work for you? You can use the Codesandbox preview project above to experiment directly in the sandbox environment or install this pull request in your project. 

Surprisingly, there was no need to change `ctx.data`, as it's never typed as `undefined`:

```ts
// Type definitions of the "graphql" package
export interface ExecutionResult<
  TData = { [key: string]: any },
  TExtensions = { [key: string]: any }
> {
  errors?: ReadonlyArray<GraphQLError>;
  // TS_SPECIFIC: TData. Motivation: https://github.com/graphql/graphql-js/pull/2490#issuecomment-639154229
  data?: TData | null;
  extensions?: TExtensions;
}
```

kettanaito:
Hey, @timdeschryver 👋 Sorry for troubling you. May I ask you for a quick review of these changes? Thanks!

timdeschryver (approved):
👍 LGTM, I think we already had fixed the issue with `ctx.data` because I also wasn't able to reproduce that.

PR #543, opened 2021-01-16
feat: multipart/form-data support

### LINKED ISSUE #215: Support for GraphQL Upload/Multipart Spec
Yes. Mocking mutations that send files results to an error.

Have support for mocking GraphQL uploads. I understand uploads aren't part of the core GraphQL spec, but [having a possibly good chunk of GraphQL users who use the upload spec](https://www.npmtrends.com/graphql-upload-vs-apollo-upload-client-vs-graphql) might be a good reason to implement it.

For our use case, we just need a way to intercept the mutation without errors so we could return a mocked response. This can be expanded to full file mocking if someone else needs it.

Our end goal is to be able to test our mutations that upload files. We currently use [MockedProvider](https://www.apollographql.com/docs/react/development-testing/testing/#mockedprovider).

**Additional context**
Looking into the code, I see that msw doesn't have parse logic for multipart requests, so req.body isn't parsed properly.

Multipart request spec - https://github.com/jaydenseric/graphql-multipart-request-spec
Package I use to upload - https://www.npmjs.com/package/apollo-upload-client
Sample repo - https://github.com/jodelasur/msw-graphql-upload-test

Notes on sample repo:
- There are three tests in App.test.js. You should look at the third one ('can upload the file').
- You should run each test individually. Running the whole App describe block results to an error. I'm assuming it's because the Apollo client doesn't reset in-between tests. I don't know how to fix this yet.

tkamenoko:
This PR fixes #215 .

It contains multipart/form-data parser that enable to parse a request body into an object.

It also has variables builder based on [this spec](https://github.com/jaydenseric/graphql-multipart-request-spec) .

In addition, rest-api can handles multipart/form-data now.


src/utils/internal/multipartParse.ts
```diff
@@ -0,0 +1,81 @@
...
+  const name = _name.slice(1, -1)
+  const filename = _filename === undefined ? undefined : _filename.slice(1, -1)
+  return { name, filename, contentType }
+}
+
+/**
+ * Parses a given string as a multipart/form-data.
+ * Does not throw an exception on an invalid multipart string.
+ */
+export function multipartParse<T extends DefaultMultipartBodyType>(
```
kettanaito: What do you think if we name this function `parseMultipartData`?

test/graphql-api/link.test.ts
```diff
@@ -48,13 +55,9 @@ test('mocks a GraphQL query to the GitHub GraphQL API', async () => {
       },
     },
   })
-
-  await runtime.cleanup()
```
kettanaito: I would not advise moving this teardown to the `beforeAll`/`afterAll` hooks. This may be a personal preference, but I prefer `runtime` to be explicitly declared and cleaned within each test. 

Take a look at [Avoid nesting when you're testing](https://kentcdodds.com/blog/avoid-nesting-when-youre-testing) by Kent C. Dodds. I think it has good points as to why I've adopted an explicit `runtime` declaration approach.

kettanaito: Removing this change will also help us keep the pull request changes scoped, which is always a good thing to do. If we discover that using jest hooks would improve our tests, we can always roll this out as a separate dedicated pull request.

test/graphql-api/multipart.mocks.ts
```diff
@@ -0,0 +1,38 @@
...
+    {
+      file1?: File
+      file2?: File
+      files?: File[]
+      plainText?: string
+    }
+  >('UploadFile', async (req, res, ctx) => {
+    const { file1, file2, files, plainText } = req.variables
+    const filesResponse: string[] = []
+    for (const file of files ?? []) {
```
kettanaito: `??` allows `files` to equal a falsy value (i.e. `false`). This would still break the `for/of` loop:

```js
for (const file of false) // false is not iterable
```

I suggest we use `||` instead. Nullish coalescing operator (`??`) is useful when dealing with possibly falsy values, which isn't the case for `files`. If `files` is undefined or falsy we should always fallback to an empty array to not break the `for`/`of` loop.

test/graphql-api/multipart.mocks.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+    {
+      file1?: File
+      file2?: File
+      files?: File[]
+      plainText?: string
+    }
+  >('UploadFile', async (req, res, ctx) => {
+    const { file1, file2, files, plainText } = req.variables
+    const filesResponse: string[] = []
```
kettanaito: What do you think if we use `Promise.all` and a plain `files.map` instead of the `for`/`of` loop?

```js
const { files = [] } = req.variables
const filesResponse = await Promise.all(files.map(file => file.text())
```

It's going to be less code, making this example more focused. `Promise.all` would also be a more recommended approach so that you don't wait for each file to be processed individually, allowing parallel transformations to text. 

I'm curious to hear your opinion.

test/graphql-api/utils/executeOperation.ts
```diff
@@ -51,21 +53,45 @@ export const executeOperation = async (
...
+      fileContents: string[] | undefined,
+    ) => {
+      const operations = JSON.stringify({
+        query,
+        variables,
+      })
+
+      const isMultipart = Boolean(map && fileContents)
+      const headers = isMultipart ? {} : { 'Content-Type': 'application/json' }
+      let body: string | FormData
```
kettanaito: I'd refactor this part as follows:

```js
const body = isMultipartData ? getMultipartGraphQLBody(operations, ...) : operations
```

This will make `executeOperations` much easier to read by delegating the logic that creates a `FormData` into a separate utility function that we can also test separately.

test/rest-api/body.test.ts
```diff
@@ -90,3 +90,49 @@ test('handles a POST request with a JSON body and "Content-Type: application/jso
...
+text content\r
+------WebKitFormBoundaryvZ1cVXWyK0ilQdab\r
+Content-Disposition: form-data; name="text2"\r
+\r
+another text content\r
+------WebKitFormBoundaryvZ1cVXWyK0ilQdab\r
+Content-Disposition: form-data; name="text2"\r
+\r
+another text content 2\r
+------WebKitFormBoundaryvZ1cVXWyK0ilQdab--\r\n`.trimStart()
```
kettanaito: Prefer using `\` (backward slash) to escape an extra new line of a formatted template literal string:

```js
const string = `\
yes
`
```

test/rest-api/body.test.ts
```diff
@@ -90,3 +90,49 @@ test('handles a POST request with a JSON body and "Content-Type: application/jso
...
+  })
+  const res = await runtime.request({
+    url: `${runtime.origin}/upload`,
+    fetchOptions: {
+      method: 'POST',
+      headers,
+      body,
+    },
+  })
+  const jsonRes = await res.json()
```
kettanaito:
```diff
-jsonRes
+body
```

kettanaito (changes requested):
Hey, @tkamenoko. Thank you for your hard work on this! It looks fantastic. 

I've left a few review comments, could you please take a look? My main suggestions were:

- Don't move `runtime` declarations to a `beforeAll`/`afterAll` hooks. I see this as an additional internal refactoring. Let's keep the changes of this pull request scoped. 
- A few naming and structuring suggestions. 

I'm excited to see this landing in the next release. Once more, huge thanks!

src/utils/internal/multipartParse.ts
tkamenoko: A function `jsonParse` already exists. Should I also rename it?

test/graphql-api/link.test.ts
tkamenoko: OK, I'll remove that change. Just my curiosity, what do you think about resource leaks when a test fails?

src/utils/internal/multipartParse.ts
kettanaito: I would leave `jsonParse` as-is. You're right to aim for consistency here. I believe `jsonParse` was named to resemble `JSON.parse`. We can address it separately if you want.

test/graphql-api/link.test.ts
kettanaito: That's a viable concern, as an exception during the test instructions (i.e. originating from a failed assertion) will short circuit the test function and the cleanup phase won't ever get called.

That being said, I don't think that before/after hooks are the solution to that. Not unless you want to ditch the "shallow" tests principle. 

That's the main reason I've written [`flat-test`](https://github.com/kettanaito/test-flat). It allows to have shallow flat tests while each test accepts a before and after steps that always execute and can share context between each other (one of the pain points of having to mutate things in a conventional before/after setup). We may consider migrating MSW test suites to use `flat-test`.

test/graphql-api/multipart.test.ts
```diff
@@ -0,0 +1,60 @@
...
+      variables: {
+        file1: null,
+        file2: null,
+        files: [null, null],
+        plainText: 'text',
+      },
+    },
+    {
+      map: {
+        '0': ['variables.file1', 'variables.files.0'],
```
kettanaito: Can this be an Array instead?

```js
map: [
  ['file1', 'file2'],
  [c, d]
]
```

test/graphql-api/utils/executeOperation.ts
```diff
@@ -51,21 +53,54 @@ export const executeOperation = async (
...
+      const getMultipartGraphQLBody = (
+        operations: string,
+        map: Record<string, string[]>,
+        fileContents: string[],
+      ): FormData => {
+        const body = new FormData()
+        body.append('operations', operations)
+        body.append('map', JSON.stringify(map))
+        const files = fileContents.map(
+          (f: string, i: number) => new File([f], `file${i}.txt`),
```
kettanaito: No need to explicitly annotate `f` and `i`. `fileContents` is a type-safe array of strings (`string[]`). Those argument types will be inferred.

kettanaito (dismissed):
This looks **solid**. Let me know if you think these changes are ready to be merged. I'll give them a test locally to get to know multipart data scenarios better.

Thank you for your work on this, @tkamenoko!

test/graphql-api/multipart.test.ts
tkamenoko: As far as I read [the spec](https://github.com/jaydenseric/graphql-multipart-request-spec#examples) , `map` should be an object, not an array.

https://github.com/graphql-python/gql/blob/c17959c453c4f9c631c31ddb987ebbd3202b971a/gql/transport/aiohttp.py#L167-L169

This is python implementation of GraphQL client. It builds `map` as `dict` type that will be encoded to object in JSON.

tkamenoko:
@kettanaito Thanks for reviewing! I think this PR is ready to merge. Please do it whenever you want.

test/graphql-api/multipart.test.ts
kettanaito: Got it 👍 No harm keeping it this way, just made me curious.

kettanaito (approved):
This is an incredible addition. Thank you, @tkamenoko! Welcome to contributors 🎉

otaciliolacerda:
Wow, great addition. Thanks for the contribution.

PR #544, opened 2021-01-18
onUnhandledRequest: Uses "console.error" instead of throwing an exception

### LINKED ISSUE #539: Show onUnhandledRequest: 'error' in stdout and fail tests
Happy new year & thanks a lot for `msw`, it's a fantastic library 🙏 

I have an app bootstrapped with `create-react-app` and various tests that I run with Jest. I would like to configure my tests to fail if `msw` detects any unhandled network request. Hence, I've implemented the following configuration:

```js
server.listen({ onUnhandledRequest: 'error' })
```

Unfortunately, doing so logs the unhandled network request errors to `stderr`, which the Jest watch mode by default doesn't show & doesn't fail the test execution. Hence, I've switched back to `warn` so that I can at least see if any network requests are unhandled when running the tests.

1. It would be great if there's a way to direct the `error` level to stdout through the options of `server.listen`
2. Likewise, maybe we could create an additional option for an event handler (e.g. `onUnhandledRequestHandler`) that one can customise to throw or run `process.exit(1)` when we encounter an unhandled network request

**Additional context**

> "bypass" | Performs an unhandled request as-is.
> -- | --
> "warn" | Prints a warning into stdout of the current process.
> "error" | Prints an error into stderr of the current process.
>
> Source: [docs](https://mswjs.io/docs/api/setup-server/listen#onunhandledrequest)

### LINKED ISSUE #573: MSW error not visible to Axios
## Describe the bug
_(I'm not sure if this is an issue in axios or in MSW - feel free to redirect me if needed)_

When using fetch to do a request to a non-existing MSW resource it throws a helpful MSW error. However this error is not visible when using axios (which just throws a `Network error`), which can cause a lot of time debugging in the wrong direction.

## Environment

- `msw: 0.26.0`
- `nodejs: 12.20.1`
- `npm: 6.14.10` (but not used)

n/a

## To Reproduce
1. `git clone https://github.com/JCB-K/msw-axios-testcase`
2. `cd msw-axios-testcase`
3. `yarn`
4. `yarn test`

## Expected behavior
All tests should pass.

## Screenshots

image

## Axios or MSW?

Like I mentioned above, I'm unsure where the issue lies. Normally I would say axios since that is doing the request and MSW is just the server which should expose a network error. But this is a bit of a different kind of error since it happens in the same environment, and is probably caught in a different way?

kettanaito:
## GitHub

- Fixes #539
- Fixes #573 

## Changes

- `onUnhandledRequest: "error"` now prints a `console.error` message instead of throwing an exception.

### DIFF
--- a/src/utils/request/onUnhandledRequest.ts
+++ b/src/utils/request/onUnhandledRequest.ts
@@ -32,11 +32,13 @@ export function onUnhandledRequest(
 
   switch (handler) {
     case 'error': {
-      throw new Error(`[MSW] Error: ${message}`)
+      console.error(`[MSW] Error: ${message}`)
+      break
     }
 
     case 'warn': {
       console.warn(`[MSW] Warning: ${message}`)
+      break
     }
 
     default:
--- a/test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
+++ b/test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
@@ -20,10 +20,14 @@ afterAll(() => {
 })
 
 test('errors on unhandled request when using the "error" value', async () => {
+  jest.spyOn(console, 'error')
+
   const getResponse = () => fetch('https://test.mswjs.io')
 
-  await expect(getResponse()).rejects.toThrow(`\
-request to https://test.mswjs.io/ failed, reason: [MSW] Error: captured a GET https://test.mswjs.io/ request without a corresponding request handler.
+  await getResponse()
+
+  expect(console.error)
+    .toHaveBeenCalledWith(`[MSW] Error: captured a GET https://test.mswjs.io/ request without a corresponding request handler.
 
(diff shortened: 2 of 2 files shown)


timdeschryver (dismissed):
This LGTM, there's a test that fails though

rssfrncs:
Hey. When will this be available in an official release? I'm on version `0.27.1` and my jest tests are not failing in watch mode. Should async tests fail?

kettanaito:
Hey, @rssfrncs. This has been released in https://github.com/mswjs/msw/releases/tag/v0.26.1. 

There's confusion by what people mean when saying "fail". In the context of this change, it means "exiting the process". That must happen in neither watch nor regular mode with testing frameworks. 

Can you share with us your test, the output of that test, and what do you expect to happen? Ideally, in a new issue so it's easier to track. Thanks.

Issue #547, opened 2021-01-18
Error while doing response patching with GraphQL


lenoirzamboni:
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.22.0   |
| browser | Chrome     |
| OS      | macOS     |

## Current behavior

While running on the **staging** environment, a response patching is failing due to the addition of a strange `map` header:

```
Access to fetch at 'https://api.somesite.com/graphql/' from origin 'https://staging.somesite.com' has been blocked by CORS policy: Request header field map is not allowed by Access-Control-Allow-Headers in preflight response.
```

![success_request_headers](https://user-images.githubusercontent.com/4825333/104938620-16f03080-598e-11eb-864d-025ca4eb21d0.png)

This `map` header, is literally a map containing:

![map_header_content](https://user-images.githubusercontent.com/4825333/104938035-4488aa00-598d-11eb-8717-0c8c2f2de316.png)
 
**OBS**: The response patch does not fail while running the same code on the **localhost** environment.

## Expected behavior

It should return the original data plus the appended mocked data.

## Request handlers

```js
import { graphql } from 'msw'

...

const allHistoryOrders = graphql.query(
  'someQuery',
  async (req, res, ctx) => {
    if (req.variables.something === SOME_DATA_X) {

      const originalResponse = await ctx.fetch(req)
      const originalResponseData = await originalResponse.json()

      const { data } = originalResponseData

      const newResults = data.someData.results.concat(
       someArrayData
      )
      data.someOtherData.results = newResults

      return res(ctx.data(data))
    }
  }
)

export { allHistoryOrders }
```

## Actual request

The actual request is made by using the [useQuery](https://www.apollographql.com/docs/react/data/queries/#executing-a-query) hook from Apollo Client (React).

---

Can you guys help me, please? Am I doing something wrong that I do not know about? Is this maybe a bug? 

Thanks!

kettanaito:
Hey, @lenoirzamboni. Thanks for reporting this.

That `map` property is an internal property of the [`Headers` polyfill](https://github.com/mswjs/headers-utils/blob/14112f6bbb0dc7f24be92f2b4afb03a6c09066b8/src/Headers.ts#L4) that MSW uses in Node to operate with headers as with a [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) instance. That API is not supported by all request modules in Node, thus the polyfill. 

[...]

PR #550, opened 2021-01-20
Adds GraphQL response patching integration test

kettanaito:
## GitHub

- Attempts to reproduce #547 

## Changes

- Adds a browser integration test for GraphQL response patching.
- Adds a NodeJS integration test for GraphQL response patching. 
- Removes any third-party GraphQL clients (i.e. Apollo) and uses a minimal specification-compliant client instead.


kettanaito:
## Removing third-party GraphQL clients

I suggest stopping to rely on third-party GraphQL clients. Those often introduce a custom abstraction over "raw" GraphQL operations and execution results. By using third-party GraphQL clients in tests we are indeed closer to how GraphQL is being used by developers, but we also expose our tests to unnecessary dependency on those abstractions that GraphQL clients introduce. 

For example, when using `@apollo/client` all the data nodes in the execution result will have a `__typename` property. This is not the behavior of running `graphql()` directly. Our tests do not need to assert that to ensure that `graphql.*` request handlers produce a spec-compliant payload. 

### What about client-related issues?

If there's a GraphQL client-related issue and we'd like to add a test to ensure it doesn't happen, I suggest putting a scenario of that issue into the [Examples repository](https://github.com/mswjs/examples) and write a necessary test there. This way the main repo is not polluted by all sorts of possible usage combinations and tests only the vitals. 

### Solution

Use a minimal custom GraphQL client that operates via `fetch` and constructs a spec-compliant request.

marcosvega91 (approved):
As said on discord I agree with this approach. Trying to cover client behaviour could not be very useful. If there will be some client-specific issue we will add a test case for it

PR #551, opened 2021-01-20
fix: add client guards in service worker

### LINKED ISSUE #536: Do not listen to all messages posted on broswer
## Describe the bug

MSW reads all posted messages as MSW messages. However, _anything_ can post a message. In particular, the react dev tools posts messages all the time. With MSW installed, MSW will throw uncaught exceptions.

image

## Environment

- `msw@0.24.4`
- `nodejs: v14.15.1`
- `npm: 6.14.4`

Google Chrome Version 87.0.4280.141 (Official Build) (x86_64)
macOS Big Sur 11.1

## To Reproduce

1. Install React Developer Tools
2. Run test suite (on a browser) with MSW
3. See error in console

## Expected behavior

There shouldn't be any errors.

## Screenshots

Here is a screenshot of the failing line in MSW:

image

And more stack traces (which all seem to be related to react-dev-tools):

image

timdeschryver:
Closes #536

This PR adds guard clauses to prevent the service worker throwing errors when the client isn't found.
Making this a draft until we can reproduce the behavior in #536

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -21,8 +21,13 @@ self.addEventListener('activate', async function (event) {
 })
 
 self.addEventListener('message', async function (event) {
+  if (!event.source.id) return
   const clientId = event.source.id
+
+  if (!event.currentTarget.clients) return
   const client = await event.currentTarget.clients.get(clientId)
+
+  if (!client) return
   const allClients = await self.clients.matchAll()
   const allClientIds = allClients.map((client) => client.id)
 
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -21,13 +21,18 @@ self.addEventListener('activate', async function (event) {
 })
 
 self.addEventListener('message', async function (event) {
-  if (!event.source.id) return
   const clientId = event.source.id
 
-  if (!event.currentTarget.clients) return
-  const client = await event.currentTarget.clients.get(clientId)
+  if (!clientId || !this.clients) {
+    return
+  }
 
-  if (!client) return
+  const client = await this.clients.get(clientId)
+
+  if (!client) {
+    return
+  }
+
   const allClients = await self.clients.matchAll()
   const allClientIds = allClients.map((client) => client.id)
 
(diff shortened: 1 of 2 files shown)


src/mockServiceWorker.js
```diff
@@ -21,8 +21,13 @@ self.addEventListener('activate', async function (event) {
 })
 
 self.addEventListener('message', async function (event) {
+  if (!event.source.id) return
```
kettanaito: Is it a possible scenario to receive an event without the client ID that sent it?

src/mockServiceWorker.js
```diff
@@ -21,8 +21,13 @@ self.addEventListener('activate', async function (event) {
...
   const clientId = event.source.id
+
+  if (!event.currentTarget.clients) return
```
kettanaito: I believe we can use `this.clients.get` instead of this to ensure there is always a global scope to resolve this client retrieval operation. As it turns out, `event.target`/`event.currentTarget` can be `null` in Safari.

src/mockServiceWorker.js
```diff
@@ -21,8 +21,13 @@ self.addEventListener('activate', async function (event) {
...
   const client = await event.currentTarget.clients.get(clientId)
+
+  if (!client) return
```
kettanaito: I believe this is the actual fix for the issue: if we didn't find the client by the given ID that's controlled by MSW's worker, then do not attempt to handle the incoming message.

src/setupWorker/setupWorker.ts
```diff
@@ -49,6 +49,11 @@ export function setupWorker(
           navigator.serviceWorker,
           'message',
           (event: MessageEvent) => {
+            // Avoid messages broadcasted from unrelated workers.
+            if (event.source !== context.worker) {
```
kettanaito, line 53: I think this is a good enhancement described [in this article](https://ponyfoo.com/articles/serviceworker-messagechannel-postmessage).

Since we establish a global listener on `navigator.serviceWorker`, it can receive messages from other workers that may control the current client. With this check we're proceeding with event handling only if the source of the event (message) is our MSW worker.

kettanaito (approved):

kettanaito:
Thank you for tackling this, @timdeschryver 🎉

PR #560, opened 2021-01-22
Suggests similar request handlers upon unhandled requests

### LINKED ISSUE #386: Give more guidelines if mocking fails

When a request is not mocked in msw the error of not mocking the request is sometimes hard to understand.
For example if in a GraphQL setup and mock is created for a query, but the tested unit actually does a mutation, msw will fail.
The error I've seen here was ```Network error: request to http://localhost:8080/api failed, reason: connect ECONNREFUSED 127.0.0.1:8080```.

What would be helpful in this case is that msw will look at the existing (setup) mocks and suggest if there are cases which could match.
Like: ```You mocked a query with name 'some-query' and there is a mock for a mutation with that name. Did you want to use that?```

### LINKED ISSUE #527: Provide information on GraphQL query when warning about unhandled requests
Loving using MSW so far, thank you!

When warning about an unhandled GraphQL query however, the warning only shows: `[MSW] Warning: captured a POST {GRAPH_QL_ENDPOINT} request without a corresponding request handler.`

It would be useful if this could at least contain the query name that is being run as well.

kettanaito:
## GitHub

- Fixes #386
- Fixes #527

## Changes

- Removes a "dummy" request handler code snippet from the unhandled request message. The library cannot reliably guess what kind of API you attempt to mock. The safest option is to link to the documentation where the developer decides on the right approach for them.
- Suggests a similar request handler based on the following criteria:
  - Similar request handler URL (REST API). Request handlers with the same method are always treated as more similar. Request handlers with a different method can still be suggested if there isn't any similar request handler of the same method.
  - Similar request handler operation name (GraphQL API). The same as with REST API handlers, GraphQL handlers are treated as more similar if their operation type matches the original request's operation type. Handlers of a different operation type can still be suggested if their operation name is similar. 

## Roadmap

- [x] Provide GraphQL suggestions.
- [x] REST: Suggest a handler of a different method if its path is similar.
- [x] GraphQL: Suggest a handler of a different operation type if its name is similar.
- [x] GraphQL: Ensure that handler suggestions are presented by the operation type. If there are multiple similar handlers, suggest the one with the same operation type as the request first. 
- [x] Suggests multiple similar handlers to increase the usefulness of the suggestions.


kettanaito:
Although `graphql.parse()` and `parseGraphQLRequest` have a similar logic, we cannot fully replace the former with the latter. The parsing logic of a GraphQL handler is far more complex: it deals with variables, the building of a multi-part/multi-operation body. We will simplify the parsing state as a part of Request Handler API refactoring (#561).

src/utils/request/onUnhandledRequest.ts
```diff
@@ -1,36 +1,162 @@
...
   | 'bypass'
   | 'warn'
   | 'error'
   | UnhandledRequestCallback
 
+type RestRequestHandler = RequestHandler<
+  MockedRequest<DefaultRequestBodyType, RequestParams>,
+  typeof restContext,
+  ParsedRestRequest
+>
```
timdeschryver: What about this? (`RestContext` should also be added to the imports)

```suggestion
type RestRequestHandler = RequestHandler<
  MockedRequest<DefaultRequestBodyType, RequestParams>,
  RestContext,
  ParsedRestRequest
>
```

src/utils/request/onUnhandledRequest.ts
```diff
@@ -1,36 +1,162 @@
...
+      return acc.concat([[score, handler]])
+    }, [])
+    .sort(([leftScore], [rightScore]) => {
+      return leftScore - rightScore
+    })
+    .find(([score]) => {
+      return score <= 2
+    })
+
+  return suggestedHandler?.[1]
```
timdeschryver: Woops, I was thrown off guard here and wanted to suggest `suggestedHandler?.[0]`.
But then I realized that 0 is the score and 1 is the handler.
Would the following make the intention more clear?

```suggestion
  const [_, suggestedHandler] = handlers
    .reduce<RequestHandlerSuggestionList>((acc, handler) => {
      const score = getScore(request, handler)
      return acc.concat([[score, handler]])
    }, [])
    .sort(([leftScore], [rightScore]) => {
      return leftScore - rightScore
    })
    .find(([score]) => {
      return score <= 2
    })

  return suggestedHandler
```

test/msw-api/setup-worker/start/on-unhandled-request/suggestions.test.ts
```diff
@@ -0,0 +1,290 @@
...
+        text,
+      )
+    })
+
+    expect(unhandledRequestWarning).toMatch(`\
+[MSW] Warning: captured a request without a matching request handler:
+
+  • GET /user-details
+
+If you still wish to intercept this unhandled request, please create a request handler for it. Read more: https://mswjs.io/docs/getting-started/mocks.`)
```
timdeschryver: Nit, should the read more paragraph has its own line?
I'm think that it's easier to catch as a user.

```suggestion
If you still wish to intercept this unhandled request, please create a request handler for it. 
Read more: https://mswjs.io/docs/getting-started/mocks.`)
```

test/msw-api/setup-worker/start/on-unhandled-request/suggestions.test.ts
```diff
@@ -0,0 +1,290 @@
...
+        text,
+      )
+    })
+
+    expect(unhandledRequestWarning).toMatch(`\
+[MSW] Warning: captured a request without a matching request handler:
+
+  • GET /users
+
+Did you mean to request "GET /user" instead?
```
timdeschryver, line 81:
```suggestion
Did you mean to intercept the "GET /user" request instead?
```

test/msw-api/setup-worker/start/on-unhandled-request/warn.mocks.ts
```diff
@@ -1,3 +1,4 @@
+import { graphqlOperation } from 'graphql-api/utils/executeOperation'
```
timdeschryver: Nit, this can be undone

```suggestion
```

src/utils/request/onUnhandledRequest.ts
```diff
@@ -1,36 +1,162 @@
...
+    parsedGraphQLQuery
+      ? getScoreForGraphQLHandler(parsedGraphQLQuery)
+      : getScoreForRestHandler(),
+  )
 
-  If you wish to intercept this request, consider creating a request handler for it:
+  if (suggestedHandler) {
+    const metaInfo = suggestedHandler.getMetaInfo()
+    handlerSuggestion = `Did you mean to request "${metaInfo.header}" instead?`
+  }
```
timdeschryver: Nit, what about declaring `handlerSuggestion`  as a const instead of reassigning it?

```suggestion
const handlerSuggestion = suggestedHandler 
  ? `Did you mean to request "${suggestedHandler.getMetaInfo().header}" instead?` 
  : ''
```

timdeschryver (commented):
This is a beauty @kettanaito !
I left a couple of subjective notes to make it a bit more readable.

test/msw-api/setup-worker/start/on-unhandled-request/suggestions.test.ts
kettanaito: That's not equivalent. The suggestion hints that you've requested a different resource. It treats an actual request as potentially invalid, not the handler. 

Given you have a `rest.get('/user')` handler but performed a `GET /users` request, the library will suggest that maybe you meant to request `GET /user`, because you have that handled.

src/utils/request/onUnhandledRequest.ts
kettanaito: It would, but `.find` can return `undefined`. One cannot destructure a possibly undefined value. Thus we would need a fallback in order to implement your suggestion:

```js
const [_, a] = [].find(predicate) || []
```

I think this is a little confusing.

kettanaito:
@timdeschryver has made a great suggestion to include _multiple_ handlers in the suggestions list. The first similar handler may not be the one a developer would be looking for. Including a list of up to 3-5 items of the suggestions with the same matching indices would be quite sensible. 

We can use `worker.use()` to provision per-test overrides to account for multiple possible combinations without polluting the same `*.mocks.ts` definition file.

src/utils/request/onUnhandledRequest.ts
timdeschryver: Oh shoot, you're right! I usually use this for `.filter` which works on an empty array, but I hadn't thought this through for `find` which can return `undefined` as you mentioned.

Sorry for the confusion!

test/msw-api/setup-worker/start/on-unhandled-request/suggestions.test.ts
timdeschryver: Thanks for taking the time to respond 😄

Issue #290, opened 2020-07-23
RFC: Request handler API


kettanaito:
## What

Need to refine an existing Request handler API to be publicly usable. 

The Request handler API is already designed with the support of creating custom request handlers for about any purpose (support of a specific communication convention, embed application's logic into handlers, etc.). However, if you take a look at the implementation example on the library's side, it uses quite a few internal functions, which may be useful to expose in the public API:

- `getCleanUrl`
- `isStringEqual`
- `match`
- The entire `log()` function.

Instead of proxying single functions from internal/external packages, MSW should expose a high-level API that can help developers build their own request handlers. Such high-level API may include:

- `matchesUrl()`, that would clean and match two given URLs just as MSW does for REST API.
- `log()`, a ready-to-use logging function (call signature preserved).

## Motivation

- #287 (support for CalDAV methods)
- Allows developers to create custom request handlers for internal and public usage (i.e. publishing in an NPM package):

```js
import { websockets } from 'msw-websockets-handlers'
import { setupWorker } from 'msw'

setupWorker(
  websockets.on('message', () => {...})
)
```

### Type declaration

https://github.com/mswjs/msw/blob/6001b790fbaafc21f8e7b922e172f4bd271f5272/src/handlers/requestHandler.ts#L54-L106

### Implementation example

https://github.com/mswjs/msw/blob/6001b790fbaafc21f8e7b922e172f4bd271f5272/src/rest.ts#L50-L147

marcosvega91:
I think that the features that you list could be enough to create custom libraries.
I was thinking about `websocket` and It would be a great feature to add to MSW. I'm only afraid that `websocket` can't be intercepted by `service worker` 😢

kettanaito:
@marcosvega91, yeap, we've got the WebSocket feature request opened here #156.

kettanaito:
With #319 the request matching API is now publicly exposed as `matchRequestUrl`. This may help developers in building their custom request handler while reusing the same request matching logic as MSW uses internally.

kettanaito:
We may have to expose a few internal functions (i.e. default loggers) for a more comfortable creation of custom handlers, but I find the current state of the request handler API sufficient for most purposes: 

https://github.com/mswjs/msw/blob/327b384938b1fdf44e77c5bd79e91b7fab0aaad4/src/utils/handlers/requestHandler.ts#L63-L122

I will document this in the recipe and close this.

PR #561, opened 2021-01-22
Re-designs the RequestHandler API

### LINKED ISSUE #489: Non-GraphQL requests with 'query' in URL params fail with AbortError
## Describe the bug

We have a setup where we talk to both a GraphQL API and some REST APIs as well, thus our server instance has both types of request handlers. If you make a regular REST request with `?query=some-not-graphql-value` in the URL params and your server also has GraphQL request handlers, the first GraphQL handler will [attempt to parse the request](https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/graphql.ts#L119-L133), which will fail and cause the request to error out.

For us, this was manifesting as `AbortError: Aborted` on those requests because it looks like the interceptor library aborts any requests where middleware throws errors. The underlying error causing it to abort the requests is a GraphQL syntax error when trying to parse: `GraphQLError: Syntax Error: Unexpected Name ...`.

We were getting this from `GET` requests, but after tracing through the code, I think the same is probably true if you tried it with `POST` and had `query` in the request body that was not a GraphQL query. We saw this behavior using the Node server in Jest unit tests.

## Environment

- `msw: 0.20.5`
- `nodejs: 14.4.0`

## To Reproduce

1. Create a server and set it to `.use` any GraphQL request handler (e.g. in Jest).
2. Create a test that sends a regular fetch `GET` request to a URL with `?query=not-graphql` in the URL.
3. See request fail when GraphQL handler parsing fails.

## Expected behavior

The request shouldn't fail in this case, although I'm unsure of what the correct fix is. It could be as simple as just doing a `try...catch` around parsing and then just returning `null` so that processing moves on to the next handler if parsing fails. If the ["predicate" step](https://github.com/mswjs/msw/blob/8a5f573aed97dadce448df7e4acfb0d23d8eb18c/src/utils/getResponse.ts#L33-L50) (i.e. URL matching I think?) happened before parsing, it would probably be possible for us to workaround it by specifying a URL for our GraphQL endpoint (i.e. use `graphql.link`) so that those GraphQL handlers wouldn't try to parse the request at all.

Happy to help with a PR or however I can.

kettanaito:
> **This pull request introduces a BREAKING CHANGE.**
> This request does not introduce any direct public-facing changes. The exception is custom request handlers authors. 

## GitHub

- Closes #489 (#535) by introducing an improved GraphQL request parsing logic. 
- Related to #290 
- Closes #579 

## Motivation

The Request Handler API can be reviewed and improved based on the usage scenarios and request handling workflow that we've learned. 

- Revisit and standardize the lifecycle of a request handler.
- Return a clear execution result upon calling `handler.run()` that includes mocked response, as well as additional request information (i.e. parsing results). 
- Simplify the creation of custom request handlers by extending a base `RequestHandler` class, or any standard request handler exported by the library (i.e. `RestHandler` or `GraphQLHandler`). 

## Example

```js
// Example of a REST API request handler that extends a base `RequestHandler` class.
const handler = new RestHandler('GET', '/user', resolver)

// Test if a given request satisfies the handler.
handler.test(req)

// Resolve the request: parses, tests, and produces a mocked response.
handler.run(req)
```

## Roadmap

- [x] Introduce a new Request Handler API.
- [x] Migrate `rest` request handlers.
- [x] Migrate `graphql` request handlers.
- [x] Clean up.
- [x] Resolve all `@fixme` comments.
- [x] Ensure that `test/msw-api/setup-worker/start/on-unhandled-request/suggestions.test.ts` doesn't error and can perform requests that intentionally fail without polluting the stdout.


kettanaito:
REST API request handlers have been successfully migrated to the new Request Handler API.

kettanaito:
GraphQL API request handlers have been successfully migrated to the new Request Handler API.

kettanaito:
@timdeschryver I'm addressing the types validation in dbd5befe1f5a0301932c5217cf465c9fe2cf61ac with the `test:ts` command. It's going to validate the usage scenarios in terms of type compliance against the built library. I've checked and it also detects type violations in `d.ts` and glossary modules (something we tried to address in #579).

cdomigan:
It looks like the docs are now out of date on creating Custom Handlers? https://mswjs.io/docs/recipes/custom-request-handler

kettanaito:
@cdomigan, looks like you're right. That recipe needs updating to be in sync with these changes. 

Please, would you be interested in adjusting that recipe? Its source is [here](https://github.com/mswjs/mswjs.io/blob/master/docs/recipes/custom-request-handler.mdx).

PR #564, opened 2021-01-24
Skips "opaque" responses when handling life-cycle events

### LINKED ISSUE #529: Failed to construct 'Response': The status provided (0) is outside the range [200, 599]
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.25.0   |
| browser | Chrome     |
| OS      | macOS     |

## Request handlers

```js
// ./mocks/handlers.js

import { graphql } from 'msw'

const handlers = [
  graphql.mutation('bookAppointment', (_req, res, ctx) => res(ctx.status(422, 'Error 422 custom'))),
]

export default handlers
```

```js
// ./mocks/browser.js

import handlers from './handlers'

const worker = setupWorker(...handlers)

export { worker }

```

```js
// ./index.js

if (process.env.NODE_ENV === 'development') {
    const { worker } = require('./mocks/browser')
    worker.start()
}
```
## Actual request

```js
import gql from 'graphql-tag'

export default gql`
  mutation bookAppointment($input: AppointmentRequestInput!) {
    bookAppointment(input: $input) {
        ...
      }
    }
  }
`

```

## Current behavior

When running the react application I'm getting the following logs:

```js
[MSW] Mocking enabled.

index.js:1034 Uncaught RangeError: Failed to construct 'Response': The status provided (0) is outside the range [200, 599].
    at index.js:1034
    at ServiceWorkerContainer.<anonymous> (index.js:1200)
```

Obviously the MSW is not intercepting said mutation...

Any ideas? I'm guessing that I'm making a silly mistake that I'm not able to see -_-

## Expected behavior

The only log I should see is the MSW getting initialise: 

```js
[MSW] Mocking enabled.
```

kettanaito:
## GitHub

- Fixes #529 

## Changes

- Response event handler now skips "opaque" responses (those originating from requests with `mode: "no-cors"`) based on their `response.type` value.
- Improves type annotation of the response object sent from the worker (it's only a partial representation of the `Response` type).

### DIFF
--- a/src/utils/worker/createResponseListener.ts
+++ b/src/utils/worker/createResponseListener.ts
@@ -21,7 +21,7 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
      * @see https://fetch.spec.whatwg.org/#concept-filtered-response-opaque
      * @see https://github.com/mswjs/msw/issues/529
      */
-    if (responseJson.type.includes('opaque')) {
+    if (responseJson.type?.includes('opaque')) {
       return
     }
 
--- a/src/utils/worker/createResponseListener.ts
+++ b/src/utils/worker/createResponseListener.ts
@@ -12,14 +12,26 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
       ServiceWorkerIncomingEventsMap['RESPONSE']
     >,
   ) => {
-    const { payload } = message
-    const response = new Response(payload.body, payload)
+    const { payload: responseJson } = message
+
+    /**
+     * CORS requests with `mode: "no-cors"` result in "opaque" responses.
+     * That kind of responses cannot be manipulated in JavaScript due
+     * to the security considerations.
+     * @see https://fetch.spec.whatwg.org/#concept-filtered-response-opaque
+     * @see https://github.com/mswjs/msw/issues/529
+     */
+    if (responseJson.type.includes('opaque')) {
+      return
+    }
+
+    const response = new Response(responseJson.body, responseJson)
     const isMockedResponse = response.headers.get('x-powered-by') === 'msw'
 
     if (isMockedResponse) {
-      context.emitter.emit('response:mocked', response, payload.requestId)
+      context.emitter.emit('response:mocked', response, responseJson.requestId)
     } else {
-      context.emitter.emit('response:bypass', response, payload.requestId)
+      context.emitter.emit('response:bypass', response, responseJson.requestId)
     }
   }
 }
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -197,6 +197,7 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
           type: 'RESPONSE',
           payload: {
             requestId,
+            type: clonedResponse.type,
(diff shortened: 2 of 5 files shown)


kettanaito:
Hey, @timdeschryver. Thank you for writing this test. If you don't mind, I've polished it but a little. I would appreciate your review if you have a spare minute. Thanks!

kettanaito:
I've replaced the image example with a local test server with a strict CORS policy. This produces the same exception without the fix, making the test fail. The main reason for doing so is to remove any external dependency (such as on an external image provider) in our tests. It should be possible to run all test suites offline, which implies no dependencies on actual servers. We can substitute actual servers with the `@open-draft/test-server` package.

kettanaito:
I believe that the Codesandbox build won't succeed with this breaking change to the Service Worker file. Without the addition of the `type` property in the response sent to the client, the following like will throw indefinitely:

```
if (responseJson.type.includes('opaque')) {
```

While this isn't nice, we are going to publish this in the next minor release, indicating a breaking change. That change will imply that you have to update your worker script by `npx msw init`.

src/utils/worker/createResponseListener.ts
```diff
@@ -12,14 +12,26 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
...
+    const { payload: responseJson } = message
+
+    /**
+     * CORS requests with `mode: "no-cors"` result in "opaque" responses.
+     * That kind of responses cannot be manipulated in JavaScript due
+     * to the security considerations.
+     * @see https://fetch.spec.whatwg.org/#concept-filtered-response-opaque
+     * @see https://github.com/mswjs/msw/issues/529
+     */
+    if (responseJson.type.includes('opaque')) {
```
marcosvega91: I could suggest adding an extra check here. Because the `type` attribute is new, some users could not have the `service worker` updated and will receive a lot of errors. 
```suggestion
    if ("type" in responseJson && responseJson.type.includes('opaque')) {
```

marcosvega91 (changes requested):
Contrary to what I have said on discord if there is a network error the service worker will not call the function `sendToClient`. I suggest only adding an extra as in the comment above, then for me is super ok :)

src/utils/worker/createResponseListener.ts
kettanaito: Yes, this is a valid concern, but as I've mentioned [in the comment above](https://github.com/mswjs/msw/pull/564#issuecomment-766327879), this is going to be released as a breaking change. So there's no need to provide a backward-compatibility hatch.

marcosvega91: okok so it's ok for me 🥳

kettanaito:
Actually, I think you're right to introduce the safety check, @marcosvega91. I've tested this once more and it produces an infinite loop with an outdated worker. I don't think that's a good experience to aim for. 

I will add a safety check that will be removed in the minor release after the next one. We'll give it 1 version to bake in.

timdeschryver (approved):
👍 Thanks for the cleanup!

marcosvega91 (approved):
Nice work 💪🏽

PR #565, opened 2021-01-24
Retrieves worker client from "ServiceWorkerGlobalScope" instead of "FetchEvent.target"

### LINKED ISSUE #558: fetchEvent.respondWith Receives an Error in Safari Only
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.25.0   |
| browser | Safari 14.0.2   |
| OS      | MacOS Catalina     |

## Request handlers

```js
// Example of declaration. Provide your code here.

const handlers = [
  graphql.query('MeQuery', (_req, res, ctx) => {
    return res(
      ctx.data({
        ...
      }),
    );
  }),
  ...
];

const worker = setupWorker(...handlers)

worker.start()
```

## Actual request

```js
// Example of making a request. Provide your code here.
export const getMe = async () => {
  try {
    const response = await fetch(MOCK_URL, {
      headers: {
      },
      method: 'POST',
      body: meQuery(),
    });
    ...
  } catch {
    ...
  }
};
```

## Current behavior

On Chrome & Firefox, everything is fine. In Safari, it seems like the request gets intercepted, as I see the mock response in the console, but then the service worker throws an error, causing the `fetch` call to catch with error
`[Error] FetchEvent.respondWith received an error: Returned response is null.` (see screenshot). Not sure why it's trying to connect to `localhost:8642`. That's the local server I run when _not_ using mocks.

## Expected behavior

The `fetch` should return the mock response and not throw an error.

## Screenshots

image

kettanaito:
## GitHub

- Fixes #558

## Motivation

https://github.com/mswjs/msw/blob/af0dada2420a31d2dc17b4d3aeae55914c9ff205/src/mockServiceWorker.js#L193

`FetchEvent.target` equals `null` in Safari only. In other browsers `FetchEvent.target` equals `self` (ServiceWorkerGlobalScope) instance. Here's a snapshot of the entire "fetch" event scope in Safari:

Screen Shot 2021-01-25 at 15 17 34

This pull request substitutes:

```diff
-event.target.clients.get(clientId)
+this.clients.get(clientId)
```

Since `clientId` comes from the `FetchEvent.clientId` and is persistent in all browsers, it matters little which ServiceWorker global scope resolves it.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -99,7 +99,7 @@ self.addEventListener('fetch', function (event) {
 
   event.respondWith(
     new Promise(async (resolve, reject) => {
-      const client = await event.target.clients.get(clientId)
+      const client = await this.clients.get(clientId)
 
       // Bypass mocking when the request client is not active.
       if (!client) {
@@ -190,7 +190,7 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
       }
     })
       .then(async (response) => {
-        const client = await event.target.clients.get(clientId)
+        const client = await this.clients.get(clientId)
         const clonedResponse = response.clone()
 
         sendToClient(client, {


src/mockServiceWorker.js
```diff
@@ -97,9 +97,11 @@ self.addEventListener('fetch', function (event) {
     return
   }
 
+  const clientPromise = event.target.clients.get(clientId)
```
marcosvega91: Because we don't have a way to test it, we can add a comment referring to the issue to remember in the future why this line has moved outside. What do you think?

kettanaito: I think that reason can be looked up in Git commit history. Now I’m wondering if we can test this somehow... the issue happened only on Safari.

marcosvega91: Yes I know and I think that we can't run puppeteer in Safari. I know that it can be found in git history but because we don't have a test for it, in case of change, it will not break anything. Of course, if we find a way to test it comment the line will be not very helpful

kettanaito: I'll try to analyze what exactly goes different in Safari and see if we can simulate that scenario in a test.

kettanaito: I was unable to create a test suite for this. `event.target` is `null` in Safari only, I don't see how I can simulate that in a test. I've tried issuing a request from ClientA, then closing that client, and observing what happens in ClientB controlled by the same worker. I'd expect that since the client was closed, the `event.target` or at least `evet.target.clients.get` will be `null`, but it's not. At least, it doesn't throw an exception.

marcosvega91: Yes, this morning I have tried to take a look at the issue but no found a way to test too. The only thing that I Have found is that the first time target is not null. It will become `null` after the first `await event.target.clients.get(clientId)`

marcosvega91 (approved):
is ready for me :)

timdeschryver (approved):
This also LGTM.
I wonder if we can do the same at `const client = await event.currentTarget.clients.get(clientId)` when we receive a message. Because this could potentially also provide a fix for #536?
Since the linked issue (#536) now includes a reproduction, I'll try again to reproduce that behavior in our test suite.

kettanaito:
@marcosvega91, @timdeschryver thank you for the review, gentlemen!

@timdeschryver that'd be a great fix indeed. I suggest we merge this and see if your suggestion could eliminate the issue. Eager to hear more about your findings.

PR #566, opened 2021-01-24
fix: getCallFrame does not need to throw

### LINKED ISSUE #562: getCallFrame does not need to throw + test in other browsers
## Describe the bug

As discussed in live stream, https://github.com/mswjs/msw/blob/5b232eb181c9b78bc0fb26954b448e5ccd26ed24/src/utils/internal/getCallFrame.ts does not need the Error to be thrown, just `new Error().stack` is enough.

---

The difference between browsers should not an issue after all, but FTR: Chrome and Node, unlike Firefox and Safari, contain the error message on the 1st line of stack (but the fist line in Firefox/Safari will be `getCallFrame` itself, so doing `frames.slice(1)` is safe). In any case, probably best to test in multiple browsers just in case.

Executing `(function a() { console.log(new Error().stack) })()` in Chrome:
```
Error
    at a (<anonymous>:1:29)
    at <anonymous>:1:50
```

in Firefox:
```
a@debugger eval code:1:28
@debugger eval code:1:1
```

Aprillion:
Fixes #562

### DIFF
--- a/src/utils/internal/getCallFrame.test.ts
+++ b/src/utils/internal/getCallFrame.test.ts
@@ -1,11 +1,18 @@
 import { getCallFrame } from './getCallFrame'
 
 const mockStack = jest.fn()
-jest.spyOn(window, 'Error').mockImplementation(() => ({
-  name: 'mockError',
-  message: '',
-  stack: mockStack(),
-}))
+
+beforeAll(() => {
+  jest.spyOn(window, 'Error').mockImplementation(() => ({
+    name: 'mockError',
+    message: '',
+    stack: mockStack(),
+  }))
+})
+
+afterAll(() => {
+  jest.restoreAllMocks()
+})
 
 test('Node on Linux and macOS error stack', () => {
   // version 1
--- a/src/utils/internal/getCallFrame.test.ts
+++ b/src/utils/internal/getCallFrame.test.ts
@@ -1,0 +1,142 @@
+import { getCallFrame } from './getCallFrame'
+
+const mockStack = jest.fn()
+jest.spyOn(window, 'Error').mockImplementation(() => ({
+  name: 'mockError',
+  message: '',
+  stack: mockStack(),
+}))
+
+test('Node error stack', () => {
+  // version 1
+  mockStack.mockImplementationOnce(() =>
+    [
+      'Error: ',
+      '    at getCallFrame (/Users/mock/github/msw/lib/umd/index.js:3735:22)',
+      '    at Object.get (/Users/mock/github/msw/lib/umd/index.js:3776:29)',
+      '    at Object.<anonymous> (/Users/mock/github/msw/test/msw-api/setup-server/printHandlers.test.ts:13:8)', // <-- this one
+      '    at Runtime._execModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:1299:24)',
+      '    at Runtime._loadModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:898:12)',
+      '    at Runtime.requireModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:746:10)',
+      '    at jasmine2 (/Users/mock/github/msw/node_modules/jest-jasmine2/build/index.js:230:13)',
+      '    at runTestInternal (/Users/mock/github/msw/node_modules/jest-runner/build/runTest.js:380:22)',
+      '    at runTest (/Users/mock/github/msw/node_modules/jest-runner/build/runTest.js:472:34)',
+    ].join('\n'),
+  )
+
+  expect(getCallFrame()).toBe(
+    '/Users/mock/github/msw/test/msw-api/setup-server/printHandlers.test.ts:13:8',
+  )
+
+  // version 2
+  mockStack.mockImplementationOnce(() =>
+    [
+      'Error: ',
+      '    at getCallFrame (/Users/mock/github/msw/lib/umd/index.js:3735:22)',
+      '    at graphQLRequestHandler (/Users/mock/github/msw/lib/umd/index.js:7071:25)',
+      '    at Object.query (/Users/mock/github/msw/lib/umd/index.js:7182:18)',
+      '    at Object.<anonymous> (/Users/mock/github/msw/test/msw-api/setup-server/printHandlers.test.ts:14:11)', // <-- this one
+      '    at Runtime._execModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:1299:24)',
+      '    at Runtime._loadModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:898:12)',
+      '    at Runtime.requireModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:746:10)',
+      '    at jasmine2 (/Users/mock/github/msw/node_modules/jest-jasmine2/build/index.js:230:13)',
+      '    at runTestInternal (/Users/mock/github/msw/node_modules/jest-runner/build/runTest.js:380:22)',
+      '    at runTest (/Users/mock/github/msw/node_modules/jest-runner/build/runTest.js:472:34)',
+    ].join('\n'),
+  )
+
+  expect(getCallFrame()).toBe(
+    '/Users/mock/github/msw/test/msw-api/setup-server/printHandlers.test.ts:14:11',
+  )
+})
+
+test('Chrome error stack', () => {
+  mockStack.mockImplementationOnce(() =>
+    [
+      'Error',
+      '    at getCallFrame (webpack:///./lib/esm/getCallFrame-deps.js?:272:20)',
+      '    at Object.eval [as get] (webpack:///./lib/esm/rest-deps.js?:1402:90)',
+      '    at eval (webpack:///./test/msw-api/setup-worker/printHandlers.mocks.ts?:6:113)', // <-- this one
+      '    at Module../test/msw-api/setup-worker/printHandlers.mocks.ts (http://localhost:59464/main.js:1376:1)',
+      '    at __webpack_require__ (http://localhost:59464/main.js:790:30)',
+      '    at fn (http://localhost:59464/main.js:101:20)',
+      '    at eval (webpack:///multi_(webpack)-dev-server/client?:4:18)',
+      '    at Object.0 (http://localhost:59464/main.js:1399:1)',
+      '    at __webpack_require__ (http://localhost:59464/main.js:790:30)',
+      '    at http://localhost:59464/main.js:857:37',
+    ].join('\n'),
+  )
+
+  expect(getCallFrame()).toBe(
+    'webpack:///./test/msw-api/setup-worker/printHandlers.mocks.ts?:6:113',
+  )
+})
+
+test('Firefox error stack', () => {
+  mockStack.mockImplementationOnce(() =>
+    [
+      'getCallFrame@webpack:///./lib/esm/getCallFrame-deps.js?:272:20',
+      'createRestHandler/<@webpack:///./lib/esm/rest-deps.js?:1402:90',
+      '@webpack:///./test/msw-api/setup-worker/printHandlers.mocks.ts?:6:113', // <-- this one
+      './test/msw-api/setup-worker/printHandlers.mocks.ts@http://localhost:59464/main.js:1376:1',
+      '__webpack_require__@http://localhost:59464/main.js:790:30',
+      'fn@http://localhost:59464/main.js:101:20',
+      '@webpack:///multi_(webpack)-dev-server/client?:4:18',
+      '0@http://localhost:59464/main.js:1399:1',
+      '__webpack_require__@http://localhost:59464/main.js:790:30',
+      '@http://localhost:59464/main.js:857:37',
+    ].join('\n'),
+  )
+
+  expect(getCallFrame()).toBe(
+    'webpack:///./test/msw-api/setup-worker/printHandlers.mocks.ts?:6:113',
+  )
+})
+
+test('Safari on macOS error stack', () => {
+  // version 1
+  mockStack.mockImplementationOnce(() =>
+    [
+      'getCallFrame',
+      '',
+      'eval code',
+      'eval@[native code]',
+      './test/msw-api/setup-worker/printHandlers.mocks.ts@http://localhost:59464/main.js:1376:5', // <-- this one
+      '__webpack_require__@http://localhost:59464/main.js:790:34',
+      'fn@http://localhost:59464/main.js:101:39',
+      'eval code',
+      'eval@[native code]',
+      'http://localhost:59464/main.js:1399:5',
+      '__webpack_require__@http://localhost:59464/main.js:790:34',
+      'http://localhost:59464/main.js:857:37',
+      'global code@http://localhost:59464/main.js:858:12',
+    ].join('\n'),
+  )
+
+  expect(getCallFrame()).toBe(
+    './test/msw-api/setup-worker/printHandlers.mocks.ts@http://localhost:59464/main.js:1376:5',
+  )
+
+  // version 2
+  mockStack.mockImplementationOnce(() =>
+    [
+      'getCallFrame',
+      'graphQLRequestHandler',
+      'eval code',
+      'eval@[native code]',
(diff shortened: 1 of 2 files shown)


src/utils/internal/getCallFrame.ts
```diff
@@ -2,25 +2,21 @@
...
 export function getCallFrame() {
-  try {
-    const inspectionError = new Error()
-    inspectionError.name = 'Inspection Error'
-    throw inspectionError
-  } catch (error) {
-    const frames: string[] = error.stack.split('\n')
+  // Error.stack is available even in IE 10, no need for nullish coalesce
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack#browser_compatibility
+  const frames: string[] = (new Error().stack as string).split('\n')
```
timdeschryver, line 5: Nit, since `stack` is a string (or can be nullable) I prefer the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) instead of a cast.

```suggestion
  const frames: string[] = new Error().stack!.split('\n')
```

timdeschryver (dismissed):
👍👍

src/utils/internal/getCallFrame.ts
Aprillion: done... 

later undone because:
a) my local git somehow did not fetch the commit made in github :(
b) `!` is forbidden by eslint in this repo

src/utils/internal/getCallFrame.ts
```diff
@@ -2,25 +2,21 @@
  * Return the stack trace frame of a function's invocation.
  */
 export function getCallFrame() {
-  try {
-    const inspectionError = new Error()
-    inspectionError.name = 'Inspection Error'
-    throw inspectionError
-  } catch (error) {
-    const frames: string[] = error.stack.split('\n')
+  // Error.stack is available even in IE 10, no need for nullish coalesce
```
kettanaito: I don't think this comment is necessary. At least, I wouldn't reference MDN link.

src/utils/internal/getCallFrame.ts
```diff
@@ -2,25 +2,21 @@
...
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack#browser_compatibility
+  const frames: string[] = new Error().stack!.split('\n')
```
kettanaito: Do we need `!` here because `Error.stack` is typed as optional in TypeScript?

src/utils/internal/getCallFrame.ts
```diff
@@ -2,25 +2,21 @@
...
-
-    // Extract file reference from the stack frame.
-    const [, declarationPath] = declarationFrame.match(/\((.+?)\)$/) || []
-    return declarationPath
+  if (!declarationFrame) {
+    return
   }
+
+  // Extract file reference from the stack frame.
+  const [, declarationPath] = declarationFrame.match(/\((.+?)\)$/) || []
```
kettanaito: Did you have a chance to examine whether Firefox stack traces match this format? 

```
  at (<FILE_PATH>)
```

kettanaito (changes requested):
Thank you for the improvement, @Aprillion! I've left a few comments, mainly questions to understand the changes.

src/utils/internal/getCallFrame.ts
Aprillion: ok, I can remove both comment lines... (probably not today, still catching up with day-job work)

src/utils/internal/getCallFrame.ts
Aprillion: yes. TS definitions for web apis are quite out of date TBH

P.S.: I don't like that `Resolve conversation` button deletes my draft reply

src/utils/internal/getCallFrame.ts
Aprillion: nope. do you have any instructions how to execute this code in browser? I can click the `here` button from CodeSandbox comment above, but what can I do to execute `getCallFrame` plz???

src/utils/internal/getCallFrame.ts
kettanaito: No pressure. I'm thankful for your contribution already!

src/utils/internal/getCallFrame.ts
kettanaito: Sure, I'll share the instructions below. 

## Executing in a browser

1. Clone the repo and install its dependencies.
2. Run a single integration test in a browser. You'd need a test that exposes the `msw` object globally. For example `yarn test:focused test/msw-api/setup-worker/printHandlers.mocks.ts`
3. Open the spawned URL in Firefox.
4. Open DevTools, switch to "Console".
5. Run `window.msw.worker.printHandlers()`.

That should print the list of handlers into the console. As a part of each message, there will be a call frame to the handler's definition.

Aprillion: Screenshot from output in Chrome (left), Safari (middle), and Firefox (right),, on macOS 10.14:

Screenshot 2021-01-29 at 19 11 22

Looks like the "Declaration" property is `undefined` => I will try to investigate.
(edit: replaced screenshot to show the problem)

src/utils/internal/getCallFrame.ts
Aprillion: now removed

src/utils/internal/getCallFrame.ts
Aprillion: looks like you guys disabled `!` in eslint => I am returning back to my original `as string` for now

Screenshot 2021-01-29 at 19 46 06

src/utils/internal/getCallFrame.ts
Aprillion: OK, I pushed a commit that will make Firefox output the same Declaration as Chrome, and Safari to output the best it can.

The unit test is necessary for TDD. As I mentioned in issue #562, I was not able to run the integration / focused test on my Windows => if someone else can add unit tests from more supported browsers or from a Deno server, feel free to tag me to help with improving the implementation.

Aprillion: arghh@!$# I need to fix `setup-server/printHandlers.test.ts` .. looks like Node has a slightly different stack than Chrome after all

edit: fixed

Aprillion:
Hmm, this will need more testing on Windows, but please review the unit test approach so far - whether it makes sense for me to continue with this approach (with my luck so far, the amount of edge cases from Windows will double the amount of unit tests).

Aprillion:
Tests on Windows are now green too. 🥳

src/utils/internal/getCallFrame.test.ts
```diff
@@ -0,0 +1,163 @@
+import { getCallFrame } from './getCallFrame'
```
kettanaito, line 1: This test looks great! 
We should also have the same test suite for NodeJS. Stack traces there may differ, it would be nice to test them.

```js
// src/utils/internal/getStackFrame.node.test.ts
/**
 * @jest-environment node
 */
```

Would you be interested in adding the NodeJS tests? If you don't have time for this I can give you a hand.

src/utils/internal/getCallFrame.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const mockStack = jest.fn()
+jest.spyOn(window, 'Error').mockImplementation(() => ({
```
kettanaito: Please put this spying in the `beforeAll` hook. Don't forget to add the `afterAll` hook that `jest.restoreAllMocks()` clears things out.

src/utils/internal/getCallFrame.test.ts
Aprillion: Hm, this is just a unit test with a mock for `new Error` ... using fake stack traces to simulate NodeJS from both macOS and Windows.

And a real native NodeJS `new Error().stack` is used in the integration test `setup-server/printHandlers.test.ts` already. IMHO no need to test mock errors in 2 different jest environments..

src/utils/internal/getCallFrame.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      '    at getCallFrame (/Users/mock/github/msw/lib/umd/index.js:3735:22)',
+      '    at Object.get (/Users/mock/github/msw/lib/umd/index.js:3776:29)',
+      '    at Object.<anonymous> (/Users/mock/github/msw/test/msw-api/setup-server/printHandlers.test.ts:13:8)', // <-- this one
+      '    at Runtime._execModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:1299:24)',
+      '    at Runtime._loadModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:898:12)',
+      '    at Runtime.requireModule (/Users/mock/github/msw/node_modules/jest-runtime/build/index.js:746:10)',
+      '    at jasmine2 (/Users/mock/github/msw/node_modules/jest-jasmine2/build/index.js:230:13)',
+      '    at runTestInternal (/Users/mock/github/msw/node_modules/jest-runner/build/runTest.js:380:22)',
+      '    at runTest (/Users/mock/github/msw/node_modules/jest-runner/build/runTest.js:472:34)',
+    ].join('\n'),
```
kettanaito, line 31: Alternatively, we could return a template string literal that preserves line breaks. Do you think that would be different from the current approach?

src/utils/internal/getCallFrame.ts
```diff
@@ -2,25 +2,22 @@
...
+  const frames: string[] = (new Error().stack as string).split('\n')
 
-    // Get the first frame that doesn't reference the library's internal trace.
-    // Assume that frame is the invocation frame.
-    const declarationFrame = frames.slice(1).find((frame) => {
-      return !/(node_modules)?\/lib\/(umd|esm)\//.test(frame)
-    })
+  // Get the first frame that doesn't reference the library's internal trace.
+  // Assume that frame is the invocation frame.
+  const rePathIsNotUseful = /(node_modules)?[\/\\]lib[\/\\](umd|esm)[\/\\]|^[^\/\\]*$/
```
kettanaito: I was thinking about utilizing `path.resolve` and its built-in system-specific path delimiters. I suppose that won't work in a browser, will it?

kettanaito: So `node_modules` part is not necessary? Interesting. Could you share more info when does this happen?

kettanaito: What do you think if we rename this variable? 
```suggestion
  const ignoreFrameRegExp= /(node_modules)?[\/\\]lib[\/\\](umd|esm)[\/\\]|^[^\/\\]*$/
```

src/utils/internal/getCallFrame.test.ts
kettanaito: Oh, I see. Since we are using an exact system-specific stack trace there's no need to re-create this test suite in a different environment. Nice catch!

src/utils/internal/getCallFrame.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      '    at runTest (/Users/mock/git/msw/node_modules/jest-runner/build/runTest.js:472:34)',
+    ].join('\n'),
+  )
+
+  expect(getCallFrame()).toBe(
+    '/Users/mock/git/msw/test/msw-api/setup-server/printHandlers.test.ts:14:11',
+  )
+})
+
+test('Node on Windows error stack', () => {
```
kettanaito, line 60: Do you mind grouping these test scenarios under their respective OS?

```js
describe('MacOS', () => {
  // ...
})

describe('Windows', () => {
  // ...
//
```

I agree that flat tests are good, but in this case grouping them would help scanning trough quicker.

kettanaito (changes requested):
I've left a few suggestions and a question. Would appreciate your help.

src/utils/internal/getCallFrame.test.ts
Aprillion: I thought that Jest spies work on the level of a single test file, but can't any documentation whether that's true or not :(

- [x] add explicit beforeAll/afterAll

src/utils/internal/getCallFrame.test.ts
Aprillion: I added `console.warn(frames)` to the function and copy&pasted the arrays from console :) Otherwise the only difference would be in indentation.

Aprillion: oh, and I couldn't do the `// <-- this one` comment .. I used it only to guide me to make the assertion, not sure how useful it is to keep those comments for future debugging 🤷‍♂️

=> not changing now, but if you tell me you really really prefer template string literal without indentation and without line comments, I can convert.

src/utils/internal/getCallFrame.ts
Aprillion: yeah, browsers use `webpack:///./test/msw-api/setup-worker/printHandlers.mocks.ts?:6:113` kind of path on all OS

Aprillion: I didn't change anything about that part, `(node_modules)?` was there already in commit 944d805

Aprillion: - [x] rename to `ignoreFrameRegExp`

src/utils/internal/getCallFrame.test.ts
Aprillion: - [ ] group tests by OS

Aprillion: hm,, what do you want me to do with `Chrome and Edge error stack` and `Firefox on macOS and Windows error stack` tests please? they have the same error stack on both

src/utils/internal/getCallFrame.ts
Aprillion: But thinking about it, it probably wouldn't make any difference if we remove that part from the regex, `/(node_modules)?[\/\\]lib[\/\\](umd|esm)[\/\\]|^[^\/\\]*$/` wouldn't match anything more than `/[\/\\]lib[\/\\](umd|esm)[\/\\]|^[^\/\\]*$/` would... I think 🤔

Aprillion: Double but, what if someone will use `/lib/esm/msw.js` file structure to set up the handlers in their repo?  While we wouldn't want to ignore that and in their case `msw` library would be inside node_modules indeed .. but I cannot think of a simple way how to make `getCallFrame` testable inside this repo as well as handle such a special edge case when msw is used from node_modules.

One idea was to detect `msw/lib` in the path to skip those frames, but there are both false positives (someone using "msw" as a directory name to set up msw handlers) and false negatives (someone cloning this repo into a folder different from "msw").

Unless someone has a better idea for more reliable detection of "internal msw" vs "user space", IMHO let's ignore that for now 😰

src/utils/internal/getCallFrame.ts
```diff
@@ -2,25 +2,22 @@
...
+  const frames: string[] = (new Error().stack as string).split('\n')
 
-    // Get the first frame that doesn't reference the library's internal trace.
-    // Assume that frame is the invocation frame.
-    const declarationFrame = frames.slice(1).find((frame) => {
-      return !/(node_modules)?\/lib\/(umd|esm)\//.test(frame)
-    })
+  // Get the first frame that doesn't reference the library's internal trace.
+  // Assume that frame is the invocation frame.
+  const ignoreFrameRegExp = /(node_modules)?[\/\\]lib[\/\\](umd|esm)[\/\\]|^[^\/\\]*$/
```
timdeschryver: I might be wrong here, but shouldn't we also at the `iife` build here?
```suggestion
  const ignoreFrameRegExp = /(node_modules)?[\/\\]lib[\/\\](umd|esm|iife)[\/\\]|^[^\/\\]*$/
```

src/utils/internal/getCallFrame.test.ts
kettanaito: Oh, a good point. I haven't noticed some of the test suites shared the stack trace across OS. I don't mind, let's keep it as-is.

src/utils/internal/getCallFrame.ts
kettanaito: @timdeschryver I think you're totally right. Ideally, we should include all the build targets (umd, esm, cjs, iife) so all source file references can be accounted for.

Aprillion: - [x] add `|iife|cjs` to regex

kettanaito (approved):
Great job, @Aprillion! Welcome to contributors!

PR #568, opened 2021-01-26
mockServiceWorker: Uses "self" instead of "this"

kettanaito:
Per @timdeschryver suggestion, it's better if we consistently used `self` in the worker file instead of context-dependant `this`.

- `self` always refers to a global Service Worker scope.
- `this` is context-dependant and may change have we re-assigned a parent function's context, for example.


timdeschryver (approved):
LGTM 👍

PR #570, opened 2021-01-28
Uses explicit null as 204 response body when handling the "response" worker event

kettanaito:
## Motivation

When handling a 204 status (actual) response, it's expected for it to have no body. However, there is no way to check if a response has no body given an arbitrary potentially 204 response:

```js
await potentially204Response.text() // "" (same as a response with the explicit "" response text)
await potentially204Response.body // ReadableStream, locked and cannot be read
```

Looks like we need a safe-guard on our side to prevent the construction of a `Response` instance with an empty string body in case of a 204 status response:

```js
new Response('', { status: 204 }) // Error
new Response(null, { status: 204 }) // OK
```

## Changes

- Uses `null` in case the response status is `204`, regardless of what its actual body is (in that case it's always `null`).
- Adds an integration test that asserts 204 status responses to be handled without page errors through the entire request/response life-cycle.

### DIFF
--- a/src/utils/worker/createResponseListener.ts
+++ b/src/utils/worker/createResponseListener.ts
@@ -25,7 +25,7 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
       return
     }
 
-    const response = new Response(responseJson.body, responseJson)
+    const response = new Response(responseJson.body || null, responseJson)
     const isMockedResponse = response.headers.get('x-powered-by') === 'msw'
 
     if (isMockedResponse) {
--- a/test/rest-api/204-response.test.ts
+++ b/test/rest-api/204-response.test.ts
@@ -1,0 +1,39 @@
+import * as path from 'path'
+import { ServerApi, createServer } from '@open-draft/test-server'
+import { runBrowserWith } from '../support/runBrowserWith'
+
+let server: ServerApi
+
+beforeAll(async () => {
+  server = await createServer((app) => {
+    app.get('/posts', (req, res) => {
+      return res.status(204).end()
+    })
+  })
+})
+
+afterAll(async () => {
+  await server.close()
+})
+
+test('handles a 204 status response without Response instance exceptions', async () => {
+  const runtime = await runBrowserWith(
+    path.resolve(__dirname, 'basic.mocks.ts'),
+  )
+  let pageError: Error
+
+  runtime.page.on('pageerror', (error) => {
+    pageError = error
+  })
+
+  const res = await runtime.request({
+    url: server.http.makeUrl('/posts'),
+  })
+
+  // There must be no such exception:
+  // Failed to construct 'Response': Response with null body status cannot have body
+  expect(pageError).toBeUndefined()
+  expect(res.status()).toBe(204)
+
+  return runtime.cleanup()
+})


kettanaito:
Smoke tests will fail until https://github.com/mswjs/examples/pull/46 is merged.

src/utils/worker/createResponseListener.ts
```diff
@@ -25,7 +25,8 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
       return
     }
 
-    const response = new Response(responseJson.body, responseJson)
+    const responseBody = responseJson.status === 204 ? null : responseJson.body
```
Aprillion: I found https://fetch.spec.whatwg.org/#statuses => IMHO let's check all null body statuses here:

```suggestion
    const responseBody = [101, 204, 205, 304].includes(responseJson.status) ? null : responseJson.body
```

if this is hot-path for perf optimization, maybe a `new Set(...)` in some util file can be used 😂

kettanaito: That's a great suggestion, but also a great example of why reflecting a spec this way is a huge redundancy. I wish to find a different way to retrieve a 204 response body as `null`.

Aprillion: yeah, no ideas about the clean way yet :/

Aprillion: well.. I do have an idea - `responseJson.body === '' ? null : responseJson.body` - but totally not sure how that would break all edge cases with empty body in `200` and other responses 😬

marcosvega91: You know why here https://github.com/mswjs/msw/commit/40920214c1c22a06d640cc3a1006528e2ba6e679  is working ?

kettanaito: What about the responses that have an empty string body? There must be a differentiation between them as those responses that have no body (`null`). With the logic above, both cases will result in `res.body` being `null`.

Aprillion: yeah, this distinction between null and empty string is really strange, no idea what was the reason to make it different in the spec :(

comatory: Can I  use this suggested change as a stop-gap for now? I'm using only `msw` for local development and it'd be nice to not have that warning in console (I'm using `patch-package` for these purposes).

kettanaito: @comatory, I'd rather recommend you edit the `mockServiceWorker.js` with a similar change instead. Changing what lies in `node_modules` isn't a reliable way to fix anything, as the modules will be pruned once you re-install the dependencies. 

If you're using some tool to ensure the persistent change in a node module across installs, then you can add this suggestion into `msw` directly.

comatory: @kettanaito I see. However with [patch-package](https://www.npmjs.com/package/patch-package) the changes are re-applied after installation so it's safe (and actually recommended way) to do it. I'm fine with this solution, once this problem is fixed, I'll just remove the patch from my codebase.

kettanaito:
After looking into the ways to differentiate between an empty string response and a null response, I've tried to read the `res.body` stream using a custom reader:

```js
function readStream(stream) {
    let result = null

    if (stream.locked) {
      return resolve(result)
    }

    const reader = stream.getReader()
    reader
      .read()
      .then(function processText({ value, done }) {
        if (done) {
          reader.releaseLock()
          return resolve(result)
        }

        const resolvedValue =
          typeof value === 'string'
            ? value
            : new TextDecoder('utf-8').decode(value)

        result = (result || '') + resolvedValue
        reader.read().then(processText)
      })
      .catch(reject)
  })
}
```

And then reading the stream:

```js
const responseBody = await readStream(clonedResponse.body)
```

While working with `ReadableStream`, I've observed that for a readable stream there is no difference whether you stream an empty string or null—stream would be marked as "done" immediately (the `result` variable won't be `""`). This got me thinking that it's possible to assume that an empty string response is a null body response, in terms of the body it gets.

We can fallback to `null` in case the response body is an empty string:

```js
new Response(responseJson.body || null, responseJson)
```

By doing so:

- `res.text()` will still produce `""` for both empty string response and a null body response.
- `res.body` will be `null` as opposed to `ReadableStream` if it's an empty string response. That's the only deviation there is.

kettanaito:
I've settled on the `body || null` fallback when constructing a `Response` instance to pass to the response life-cycle events. That produces a response instance compatible with an actual instance. The only difference is that instead of a `ReadableStream` that streams nothing it sets `res.body` to `null`. Imho, that's a more reproducible behavior:

```js
const res = new Response(null, { status: 204 })
res.body // null, as expected
```

kettanaito:
@Aprillion, @marcosvega91, may I please ask for your review on this? GitHub's "re-request review" is broken for me. Thanks.

marcosvega91 (approved):
Your observation is right, I think is great 💪  Great!!

timdeschryver (approved):

Aprillion (approved):

Aprillion:
> I've settled on the `body || null` fallback when constructing a `Response` instance

I just checked `await new Response(null, {status: 200}).text()` is an empty string and not an exception, so it should be fine 👍

src/utils/worker/createResponseListener.ts
```diff
@@ -25,7 +25,7 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
       return
     }
 
-    const response = new Response(responseJson.body, responseJson)
+    const response = new Response(responseJson.body || null, responseJson)
```
kettanaito, line 28: This is rather brave, but when handling a `Response` instance there is no difference between working with a `new Response(null)` and `new Response('')` in terms of the operable response body (i.e. `.json`/`.text`/etc.).

PR #577, opened 2021-02-02
fix renamed type (OnUnhandledRequest -> UnhandledRequestStrategy)

robbtraister:
This type seems to have been renamed in the most recent release.

PR #579, opened 2021-02-03
build: validate types

timdeschryver:
From #577 (and #577 should be merged first to complete this PR)

It seems that the rollup plugin doesn't have an option to validate the types.
Some even created their [own plugin](https://github.com/ReactTraining/react-router/blob/dev/scripts/rollup/tsc-plugin.js) to make this work, which seemed like an overkill to me and something we don't want to maintain.

I started looking at `dtslint` and `tsd` to validate the types **after** the build but this had some drawbacks.
`dtslint`  expects a tsconfig at the root of the types.
For `tsd` we needed to copy-paste most of the `compilerOptions` `tsconfig` into our `package.json` and make some tweaks to the libs, and `tsd` also included the `node_modules` for some reason and there were also invalid types there...

Then I stumbled upon the `--noemit` option of `tsc` which does what it needs to do and is simple to implement and doesn't require an additional package. In this PR, we run `tsc` before the rollup build (without generating the output files).

### DIFF
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -118,10 +118,10 @@ export interface SetupWorkerInternalContext {
      * Adds an event listener on the given target.
      * Returns a clean-up function that removes that listener.
      */
-    addListener<E extends Event>(
+    addListener(
       target: EventTarget,
       eventType: string,
-      listener: (event: E) => void,
+      listener: ((event: MessageEvent) => void) | EventListener,
     ): () => void
     /**
      * Removes all currently attached listeners.
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
     "resolveJsonModule": true,
     "declaration": true,
     "declarationDir": "lib/types",
+    "skipLibCheck": true,
     "lib": ["es2017", "ESNext.AsyncIterable", "dom", "webworker"]
   },
   "include": ["global.d.ts", "src/**/*.ts"],
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,8 +9,7 @@
     "resolveJsonModule": true,
     "declaration": true,
     "declarationDir": "lib/types",
-    "skipLibCheck": true,
-    "lib": ["es2017", "ESNext.AsyncIterable", "dom", "webworker"]
+    "lib": ["es2017", "ESNext.AsyncIterable", "dom"]
   },
   "include": ["global.d.ts", "src/**/*.ts"],
   "exclude": ["node_modules", "**/*.spec.ts", "**/*.test.ts"]
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "start": "cross-env NODE_ENV=development rollup -c rollup.config.ts -w",
     "clean": "rimraf lib node/**/*.js",
     "lint": "eslint \"{cli,config,src,test}/**/*.ts\"",
-    "build": "yarn clean && cross-env NODE_ENV=production rollup -c rollup.config.ts",
+    "build": "yarn clean && tsc --noemit && cross-env NODE_ENV=production rollup -c rollup.config.ts",
     "test": "yarn test:unit && yarn test:integration",
     "test:unit": "cross-env BABEL_ENV=test jest --runInBand",
     "test:integration": "node --max_old_space_size=8000 node_modules/jest/bin/jest.js --config=test/jest.config.js --runInBand",
--- a/src/setupWorker/setupWorker.ts
+++ b/src/setupWorker/setupWorker.ts
@@ -76,13 +76,13 @@ export function setupWorker(
       addListener(
         target: EventTarget,
         eventType: string,
-        callback: EventListener,
+        listener: EventListener,
       ) {
-        target.addEventListener(eventType, callback)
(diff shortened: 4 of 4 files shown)


tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "resolveJsonModule": true,
     "declaration": true,
     "declarationDir": "lib/types",
+    "skipLibCheck": true,
```
timdeschryver: Skip checking `node_modules`, there's a problem with it because we include `dom` and `webworker`, both have identical types that clash. For example:

```
node_modules/typescript/lib/lib.webworker.d.ts:5951:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'onrejectionhandled' must be of type '((this: Window, ev: PromiseRejectionEvent) => any) | null', but here has type '((this: DedicatedWorkerGlobalScope, ev: PromiseRejectionEvent) => any) | null'.

5951 declare var onrejectionhandled: ((this: DedicatedWorkerGlobalScope, ev: PromiseRejectionEvent) => any) | null;
                 ~~~~~~~~~~~~~~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:19933:13
    19933 declare var onrejectionhandled: ((this: Window, ev: PromiseRejectionEvent) => any) | null;
                      ~~~~~~~~~~~~~~~~~~
    'onrejectionhandled' was also declared here.
```

kettanaito: Doesn't `skipLibCheck: true` negates all type checking originating from third-party libraries?

Can you please verify that we still get type violations if we use some third-party package against its defined types?

timdeschryver: You're right @kettanaito, third party type violations will not be caught. 
[See docs](https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#new---skiplibcheck)

If we want to keep this enabled, I think the only solution would be to create multiple `tsconfig`'s.
One for each build, so they can have their own included libs.
This way they won't interfere with each other.

Or perhaps we could exclude typescript's definition files.

kettanaito: TS configs per each build target sound like a sensible idea. We won't have to compose TS configs in `rollup.config.ts`. I think it's still possible to provide `tsc` with a custom config like:

```bash
$ tsc --config <CONFIG_PATH>
```

timdeschryver: Adding those declaration files to our excludes doesn't help.
How do you want to proceed with this? Should we try a setup with multiple `tsconfig` files? 
EDIT: I didn't notice your comment 😅

kettanaito: Let's try multiple `tsconfig.json` configurations. If you feel it starts to get messy and takes too much time, we can always discuss this and find a different solution. Thank you for your investigation on this, by the way!

timdeschryver: I'm confused.

I thought it wouldn't be able to add multiple `tsconfig`s because they are both (dom and webworker) included in our bundle (both are exported via the index file). Now the weird part is that I was able to exclude the webworker types after I modified some types in 2b59cd94f5894f2389588408a28534a5dd8cc51e. 

After this, everything worked, but I don't exactly know why...

src/setupWorker/glossary.ts
```diff
@@ -118,10 +118,10 @@ export interface SetupWorkerInternalContext {
      * Adds an event listener on the given target.
      * Returns a clean-up function that removes that listener.
      */
-    addListener<E extends Event>(
+    addListener(
       target: EventTarget,
       eventType: string,
-      listener: (event: E) => void,
+      listener: ((event: MessageEvent) => void) | EventListener,
```
kettanaito, line 124: The `listener` doesn't have to accept `MessageEvent` only. This is a general shorthand function to create/remove event listeners on any target:

```js
context.events.addListener(window, 'beforeunload')
```

I'd expect this call signature to throw type violations.

tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "resolveJsonModule": true,
     "declaration": true,
     "declarationDir": "lib/types",
-    "lib": ["es2017", "ESNext.AsyncIterable", "dom", "webworker"]
+    "lib": ["es2017", "ESNext.AsyncIterable", "dom"]
```
kettanaito, line 12: So removing the `webworker` lib `tsc --no-emit` now validates the types? 

I wonder what types we lose by removing that lib. I'd suspect it has something to do with web workers 😛

timdeschryver: > So removing the webworker lib tsc --no-emit now validates the types

Correct. That command was also validating our types before this change, but it was throwing errors due to duplicate types in `dom` and `webworker`. See https://github.com/microsoft/TypeScript/issues/20595 for more info.

In that issue, someone mentioned to add the types of `webworker` manually.  I tried to do that, but the issue was already resolved by removing the `webworker` lib.

My guess is that we don't use any of the `webworker` specific types, although I'm not 100% sure of that.
Do you think this is better than using the `skipLibCheck` option?

kettanaito:
We should be able to validate the typings via the `yarn test:ts` command. It compiles the usage examples (in `test/typings`) and the library's output with it.

timdeschryver:
Thanks @kettanaito 👨‍🔬🧪

kettanaito:
Thank you for the initial effort behind this, @timdeschryver. Honored to have you on the team!

PR #583, opened 2021-02-06
Skip one-time request handlers inside the response reducer so it handles a scenario with parallel requests

fnmunhoz:
I came across an issue where a `one-time` request handler is used more than once.

After some investigation, I created an automated test where it fails (actually, it failed only with `setupServer` and worked fine with `setupWorker`).

Apparently, the problem is related to multiple requests running in parallel (i.e., not awaited).

The test passes by moving the request handler "skip process"  to inside the response reducer function.

I've also tried the MSW build with the change on the project where I initially identified the problem, and it also worked.

Please review and see if it's a viable solution or if you have other suggestions!

Ps: I've read the contribution guidelines and hope to have followed them, but I'm happy to update the PR if I missed something.

### DIFF
--- a/src/utils/getResponse.ts
+++ b/src/utils/getResponse.ts
@@ -85,7 +85,8 @@ export const getResponse = async <
 
       const mockedResponse = await resolver(publicRequest, response, context)
 
-      if (!mockedResponse) {
+      // Check for the handler relevancy in case of multiple parallel one-time handlers.
+      if (!mockedResponse || requestHandler.shouldSkip) {
         return acc
       }
 
--- a/test/msw-api/setup-server/use.test.ts
+++ b/test/msw-api/setup-server/use.test.ts
@@ -72,3 +72,25 @@ test('returns a mocked response from a one-time request handler override only up
   expect(anotherBookResponse.status).toBe(200)
   expect(anotherBookBody).toEqual({ title: 'Original title' })
 })
+
+test('returns a mocked response from a one-time request handler override only upon first request match with parallel requests', async () => {
+  server.use(
+    rest.get('https://test.mswjs.io/book/:bookId', (req, res, ctx) => {
+      const { bookId } = req.params
+      return res.once(ctx.json({ title: 'One-time override', bookId }))
+    }),
+  )
+
+  const bookRequestPromise = fetch('https://test.mswjs.io/book/abc-123')
+  const anotherBookRequestPromise = fetch('https://test.mswjs.io/book/abc-123')
+
+  const bookResponse = await bookRequestPromise
+  const bookBody = await bookResponse.json()
+  expect(bookResponse.status).toBe(200)
+  expect(bookBody).toEqual({ title: 'One-time override', bookId: 'abc-123' })
+
+  const anotherBookResponse = await anotherBookRequestPromise
+  const anotherBookBody = await anotherBookResponse.json()
+  expect(anotherBookResponse.status).toBe(200)
+  expect(anotherBookBody).toEqual({ title: 'Original title' })
+})
--- a/test/msw-api/setup-worker/use.test.ts
+++ b/test/msw-api/setup-worker/use.test.ts
@@ -108,3 +108,40 @@ test('returns a mocked response from a one-time request handler override only up
 
   await runtime.cleanup()
 })
+
+test('returns a mocked response from a one-time request handler override only upon first request match with parallel requests', async () => {
+  const runtime = await runBrowserWith(path.resolve(__dirname, 'use.mocks.ts'))
+
+  await runtime.page.evaluate(() => {
+    const { msw } = window
+
+    msw.worker.use(
+      msw.rest.get('/book/:bookId', function oneTimeOverride(req, res, ctx) {
+        const { bookId } = req.params
+        return res.once(ctx.json({ title: 'One-time override', bookId }))
+      }),
+    )
+  })
+
+  const bookPromise = runtime.request({
+    url: runtime.makeUrl('/book/abc-123'),
+  })
+
+  const anotherBookPromise = runtime.request({
+    url: runtime.makeUrl('/book/abc-123'),
+  })
+
+  const bookResponse = await bookPromise
+  const bookStatus = bookResponse.status()
+  const bookBody = await bookResponse.json()
+  expect(bookStatus).toBe(200)
+  expect(bookBody).toEqual({ title: 'One-time override', bookId: 'abc-123' })
+
+  const anotherBookResponse = await anotherBookPromise
+  const anotherBookStatus = anotherBookResponse.status()
+  const anotherBookBody = await anotherBookResponse.json()
+  expect(anotherBookStatus).toBe(200)
+  expect(anotherBookBody).toEqual({ title: 'Original title' })
+
+  await runtime.cleanup()
+})


src/utils/getResponse.ts
```diff
@@ -31,10 +31,6 @@ export const getResponse = async <
   handlers: H,
 ): Promise<ResponsePayload> => {
   const relevantHandlers = handlers
-    .filter((requestHandler) => {
```
timdeschryver: I think we can keep this here as a fast return?

src/utils/getResponse.ts
```diff
@@ -90,6 +86,11 @@ export const getResponse = async <
       }
 
       if (mockedResponse && mockedResponse.once) {
+        // Skip a handler if it has been already used for a one-time response.
+        if (requestHandler.shouldSkip) {
```
timdeschryver: What about moving this to line 84?
Reasoning,  it would also catch other skip conditions in the future.

```ts
if (!mockedResponse || requestHandler.shouldSkip) {
        return acc
}
```

test/msw-api/setup-server/use.test.ts
```diff
@@ -72,3 +72,24 @@ test('returns a mocked response from a one-time request handler override only up
   expect(anotherBookResponse.status).toBe(200)
   expect(anotherBookBody).toEqual({ title: 'Original title' })
 })
+
+test('returns a mocked response from a one-time request handler override only upon first request match with parallel requests', async () => {
+  server.use(
+    rest.get('https://test.mswjs.io/book/:bookId', (req, res, ctx) => {
+      return res.once(ctx.json({ title: 'One-time override' }))
```
timdeschryver: What if we add the `bookId` in the title, if we do this it's clear which request is caught here?

test/msw-api/setup-worker/use.test.ts
```diff
@@ -108,3 +109,45 @@ test('returns a mocked response from a one-time request handler override only up
...
+
+test('returns a mocked response from a one-time request handler override only upon first request match with parallel requests', async () => {
+  const runtime = await runBrowserWith(path.resolve(__dirname, 'use.mocks.ts'))
+
+  await runtime.page.evaluate(() => {
+    const { msw } = window
+
+    msw.worker.use(
+      msw.rest.get('/book/:bookId', function oneTimeOverride(req, res, ctx) {
+        return res.once(ctx.json({ title: 'One-time override' }))
```
timdeschryver: Same here

timdeschryver (commented):
Good catch! I added some comments

src/utils/getResponse.ts
fnmunhoz: Sounds good, will do.

src/utils/getResponse.ts
fnmunhoz: Agreed.

src/utils/getResponse.ts
```diff
@@ -85,7 +85,9 @@ export const getResponse = async <
 
       const mockedResponse = await resolver(publicRequest, response, context)
 
-      if (!mockedResponse) {
+      // Skip a handler if there is no response or if it has been already used for
```
fnmunhoz: @timdeschryver added a comment here similar to L34, let me know your thoughts, please.

test/msw-api/setup-worker/use.test.ts
```diff
@@ -108,3 +109,45 @@ test('returns a mocked response from a one-time request handler override only up
...
+      }),
+    )
+  })
```
fnmunhoz: @timdeschryver not sure if I got the suggestion, would it be that:

```suggestion
  msw.worker.use(
    msw.rest.get('/book/:bookId', function oneTimeOverride(req, res, ctx) {
      const { bookId } = req.params
      return res.once(
        ctx.json({ title: `One-time override; bookId = '${bookId}'` }),
      )
    }),
  )
```

And then an assertion like that:

```ts
expect(bookBody).toEqual({ title: `One-time override; bookId = 'abc-123'` })
```

test/msw-api/setup-server/use.test.ts
fnmunhoz: I added a question about that, please have a look.

test/msw-api/setup-worker/use.test.ts
```diff
@@ -108,3 +109,45 @@ test('returns a mocked response from a one-time request handler override only up
...
+  const bookStatus = bookResponse.status()
+  const bookBody = await bookResponse.json()
+  expect(bookStatus).toBe(200)
+  expect(bookBody).toEqual({ title: 'One-time override' })
+
+  const anotherBookResponse = await anotherBookPromise
+  const anotherBookStatus = anotherBookResponse.status()
+  const anotherBookBody = await anotherBookResponse.json()
+  expect(anotherBookStatus).toBe(200)
+  expect(anotherBookBody).toEqual({ title: 'Original title' })
```
fnmunhoz, line 144: One thing that was tricky to create the test for the `setupWorker` case was that the test would get the wrong title here (`One-time override`) instead of the expected one (`Original title`). 

The reason for that was because `anotherBookPromise` was being set to the same promise as `bookPromise` since they had the same URL and were using the default `responsePredicate` inside `runtime.request()`.

So to overcome this, I added a query param to the URL with a different value for each request.

Then I passed a custom `responsePredicate` that matches the full URL, including the query params, which is different than the default one, which removes any query params before checking for matches.

I'm not sure if there is a better way to do that. Suggestions?

fnmunhoz: Considering that I was wondering if would be clearer if the URL was set to something like that:

```ts
runtime.makeUrl('/book/abc-123?promiseId=anotherBook')
```

```ts
runtime.makeUrl('/book/abc-123?requestId=anotherBook')
```

It would not make a difference for the test (the test only needs a different param for each request), I'm just thinking if it would make the test more readable.

src/utils/getResponse.ts
```diff
@@ -85,7 +85,9 @@ export const getResponse = async <
...
+      // a one-time response.
+      if (!mockedResponse || requestHandler.shouldSkip) {
```
kettanaito, line 89: So because there can be multiple one-time responses running in parallel, we can't guarantee the request handler's relevancy in the preceding `.filter` step, right? Thus an extra check here.

test/msw-api/setup-worker/use.test.ts
kettanaito: You can just add a `bookId` property on the response JSON body directly:

```js
return res.once(ctx.json({ title: 'One-time override', bookId }))
```

No need to put all that data into a single string. You can later ensure that the right one-time handler has resolved by adding the `bookId` to the expected object:

```js
fetch('/book/abc-123')
fetch('/book/def-456')

// Checking the `bookId` value so we're sure which response we got. 
expect(bookBody).toEqual({ title: 'One-time override', bookId: 'abc-123' })
```

test/msw-api/setup-worker/use.test.ts
kettanaito: That's a great concern about the internal response awaiting utility we use for testing. There can be cases where multiple requests are made to the same URL, and they should resolve properly. 

Adding a query parameter looks like a workaround rather than a solution. Since the issue lies in how we map request/response in tests, I suggest we fix it there. While adding distinct information like different query parameter will work, I think it creates a confusing test because requests now contain data they don't need to showcase a usage example. That data is used solely to compensate for the internal testing setup.

kettanaito: I will provide the fix for this issue so you don't have to worry about the right response being awaited.

kettanaito: This is addressed in https://github.com/mswjs/msw/pull/584.

src/utils/getResponse.ts
timdeschryver: Yea, correct. At the time when the `filter` is invoked the two requests are still pending, thus the `filter` doesn't know it has to be skipped.
At least that's how I understand it. 

It does bring the following question/problem tho, how do we decide which response handler handles the first request? The one that receives the request first, or the one that returns the response first. 
With this implementation it's the handler that returns first, which is fine for me.

test/msw-api/setup-server/use.test.ts
timdeschryver: I think this got answered, right?

src/utils/getResponse.ts
timdeschryver: Looks good to me! 👍

src/utils/getResponse.ts
kettanaito: It would be nice to include this short motivation note above the logic. Things like these usually take some playing around to figure out.

kettanaito (changes requested):
Hey, @fnmunhoz. Thank you for an amazing contribution!

Everything looks good. There's one comment left regarding the `bookId` assertions in the responses. Also, please, rebase your branch against the latest `master` to get the response awaiting fix so there would be no need to include query parameters in 2 requests to the same URL. Those will be awaited properly now by default.

test/msw-api/setup-server/use.test.ts
fnmunhoz: Indeed! I'll give it a go, thanks @timdeschryver

test/msw-api/setup-worker/use.test.ts
fnmunhoz: Got it, will give it a try!

fnmunhoz: Ok, so I've added the `bookId` to the response body as suggested. It looks good!

I've got two related questions:

1. Should I do the same for the one-time handler of the other existing tests (the ones with sequential execution) so the tests are more consistent? Or it's out of the scope of this PR?
2. Should I add the `bookId` param to the default handler's body as well? Or it's better to keep it only in the one-time handler, so they have two differences between each other (the `title` and the `bookId` param)?

src/utils/getResponse.ts
fnmunhoz: How about that? Too verbose? Or it's ok?

```ts
  const relevantHandlers = handlers
    .filter((requestHandler) => {
      // Skip a handler if it has been already used for a one-time response.
      // Because there can be multiple one-time responses running in parallel,
      // we can't guarantee the request handler's relevancy here.
      // Thus, an extra check in succeeding `relevantHandlers` reducer.
      // The filter is kept here as well, so the handler is skipped sooner
      // for sequential requests.
      return !requestHandler.shouldSkip
    })
```

```ts
      // Skip a handler if one of the following is true:
      // * There is no response
      // * It has been already used for a one-time response. Because there can be multiple
      //   one-time responses running in parallel, we can't guarantee the request handler's
      //   relevancy in the preceding `.filter` step. Thus an extra check here.
      if (!mockedResponse || requestHandler.shouldSkip) {
        return acc
      }
```

kettanaito: I like that, but it's too verbose. We can leave the `filter` comment out and only comment on the part we've added. 

```js
// Check for the handler relevancy
// in case of multiple parallel one-time handlers.
```

It doesn't need to be an in-detail explanation, as we have Git history to help us. A quick comment is still useful, so the logic won't get mistaken by a repetition, and to stress that it's intentional.

test/msw-api/setup-worker/use.test.ts
kettanaito: Thank you for adding that!

1. No, I think there's no need. Unless those tests are related to the issue we're fixing, we'd better leave them as-is.
2. Let it be only in the one-time handler. The default handler won't have the `bookId`, which would fail the assertion in case of an issue. That's what we want. At this point there is no difference if the assertion fails because the key is missing, or because it has a wrong value.

fnmunhoz: Sounds good, I've updated it based on the above, please have a look!

src/utils/getResponse.ts
fnmunhoz: Cool, I've updated it, please have another look!

kettanaito (approved):
The changes look great! Thank you for working on this, @fnmunhoz! It's an exemplary contribution and we're happy to receive it from you. You're awesome!

fnmunhoz:
Yay! Thanks for all your help in the review @kettanaito and @timdeschryver  🚀 

I learned a lot about MSW, and I hope I can contribute more soon!

I appreciate the fantastic work you have been doing for this project. It makes my automated tests at work way better. 🤓

kettanaito:
We will be happy to collaborate more with you, @fnmunhoz! Feel free to tackle any issues you find interesting.

PR #584, opened 2021-02-07
Awaits responses in tests by request UUID

kettanaito:
## GitHub

- Originates from the discussion on #583 

## Changes

- The default response predicate used for `page.waitForResponse` is now based on the UUID generated for each individual request. No need to match URL, query parameters, and such. Guarantees reliable request/response pairing when using `runtime.request()` utility. 
- Stores the request UUID in the `Accept-Language` header. Seems random, but that's one of the few headers that don't get stripped during `no-cors` requests, as well as don't crash requests to actual API (i.e. GitHub API). 

Tests to check if the header name works:

- `rest-api/basic.test.ts` (general request)
- `rest-api/cors.test.ts` (headers that are stripped in CORS requests)
- `rest-api/response-patching.test.ts`

### DIFF
--- a/test/support/runBrowserWith.ts
+++ b/test/support/runBrowserWith.ts
@@ -1,9 +1,8 @@
 import * as puppeteer from 'puppeteer'
 import { headersToObject } from 'headers-utils'
-import { match } from 'node-match-path'
-import { getCleanUrl } from 'node-request-interceptor'
 import WebpackDevServer from 'webpack-dev-server'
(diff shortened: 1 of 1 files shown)


timdeschryver (approved):
👍

fnmunhoz:
Thanks for that @kettanaito will make the tests in #583 much better!

kettanaito:
@fnmunhoz Thank you for discovering that issue with the setup! You rock.

PR #585, opened 2021-02-07
fix: abort pending requests

### LINKED ISSUE #506: Vue.js throws warning when I run Jest test
## Describe the bug
Out-of-the-box Vue CLI based project. Added Jest and mswjs.
I tried 3 versions of node-based fetchers (axios, node-fetch, whatwf-fetch).
whatwf-fetch works fine, but tests based on axios and node-fetch throws a warning/error in the console.
```
[Vue warn]: Error in nextTick: "TypeError: Cannot read property 'createElement' of null"
```

Browser works without any issue. Only server implementation has this problem.

## Environment

- `msw: 0.24.2`
- `nodejs: 12.19.0`
- `npm: 6.14.8`
- `Chrome: 87.0.4280.88 (Official Build) (64-bit)`

## To Reproduce
1. Clone https://github.com/sumerokr/vue-mswjs
2. Checkout branch `axios` or `node-fetch`
3. Run `npm run test:unit`

## Expected behavior
No warning/error in the console after the tests

P. S.
Try to uncomment one more test in spec file. Both works!
Try to remove `v-if` condition in Vue template. Works!
Try to avoid mswjs based endpoint and use the real one. Works!

I can't figure out, what is wrong here. But this error makes it uncomfortable to use mswjs in Vue projects.

### LINKED ISSUE #580: Browser refresh and hot-reloads cause error "TypeError: Cannot read property 'postMessage' of undefined"
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.26.1   |
| browser | Chrome 87    |
| OS      | Win10     |

## Request handlers

[Handler from MSW examples](https://github.com/mswjs/examples/blob/master/examples/rest-react/src/mocks/handlers.js) with additional delay `ctx.delay(1000)`.

```js
import { rest } from "msw";

export const handlers = [
    rest.post("/login", (req, res, ctx) => {
        const { username } = req.body;

        return res(
            ctx.json({
                id: "f79e82e8-c34a-4dc7-a49e-9fadc0979fda",
                username,
                firstName: "John",
                lastName: "Maverick",
            }),
            ctx.delay(1000), // Increase if unable to reproduce the issue
        );
    }),
];
```

## Actual request

https://github.com/mswjs/examples/blob/83603a4619abb09447322d88bb88a7e4d6a2c91c/examples/rest-react/src/LoginForm.js#L18-L36

```js
fetch("/login", {
    method: "POST",
    body: JSON.stringify({ username })
});
```

## Current behavior

Seen in browser console. 

```
mockServiceWorker.js:256 Uncaught (in promise) TypeError: Cannot read property 'postMessage' of undefined
    at mockServiceWorker.js:256
    at new Promise (<anonymous>)
    at sendToClient (mockServiceWorker.js:245)
    at mockServiceWorker.js:206
```

https://github.com/mswjs/msw/blob/1769a1e2647a35ace2485837cd7622b8a5e5d545/src/mockServiceWorker.js#L244-L258

## Expected behavior

No errors should be visible. Maybe MSW should expect `client` to be falsy in some edge cases.

## Repro:

This issue can be reproduced by navigating back in browser before request resolves. In my real use case this is triggered by hot-reloads of `webpack-dev-server` and/or `react-refresh`. I think these might also come from full page reloads. 

1. Clone https://github.com/mswjs/examples
2. Prepare project as instructed (+ `npx msw init public` for latest mockServiceWorker)
3. Add delay to handler as shown above
4. `npm run start` in rest-react example
5. Once app has started change URL to `http://localhost:3001/?temp`
6. Type user to text input. Wait for request to start and press browsers back button before request resolves
7. Error is shown in browser JS console

Edit: Also reproduced without changing URL:
1-4: as above
5. Type user to text input. Wait for request to start and refresh page.
6. Error is shown in browser JS console

timdeschryver:
This PR keeps track of pending requests and cancels them when needed.
For browsers this is when the page is refreshed, while for a Node environment this is when `resetHandlers` is invoked (thus in most cases after each test).

TODO: 
- [ ] Writing tests, I'll have to work on a reproduction first (I already tried this in the reproductions and it seems to work there)

Closes #506 
Closes #580

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -11,6 +11,7 @@ const INTEGRITY_CHECKSUM = '<INTEGRITY_CHECKSUM>'
 const bypassHeaderName = 'x-msw-bypass'
 
 let clients = {}
+let pendingRequests = {}
 
 self.addEventListener('install', function () {
   return self.skipWaiting()
@@ -78,6 +79,11 @@ self.addEventListener('message', async function (event) {
       if (remainingClients.length === 0) {
         self.registration.unregister()
       }
+
+      // Clear pending requests so they don't resolve
+      // Otherwise this causes previous requests to resolve after a refresh
+      Object.values(pendingRequests).forEach(clearTimeout)
+      pendingRequests = {}
 
       break
     }
@@ -158,10 +164,13 @@ self.addEventListener('fetch', function (event) {
 
       switch (clientMessage.type) {
         case 'MOCK_SUCCESS': {
-          setTimeout(
-            resolve.bind(this, createResponse(clientMessage)),
-            clientMessage.payload.delay,
-          )
+          const t = setTimeout(() => {
+            if (pendingRequests[requestId]) {
+              resolve(createResponse(clientMessage))
+              delete pendingRequests[requestId]
+            }
+          }, clientMessage.payload.delay)
+          pendingRequests[requestId] = t
           break
         }
 
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -11,7 +11,7 @@ const INTEGRITY_CHECKSUM = '<INTEGRITY_CHECKSUM>'
 const bypassHeaderName = 'x-msw-bypass'
 
 let clients = {}
-let pendingRequests = {}
+const pendingRequests = new Map()
 
 self.addEventListener('install', function () {
   return self.skipWaiting()
@@ -80,10 +80,10 @@ self.addEventListener('message', async function (event) {
         self.registration.unregister()
       }
 
-      // Clear pending requests so they don't resolve
+      // Clear pending requests
       // Otherwise this causes previous requests to resolve after a refresh
-      Object.values(pendingRequests).forEach(clearTimeout)
-      pendingRequests = {}
+      pendingRequests.forEach(clearTimeout)
+      pendingRequests.clear()
 
       break
     }
@@ -164,13 +164,13 @@ self.addEventListener('fetch', function (event) {
 
       switch (clientMessage.type) {
         case 'MOCK_SUCCESS': {
-          const t = setTimeout(() => {
-            if (pendingRequests[requestId]) {
+          const responseTimeout = setTimeout(() => {
+            if (pendingRequests.has(requestId)) {
               resolve(createResponse(clientMessage))
-              delete pendingRequests[requestId]
+              pendingRequests.delete(requestId)
             }
           }, clientMessage.payload.delay)
-          pendingRequests[requestId] = t
+          pendingRequests.set(requestId, responseTimeout)
           break
         }
 
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -39,7 +39,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
         )
     })
     const interceptor = new RequestInterceptor(interceptors)
-    let pendingRequests: { [requestId: string]: NodeJS.Timeout } = {}
+    const pendingRequests = new Map<string, NodeJS.Timeout>()
 
     // Error when attempting to run this function in a browser environment.
     if (!isNodeProcess()) {
@@ -151,14 +151,14 @@ export function createSetupServer(...interceptors: Interceptor[]) {
 
             // the node build will use the timers module to ensure @sinon/fake-timers or jest fake timers
             // don't affect this timeout.
-            const t = setTimeout(() => {
+            const responseTimeout = setTimeout(() => {
               // Only resolve when the request is still relevant
-              if (pendingRequests[mockedRequest.id]) {
+              if (pendingRequests.has(mockedRequest.id)) {
                 resolve(mockedResponse)
-                delete pendingRequests[mockedRequest.id]
+                pendingRequests.delete(mockedRequest.id)
               }
             }, response.delay ?? 0)
-            pendingRequests[mockedRequest.id] = t
+            pendingRequests.set(mockedRequest.id, responseTimeout)
             emitter.emit('request:end', mockedRequest)
           })
         })
@@ -173,9 +173,10 @@ export function createSetupServer(...interceptors: Interceptor[]) {
       },
 
       resetHandlers(...nextHandlers) {
-        // Clear pending requests so they don't resolve
-        Object.values(pendingRequests).forEach(clearTimeout)
-        pendingRequests = {}
+        // Clear pending requests
+        // Otherwise this causes previous requests to resolve when a test has already finished
+        pendingRequests.forEach(clearTimeout)
+        pendingRequests.clear()
 
         currentHandlers = requestHandlerUtils.resetHandlers(
           requestHandlers,
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -39,6 +39,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
         )
     })
     const interceptor = new RequestInterceptor(interceptors)
+    let pendingRequests: { [requestId: string]: NodeJS.Timeout } = {}
 
     // Error when attempting to run this function in a browser environment.
     if (!isNodeProcess()) {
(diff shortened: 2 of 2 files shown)


src/mockServiceWorker.js
```diff
@@ -158,10 +163,11 @@ self.addEventListener('fetch', function (event) {
 
       switch (clientMessage.type) {
         case 'MOCK_SUCCESS': {
-          setTimeout(
-            resolve.bind(this, createResponse(clientMessage)),
-            clientMessage.payload.delay,
-          )
+          const t = setTimeout(() => {
```
timdeschryver: @kettanaito what was the reason to add the timeout in this service worker?
I was thinking some abort logic could be centralized if the delay would've  been handled in the library.

src/node/createSetupServer.ts
```diff
@@ -39,6 +39,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
         )
     })
     const interceptor = new RequestInterceptor(interceptors)
+    let pendingRequests: { [requestId: string]: NodeJS.Timeout } = {}
```
kettanaito: Perhaps we could consider using `Set` for storing the pending requests?

```js
const pendingRequests = new Set<string, NodeJS.Timeout>()

pendingRequests.has(mockedRequest.id)
pendingRequests.delete(mockedRequest.id)
pendingRequests.clear()
```

- Don't mutate variables.
- Don't mutate objects with `delete`.

src/node/createSetupServer.ts
```diff
@@ -150,10 +151,14 @@ export function createSetupServer(...interceptors: Interceptor[]) {
 
             // the node build will use the timers module to ensure @sinon/fake-timers or jest fake timers
             // don't affect this timeout.
-            setTimeout(() => {
-              resolve(mockedResponse)
+            const t = setTimeout(() => {
+              // Only resolve when the request is still relevant
+              if (pendingRequests[mockedRequest.id]) {
```
timdeschryver: Having this `if` statement was needed in node environments, but not in the service worker.
To keep things consistent (and just to be sure), I've also added it to the service worker.

src/node/createSetupServer.ts
```diff
@@ -168,6 +173,10 @@ export function createSetupServer(...interceptors: Interceptor[]) {
       },
 
       resetHandlers(...nextHandlers) {
+        // Clear pending requests so they don't resolve
+        Object.values(pendingRequests).forEach(clearTimeout)
```
kettanaito: Technically, we are only clearing the timeouts that resolve the responses. However, that still leaves the requests pending, as what it means is that the request middleware just never resolves, as opposed to _aborts_ a request. 

I think if we wish to support this in NodeJS we need to extend `node-request-interceptor` to provide a client-agnostic API to abort requests. Both `ClientRequest` and `XMLHttpRequest` should support aborting. 

```js
intereceptor.use((req, mod, utils) => {
  utils.abort()
})
```

src/node/createSetupServer.ts
```diff
@@ -150,10 +151,14 @@ export function createSetupServer(...interceptors: Interceptor[]) {
 
             // the node build will use the timers module to ensure @sinon/fake-timers or jest fake timers
             // don't affect this timeout.
-            setTimeout(() => {
-              resolve(mockedResponse)
+            const t = setTimeout(() => {
```
kettanaito: Shall we name it `responseTimeout`? I think there should be a more descriptive name than `t` :)

src/mockServiceWorker.js
kettanaito: The timeout was added here to push the delaying as much up as possible (service worker being the last link in the request handling chain). We should be able to move the delay to the library's code, but we need to inspect how that delay will be reflected in the Network tab. I have a suspicion there's a chance the response time will be instant if the delay is on the library side, but I'm not sure.

src/node/createSetupServer.ts
kettanaito: `setupWorker` has the same issue, but since in that case, the loss of the context means closing a tab, unresolved Promise will be dumped and forgotten (at least, I hope it would be). In NodeJS that's not the case, as the interception happens on the entire process.

kettanaito:
It would be nice to release this alongside #589, so we introduce all the changes to the worker script under a single release.

timdeschryver:
@kettanaito I'll try to resolve the comments (thanks for the review) tomorrow. 
I got a bit sidetracked the last days with other projects 😅

src/node/createSetupServer.ts
timdeschryver: Yea, that's also probably why I needed the extra check in `setupServer` and not in `setupWorker`.

```ts
if (pendingRequests.has(mockedRequest.id)) {
                resolve(mockedResponse)
                pendingRequests.delete(mockedRequest.id)
}
```

src/node/createSetupServer.ts
timdeschryver: I also updated the service worker (it seems like `Map` is supported accross older browsers).
Fyi, `Set` only holds values, so I used a `Map` which hold key-value pairs.

timdeschryver:
I've addressed the review, and I also rebased.
I'll try to write some tests for this later.

src/mockServiceWorker.js
```diff
@@ -79,6 +80,11 @@ self.addEventListener('message', async function (event) {
         self.registration.unregister()
       }
 
+      // Clear pending requests
+      // Otherwise this causes previous requests to resolve after a refresh
+      pendingRequests.forEach(clearTimeout)
+      pendingRequests.clear()
```
kettanaito, line 86: I'm curious to see how the worker/clients will behave with this. Technically, even if we clear the response resolving timeouts, the response promise that we pass to `event.respondWith` till still be pending. I may be wrong on this, though. 

Giving a little heads up: I'm reworking our integration tests to use `page-with`. Planning to merge that today/tomorrow. That should make the integration test writing a slightly better experience.

timdeschryver: While I'm trying to write tests for this, I'm thinking of a different solution that might address your valid concerns because that promise will still be pending.
Instead of clearing the timeout, we could also invoke the `reject` method of the promise.

In the test itself this does give the error `FetchError: request to https://test.api/ failed, reason: getaddrinfo ENOTFOUND test.api` tho :( . I'll also try this approach in the reproduction repo's.

timdeschryver: Seems like rejecting the promise won't work, this gives the following result in the vue repo (https://github.com/sumerokr/vue-mswjs)

```
(node:11268) UnhandledPromiseRejectionWarning: FetchError: invalid json response body at https://example.com/user reason: Unexpected token < in JSON at position 0
(node:11268) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 3)
```

timdeschryver:
There could be a better way of doing this, but I don't see it.
As soon as the promise is rejected (or fulfilled) the client (the one making the request) receives the response.
In other words, if we abort the promise the client receives an error.
The proposed solution kinda swallows the request so it never completes.

src/mockServiceWorker.js
kettanaito: Perhaps we should `resolve` it instead? We'd have to provide it with a valid `Response` instance, otherwise the `event.respondWith` will raise an exception. Providing it with an empty response should suffice:

```js
// When clearing pending response promises:
resolve(new Response())
```

kettanaito:
> In other words, if we abort the promise the client receives an error.

That's true. I suggest we try to resolve the pending promise instead of rejecting it. Since in both browser and node scenarios the execution context is destroyed (page close/reload, handlers reset), the dependent code doesn't care about the resolved response, as there isn't any context left to evaluate that response. We can go even as to resolve the pending promise with an empty response and it should be alright. 

I'm still concerned with this approach in NodeJS, as I don't think that resolving the promise created in MSW will suffice. Looks like we'd have to abort/prematurely resolve the response on the `node-request-interceptor` side for things to be consistent. This needs investigation.

timdeschryver:
@kettanaito even if we resolve the promise, it needs a valid response body because the client still receives the response if I remember correctly. I'm not familiar with node-request, we could split this PR into two, one for the browser and the other one for node (that needs more investigation).

kettanaito:
> we could split this PR into two, one for the browser and the other one for node (that needs more investigation).

This is a great suggestion. If we have a working implementation and tests for the browser we can merge that first.

timdeschryver:
I'll update the PR during this week 
Do you think we can write a test for this case? 
I failed all my attempts

kettanaito:
Thank you, @timdeschryver! I'll try to write a test that would make a request and close the client while the request is still pending. Let's see if there's any reliable way to assert what happens to such a request.

timdeschryver:
I'm closing this issue. we'll tackle the node timeout in a seperate PR.
For browsers, #589 resolves this issue.

PR #587, opened 2021-02-07
Adds type checking for GraphQL data payload

### LINKED ISSUE #555: Typechecking of graphql data response doesn't work in versions >0.24.0
## Describe the bug
Typechecking of graphql data response doesn't work in versions >0.24.0. See example below

## Environment

- `msw: 0.25.0`
- `nodejs: 15.6.0`
- `npm: 7.4.1`

## To Reproduce
Consider this code in `msw@0.24.0`:

```ts
interface GetUsers {
  users: { id: number; login: string }[]
}
// ...
server.use(
  graphql.query<GetUsers>('GetUsers', (req, res, ctx) => res(ctx.data({}))),
)
```

This will produce a typechecking error, which is expected:
```
Argument of type '{}' is not assignable to parameter of type 'GetUsers'.
```

In `msw@0.25.0` I see no error in the above code. 

## Expected behavior
I expected to see a typechecking error in `msw@0.25.0`

### LINKED ISSUE #574: graphql.query response type is not checked by TypeScript
## Describe the bug
According to [the documentation](https://mswjs.io/docs/api/graphql/query#usage-with-typescript), we should be able to annotate queries in TypeScript to get validation for query data and variables. This works for query variables, but not for query data. Provided `QueryType` is not used in resolver types.

## Environment

- `msw: 0.26.0`

## To Reproduce
See [this example](https://codesandbox.io/s/xenodochial-lake-rqxuc?file=/src/index.ts) - query data is different from the type, and there is no TypeScript error.

## Expected behavior
TypeScript should raise an error if query data type does not correspond to the provided `QueryType`.

dankremniov:
Basically, this PR partially reverts changes to the types made in 8f989ef5fdc1680aac1d78f2e4d785ab93767bc1.

Moreover, I added `graphql.test-d.ts` file for type definition tests using [ts-expect-error](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) which should help to ensure that graphql data response will not be broken by other refactorings in the future. Unfortunately, type checking (such as `tsc --noEmit`) is not a part of build pipeline at the moment, and, thus, the added tests will not stop build even if they are broken. However, this should be fixed by #579. But for now, they are only useful for TS errors in IDE. Let me know if you think that it does not make sense to have them in this PR and I will remove them.

- Closes #555
- Closes #574 (duplicate).

### DIFF
--- a/src/context/data.ts
+++ b/src/context/data.ts
@@ -3,15 +3,15 @@ import { jsonParse } from '../utils/internal/jsonParse'
 import { mergeRight } from '../utils/internal/mergeRight'
 import { json } from './json'
 
+export type DataContext<T> = (payload: T) => ResponseTransformer
+
 /**
  * Sets a given payload as a GraphQL response body.
  * @example
  * res(ctx.data({ user: { firstName: 'John' }}))
  * @see {@link https://mswjs.io/docs/api/context/data `ctx.data()`}
  */
-export const data = <T extends Record<string, unknown>>(
-  payload: T,
-): ResponseTransformer => {
+export const data: DataContext<Record<string, unknown>> = (payload) => {
   return (res) => {
     const prevBody = jsonParse(res.body) || {}
     const nextBody = mergeRight(prevBody, { data: payload })
--- a/src/graphql.test-d.ts
+++ b/src/graphql.test-d.ts
@@ -1,0 +1,12 @@
+import { graphql } from './graphql'
+
+// @ts-expect-error
+graphql.query<{ key: string }, null>('', (req, res, ctx) => res(ctx.data({})))
+
+graphql.mutation<{ key: string }, null>('', (req, res, ctx) =>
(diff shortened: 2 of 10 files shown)


kettanaito:
Hey, @dankremniov. Thank you for providing a fix for the GraphQL typings! 

It looks great, I will check out your branch and give it a test tomorrow. That definition test looks amazing, thank you for adding it also. I believe #579 can help us in running that definition test as well.

kettanaito:
I took this opportunity to annotate all `graphql.*` usage examples in our integration tests. This should help us catch at least some percentage of type-related violations upon test runs. Plus, it gives a nice example of how to annotate your GraphQL mocks.

kettanaito (approved):
Thank you for fixing this, @dankremniov! Welcome to contributors!

PR #588, opened 2021-02-08
Groups response body tests

kettanaito:
Sorts the response body types into separate tests.

### DIFF
--- a/test/rest-api/body/body-json.test.ts
+++ b/test/rest-api/body/body-json.test.ts
@@ -1,0 +1,20 @@
+import * as path from 'path'
+import { runBrowserWith } from '../../support/runBrowserWith'
+
+test('responds with a JSON response body', async () => {
+  const runtime = await runBrowserWith(
+    path.resolve(__dirname, 'body-json.mocks.ts'),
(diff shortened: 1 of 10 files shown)


test/rest-api/body/body-json.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    path.resolve(__dirname, 'body-json.mocks.ts'),
+  )
+
+  const res = await runtime.request({
+    url: runtime.makeUrl('/json'),
+  })
+  const headers = res.headers()
+  const json = await res.json()
+
+  expect(headers).toHaveProperty('content-type', 'application/json')
```
timdeschryver, line 14: Nit, should these be asserted before the content is read?
Otherwise, I think the behavior would be that `res.json` throws.

timdeschryver (dismissed):
LGTM 👍

test/rest-api/body/body-json.test.ts
kettanaito: I don't mind changing the assertions order. Thank you!

PR #589, opened 2021-02-08
Supports the interception of requests issued in an iframe

### LINKED ISSUE #374: Question: Cypress MSW started, but not intercepted

### LINKED ISSUE #578: Infinite loading when refreshing the page in NextJS
## Describe the bug
Hello @kettanaito 
I add msw to nextjs with server side rendering, which make hot reloading in Nextjs not work.

I tested and saw that at my local, when editting code then waiting it display on screen, but not. I have to Ctrl + f5.
If it's lucky it will show up, otherwise the machine will hang with the words '**waiting for existing socket**' .

But when I close tab and open new tab, it's ok, But that's really inconvenient, if I comment require('../mocks') in _app.js file. It's oke

## Environment

- `msw: 0.26.1`
- `nodejs: 12.18.2`
- `npm: 6.14.5`

> Version chrome: 88.0.4324.104

## To Reproduce
Link repo: [https://codesandbox.io/s/serverless-fog-mex0z?file=/package.json:352-358](url)

## Screenshots
image

kettanaito:
> This pull request introduces a BREAKING CHANGE to the service worker script. 

## GitHub

- Fixes #374
- Potentially fixes #578 
- Closes #383 
- Closes #594

## Changes

- **Enables interception of requests issued in a nested client (iframe).**
- Worker now stores all the "master" clients (clients that activate MSW) in a `Map` instead of an object.
- Worker now resolves a "master" client as a part of the "fetch" event. The client that issues a request doesn't necessarily equal the client that registered MSW. The latter is to whom the worker should communicate the request and await the response. 
- **Side-effect:** worker now checks if the "master" client is activated and ready before attempting to `sendToClient` a "response:* life-cycle events. Previously, this resulted in an infinite page loading due to the pending message as there was no JavaScript code to handle that message.

## Roadmap

- [x] Support intercepting of requests given nested iframes:
```
parent (registers the worker) 
  - child iframe
    - nested iframe --> request!
```
- [x] Support intercepting of requests from a deeply nested iframe, while the top-level iframe is the master client:
```
parent (registers the worker)
  - child iframe
    - deeply nested iframe (makes a request)
```
- [x] Support intercepting of requests given the "main" client and the request-issuing client are both nested clients (Cypress):
```
parent
  - child iframe (registers the worker)
    - nested iframe (makes a request)
```


kettanaito:
Experienced a flaky Puppeteer behavior during this test:

```
 FAIL  msw-api/setup-worker/start/on-unhandled-request/suggestions.test.ts (26.805 s)
  ● REST API › suggest a handler of a different method if its URL is similar

    Protocol error (Runtime.callFunctionOn): Target closed.

      at ../node_modules/puppeteer/lib/cjs/puppeteer/common/Connection.js:208:63
      at CDPSession.send (../node_modules/puppeteer/lib/cjs/puppeteer/common/Connection.js:207:16)
      at ExecutionContext._evaluateInternal (../node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:200:50)
      at ExecutionContext.evaluate (../node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:106:27)
      at DOMWorld.evaluate (../node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:90:24)
```

The test is confirmed stable locally. Must be either CI (Puppeter in CI halts sometimes), or the OS that CI uses to run Puppeteer.

src/mockServiceWorker.js
```diff
@@ -84,150 +82,181 @@ self.addEventListener('message', async function (event) {
...
+async function resolveClient(event) {
+  const client = await self.clients.get(event.clientId)
+  const allClients = await self.clients.matchAll()
+  return client.frameType === 'top-level'
+    ? client
+    : allClients.find((parentClient) => {
+        return (
+          parentClient.frameType === 'top-level' &&
+          parentClient.visibilityState === 'visible' &&
+          client.url.startsWith(parentClient.url)
```
kettanaito: Here we are assuming that the first visible `top-level` that stands at the beginning of the nested client's URL is its parent window.

src/mockServiceWorker.js
```diff
@@ -241,39 +270,35 @@ function serializeHeaders(headers) {
   return reqHeaders
 }
 
-function sendToClient(client, message) {
+async function sendToClient(client, message) {
```
kettanaito: No need to convert it to an `async` function, this is a left-over.

test/fixtures/iframe.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <h2>Iframe</h2>
+    <script>
+      window.request = () => {
```
kettanaito: I wasn't successful in writing a reliable Puppeteer await logic if the request happens on load. Waiting for elements is extremely flaky.

kettanaito:
Hey, @deshiknaves. May I please ask you to build this feature branch and test it with your Cypress setup? 

Let me know if you need any helping to provision a local build. Thanks.

src/mockServiceWorker.js
deshiknaves: This is a much better solution. Let refactor a bit of code in my test harness and then test this out. I'll get back to you soon.

deshiknaves: From initial looks, I'm getting an error:

```
[MSW] Failed to mock a "GET" request to "http://localhost:8080/dist/index.js": TypeError: Cannot read property 'id' of undefined
```

Let me investigate it a bit more and get back to you 😊

deshiknaves: `client.url.startsWith(parentClient.url)` — the client URL could be anything here:

- Parent URL for cypress is `http://localhost:8080/__/#/tests/integration/msw-requests.spec.js`
- Child URL is `http://localhost:3000`

But this URL could be anything at all

deshiknaves: In the case of Cypress, the parent iframe also has a type of `nested`

kettanaito: Thank you for the insights, @deshiknaves!

Please, is there a way we can reproduce a Cypress iframe layout without actually using Cypress? This scenario needs to be put into an integration test for the library. Maybe you can help us out laying out the iframes structure to reproduce it?

Don't worry if you're not sure about the Cypress iframes. If you can post a reproduction repo with Cypress it'd be sufficient also, we could dig into the page structure and attempt to replicate it in an isolated test. 

> But this URL could be anything at all

Wonder how come those 2 URLs have different ports. Does it mean Cypress establishes the runner on port A and the tests themselves on port B?

deshiknaves: Yeah. I'll replicate this in a test. I also have some ideas on how we might be able to identify this. I'll give a few a shot tomorrow and get back.

Yes, Cypress stats an application on a port and the app you are testing might be on another port, or even a completely different domain (it doesn't have to be local).

deshiknaves: It's basically multiple nested iframe. Inception like — 3 levels deep.

kettanaito: Sounds promising! Looking forward to seeing your test setup. We can then move it to the library to ensure this kind of nested client requests is also intercepted. Thank you!

arthurdenner:
@kettanaito, I'm quite new to `msw` so sorry in advance in case something is wrong in my message/guess below.

I'm facing the error this PR is aiming to solve. I have an Electron app which opens Storybook inside an iframe and the requests are not intercepted.

I've built and tried this PR locally, but I've got `Cannot read property 'id' of undefined`. 

I think the reason is the check below, as when using Storybook, the query params added to the URL break the check, and the returned `client` is `undefined`.

Some screenshots that may help visualize what I've described:

image

image

Let me know if I can help with more information.

kettanaito:
Hey, @arthurdenner 👋 That's a great discovery. The current implementation doesn't take into account that the URL of the parent client may contain additional info (query parameters, hashes, even nested paths). Looks like we'd have to parse the parent URL and then check if it's relevant to the nested client. 

This may be a little more complicated given nested paths... Now the actual challenge is to write a test that creates such a scenario.

src/mockServiceWorker.js
deshiknaves: @kettanaito I did end up setting up a nested iframe structure in a unit test and it passed, so I had to go investigate what I looked at the other day.

So, the nesting isn't the problem, but the fact that that the URLs don't match. Removing `client.url.startsWith(parentClient.url)` will make the parent mock the requests of the child application. We could:
- Remove this check, let it mock requests that match regardless of the iframe src URL
- Pass a way for the end user to be able to pass something that will determine if the URL is valid. Maybe a RegExp, but that will have to be serialized and passed though the `event.data` to the service worker.

I can work on polishing this off if you like either of these options?

deshiknaves: On more investigation on the Cypress side I now have a better understanding of what the problem is in this scenario. The worker is started on a child iframe. So the `top-level` iframe isn't the one that is mocking requests. It needs to use one of the `nested` iframes instead.

If the logic is something like this:

```javascript
return client.frameType === 'top-level'
    ? client
    : allClients.find(parentClient => {
        return (
          parentClient.visibilityState === 'visible' &&
          /__cypress/.test(parentClient.url)
        )
      })
```

Then it will work. Obviously, I've hardcoded this RegExp. But if we allow the user to specify a way to determine the ancestor to use, then it will work. What do you think?

kettanaito: Interesting. If the worker is registered in an iframe, then any request originating from that iframe is caught by the worker automatically (iframe client equals worker controlled client). I see room for issues only in this scenario:

```
parent
  - iframe (Cypress, MSW worker)
    - nested iframe -- request --!
```

deshiknaves: That's exactly the situation here. The request is in a different iframe but worker is also in a nested iframe.

deshiknaves:
@kettanaito so I can get a unit test written with the right level of nesting, but don't have a way to run `worker.start` on the middle iframe.

I was thinking it could be possible to add something like this to the window:

```javascript
// @ts-ignore
window.msw = {
  worker,
  createIframe: () => {
    // Append an iframe to the body and issue a request from it.
    const iframe = document.createElement('iframe')
    iframe.id = 'middle'
    iframe.setAttribute('src', '/test/support/template/index.html')
    document.body.appendChild(iframe)
  },
}
```

Then on the top level page call:

```javascript
runtime.page.evaluate(() => {
    // @ts-ignore
    window.msw.createIframe()
  })

const middle = await runtime.page.waitForSelector('#middle')
  const middleFrame = await middle.contentFrame()
  await middleFrame.evaluate(() => {
    // @ts-ignore
    window.msw.worker.start()
  })
```

Essentially loading the same page in the top frame and the middle frame and only starting the server on the middle frame. However, looks like `window.msw` is only available on the top level page, even though they load the same page. I'm going to start digging into `runBrowserWith` and see how it get's this mock into the context of the page.

I thought I'd put this down here incase you have some suggestions on this.

If this were possible, then we could run:

```javascript
await middleFrame.evaluate(() => {
    // @ts-ignore
    window.msw.worker.start()
    const iframe = document.createElement('iframe')
    iframe.id = 'iframe'
    iframe.src = '/test/fixtures/iframe.html'
    document.body.appendChild(iframe)
  })

  const iframe = await frame.waitForSelector('#iframe')
  const childFrame = await iframe.contentFrame()
  await childFrame.evaluate(() => window.request())

  const firstNameElement = await childFrame.waitForSelector('#first-name')
  const firstName = await firstNameElement.evaluate((node) => node.textContent)

  expect(firstName).toBe('John')
```

## Update:
I realized that the webpack build is just `/` instead of the the template.html. ~So now I'm loading with `window.location.href` so that it loads the same page. Getting the `window.msw` but not worker there. But making some progress 😊~

I've got this working and can be seen in the draft PR https://github.com/mswjs/msw/pull/594

kettanaito:
Thanks for the help in preparing the integration test for this, @deshiknaves. Please note that we no longer use `runBrowserWith` as it's been replaced with the `page-with` library. If you still see that old utility, double-check if you're on the latest state of the repository. Thanks.

src/mockServiceWorker.js
```diff
@@ -82,22 +82,26 @@ self.addEventListener('message', async function (event) {
...
-        return (
-          parentClient.frameType === 'top-level' &&
-          parentClient.visibilityState === 'visible' &&
-          client.url.startsWith(parentClient.url)
-        )
-      })
+
+  return allClients
+    .filter((client) => {
+      return client.visibilityState === 'visible'
```
kettanaito, line 101: By filtering only the visible clients we get a list of clients on the current page. This way we omit all the other possibly open clients.

src/mockServiceWorker.js
```diff
@@ -82,22 +82,26 @@ self.addEventListener('message', async function (event) {
...
+    })
+    .find((client) => {
+      return activeClientIds.has(client.id)
```
kettanaito, line 106: No matter the number of nested clients (i.e. iframes), one of them is the one that registered the worker (master client). We can look up if its ID is in the stored active clients' Map and resolve if it is, otherwise, it means that the master client hasn't sent the activation message to the worker yet and the requests must be bypassed.

kettanaito:
Hi, @deshiknaves. I've added an integration test for the scenario when MSW is registered in a nested iframe. The tests are passing now, so I hope this PR adds the needed support for the Cypress use case. Could you please try out these latest changes in your setup and let us know? Thanks.

deshiknaves:
@kettanaito Yep, I will check it out in a couple of hours 😊

deshiknaves:
@kettanaito this is perfect 👏🏽. Works like a charm, Thanks for looking into this.

test/msw-api/setup-worker/scenarios/iframe/iframe.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  const firstNameElement = await frame.waitForSelector('#first-name')
+  const firstName = await firstNameElement.evaluate((node) => node.textContent)
+
+  expect(firstName).toBe('John')
+})
+
+test('intercepts a request from a deeply nested iframe', async () => {
+  const { page } = await pageWith({
+    example: path.resolve(__dirname, 'iframe.mocks.ts'),
+    markup: path.resolve(__dirname, 'iframe-deep.page.html'),
```
Aprillion: 😱😱 I find this naming slightly headache-inducing.. what about following?
* page-without-iframe.html
* page-with-one-iframe.html
* page-with-two-nested-iframes.html
(leaving space for more levels in the future if needed and/or for page-with-two-parallel-iframes if you will need to test multiple iframes side by side)

Aprillion (dismissed):
https://github.com/mswjs/msw/pull/589#issuecomment-778127983 mentions Storybook and this PR looks very promising in that direction, but there is no Storybook example yet => do you track the idea to add Storybook example in a separate issue perhaps?

test/msw-api/setup-worker/scenarios/iframe/iframe.test.ts
kettanaito: Good point about the naming. It's a little cryptic, I agree. The challenge comes from the way the in-browser tests are constructed: the compiled example code is always appended to the main page. That's why in the deeply nested iframe test suite I create an empty page and then load what is effectively the first testing scenario (`top -> nested iframe`) in _another_ iframe.

kettanaito:
Thank you for the review, @Aprillion. 

I think Storybook is covered by the first/second integration tests, is it not? There may be multiple points of integration, but the one we recommend in the examples repo was already tested.

test/msw-api/setup-worker/scenarios/iframe/iframe.test.ts
kettanaito: I've renamed them to:

- `app.html` where the request logic lives.
- `page-in-iframe` where the's an iframe for the `app.html`.
- `page-in-nested-iframe` where the `page-in-iframe` is nested in another iframe.

Aprillion:
> I think Storybook is covered by the first/second integration tests, is it not?

I meant a fully working copy&pastable example of a `msw.story.js` file - but I just discovered it already exists in a separate repo https://github.com/mswjs/examples/tree/master/examples/with-storybook => so don't mind my suggestion, nothing new to add (probably?) :)

src/mockServiceWorker.js
kettanaito: Thanks once more for the detailed investigation about how Cypress iframes are structured. This scenario should be covered in the integration tests now.

test/fixtures/iframe.html
```diff
@@ -0,0 +1,20 @@
...
+          .then((data) => {
+            const node = document.createElement('p')
+            node.setAttribute('id', 'first-name')
+            node.innerText = data.firstName
+            document.body.appendChild(node)
+          })
+      }
+
+      document.addEventListener('click', window.request)
+    </script>
```
marcosvega91: We still need this file ?

marcosvega91 (dismissed):
Helloooo the work is super awesome 🚀 , I have left only one comment

test/fixtures/iframe.html
kettanaito: True, we don't use this file anymore :) Will remove.

PR #590, opened 2021-02-11
Uses "page-with" for example-based testing

kettanaito:
## Changes
- Integration tests now use the `page-with` library instead of a custom made setup. The library is inspired by that setup we used to have.
- The call signature of the `request` utility now resembles `fetch`.
- No need to clean up after each test.
- Removes the focused test mode in favor of debug mode in `page-with`.

## Benchmarks
Unlike a new browser and WDS process per test previously, `page-with` spawns 1 browser and 1 webpack server, decreasing the build time and making tests faster.

Below you can see some benchmarks of test runs on the same machine and test command.

- `/test/rest-api`
```diff
-77.556 s
+70.626 s
```
**Delta:** ~8s faster.

- `/test/graphql-api`
```diff
-33.465 s
+28.791 s
```
**Delta:** ~5s faster.

- `/test` (all integration tests)
```diff
-216.11 s
+174.514 s
```
Delta: ~43s faster.


kettanaito:
The next step will be to run the integration tests in parallel (currently we run in band) and verify that the RAM consumption issue is still relevant (shouldn't be as we are now spawning a single automated Chromium).

marcosvega91 (dismissed):
nothing to say, nice improvements 💯

PR #594, opened 2021-02-14
tests(iframe): Load the mocks in a nested iframe

deshiknaves:
@kettanaito this is how I think we can start the worker in a `nested` frame instead of the `top-level` frame.

- Top level loads `/` which uses `window.createIframe(id, src)` to create a child iframe with the same URL
- That iframe starts the worker creates another iframe which loads `iframe.html` and then makes a request
- This is not intercepted because the worker is on the middle iframe and not the parent

This mimics the document structure of Cypress with one exception — the URLs can be completely different on these iframes.

I'm going to look at a few options on how the user can specify which frame to use. Maybe some way to let them use regex to locate the page?

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -270,7 +270,7 @@ function serializeHeaders(headers) {
   return reqHeaders
 }
 
-async function sendToClient(client, message) {
+function sendToClient(client, message) {
   return new Promise((resolve, reject) => {
     const channel = new MessageChannel()
 
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -239,6 +239,13 @@ self.addEventListener('fetch', function (event) {
     return
   }
 
+  // Bypass all requests when there are no active client IDs.
+  // Prevents the self-unregistered worked from handling requests
+  // after it's been deleted (still remains active until the next reload).
+  if (activeClientIds.size === 0) {
+    return
+  }
+
   const requestId = uuidv4()
 
   return event.respondWith(
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -9,8 +9,7 @@
 
 const INTEGRITY_CHECKSUM = '<INTEGRITY_CHECKSUM>'
 const bypassHeaderName = 'x-msw-bypass'
-
-let clients = {}
+const activeClientIds = new Set()
 
 self.addEventListener('install', function () {
   return self.skipWaiting()
@@ -34,7 +33,6 @@ self.addEventListener('message', async function (event) {
   }
 
   const allClients = await self.clients.matchAll()
-  const allClientIds = allClients.map((client) => client.id)
 
   switch (event.data) {
     case 'KEEPALIVE_REQUEST': {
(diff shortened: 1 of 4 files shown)


kettanaito:
Hey 👋 Thanks for writing a test scenario for this!

> I'm going to look at a few options on how the user can specify which frame to use. Maybe some way to let them use regex to locate the page?

I don't think the end-developer should be in charge of which iframe to treat as the right client. For a developer that matters little: they see a page making a request, and they expect that request to be intercepted and mocked. Instead, it's the worker (or client) that needs to analyze the request's client and go from there. 

I may be mistaken, but I see the only reason for any issue to happen is when the request-issuing client (i.e. nested iframe) is not the client _that registered the worker_. We should try building our logic around that and see whether that statement holds. 

Please correct me if I'm mistaken, it looks like this is what happens with scenarios like Cypress:

```
(Top-level) Observable application
  -- (Nested client) Cypress shell. This is where the MSW worker is registered.
    -- (Nested client) Nested iframe (your app under test). This is where request happens. 
 ```

Since both the main client (where the worker is registered) and the request-issuing client are nested clients, the currently suggested logic in #589 doesn't resolve the end client to communicate with.

**Comment by deshiknaves** [comment]
> Hey 👋 Thanks for writing a test scenario for this!
> 
> > I'm going to look at a few options on how the user can specify which frame to use. Maybe some way to let them use regex to locate the page?
> 
> I don't think the end-developer should be in charge of which iframe to treat as the right client. For a developer that matters little: they see a page making a request, and they expect that request to be intercepted and mocked. Instead, it's the worker (or client) that needs to analyze the request's client and go from there.
> 
> I may be mistaken, but I see the only reason for any issue to happen is when the request-issuing client (i.e. nested iframe) is not the client _that registered the worker_. We should try building our logic around that and see whether that statement holds.

Makes sense, however, I'm not seeing much context that's available to know which frame would have the the mocks in the hierarchy. I was suggesting by default it looks at the `top-level`. However, if that's not the case, then we need something else.

Do you know of any way to know if the `client` is in the hierarchy of current context? And if there is a server on that client?

**Comment by kettanaito** [comment]
It should come down to figuring out if a client is a parent of another client. That would solve any address-related issues. 

When a request originates from a nested iframe, its `event.clientId` comes to the `fetch` event. Here we block unknown client IDs for two reasons:

1. Prevent worker from handling initial page load assets (html/js/css). Those must _always_ be bypassed.
2. Enable per-client mock pausing via `worker.stop()`. Worker can have multiple clients (i.e. tabs) and have mocking enable in one, while disabled in another. 

The first point is far more important than the second one. If we omit any kind of client ID checks, I believe the nested clients won't be an issue at all, and all requests will be captured and mocked as expected. However, when reloading the page, those initial assets will hit the worker and the following thing will happen:

```
request (asset)
  -> worker fetch
  -> sentToClient (are there any handlers for this request?)
  -> message will pend infinitely, as the client lives in JavaScript that's not yet downloaded/parsed. 
```

I'm voicing this out because the actual root of the problem may lie elsewhere than nested clients handling.

deshiknaves:
@kettanaito thanks for the info. That give me a lot of background 😊

PR #598, opened 2021-02-15
TypeScript: Fixes declarations auto-import

### LINKED ISSUE #596: Types autocomplete from the "lib/types" module
## Steps to reproduce

1. Create a blank TypeScript file in a project that has `msw` installed.
2. Type `setupWorker` and press Enter to auto-import that function.
3. See the `setupWorker` function imported from `msw/lib/types`.

## Expected behavior

API that MSW exposes must be auto-imported from the right path. In the case above, `import { setupWorker } from 'msw'`. Account for the functions that are in `msw/node` and `msw/native`.

kettanaito:
## GitHub

- Fixes #596

## Changes

- Type declarations are not emitted next to each build target directory (esm, umd, cjs, iife).

## Roadmap

- [x] Ensure auto-imports from the `msw` package (defaults to `lib/umd`).
- [x] Ensure auto-imports from the `msw/node` package.

### DIFF
--- a/node/package.json
+++ b/node/package.json
@@ -1,3 +1,4 @@
 {
+  "main": "lib",
   "types": "../lib/types/node"
 }
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -26,9 +26,7 @@ const plugins = [
     input: SERVICE_WORKER_SOURCE_PATH,
     output: SERVICE_WORKER_BUILD_PATH,
   }),
-  typescript({
-    useTsconfigDeclarationDir: true,
-  }),
+  typescript(),
   commonjs(),
 ]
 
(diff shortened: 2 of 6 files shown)


rollup.config.ts
```diff
@@ -96,6 +94,10 @@ const buildNode = {
       extensions: ['.js', '.jsx', '.ts', '.tsx'],
     }),
     typescript({
+      /**
+       * @fixme Export only "msw/node" relevant type declarations
+       * to enable auto-imports from "msw/node".
+       */
       useTsconfigDeclarationDir: true,
```
kettanaito, line 99: If this is set to `false` the entire library declarations will be emitted to `node/`:

```
node/
  handlers/
    RestHandler.d.ts
  context/
   json.d.ts
  ...
  index.d.ts
```

This isn't correct. Only the node-related declarations (those in `src/node`) must be emitted to the `node/` directory.

node/package.json
```diff
@@ -1,3 +1,4 @@
 {
+  "main": "lib",
```
kettanaito, line 2: Nesting the entire `node`/`native` builds in a `lib` directory so it's easier to perform a cleanup.

kettanaito:
I've tested this on an empty CRA+TS project and the auto-imports of `setupWorker`/`setupServer`/`rest`/`graphql` worked correctly.

PR #600, opened 2021-02-15
Adds FormData request body integration test

### LINKED ISSUE #599: FormData body is parsed as "[object FormData]" string
## Describe the bug

When sending a POST request with a `FormData` body, the `req.body` contains a string `"[object FormData]"` and I can't get the data that was sent.

## Environment

- `msw: 0.26.2`
- `nodejs: 14.15.4`
- `npm: 6.14.11`

Browser version is not relevant as I'm using msw in Jest tests context. Jest version is 26.6.3. 

## To Reproduce

1. Send a request from the application with POST method and a `FormData` object as the body

```js
const form = document.querySelector(".my-form")

fetch("http://localhost/user/register", {
  method: "POST",
  body: new FormData(form)
})
```

2. Implement a POST handler in msw and log `req.body`:

```js
import { rest } from "msw"

export const handlers = [
    rest.post("/user/register", (req, res, ctx) => {
    console.log(req.body, typeof req.body)

    return res(ctx.status(200))
  }),
]
```

3. See that you get `[object FormData] string` as the `console.log` result

## Expected behavior

I expected `req.body` to be an object like it is in the browser.

kettanaito:
## GitHub

- Closes #599


test/rest-api/request/body/body-form-data.page.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <form id="sign-in-form">
```
Aprillion, line 4: [idea only slightly related to the purpose of PR]:

`<form action="/sign-in" method="POST">` should do the same thing as you do manually in `<script>`, but not sure how `pageWith` handles submitting real forms

Aprillion (dismissed):

test/rest-api/request/body/body-form-data.page.html
kettanaito: That's a good point, but I'd rather leave the `FormData` construction to be explicit. People can then see where and how that form data is created and how it's passed to the request. It's not always you submit a form in that conventional way like you've described. I'd even say in these modern times you rarely do so.

PR #602, opened 2021-02-18
fix: not await for .text() when receiving a response

marcosvega91:
This PR should fix an issue on browser that causing some timeout using streams.

The problem was that in the service worker the response was delayed to send the response to the client.

This behavior seems to be happening also in the browser as for #636

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -115,21 +115,23 @@ async function handleRequest(event, requestId) {
   // Ensure MSW is active and ready to handle the message, otherwise
   // this message will pend indefinitely.
   if (activeClientIds.has(client.id)) {
-    const clonedResponse = response.clone()
-
-    sendToClient(client, {
-      type: 'RESPONSE',
-      payload: {
-        requestId,
-        type: clonedResponse.type,
-        ok: clonedResponse.ok,
-        status: clonedResponse.status,
-        statusText: clonedResponse.statusText,
-        body: clonedResponse.body === null ? null : await clonedResponse.text(),
-        headers: serializeHeaders(clonedResponse.headers),
-        redirected: clonedResponse.redirected,
-      },
-    })
+    ;(async function () {
+      const clonedResponse = response.clone()
+      sendToClient(client, {
+        type: 'RESPONSE',
+        payload: {
+          requestId,
+          type: clonedResponse.type,
+          ok: clonedResponse.ok,
+          status: clonedResponse.status,
+          statusText: clonedResponse.statusText,
+          body:
+            clonedResponse.body === null ? null : await clonedResponse.text(),
+          headers: serializeHeaders(clonedResponse.headers),
+          redirected: clonedResponse.redirected,
+        },
+      })
+    })()
   }
 
   return response
--- a/test/support/utils.ts
+++ b/test/support/utils.ts
@@ -3,6 +3,9 @@ import { MockedRequest } from './../../src'
 import { uuidv4 } from '../../src/utils/internal/uuidv4'
 import { ChildProcess } from 'child_process'
 
+const TIMEOUT = 3000
+const RETRY_INTERVAL = 1000
+
 export function sleep(duration: number) {
   return new Promise((resolve) => {
     setTimeout(resolve, duration)
@@ -57,5 +60,28 @@ export function promisifyChildProcess(
     child.addListener('exit', () => {
       resolve({ stdout, stderr })
     })
+  })
+}
+
+export function waitUntil(expectation: () => void) {
+  const maxTries = Math.ceil(TIMEOUT / RETRY_INTERVAL)
+  let tries = 0
+  return new Promise((resolve, reject) => {
+    const rejectOrRerun = (error: unknown) => {
+      if (tries > maxTries) {
+        reject(error)
+        return
+      }
+      setTimeout(runExpectation, RETRY_INTERVAL)
+    }
+    function runExpectation() {
+      tries += 1
+      try {
+        Promise.resolve<void>(expectation()).then(resolve).catch(rejectOrRerun)
+      } catch (error) {
+        rejectOrRerun(error)
+      }
+    }
+    setTimeout(runExpectation, 0)
   })
 }
--- a/test/msw-api/life-cycle-events/life-cycle-events.test.ts
+++ b/test/msw-api/life-cycle-events/life-cycle-events.test.ts
@@ -1,7 +1,7 @@
 import * as path from 'path'
 import { SetupWorkerApi } from 'msw'
 import { ScenarioApi, pageWith } from 'page-with'
-import { sleep } from '../../support/utils'
+import { sleep, waitUntil } from '../../support/utils'
 
 declare namespace window {
   export const msw: {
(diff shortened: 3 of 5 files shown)


src/mockServiceWorker.js
```diff
@@ -115,26 +115,29 @@ async function handleRequest(event, requestId) {
...
-        redirected: clonedResponse.redirected,
-      },
-    })
+    processResponse(client, requestId, response)
   }
 
   return response
 }
 
+async function processResponse(client, requestId, response) {
```
kettanaito: Looks like this only moves the block outside. Is there a reason to do so?

We should be cautious when it comes to the service worker changes, as each such change is a breaking change. I'd say we can keep the worker file as-is without this change, it should produce the same result.

test/msw-api/handle-stream.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        setTimeout(() => {
+          res.write(' last-chunk')
+          res.end()
+        }, 1000 * 6)
+      })
+    },
+  })
+}
+
+test('should handle stream with no timeout', async () => {
```
kettanaito: Nitpick: should the _stream_ have no timeout, or should MSW produce no timeouts?

Based on the code below, stream does have the `abortTimeout`, so I believe the test name can be:

```
test('handles a stream response without throwing a timeout error')
```

src/mockServiceWorker.js
marcosvega91: in this way we are not awaiting the `.text` when handling the response

test/support/utils.ts
```diff
@@ -59,3 +59,32 @@ export function promisifyChildProcess(
     })
   })
 }
+
+export function waitUntil(
+  expectation: () => void,
+  { timeout = 3000, interval = 1000 } = {},
```
kettanaito, line 65:
```diff
-interval = 1000
+retryInterval = 1000
```

test/msw-api/handle-stream.test.ts
marcosvega91: yes the test name is better 👍

test/support/utils.ts
marcosvega91: nice !! :)

test/support/utils.ts
```diff
@@ -59,3 +59,32 @@ export function promisifyChildProcess(
...
+        reject(error)
+        return
+      }
+      setTimeout(runExpectation, interval)
+    }
+    function runExpectation() {
+      tries += 1
+      try {
+        Promise.resolve(expectation())
+          .then(() => resolve(undefined))
```
kettanaito: You can chain the resolve directly: `.then(resolve`). 

To prevent any type violations with this, pass `void` as the generic to the upper `Promise`:

```ts
return new Promise<void>(...)
```

test/support/utils.ts
```diff
@@ -59,3 +59,32 @@ export function promisifyChildProcess(
...
+          .catch(rejectOrRerun)
+      } catch (error: unknown) {
```
kettanaito: Does TypeScript require we type the `error` here? Afaik, it should be inferred as a general `Error`. If it's not, we should annotate it as such:

```ts
} catch (error: Error) {
```

test/support/utils.ts
kettanaito: We are not using these options at the moment, so we can drop them from being in the arguments.

```ts
const TIMEOUT = 3000
const RETRY_INTERAL = 1000

export function waitUntil(expectation) { ... }
```

test/msw-api/handle-stream.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  return pageWith({
+    example: path.resolve(__dirname, 'handle-stream.mocks.ts'),
+    routes(app) {
+      app.get('/stream', (req, res) => {
+        res.write('first-chunk')
+
+        setTimeout(() => {
+          res.write(' last-chunk')
+          res.end()
+        }, 1000 * 6)
```
kettanaito: 6s is a rather long timeout. Can we make it lower while still preserving the async nature of the stream in the test?

Perhaps `250 * 6`?

test/msw-api/handle-stream.mocks.ts
```diff
@@ -0,0 +1,10 @@
+import { setupWorker } from 'msw'
+
+const worker = setupWorker()
+
+worker.on('response:bypass', async (res) => {
+  const textResponse = await res.text()
```
kettanaito: This setup (and its test) are relevant in the context of life-cycle events. Maybe let's move them both into `test/msw-api/life-cycle-events`? 

Ideally, it would be nice to introduce this test in `test/rest-api/response/body/stream` and add an assertion that MSW is able to read that stream (which I think you already have).

kettanaito (changes requested):
Those are some fine-grained improvements, @marcosvega91. Thank you!

I've left a few comments, mainly some minor suggestions and a few questions.

test/support/utils.ts
marcosvega91: no we don't need

test/msw-api/handle-stream.test.ts
marcosvega91: yes of course

test/msw-api/handle-stream.mocks.ts
marcosvega91: The behavior is a regression of `life-cycle` apis. So as you said I think that is better to go there. 

Do you prefer to have in `test/rest-api/response/body/stream` or `test/msw-api/life-cycle-events` ?

kettanaito: I think there's little difference, to be honest. As it's a regression, let's keep it close to the problematic surface, that being life-cycle events API.

src/mockServiceWorker.js
kettanaito: Could we achieve the same by wrapping it in a self-invoking function?

```js
if (activeClientIds.has(client.id) {
  (async function handleResponse() {
    // ....
  })()
}
```

The `handleResponse` isn't used anywhere outside of this particular moment, so I feel it's natural to keep it where it's used, adding that extra scope to prevent await. 

It'd be useful to comment next to that function why it's structured that way (that awaiting on `.text` may cause the response handling to timeout).

marcosvega91: Yes we will have the same result

test/msw-api/handle-stream.mocks.ts
marcosvega91: I agree with you 👍

test/support/utils.ts
```diff
@@ -3,6 +3,9 @@ import { MockedRequest } from './../../src'
 import { uuidv4 } from '../../src/utils/internal/uuidv4'
 import { ChildProcess } from 'child_process'
 
+const TIMEOUT = 3000
+const RETRY_INTERAL = 1000
```
kettanaito:
```suggestion
const RETRY_INTERVAL = 1000
```

kettanaito (dismissed):
Amazing!
Just one nitpick about the `RETRY_INTERVAL` typo.

Thank you for this fix!

test/support/utils.ts
marcosvega91: 🤦  hahahahah

kettanaito:
I've rebased the branch against the latest master and tidied commits a little, I hope you don't mind.

kettanaito (approved):

PR #606, opened 2021-02-21
Automatically updates the worker script if the "msw" property is set in "package.json"

### LINKED ISSUE #605: Make the worker updates half-automatic
## Current behavior

When the worker registers, the client-side of the library sends a message to it with the latest worker's checksum to know if your app is running an outdated worker (worker has some changes in the latest version vs. the worker that's currently running). 

### Disadvantages

- Worker updates are not apparent. Users find out about them (if they do) only by seeing the console error in a browser.
- Worker updates are manual, as you need to re-run the `init` command again.

## Suggested behavior

The library can check if there's a designated property set in the project's `package.json` that indicates the public directory where the worker script is located (and has been previously instantiated).

```json
// package.json
{
  "msw": {
    "workerDirectory": "./public"
  }
}
```

> Inferring the public directory from other possible settings is a guessing game, as there's no guarantee that the public directory is otherwise defined or equals to the directory where the user serves the worker script. 

When such a relative directory path is provided, MSW will automatically excute the `init <WORKER_DIR>` command on the `postinstall` hook of the library. That way any updates (or downgrades) of the `msw` package are followed by updating the worker script automatically. 

If the special property is not specified in `package.json`, nothing happens. Developer will still see an integrity mismatch error in a browser's console and would be asked to update the worker script manually (via the CLI). 

### Advantages

- Worker script updates become automatic.

### Disadvantages

- Developer needs to have a special field in `package.json` to enable this.
- The worker script update becomes implicit, even despite the message printed in stdout.

kettanaito:
## GitHub

- Closes #605

### Worker directory auto-detection on `postinstall`

**This will have no effect when installing `msw <= 0.26.x`**.

Starting from the next version, MSW will try to detect the `msw.workerDirectory` property in your `package.json` that indicates the public directory where the worker script is. If that property is set, MSW will automatically update the worker script upon installation. If the property is not set, nothing will happen. 

We should recommend including that property to ensure more seamless updates to the worker script in the future:

```js
{
  "msw": {
    "workerDirectory": "./public"
  }
}
```

### New `--save` option for the `msw init` command

```bash
$ msw init <PUBLIC_DIR> --save
```

When the `--save` option is provided, save the given `PUBLIC_DIR` to the `msw.workerDirectory` in package.json automatically (no prompt).


package.json
```diff
@@ -19,7 +19,8 @@
     "test:smoke": "config/scripts/smoke.sh",
     "test:ts": "tsc -p test/typings/tsconfig.json",
     "test:focused": "node node_modules/ts-node/dist/bin.js --project=test/tsconfig.json test/focusedTest.ts",
-    "prepublishOnly": "yarn lint && yarn test:unit && yarn build && yarn test:integration"
+    "prepublishOnly": "yarn lint && yarn test:unit && yarn build && yarn test:integration",
+    "postinstall": "node -e \"try{require('./config/scripts/postinstall')}catch(e){}\""
```
kettanaito, line 23: Executing a script via `node` eliminates the `Permission denied` issues when running `postinstall.js` as an executable.

timdeschryver (dismissed):
LGTM.
What if the `init` script also adds the `msw` entry to the `package.json`?

kettanaito:
Thank you for the review, @timdeschryver.

> What if the init script also adds the msw entry to the package.json?

Not sure how ethical it is to change the `package.json` without explicit consent. I was thinking that perhaps we could spawn a prompt after `npx init` that would ask if the user wants to add this directory to `package.json` to make the worker script updates easier:

```bash
$ npm install msw
$ npx msw init ./public

Successfully created the Service Worker script!

In order to ease future updates to the worker script, we highly recommend
storing the path to the worker directory in your "package.json".

> Do you wish to add "./public" as the worker public directory? [Y/n]
```

timdeschryver:
👍  That's also how I see it working

cli/index.js
```diff
@@ -15,6 +15,10 @@ yargs
           normalize: true,
         })
         .example('init', 'msw init public')
+        .option('save', {
```
kettanaito, line 18: Adding this option to be able to skip the prompt during integration tests.

cli/init.js
```diff
@@ -1,40 +1,136 @@
 const fs = require('fs')
 const path = require('path')
 const chalk = require('chalk')
+const inquirer = require('inquirer')
 const invariant = require('./invariant')
 const { SERVICE_WORKER_BUILD_PATH } = require('../config/constants')
 
-const cwd = process.cwd()
+const CWD = process.env.INIT_CWD || process.cwd()
```
kettanaito, line 8: `process.env.INIT_CWD` is an environmental variable exposed by NPM that indicates the parent project's CWD when running scripts (i.e. `postinstall`). `process.cwd()` in a `postinstall` script equals the _library's_ directory, not the parent project.

KnisterPeter:
@kettanaito This does not work that well for bigger monorepos. You need to put the "msw" config key in the root `package.json` which is most likely not the one installing the dependency. That means configuration and library need to be configured in different places.

kettanaito:
Hey, @KnisterPeter. That's a valid point. Could you please open an issue about the monorepo support for the worker auto-updates? Thanks.

PR #607, opened 2021-02-21
refactor: throw on start fail, vs null return

### LINKED ISSUE #603: rfc: update `start()` API to resolve with registration or reject
## Environment

n/a

## Request handlers

...any handlers

```js
worker.start()
```

## Actual request

n/a

## Current behavior

`worker.start()` returns `null` if the worker does not register.

## Expected behavior

if the worker does not start, i expect a rejection with a failure message.

without it, i regularly get silent failures, or handle the failure poorly by throwing a generic, non-indicative error message.

## Screenshots

n/a

cdaringe:
# Problem

Fixes #603. A failure currently just returns `null`

# Solution

Failing to start the worker now results in a `throw`, as the expressed intent of `start(...)` failed

### DIFF
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -1,11 +1,7 @@
 import { until } from '@open-draft/until'
 import { getWorkerInstance } from './utils/getWorkerInstance'
 import { activateMocking } from './utils/activateMocking'
-import {
-  SetupWorkerInternalContext,
-  ServiceWorkerInstanceTuple,
-  StartOptions,
-} from '../glossary'
+import { SetupWorkerInternalContext, StartOptions } from '../glossary'
 import { createRequestListener } from '../../utils/worker/createRequestListener'
 import { requestIntegrityCheck } from '../../utils/internal/requestIntegrityCheck'
 import { deferNetworkRequestsUntil } from '../../utils/deferNetworkRequestsUntil'
--- a/src/setupWorker/start/createStart.ts
+++ b/src/setupWorker/start/createStart.ts
@@ -37,10 +37,9 @@ export const createStart = (context: SetupWorkerInternalContext) => {
 
     const startWorkerInstance = async () => {
       if (!('serviceWorker' in navigator)) {
-        console.error(
+        throw new Error(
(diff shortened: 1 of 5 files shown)


src/setupWorker/start/createStart.ts
```diff
@@ -65,29 +64,25 @@ export const createStart = (context: SetupWorkerInternalContext) => {
       )
 
       if (!instance) {
-        return null
+        throw new Error('[MSW] Failed to find locate service worker instance')
```
kettanaito, line 67:
```suggestion
        throw new Error('[MSW] Failed to find a service worker instance.')
```

kettanaito: Should we make the `getWorkerInstance` throw exceptions when failing to locate/create an instance as well? I'm speaking about these cases: 

https://github.com/mswjs/msw/blob/e23dca895156e944b4eb80ce24662711592fb227/src/setupWorker/start/utils/getWorkerInstance.ts#L77-L82

https://github.com/mswjs/msw/blob/e23dca895156e944b4eb80ce24662711592fb227/src/setupWorker/start/utils/getWorkerInstance.ts#L87-L90

If we throw exceptiont in the mentioned cases, we will be able to get that exception as the first argument of the `until`'s return tuple:

```js
const [instanceError, instance ] = await until(() => getWorkerInsance(...))

if (instanceError) {
  throw instanceError
}
```

Note that `until` gracefully handles promise exceptions and will not propagate them to the parental scope automatically. Instead, it will return them as the first item in the tuple.

kettanaito (changes requested):
Hey, @cdaringe! Thank you for picking this improvement up!

It looks great. I've left one comment regarding error propagation for `getWorkerInstance` to make the internal errors more evident (right now it would `console.error` + throw a neutral "Cannot find instance" exception).

src/setupWorker/start/createStart.ts
cdaringe: agreed. patched in 70ab1a4.  i saw a few places where `until` was warranted, but ended up pruning two uses where, per the above, we just plain wanted to throw, rather than fail, detect, and rethrow

kettanaito: @cdaringe, makes sense. Thank you for the follow up on this!

kettanaito (approved):
Great change! Thank you for this, @cdaringe. Welcome to contributors 🎉

PR #613, opened 2021-02-24
Adds tests for "GraphQLHandler" class

kettanaito:
## Changes

- Changes the `request` argument type from `Request` to `MockedRequest` in the GraphQL handler, because it gets a general `MockedRequest` and only then parses/transforms it. 
- Adds tests for the `GraphQLHandler` class.

### DIFF
--- a/src/utils/internal/parseGraphQLRequest.ts
+++ b/src/utils/internal/parseGraphQLRequest.ts
@@ -142,11 +142,11 @@ function getGraphQLInput(request: MockedRequest<any>): GraphQLInput | null {
  */
 export function parseGraphQLRequest(
   request: MockedRequest<any>,
-): ParsedGraphQLRequest | null {
+): ParsedGraphQLRequest {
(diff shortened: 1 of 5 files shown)


src/utils/internal/parseGraphQLRequest.ts
```diff
@@ -161,7 +161,7 @@ export function parseGraphQLRequest(
       `[MSW] Failed to intercept a GraphQL request to "${request.method} ${requestPublicUrl}": cannot parse query. See the error message from the parser below.`,
     )
     console.error(parsedResult)
-    return null
+    return undefined
```
kettanaito, line 164: Returning `undefined` from the `parseGraphQLRequest` function so its return type is compatible with the `ParsedGraphQLRequest` type. No need to keep `Parsed | null | undefined`, that's confusing. We're eliminating `null` by replacing it with `undefined` here.

PR #616, opened 2021-02-25
Updates dependencies

kettanaito:


.eslintrc.js
```diff
@@ -6,7 +6,6 @@ module.exports = {
   },
   extends: [
     'plugin:@typescript-eslint/recommended',
-    'prettier/@typescript-eslint',
```
kettanaito, line 9: No longer necessary ([ref](https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#version-800-2021-02-21)).

rollup.config.ts
```diff
@@ -157,6 +163,7 @@ const buildIife = {
   },
   plugins: [
     replace({
+      preventAssignment: true,
```
kettanaito, line 166: The next major version of the replace plugin will set this to `true` anyway. This is a recommended action from the plugin.

PR #618, opened 2021-02-26
ctx.fetch: Annotates return type

### LINKED ISSUE #608: Return type in Typescript of context.fetch should be Response
Currently the typings for the `context.fetch` declare a return type of `any`. Instead this function should return `Response` since it calls `window.fetch` (or the equivalent on each platform) under the hood.

Explicitly define the return type as `Response` here: https://github.com/mswjs/msw/blob/e23dca895156e944b4eb80ce24662711592fb227/src/context/fetch.ts#L40

You could define your own typing but it's easier to just have this defined upstream.

kettanaito:
## GitHub

- Fixes #608

## Changes

- The `fetch` context utility now annotates an exact return type.
- Removes unnecessary type guards from internal `fetch` context utility functions.
- Annotates the `useFetch` isomorphic fetch function.
- Annotates `ContextMap` of the RequestHandler to have `any` return type: not all context utilities return a `MockedResponse` (i.e. `ctx.fetch` doesn't).

### DIFF
--- a/src/context/fetch.ts
+++ b/src/context/fetch.ts
@@ -2,7 +2,10 @@ import { Headers } from 'headers-utils'
 import { MockedRequest } from '../handlers/RequestHandler'
 import { isNodeProcess } from '../utils/internal/isNodeProcess'
 
-const useFetch = isNodeProcess() ? require('node-fetch') : window.fetch
+const useFetch: (
+  input: RequestInfo,
+  init?: RequestInit,
+) => Promise<Response> = isNodeProcess() ? require('node-fetch') : window.fetch
 
 export const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
   const headers = new Headers(requestInit.headers)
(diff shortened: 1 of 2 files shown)


src/context/fetch.ts
```diff
@@ -2,7 +2,10 @@ import { Headers } from 'headers-utils'
 import { MockedRequest } from '../handlers/RequestHandler'
 import { isNodeProcess } from '../utils/internal/isNodeProcess'
 
-const useFetch = isNodeProcess() ? require('node-fetch') : window.fetch
+const useFetch: (
+  input: RequestInfo,
+  init?: RequestInit,
+) => Promise<Response> = isNodeProcess() ? require('node-fetch') : window.fetch
```
kettanaito, line 8: This cannot be injected at build time because `ctx.fetch` is _shared_ between the browser and NodeJS bundles.

PR #629, opened 2021-03-03
Sets "process.env.NODE_ENV" to "development" during the build

### LINKED ISSUE #609: worker.start() fails if there is no process or process.env
## Describe the bug
msw is not working with the latest version, it works if I revert back to 0.25.
apparently a process.env variable is needed, and I do not have one.

## Environment
- `msw: 0.26.2`

latest Chrome

## To Reproduce
```worker.start()``` fails with following error:
```
Uncaught ReferenceError: process is not defined
    at graphql-deps.js:790
    at browser.js:8
```
inspection shows the that `process` is required,  see this part of the code found in graphql-deps.js at line 790:
```
/**
 * A replacement for instanceof which includes an error warning when multi-realm
 * constructors are detected.
 */
// See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production
// See: https://webpack.js.org/guides/production/
var instanceOf = process.env.NODE_ENV === 'production' ? // istanbul ignore next (See: 'https://github.com/graphql/graphql-js/issues/2317')
// eslint-disable-next-line no-shadow
function instanceOf(value, constructor) {
  return value instanceof constructor;
} : // eslint-disable-next-line no-shadow
function instanceOf(value, constructor) {
  if (value instanceof constructor) {
    return true;
  }

  if (value) {
    var valueClass = value.constructor;
    var className = constructor.name;

    if (className && valueClass && valueClass.name === className) {
      throw new Error("Cannot use ".concat(className, " \"").concat(value, "\" from another module or realm.\n\nEnsure that there is only one instance of \"graphql\" in the node_modules\ndirectory. If different versions of \"graphql\" are the dependencies of other\nrelied on modules, use \"resolutions\" to ensure only one version is installed.\n\nhttps://yarnpkg.com/en/docs/selective-version-resolutions\n\nDuplicate \"graphql\" modules cannot be used at the same time since different\nversions may have different capabilities and behavior. The data from one\nversion used in the function from another could produce confusing and\nspurious results."));
    }
  }

  return false;
};
```

## Expected behavior
I would expect the mocking to work as it did with version 0.25
using a serviceworker on a web project, I would not expect to need process.env.

### PR DESCRIPTION
**Pull request by kettanaito**
## GitHub

- Fixes #609

## Changes

- The `process.env.NODE_ENV` expression is evaluated to "development" during all builds. This way the third-party code that relies on that expression knows that we're bundling for the development environment.

### DIFF
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -21,6 +21,10 @@ const plugins = [
     mainFields: ['module', 'main', 'jsnext:main', 'browser'],
     extensions: ['.js', '.jsx', '.ts', '.tsx'],
   }),
+  replace({
+    preventAssignment: true,
+    'process.env.NODE_ENV': JSON.stringify('development'),
+  }),
   integrityCheck({
     checksumPlaceholder: '<INTEGRITY_CHECKSUM>',
     input: SERVICE_WORKER_SOURCE_PATH,
--- a/src/handlers/RequestHandler.ts
+++ b/src/handlers/RequestHandler.ts
@@ -1,10 +1,5 @@
 import { Headers } from 'headers-utils'
-import {
-  ResponseTransformer,
-  MockedResponse,
-  response,
-  ResponseComposition,
-} from '../response'
+import { MockedResponse, response, ResponseComposition } from '../response'
 import { getCallFrame } from '../utils/internal/getCallFrame'
 import { status } from '../context/status'
 import { set } from '../context/set'

### CODE REVIEW

File: rollup.config.ts [hunk 1]
```diff
@@ -21,6 +21,10 @@ const plugins = [
     mainFields: ['module', 'main', 'jsnext:main', 'browser'],
     extensions: ['.js', '.jsx', '.ts', '.tsx'],
   }),
+  replace({
+    preventAssignment: true,
+    'process.env.NODE_ENV': JSON.stringify('development'),
+  }),
```
marcosvega91, line 27: If we are building why we use `development` as NODE_ENV and not `production` ?

kettanaito: I consider MSW to be a development tool. Since it never runs in production, developers should get the benefit of warnings from third-party packages visible only in development. In other words, "development" target is for "yes, please, I'm potentially debugging this code."

marcosvega91: ok fine, makes sense :)

marcosvega91 (approved):
niceeee, thanks for the explanation above :)

PR #645, opened 2021-03-12
fix: handle no active "client" in the worker script

### LINKED ISSUE #643: msw fails to find the correct client in iframe scenarios (like with Cypress)
## Describe the bug
This is a tricky one. I suppose the root error is failing to find the correct client that emitted a specific event, which then throws an error in `./mockServiceWorker.js#117` on `if (activeClientIds.has(client.id))` because `client` is undefined.

It only happens in Cypress tests, which changes how the app is rendered I suppose, and has to do with iframes I guess.

More info in screenshots.

## Environment

- `msw: 0.27.1`
- `nodejs: 14.13.0`
- `npm: 7.6.1`
- `cypress: 6.6.0`
> msw@0.26.2 does not have this issue

Chrome: `89.0.4389.82`
> but also happens on Electron and Firefox, doesn't seem to be browser related

## To Reproduce
1. Run `msw` on app start
2. Run Cypress to test app (or another way to put app into an iframe?)
3. Re-run tests until error (usually first or second is enough)

> I realise this is bad, please take a look at the screenshots, perhaps it will become clearer how to reproduce

## Expected behavior
`msw` can be used in Cypress/iframes without issues.

## Screenshots

It starts by intercepting this request, I don't know why because it's definitely not a path I mock.
I suppose it should simply pass through.

image

It then finds the client, but it's `nested` so on to trying to find the parent I suppose?

image

This is where it goes wrong. `allClients` has no id that matches one in `activeClientIds`.

image

Therefore `client` is now `undefined` here, and `client.id` throws.

image

I hope that helps, I'm sure it has to do with iframes but I don't know why the ids don't match.

Curiously, the first Cypress test always passes (ids match and all), but the second on always fail.

tget4:
This fixes: https://github.com/mswjs/msw/issues/643

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -114,7 +114,7 @@ async function handleRequest(event, requestId) {
   // Send back the response clone for the "response:*" life-cycle events.
   // Ensure MSW is active and ready to handle the message, otherwise
   // this message will pend indefinitely.
-  if (activeClientIds.has(client.id)) {
+  if (client && activeClientIds.has(client.id)) {
     ;(async function () {
       const clonedResponse = response.clone()
       sendToClient(client, {
--- a/test/msw-api/setup-worker/scenarios/iframe/iframe.test.ts
+++ b/test/msw-api/setup-worker/scenarios/iframe/iframe.test.ts
@@ -68,3 +68,45 @@ test('intercepts a request from a deeply nested iframe given MSW is registered i
 
   expect(firstName).toBe('John')
 })
+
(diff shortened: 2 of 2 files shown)


kettanaito (dismissed):
Thank you for fixing this, @tget4! Welcome to contributors 🎉

kettanaito:
I've added an integration test that asserts request interception given two _sibling_ iframes:

- One that dispatches a request.
- The second one that registers MSW.

That seems to pass even on the current `master`, so it's not the error scenario we are fixing here. Still need to find a way to reproduce the issue in a test.

PR #647, opened 2021-03-12
Fix IE11 string split issue with error stack

### LINKED ISSUE #646: MSW throws a string parse error when an undefined or null stack trace is encountered
## Describe the bug
In IE11 when a new Error is thrown with an `undefined` or a `null` `stack`, a parse error is thrown.

```
Queuing uncaught exception from window.onerror Unable to get property 'split' of undefined or null reference  349 3 TypeError: Unable to get property 'split' of undefined or null reference
SCRIPT5007: Unable to get property 'split' of undefined or null reference
getCallFrame-deps.js (349,3)
```

## Environment

- `msw: 0.26.2`
- `nodejs: 14.15.4`
- `npm: 6.14.10`

image
- `IE 11.15.16299.0`

## To Reproduce
1. Go to the console in IE11
2. Throw a new error and output the stack
3. Inline the MSW `getCallFrame()` `frame` logic
4. See the parse `TypeError`

image

## Expected behavior
MSW should gracefully handle an `undefined` or `null` stack by converting to an empty string.

seriouslysean:
Solves #646.

### DIFF
--- a/src/utils/internal/getCallFrame.test.ts
+++ b/src/utils/internal/getCallFrame.test.ts
@@ -169,14 +169,10 @@ test('Safari on macOS error stack', () => {
   )
 })
 
-test('handles the undefined stack trace', () => {
-  // version 1
+test('Handles the undefined stack trace', () => {
   mockStack.mockImplementationOnce(() => undefined)
-
   expect(() => getCallFrame()).not.toThrow(TypeError)
 
-  // version 2
   mockStack.mockImplementationOnce(() => null)
-
   expect(() => getCallFrame()).not.toThrow(TypeError)
 })
--- a/src/utils/internal/getCallFrame.test.ts
+++ b/src/utils/internal/getCallFrame.test.ts
@@ -169,7 +169,7 @@ test('Safari on macOS error stack', () => {
   )
 })
 
-test('IE11 undefined or null error stack', () => {
+test('handles the undefined stack trace', () => {
   // version 1
   mockStack.mockImplementationOnce(() => undefined)
 
--- a/src/utils/internal/getCallFrame.test.ts
+++ b/src/utils/internal/getCallFrame.test.ts
@@ -168,3 +168,15 @@ test('Safari on macOS error stack', () => {
     './test/msw-api/setup-worker/printHandlers.mocks.ts@http://localhost:56460/main.js:1376:5',
   )
 })
+
+test('IE11 undefined or null error stack', () => {
+  // version 1
+  mockStack.mockImplementationOnce(() => undefined)
+
+  expect(() => getCallFrame()).not.toThrow(TypeError)
+
+  // version 2
+  mockStack.mockImplementationOnce(() => null)
+
+  expect(() => getCallFrame()).not.toThrow(TypeError)
+})
--- a/src/utils/internal/getCallFrame.ts
+++ b/src/utils/internal/getCallFrame.ts
@@ -3,7 +3,8 @@
  */
 export function getCallFrame() {
   // In <IE11, new Error may return an undefined stack
-  const frames: string[] = ((new Error().stack || '') as string).split('\n')
+  const stack = (new Error().stack || '') as string
+  const frames: string[] = stack.split('\n')
 
   // Get the first frame that doesn't reference the library's internal trace.
   // Assume that frame is the invocation frame.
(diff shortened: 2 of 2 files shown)


src/utils/internal/getCallFrame.ts
```diff
@@ -2,7 +2,7 @@
  * Return the stack trace frame of a function's invocation.
  */
 export function getCallFrame() {
-  const frames: string[] = (new Error().stack as string).split('\n')
+  const frames: string[] = ((new Error().stack || '') as string).split('\n')
```
seriouslysean: Full disclosure, I haven't done much TS at all-- mostly only reading so if this needs to change or there is a better way to write it I'm all ears.

msutkowski: The only other option would be `??` it, but it doesn't really matter here and the `||` makes sense. I think for future readability, we should add a comment above this explaining why this is here (ex: < IE11 returns `undefined` instead of a string).

msutkowski (commented):
Wild catch! IE is the gift that keeps on giving :)

src/utils/internal/getCallFrame.ts
seriouslysean: Thanks @msutkowski, added in 1b3873afba671d3fe707f6e019264a1c6fa931f8.

src/utils/internal/getCallFrame.test.ts
```diff
@@ -168,3 +168,15 @@ test('Safari on macOS error stack', () => {
     './test/msw-api/setup-worker/printHandlers.mocks.ts@http://localhost:56460/main.js:1376:5',
   )
 })
+
+test('IE11 undefined or null error stack', () => {
```
kettanaito: Could you please rename this test title to something like

```suggestion
test('handles the undefined stack trace', () => {
```

src/utils/internal/getCallFrame.test.ts
```diff
@@ -168,3 +168,15 @@ test('Safari on macOS error stack', () => {
...
+  // version 1
+  mockStack.mockImplementationOnce(() => undefined)
+
+  expect(() => getCallFrame()).not.toThrow(TypeError)
+
+  // version 2
```
kettanaito: You can drop these comments, I think you're doing a great job with the test so the things are self-explanatory.

src/utils/internal/getCallFrame.ts
kettanaito: We should be careful about `??`, since that wouldn't handle `false`, which is still an unexpected value that mustn't be treated as an error's stack trace. I think `||` is the way to go here.

src/utils/internal/getCallFrame.ts
```diff
@@ -2,7 +2,8 @@
  * Return the stack trace frame of a function's invocation.
  */
 export function getCallFrame() {
-  const frames: string[] = (new Error().stack as string).split('\n')
+  // In <IE11, new Error may return an undefined stack
+  const frames: string[] = ((new Error().stack || '') as string).split('\n')
```
kettanaito: This line is getting quite complex, what do you think if we move it into its own variable?

```ts
const stack = (new Error().stack || '') as string
const frames: string[] = stack.split('\n')
```

It's subjective, but I find it to read much better.

kettanaito (changes requested):
Hey, @seriouslysean! Thank you so much for working on this improvement. I've left only a few nitpick comments to push this pull request from good to awesome. Looking forward to seeing this merged.

src/utils/internal/getCallFrame.test.ts
seriouslysean: Copy and paste was not my friend today, haha. 😬

src/utils/internal/getCallFrame.ts
seriouslysean: I actually had it this way first, adjusted in 895af0f05cbd9f2667ed02dc4622e026bd15299f.

seriouslysean:
Thanks @kettanaito, should be good to go now-- thanks for the quick feedback. I can tell that we're all SUPER passionate about IE11 support, haha. 😎

msutkowski (approved):
LGTM, thanks! :rocket:

kettanaito (approved):
Thank you for fixing this, @seriouslysean. Welcome to contributors! 🎉

Issue #641, opened 2021-03-11
TypeScript 3.9.9 can't type check lib/utils/internal/compose.d.ts


someden:
Same bug as was in #615, but with typescript version 3.9.9

## Environment

- `msw: 0.27.1`
- `nodejs: 12.18.3`
- `npm: 6.14.6`
- `typescript: 3.9.9`

## Steps to reproduce the behavior:
```bash
$ yarn add msw typescript@3
$ echo "import { setupWorker } from 'msw'" >> index.ts
$ yarn tsc --noEmit index.ts
```
Error:
```
node_modules/msw/lib/types/utils/internal/compose.d.ts:6:8 - error TS1005: ',' expected.
6     arg: any,
         ~

node_modules/msw/lib/types/utils/internal/compose.d.ts:7:12 - error TS1005: ',' expected.
7     ...rest: infer LastArg
             ~

Found 2 errors.
```

timdeschryver:
Hi, I don't think that this is the same issue.
I think the issue here is that we're using [variadic tuple types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types), which were introduced in TS 4.0.
See #491 and #512 for more info.

kettanaito:
Thanks for reporting this. 

MSW uses type definitions that rely on TypeScript >=4.0 features. Please upgrade your TypeScript versions to use the latest types of MSW. Thanks.

Issue #512, opened 2020-12-15
Typescript error: expecting "," in type PickLastInTuple (utils/internal/compose)


maurealtrends:
## Describe the bug
After installing and following the steps provided in the documentation, I tried to run my app (CRA) but a Typescript compilation error arise (see screenshot attached).

## Environment

- `msw: ^0.24.2`
- `nodejs: 15.3.0`
- `npm: 7.0.14`
- `typescript: ^3.9.5`
- `react: ^16.13.1"`
- `react-dom: ^16.13.1"`
- `react-scripts: 3.2.0`

## `tsconfig.json`
```json
{
  "compilerOptions": {
    "target": "es2016",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "noEmit": true,
    "jsx": "preserve"
  },
  "include": ["src"]
}
```

## To Reproduce
1. Follow the complete guide up until [Verify and Inspect](https://mswjs.io/docs/getting-started/integrate/browser#verify--inspect)
2. Run your CRA app
3. See error

## Expected behavior
I expected the app to build succesfully.

## Screenshots
![image](https://user-images.githubusercontent.com/47781631/102217006-bbc4ad00-3eba-11eb-9d73-5a758fe9b56c.png)

chrisguttandin:
Hi @maurealtrends, I believe this is a duplicate of #491. Is it possible for you to update TypeScript to v4?

maurealtrends:
> Hi @maurealtrends, I believe this is a duplicate of #491. Is it possible for you to update TypeScript to v4?

I'll try right now and get back to you as ASAP 😉

maurealtrends:
It worked!

Thank you very much and sorry for the inconvenience.

chrisguttandin:
No problem. I'm happy that we could solve it so quickly.

PR #658, opened 2021-03-19
change `DropFirstInTuple` type for backwards compatibility with older TS versions

phryneas:
Right now, just having msw installed in TS versions <= 3.8 will result in a parsing error, completely halting tsc:

[image](https://github.com/rtk-incubator/rtk-query/runs/2149226738?check_suite_focus=true)

This should solve that behaviour.

- Related to #512, #641

### DIFF
--- a/src/utils/internal/compose.ts
+++ b/src/utils/internal/compose.ts
@@ -4,10 +4,10 @@ type LengthOfTuple<Tuple extends any[]> = Tuple extends { length: infer L }
   ? L
   : never
 
-type DropFirstInTuple<Tuple extends any[]> = Tuple extends [
+type DropFirstInTuple<Tuple extends any[]> = ((...args: Tuple) => any) extends (
   arg: any,
   ...rest: infer LastArg
-]
+) => any
   ? LastArg
   : Tuple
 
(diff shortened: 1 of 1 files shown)


src/utils/internal/compose.ts
```diff
@@ -4,10 +4,10 @@ type LengthOfTuple<Tuple extends any[]> = Tuple extends { length: infer L }
...
   : never
 
-type DropFirstInTuple<Tuple extends any[]> = Tuple extends [
+type DropFirstInTuple<Tuple extends any[]> = ((...args: Tuple) => any) extends (
   arg: any,
   ...rest: infer LastArg
-]
+) => any
   ? LastArg
   : Tuple
```
phryneas, line 12: Also I think that this should be `[]` here but I just wanted to fix the parsing problem, this might change behaviour (but I think this is a bug right now)
```suggestion
  : []
```

src/utils/internal/compose.ts
```diff
@@ -34,7 +34,7 @@ export function compose<
 >(
   ...fns: Functions
 ): (
-  ...args: LeftReturnType extends never ? never[] : [LeftReturnType]
+  ...args: [LeftReturnType] extends [never] ? never[] : [LeftReturnType]
```
phryneas, line 37: This gave me problems in TS 3.7.7, boxing the type solves the problem.

kettanaito (approved):
This looks fantastic. Thank you for the improvement, @phryneas! Welcome to contributors 🎉

src/utils/internal/compose.ts
kettanaito: We can address that in a separate pull request. I'd like to know more what bug this causes at the moment.

PR #648, opened 2021-03-13
Updates "headers-utils" to 2.0

kettanaito:
- Includes the `Headers.forEach()` callback arguments order fix. 
- Ensures the `Headers` polyfill is compatible with the standard `window.Headers` class (adds iterators).

PR #660, opened 2021-03-23
Migrates to "@mswjs/interceptors"

kettanaito:
> BREAKING CHANGE (public).

## Changes

- Updates to @mswjs/interceptors.
- Updates to headers-utils v3.0.x

## Breaking changes

- Calling `setupServer` no longer enables the interception. Calling `server.listen()` does.
- The `response` instance in the life-cycle events now has `response.headers` as the `HeadersPolyfill` instance.

## GitHub

- Closes #648

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
   "sideEffects": false,
   "dependencies": {
     "@mswjs/cookies": "^0.1.4",
+    "@mswjs/interceptors": "^0.8.0",
     "@open-draft/until": "^1.0.3",
     "@types/cookie": "^0.4.0",
     "@types/inquirer": "^7.3.1",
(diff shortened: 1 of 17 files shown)


timdeschryver (approved):

patrickmcdougle-okta:
Should this package be externalized so it uses the dependency as listed? Right now, it bundles the installed version into this package, which doesn't allow the version numbers from incrementing independently. Each bug fix in interceptors will REQUIRE and version bump here, unless it's externalized in the build.

kettanaito:
That's a great point, @patrickmcdougle-okta. I believe that `@mswjs/interceptors` must be externalized so that the version installed on the users' machines is used with MSW (per semver). 

Please, would you be willing to submit a pull request with this change? This should be a matter of listing `@mswjs/interceptors` in the `external` property of our Rollup config.

patrickmcdougle-okta:
@kettanaito Yes, I can do that. Should it be externalized in ALL builds or just the esm build?

kettanaito:
@patrickmcdougle-okta, I've left the comment regarding your question in #913.

PR #662, opened 2021-03-24
fix: not print error when request is handled with no response

### LINKED ISSUE #655: getting an unhandled request error for requests that are handled but intended to be forwarded on
## Describe the bug

When using `server.listen({onUnhandledRequest: 'error'})`, I expect to see errors when there's an unhandled request, but not when the request is handled with a `return undefined`.

## Environment

- `msw: 0.27.1`
- `nodejs: 14`
- `npm: 6`

## To Reproduce

1. Clone https://github.com/kentcdodds/msw-handler-error-bug
2. Install
3. Run `node index.js`
4. See error when `mock = false` but no error when `mock = true`

## Expected behavior

There should be no error

## Screenshots

image

marcosvega91:
fix #655.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "test:integration": "node --max_old_space_size=8000 node_modules/jest/bin/jest.js --config=test/jest.config.js --runInBand",
     "test:smoke": "config/scripts/smoke.sh",
     "test:ts": "tsc -p test/typings/tsconfig.json",
-    "test:focused": "node node_modules/ts-node/dist/bin.js --project=test/tsconfig.json test/focusedTest.ts",
     "prepublishOnly": "yarn lint && yarn test:unit && yarn build && yarn test:integration",
     "postinstall": "node -e \"try{require('./config/scripts/postinstall')}catch(e){}\""
   },
--- a/src/utils/getResponse.ts
+++ b/src/utils/getResponse.ts
@@ -44,8 +44,12 @@ export const getResponse = async <
 
     const result = await handler.run(request)
 
-    if (result === null || !result.response || result.handler.shouldSkip) {
+    if (result === null) {
       return null
+    } else {
+      if (!result.response || result.handler.shouldSkip) {
+        return result
+      }
     }
 
     if (result.response.once) {
--- a/src/utils/getResponse.ts
+++ b/src/utils/getResponse.ts
@@ -1,5 +1,9 @@
 import { MockedResponse } from '../response'
-import { MockedRequest, RequestHandler } from '../handlers/RequestHandler'
+import {
+  MockedRequest,
+  RequestHandler,
+  RequestHandlerExecutionResult,
+} from '../handlers/RequestHandler'
 
 interface ResponsePayload {
   handler?: RequestHandler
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -18,7 +18,6 @@
     "test:integration": "node --max_old_space_size=8000 node_modules/jest/bin/jest.js --config=test/jest.config.js --runInBand",
     "test:smoke": "config/scripts/smoke.sh",
     "test:ts": "tsc -p test/typings/tsconfig.json",
-    "test:focused": "node node_modules/ts-node/dist/bin.js --project=test/tsconfig.json test/focusedTest.ts",
```
kettanaito, line 21: Deprecated in favor of the debug mode in `page-with`:

```sh
DEBUG=pageWith:* yarn test:integration path/to/suite.test.ts
```

```js
// path/to/suite.test.ts
it('some test', async () => {
  const context = await pageWith({...})
  await context.debug() // creates a breakpoint
})
```

test/msw-api/setup-worker/start/on-unhandled-request/error.mocks.ts
```diff
@@ -1,7 +1,10 @@
 import { setupWorker, rest } from 'msw'
 
 const worker = setupWorker(
-  rest.get('/user', (req, res, ctx) => {
+  rest.get('/user/:id', (req, res, ctx) => {
```
kettanaito: I think we should add a new handler just for this particular scenario. It's always better to keep them separated: easier to reason and refactor. 

```js
rest.get('/user') // leave the previous handler
rest.get('/new-name') // a new handler for this use case
```

src/utils/getResponse.ts
```diff
@@ -33,27 +49,7 @@ export const getResponse = async <
...
-      return null
-    }
-
-    if (result.response.once) {
-      handler.markAsSkipped(true)
-    }
-
-    return result
-  }, Promise.resolve(null))
+  const result = await getResult(relevantHandlers, request)
```
kettanaito: This is a great intention, but I would leave the previous logic as-is. This part has some history behind it, changing it may introduce regressions not related to the issue we are solving. 

Let's leave everything as-is, add a new failing test, and then address the issue. What do you think?

kettanaito (changes requested):
Hey, @marcosvega91! Thank you for the new test. I've left a few comments, could you please take a look at them when you have time?

marcosvega91:
I see that there is a test that does the opposite of mine. https://github.com/mswjs/msw/blob/655c93f83b7fcebe9a9093cacd7d891939028d29/test/msw-api/life-cycle-events/life-cycle-events.node.test.ts#L77-L90. It's the test that is failing

test/msw-api/setup-server/scenarios/on-unhandled-request/error.test.ts
```diff
@@ -34,3 +37,13 @@ test('errors on unhandled request when using the "error" value', async () => {
 If you still wish to intercept this unhandled request, please create a request handler for it.
 Read more: https://mswjs.io/docs/getting-started/mocks`)
 })
+
+test('not erroros on handled request with no response', async () => {
```
timdeschryver:
```suggestion
test('not errors on handled request with no response', async () => {
```

timdeschryver (dismissed):

kettanaito:
@marcosvega91, it looks like that test suite has a misleading title. The `request:unhandled` is emitted only when there is no handler associated with the request:

https://github.com/mswjs/msw/blob/655c93f83b7fcebe9a9093cacd7d891939028d29/src/node/createSetupServer.ts#L129-L142

The test is correct, its title is wrong. You can see from the file above that the `onUnhandledRequest` callback will be called when there is a handler but there was no response. _That_ is actually the cause of the issue. I think the callback should be moved into the previous `if (!handler)` block to behave as expected. 

I don't think that the existing test asserts the current issue, so your test suite is definitely something great to add.

marcosvega91:
Yes you're right but I think that the test had an extra check because it should emit the event only when there is a response right? btw I have updated the branch with the fix
```diff
--- a/test/msw-api/life-cycle-events/life-cycle-events.node.test.ts
+++ b/test/msw-api/life-cycle-events/life-cycle-events.node.test.ts
@@ -81,6 +81,7 @@ test('emits events for a handled request with no response', async () => {
 
   expect(listener.mock.calls).toEqual([
     [`[request:start] POST https://mswjs.io/no-response ${requestId}`],
-    [`[request:unhandled] POST https://mswjs.io/no-response ${requestId}`],
     [`[request:end] POST https://mswjs.io/no-response ${requestId}`],
     [`[response:bypass] Vercel ${requestId}`],
```

kettanaito:
>  That is actually the cause of the issue. I think the callback should be moved into the previous if (!handler) block to behave as expected.

I'm sorry, I misunderstood the issue. I wouldn't jump to that conclusion of mine so fast. Need to think more about what is the expected behavior here.

Right now any request that happens as-is, either if there's no handler or a handler that explicitly/implicitly returns nothing, is considered unhandled. In other words, any request that doesn't use a mocked response is unhandled.

PR #671, opened 2021-03-31
content-type header parsing: be more forgiving

matthewleon:
The space after the semicolon delimiter in HTTP header content is optional according to the standard, and a number of implementations will not add such whitespace.

I ran into this problem when mocking form data using the form-data lib (https://www.npmjs.com/package/form-data), which does not add the optional space when serializing headers.

Please let me know if it's worth adding a test _just_ to demonstrate the tolerance of "no-whitespace" headers. I figured that passing the current test should be enough.

### DIFF
--- a/src/utils/internal/parseMultipartData.test.ts
+++ b/src/utils/internal/parseMultipartData.test.ts
@@ -1,6 +1,24 @@
 import { parseMultipartData } from './parseMultipartData'
 
 test('parses a given valid multipart string', async () => {
+  expect.assertions(3)
+  await testMultipartDataWithContentType(
+    'multipart/form-data; boundary=WebKitFormBoundaryvZ1cVXWyK0ilQdab',
+  )
+})
+
+test('parses a given valid multipart string give non-pretty content-type', async () => {
+  expect.assertions(3)
+  // node-fetch will serialize content-type in this format, which is valid according to HTTP
+  // https://github.com/node-fetch/node-fetch/blob/d8fc32d6b29bd43d1ad377e80b3e439fe37f2904/test/main.js#L1438
+  await testMultipartDataWithContentType(
+    'multipart/form-data;boundary=WebKitFormBoundaryvZ1cVXWyK0ilQdab',
+  )
+})
+
+async function testMultipartDataWithContentType(
+  contentType: string,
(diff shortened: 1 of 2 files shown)


matthewleon:
@kettanaito this test failure appears to be non-deterministic. At least, I can't replicate locally. Running the integration tests on this branch results in no failures.

src/utils/internal/parseMultipartData.ts
```diff
@@ -53,7 +53,7 @@ export function parseMultipartData<T extends DefaultRequestMultipartBody>(
     return undefined
   }
 
-  const [, ...directives] = contentType.split('; ')
+  const [, ...directives] = contentType.split(/; */)
```
kettanaito, line 56: This is a great change! Let's add a test for handling such headers in `src/utils/internal/parseMultipartData.test.ts`. Please, would you be comfortable with adding such test?

kettanaito (commented):
Thank you for this improvement, @matthewleon! 
No worries about the failing test, it doesn't seem to be related to your changes. I've restarted the CI so it should pass now. 

I've left just one comment about the need to test this.

matthewleon:
@kettanaito This should be good to go.

src/utils/internal/parseMultipartData.test.ts
```diff
@@ -1,6 +1,24 @@
 import { parseMultipartData } from './parseMultipartData'
 
 test('parses a given valid multipart string', async () => {
+  expect.assertions(3)
+  await testMultipartDataWithContentType(
+    'multipart/form-data; boundary=WebKitFormBoundaryvZ1cVXWyK0ilQdab',
+  )
+})
+
+test('parses a given valid multipart string give non-pretty content-type', async () => {
```
kettanaito: Looks like we need one more scenario when there are multiple spaces between the content type and its additional information: `multipart/form-data;___boundary=xxx` (replace "_" by spaces, markdown trims that).

matthewleon: done

kettanaito (approved):
Thank you for shipping this improvement, @matthewleon!

kettanaito:
Welcome to contributors 🎉

PR #672, opened 2021-04-01
Consider request unhandled only when it has no handler

### LINKED ISSUE #655: getting an unhandled request error for requests that are handled but intended to be forwarded on

kettanaito:
## GitHub

- Fixes #655
- Closes #662

## Changes

- setupServer: Request is considered unhandled only when it has no corresponding request handler. 
- No matter what the corresponding request handler returns or not, the request is still considered handled.

### DIFF
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -103,16 +103,16 @@ export function createSetupServer(...interceptors: Interceptor[]) {
 
         if (!handler) {
           emitter.emit('request:unhandled', mockedRequest)
-        }
-
-        if (!response) {
-          emitter.emit('request:end', mockedRequest)
 
           onUnhandledRequest(
             mockedRequest,
             currentHandlers,
             resolvedOptions.onUnhandledRequest,
           )
+        }
+
+        if (!response) {
+          emitter.emit('request:end', mockedRequest)
           return
         }
 
--- a/src/utils/getResponse.ts
+++ b/src/utils/getResponse.ts
@@ -44,8 +44,17 @@ export const getResponse = async <
 
     const result = await handler.run(request)
 
-    if (result === null || !result.response || result.handler.shouldSkip) {
+    if (result === null || result.handler.shouldSkip) {
       return null
+    }
+
+    if (!result.response) {
+      return {
+        request: result.request,
+        handler: result.handler,
+        response: undefined,
+        parsedResult: result.parsedResult,
+      }
     }
 
     if (result.response.once) {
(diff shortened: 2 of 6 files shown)


src/node/createSetupServer.ts
```diff
@@ -103,16 +103,16 @@ export function createSetupServer(...interceptors: Interceptor[]) {
 
         if (!handler) {
           emitter.emit('request:unhandled', mockedRequest)
-        }
-
-        if (!response) {
-          emitter.emit('request:end', mockedRequest)
 
           onUnhandledRequest(
```
kettanaito, line 107: I don't like how the worker and the server handle the same request/response flow in separate places. This opens a room to bugs, just as this one. It would be good to align this post-resolve handling into some sort of reused chain, the same for the worker and the server.

kettanaito:
Hey, @marcosvega91 👋🏻 Would you have a minute to look at these changes? Vastly inspired by your work on #662, we are addressing the issue of that unhandled requests here.

src/utils/getResponse.ts
```diff
@@ -44,10 +44,19 @@ export const getResponse = async <
 
     const result = await handler.run(request)
 
-    if (result === null || !result.response || result.handler.shouldSkip) {
+    if (result === null || result.handler.shouldSkip) {
       return null
     }
```
marcosvega91, line 49: Hello, I don't know if it's really correct here.

I could have a handler with a `once` response that I think should be treated as handled, or not ?

marcosvega91 (changes requested):
Hello @kettanaito :) It sounds good but I have one doubt

src/utils/getResponse.ts
kettanaito: That's a valid concern. Note that `handler.shouldSkip` checks if the handler should be skipped, not that it's a one-time handler. If you have a `res.once` running _the first time_, `result.handler.shouldSkip` will be `false`. It's only later that it will be marked as skipped:

https://github.com/mswjs/msw/blob/12bdd274c2e283bdd16448813e58b510ec676ffc/src/utils/getResponse.ts#L51-L53

On the subsequent calls that one-time handler will have `result.handler.shouldSkip` as `true`.

marcosvega91: yes right, in case I have a skipped request and I don't have any other handler rather than the one skipped I will have an error of unhandled?

kettanaito: @marcosvega91 exactly. After `res.once` has executed once, it's ignored until you `.resetHandlers`, so it won't be taken into account when deciding if a request is handled or not.

Issue #474, opened 2020-11-24
Support concurrent runs in Node.js


kettanaito:
## What

I suggest for the `setupServer` (`node-request-interceptor`) API to support execution in a concurrent mode (i.e. multiple test suite at the same time). 

## Why

Concurrency is crucial in large projects, allowing for faster execution of tests.

## Current behavior

Currently `node-request-interceptor` patches native request issuing modules _per process_, which allows irrelevant mocking rules to leak to another test suites, as they all run in a single process.

## How

Look into how [`nock`](https://github.com/nock/nock) operates. It performs modules patching in a similar fashion, and I'd suspect it to support concurrent mode. We can learn how they did it and evaluate if we can adopt a similar solution.

marcosvega91:
`concurrency` problem should happen only with multiple instances of `setupServer`. Under the hood when `setupServer` is called `http`, `https` and `XMLHttpRequest` are patched to start intercepting requests.
So if I use a single instance across my application I'll not encourage any problem.

If I use multiple instances of `setupServer` could be possible that a test restores the patched modules before other tests, causing the issue.

It's very strange that nock is not having that problem, their [restore function](https://github.com/nock/nock/blob/740b862305881b2b6f05b7d9a8d8ec3c88376204/lib/recorder.js#L366) is very simple and don't have any kind of check 🤔

kettanaito:
Thank you for the insights, @marcosvega91.

I suppose we can dedupe modules patching in `node-request-interceptor`, but that won't solve the issue of unrelated modules (i.e. test suite) being affected by the interceptors declared in other modules, as they still share the same process. 

What we can investigate is how exactly modern test frameworks execute tests in parallel. For instance, do they provision forked processes? If so, perhaps we could bound request interception to the forked process instead of the parent process.

marcosvega91:
I have tried to investigate today a little bit. it seems that `nock` is not having this problem simply resetting modules between tests. It is strange because patching `native` modules are shared between tests. 
I have tried also to run jest printing memory usage and it seems that we have some memory leak in `node-request-interceptor`. 

![image](https://user-images.githubusercontent.com/5365582/100780700-54aeef00-340a-11eb-9a41-3f9f06742144.png)

kettanaito:
[...]

PR #673, opened 2021-04-01
Enables parallel test runs

kettanaito:
## Changes

- Removes `--runInBand` flag from unit and integration test commands.
- Removes the `--max_old_space_size=8000` cap on the integration tests (previously added due to spawning WDS on each test and ts compiler causing a memory heap).

## Motivation

Current CI takes a lot of time to run, decreasing the iteration speed for the changes. Integration tests alone take about 6m 30s to run, not taking the workspace attachment into account ([ref](https://app.circleci.com/pipelines/github/mswjs/msw/1508/workflows/72fbec38-2574-4382-b94c-0788eba9bdf5/jobs/4578)). 

## History

The `--runInBand` flag was added to prevent some leaking state between tests. It might have been related to the old testing setup that we don't have anymore, or to #474. I didn't see any problems running parallel tests locally. 

## Benchmarks

- [Before](https://app.circleci.com/pipelines/github/mswjs/msw/1508/workflows/72fbec38-2574-4382-b94c-0788eba9bdf5/jobs/4578)
- After (see CI for the latest commit)


kettanaito:
CI fails spectacularly, reaching out of memory state, and taking a tremendous amount of time to run.

fnmunhoz:
Hey @kettanaito I have experienced this problem before and I'm almost sure it's related to that:

https://discuss.circleci.com/t/memory-problems-with-jest-and-workers/10297/4

The CircleCI job exposes all the VM CPUs to Jest, but it actually can only use 2 in this case, which is the number of available CPUs to the job (via docker).

I would suggest playing with [this option](https://jestjs.io/docs/cli#--maxworkersnumstring), starting with 2: `--maxWorkers=2`.

Jest recommends one worker per CPU core, minus 1 for the main thread, but it might be possible to spawn more in CircleCI, until the tests run consistently.

Hope it helps :)

msutkowski:
Yup, circleci isn't very friendly with Jest depending on the instance size. We might want to consider switching CI to github actions for fun :)

PR #684, opened 2021-04-07
fix: empty request parameters to be object

### LINKED ISSUE #682: `req.params` undocumented breaking change + typings not matching: Regression 0.26.x-0.27.x
Kind of related to https://github.com/mswjs/msw/issues/639 but this is an actual bug in the code. 

## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.28.1   |
| browser | Chrome 88     |
| OS      | win10     |

## Request handlers

[Handler from MSW examples](https://github.com/mswjs/examples/blob/master/examples/rest-react/src/mocks/handlers.js) with `req.params` logging.

```js
import { rest } from "msw";

export const handlers = [
    rest.post("/login", (req, res, ctx) => {
        console.log("req.params", req.params);
        const { username } = req.body;

        return res(
            ctx.json({
                id: "f79e82e8-c34a-4dc7-a49e-9fadc0979fda",
                username,
                firstName: "John",
                lastName: "Maverick",
            }),
        );
    }),
];
```

## Actual request

https://github.com/mswjs/examples/blob/83603a4619abb09447322d88bb88a7e4d6a2c91c/examples/rest-react/src/LoginForm.js#L18-L36

```js
fetch("/login", {
    method: "POST",
    body: JSON.stringify({ username })
});
```

## Current behavior

```js
export const handlers = [
    rest.post("/login", (req, res, ctx) => {
        console.log("req.params", req.params);
                ^^^  req.params {}   // msw@0.26.2
                ^^^  req.params null // msw@0.27.2
                ^^^  req.params null // msw@0.28.1
```

## Expected behavior

`req.params` should never be `null`. 

https://github.com/mswjs/msw/blob/443ce4535a61efb01a9f7990ebfc230a2ba59541/src/handlers/RestHandler.ts#L82

I have a custom request handler which is utilized across multiple projects. This handler is accessing `req.params` with `Object.entries` method. When no params are present the call will throw with `msw@0.27.x`. With `msw@0.26.x` the `req.params` always had value. 

If `null` is expected behaviour (👎 ) the typings should be fixed:

```diff
- params: ParamsType
+ params?: ParamsType
```

If `req.params` should always hold a value (👍 ) the code should be fixed.

AriPerkkio:
Fixes https://github.com/mswjs/msw/issues/682.

### What

When matched URL does not contain path parameters the `req.params` is `null`. It should always hold a value.

### How

Applied https://github.com/mswjs/node-match-path/pull/39 locally and fixed errors reported by typescript. 

### Testing

- Added a simple unit test to validate `req.params` is an empty object. 
- I don't think integration test is required for this but I'm ready to add one if wanted. 
- Manually verified the fix by applying it locally with repro of the bug issue.

### DIFF
--- a/src/handlers/RestHandler.test.ts
+++ b/src/handlers/RestHandler.test.ts
@@ -163,4 +163,16 @@ describe('run', () => {
 
     expect(result).toBeNull()
   })
+
+  test('request params defaults to empty object', async () => {
+    const handler = new RestHandler('GET', '/users', (req, res, ctx) => {
+      res(ctx.status(200))
+    })
+    const request = createMockedRequest({
+      url: new URL('/users', location.href),
+    })
+    const result = await handler.run(request)
+
+    expect(result?.request.params).toEqual({})
+  })
 })
--- a/src/handlers/RestHandler.test.ts
+++ b/src/handlers/RestHandler.test.ts
@@ -164,14 +164,13 @@ describe('run', () => {
     expect(result).toBeNull()
   })
 
-  test('request params defaults to empty object', async () => {
-    const handler = new RestHandler('GET', '/users', (req, res, ctx) => {
-      res(ctx.status(200))
-    })
-    const request = createMockedRequest({
-      url: new URL('/users', location.href),
-    })
-    const result = await handler.run(request)
+  test('returns an empty object as "req.params" given request with no URL parameters', async () => {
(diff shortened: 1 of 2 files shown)


kettanaito (approved):
Thank you for fixing this, @AriPerkkio! Welcome to contributors 🎉

PR #688, opened 2021-04-12
Fallback mode

### LINKED ISSUE #663: RFC: Fallback mode
## Description

_Fallback mode_ is an alternative request interception mechanism applied in a browser. Fallback mode is automatically enabled by the library when either of the following conditions is met:

- The browser doesn't support the Service Worker API;
- The user has `navigator.cookieEnabled` set to `false` that forbids access to the worker;
- The application is loaded using the `file://` protocol.

## Motivation

Fallback mode is aimed at improving the browser support and enabling the library in static builds of various tools (i.e. Storybook).

## Statements

1. Fallback mode is enabled automatically by the library.
1. There is no way to opt-out of the fallback mode.
1. There is no way to enable the fallback mode when the worker can be used instead.
1. Requests intercepted via the fallback mode do not appear in the "Network" tab and don't actually happen (see technical details).

## Technical details

Fallback mode is powered by the recent addition of the [fetch and XMLHttpRequest](https://github.com/mswjs/interceptors/pull/93) interceptors to the `@mswjs/interceptors` library. Under the hood, fallback mode patches `window.fetch` and `window.XMLHttpRequest` to provision the interception of requests. 

> Fallback mode is similar to the way most other API mocking solutions work: patching the in-browser request modules. In Mock Service Worker fallback mode will only affect the library's behavior in the environments where Service Worker API cannot be otherwise used.

### Pre-requisites 

- https://github.com/mswjs/interceptors/pull/93
- https://github.com/mswjs/msw/pull/660

## Public API

There are no changes to the public API related to this feature. Fallback mode is enabled automatically, when necessary.

The only user-facing change would be the change of the library's activation message:

```
[MSW] Mocking enabled (fallback-mode).
```

In addition to the activation message change, the library should print the warning message indicating that the worker couldn't be registered in the current environment:

```
[MSW] Cannot register a Service Worker in the current environment. Using the fallback mode instead.

Read more about the fallback mode: <DOCS_LINK>
```

### LINKED ISSUE #689: Unify response retrieval and the related event handling
## GitHub

- Originates from https://github.com/mswjs/msw/pull/672

## What

I suggest abstracting the response retrieval and all the related logic into a separate function. Reuse that function later for both browser and server API. 

## Why

Currently, the browser and server API reuse the `getResponse()` to retrieve a mocked response from the list of request handlers. However, all the surrounding logic (i.e. life-cycle events, `onUnhandledRequest`) is _duplicated_ in both APIs, opening possibilities to bugs and misaligned behavior.

kettanaito:
## Changes

- Unifies response retrieval by introducing a new `handleRequest` function that encapsulates:
  - Response lookup via `getResponse`.
  - Life-cycle events emitting in the proper phases.
  - Unhandled request callback invocation.
- Removes mocked response headers serialization into a `HeadersList`. Instead, serializes mock response headers into a `FlatHeadersObject` so that their shape is the same across regular `start` (where serialization is necessary to transmit the headers to the worker) and the fallback `start` (where serialization is only necessary to respond properly). 
- Adds support for the "fallback mode"—alternative interception mechanism that's automatically enabled in _browser_ environments when Service Worker API isn't supported (legacy browsers) or cannot run (static local builds).
- Splits the Rollup configuration so each build target uses its own set of plugins. This way the browser bundles are configured properly (see the note below). 

## Motivation

- Add support for the browsers that don't support Service Workers.
- Allows MSW execution in statically built apps (i.e. static Storybook build), where running a Service Worker from a `file://` is forbidden. 

## GitHub

- Closes #663 
- Closes #688 
- Closes #689 

## Roadmap

- [x] Find a way to enable and control the `interceptor` when calling `worker.start()` and `worker.stop()` without disrupting the public API while keeping the internals clean. 
- [x] Bundle the `debug` transient dependency of `interceptors` property (currently pulls in `util` and `tty` dependencies being bundled for UMD and ESM targets).


src/utils/handleRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+export async function handleRequest<
+  ResponseType extends Record<string, any> = MockedResponse<string>
+>(
+  request: MockedRequest,
+  handlers: RequestHandler[],
+  options: SharedOptions,
+  emitter: StrictEventEmitter<ServerLifecycleEventsMap>,
+  handleRequestOptions?: HandleRequestOptions<ResponseType>,
+): Promise<ResponseType | undefined> {
+  emitter.emit('request:start', request)
```
kettanaito, line 49: Request cookies _must not_ be set here, as they are already set in `parseIsomorpchicRequest`.

If cookies are set here once more, they will get overridden in Node.js, as request cookies in Node.js are resolved to `{}` (no location to check if the cookies should be inherited).

src/utils/request/parseIsomorphicRequest.ts
```diff
@@ -0,0 +1,58 @@
...
+  setRequestCookies(mockedRequest)
+
+  const requestCookies = requestCookiesString
+    ? cookieUtils.parse(requestCookiesString)
+    : {}
+
+  // Merge both direct request cookies and the cookies inherited
+  // from other same-origin requests in the cookie store.
+  mockedRequest.cookies = {
+    ...mockedRequest.cookies,
```
kettanaito, line 53: Surprisingly, there was no test that checks if the request cookies are properly set given the combination of its own cookies and those inherited from the cookie store. Previously, such combination resulted in the own request cookies being overridden, with the store cookies taking priority. 

Here, I swap the cookies priority, with the request's own cookies potentially overriding the store (inherited) cookies, as that seems like something you'd expect.

src/utils/worker/createRequestListener.ts
```diff
@@ -31,71 +29,44 @@ export const createRequestListener = (
...
+              headers: headersToList(response.headers),
+            }
+          },
+          onBypassResponse() {
+            return channel.send({
+              type: 'MOCK_NOT_FOUND',
+            })
+          },
+          onMockedResponse(response) {
+            channel.send({
```
kettanaito, line 49: It's important to signal the worker _ignoring the response's delay_, as the worker resolves the response respecting that delay. Otherwise, the delay would get doubled.

test/msw-api/setup-server/scenarios/cookies-request.test.ts
```diff
@@ -1,37 +1,51 @@
 import * as https from 'https'
 import { rest } from 'msw'
-import { setupServer } from 'msw/node'
+import { setupServer, SetupServerApi } from 'msw/node'
+import { ServerApi, createServer } from '@open-draft/test-server'
```
kettanaito, line 4: I'm using this pull request as an opportunity to remove external HTTP servers dependency from tests and replace it with the locally running server that represents a production one (which mustn't be hit in most test cases).

kettanaito:
## Bundling `debug`

The `debug` dependency has a dynamic import that loads either Browser and Node versions of the library on runtime:

https://github.com/visionmedia/debug/blob/d177f2bc36d3b8b8e9b1b006727ef5e04f98eac7/src/index.js#L6-L10

When being bundled, though, it loaded the Node version all the time, resulting in broken browser builds. 

### Solution

- Configure browser builds (UMD, IIFE) to prioritize the "browser" field in `mainFields` option (decides which dependency entries are loaded first).
- Set `browser: true` and `preferBuiltins: false` for the browser builds.

## Links

- https://github.com/visionmedia/debug/issues/468

PR #693, opened 2021-04-13
Move lint ignore comments to top of file

dallonf:
At my company, there are lint rules for JSDoc formatted comments. In this particular case, they're requesting a blank line somewhere in the description. Since it seems like the intent is for this file to be completely ignored by linters, I moved the linter disable comments to the very top of the file so that the whole thing will be ignored.

### DIFF
--- a/test/msw-api/cli/init.test.ts
+++ b/test/msw-api/cli/init.test.ts
@@ -157,14 +157,11 @@ test('does not cause ESLint errors or warnings', async () => {
     exec(`node cli/index.js init ${getPath('public')} --no-save`),
   )
   expect(initStderr).toBe('')
-  expect(fs.existsSync(getPath('public/mockServiceWorker.js'))).toBe(true)
-
-  console.log(fs.readFileSync(getPath('public/mockServiceWorker.js'), 'utf-8'))
 
   const {
     stdout: eslintStdout,
     stderr: eslintStderr,
-  } = await promisifyChildProcess(exec(`npx --no-install eslint ${getPath()}`))
+  } = await promisifyChildProcess(exec(`node_modules/.bin/eslint ${getPath()}`))
   expect(eslintStdout).toBe('')
   expect(eslintStderr).toBe('')
 
--- a/test/msw-api/cli/init.test.ts
+++ b/test/msw-api/cli/init.test.ts
@@ -133,3 +133,40 @@ test('warns when current public directory does not match the saved directory fro
 
   await cleanup()
 })
+
+test('does not cause ESLint errors or warnings', async () => {
+  const { prepare, getPath, cleanup } = createTeardown(
+    './tmp/cli/init/eslint',
+    addFile('package.json', {
+      name: 'package-name',
+    }),
+    // intentionally degenerate config to cause a warning in the initial comment
+    addFile('.eslintrc.json', {
+      rules: {
+        'no-warning-comments': [
+          'warn',
+          { terms: ['mock'], location: 'anywhere' },
+        ],
+      },
+    }),
+    addDirectory('public'),
+  )
+  await prepare()
+
+  const { stderr: initStderr } = await promisifyChildProcess(
+    exec(`node cli/index.js init ${getPath('public')} --no-save`),
+  )
+  expect(initStderr).toBe('')
+  expect(fs.existsSync(getPath('public/mockServiceWorker.js'))).toBe(true)
+
+  console.log(fs.readFileSync(getPath('public/mockServiceWorker.js'), 'utf-8'))
+
+  const {
+    stdout: eslintStdout,
+    stderr: eslintStderr,
+  } = await promisifyChildProcess(exec(`npx --no-install eslint ${getPath()}`))
+  expect(eslintStdout).toBe('')
+  expect(eslintStderr).toBe('')
+
+  await cleanup()
+})
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -1,11 +1,12 @@
+/* eslint-disable */
+/* tslint:disable */
+
 /**
  * Mock Service Worker.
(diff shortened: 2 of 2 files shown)


kettanaito:
Thank you for the improvement, @dallonf! 

This looks great. I'd like to take this opportunity to add an automated test that checks the `mockServiceWorker.js` being properly ignored by ESLint. It can be a matter of running ESLint in Node.js, providing it with the built worker script, applying some linting rules that intentionally contradict the worker script's contents, and asserting that there are no warnings/errors from the linter. 

Please, if that sounds interesting to you, please commit such a test. I'd be glad to help you in any way necessary to make it concise and useful. If not, I will try adding it sometime this/next week. Thank you for your contribution!

dallonf:
I'll see what I can do over the next couple of days!

dallonf:
Let me know how that looks 😁

test/msw-api/cli/init.test.ts
```diff
@@ -133,3 +133,40 @@ test('warns when current public directory does not match the saved directory fro
...
+    }),
+    addDirectory('public'),
+  )
+  await prepare()
+
+  const { stderr: initStderr } = await promisifyChildProcess(
+    exec(`node cli/index.js init ${getPath('public')} --no-save`),
+  )
+  expect(initStderr).toBe('')
+  expect(fs.existsSync(getPath('public/mockServiceWorker.js'))).toBe(true)
```
kettanaito: Checking this isn't part of this test's intention. I think we can remove this assertion.

test/msw-api/cli/init.test.ts
```diff
@@ -133,3 +133,40 @@ test('warns when current public directory does not match the saved directory fro
...
+
+  console.log(fs.readFileSync(getPath('public/mockServiceWorker.js'), 'utf-8'))
+
+  const {
+    stdout: eslintStdout,
+    stderr: eslintStderr,
+  } = await promisifyChildProcess(exec(`npx --no-install eslint ${getPath()}`))
```
kettanaito: We already have `eslint` installed, so perhaps let's use the locally installed version instead of pulling it from npm?

dallonf: That's exactly what `npx --no-install` does. I suppose I could try to pull it from `node_modules/.bin` for more clarity, though

test/msw-api/cli/init.test.ts
```diff
@@ -133,3 +133,37 @@ test('warns when current public directory does not match the saved directory fro
...
   await cleanup()
 })
+
+test('does not cause ESLint errors or warnings', async () => {
+  const { prepare, getPath, cleanup } = createTeardown(
+    './tmp/cli/init/eslint',
+    addFile('package.json', {
+      name: 'package-name',
+    }),
+    // intentionally degenerate config to cause a warning in the initial comment
```
kettanaito, line 143: "intentionally degenerate" love this 😆

test/msw-api/cli/init.test.ts
```diff
@@ -133,3 +133,37 @@ test('warns when current public directory does not match the saved directory fro
...
+    addFile('.eslintrc.json', {
+      rules: {
+        'no-warning-comments': [
+          'warn',
+          { terms: ['mock'], location: 'anywhere' },
```
kettanaito, line 148: So this triggers when a comment block has a "mock" phrase?

I was thinking maybe we can design these contradictory rules to stay agnostic (as possible) of the script's content? For example, forbidding the usage of `self`, which we cannot drop in a worker script.

test/msw-api/cli/init.test.ts
```diff
@@ -133,3 +133,37 @@ test('warns when current public directory does not match the saved directory fro
...
+  const { prepare, getPath, cleanup } = createTeardown(
+    './tmp/cli/init/eslint',
+    addFile('package.json', {
+      name: 'package-name',
+    }),
+    // intentionally degenerate config to cause a warning in the initial comment
+    addFile('.eslintrc.json', {
+      rules: {
+        'no-warning-comments': [
+          'warn',
```
kettanaito, line 147: Do you see any implications of choosing "error" instead?

test/msw-api/cli/init.test.ts
dallonf: Depends on how thoroughly you want to test _this_ change of moving the `/* eslint-ignore */` line to the top; if our contradictory rules don't target the comment at all, this test will pass even without that change!

@kettanaito how do you feel about that?

test/msw-api/cli/init.test.ts
dallonf: oh... actually it turns out "degenerate" doesn't mean exactly what I thought it did, heh :) "antagonistic" and/or "contrived" is probably more what I'm looking for

test/msw-api/cli/init.test.ts
dallonf: The main thing I can think of is that warnings are considered less severe than errors by ESLint and are sometimes ignored or suppressed in reports. By using an error in our test case, the test itself could develop a bug where warnings are suppressed, causing the test to pass even though there is a warning that would be seen by an end user.

test/msw-api/cli/init.test.ts
kettanaito: You make a point. Let's keep the rule as-is now, as it indeed targets the introduced change closely. We can always iterate on this in the future if we see any problems (although it's unlikely).

test/msw-api/cli/init.test.ts
kettanaito: All words are fine!

kettanaito (approved):
Thank you for adding this, @dallonf! A great change.

kettanaito:
Welcome to contributors 🎉

PR #699, opened 2021-04-17
feat: create directory in cli init if it doesn't exist

tbergquist-godaddy:
---
name: Feature request
about: create directory in cli init if it doesn't exist
title: ''
labels: feature
assignees: ''

---

If you try to run `npx msw init public` the command will fail. It would be better DX to try to create it. 

Create the directory if possible. 

**Additional context**


test/msw-api/cli/init.test.ts
```diff
@@ -44,18 +45,39 @@ test('copies the worker script into an existing directory without errors', async
...
+  await cleanup()
+})
+
+test('throws an exception and does not copy the worker script if creating the non-existing directory fails', async () => {
   const { prepare, getPath, cleanup } = createTeardown(
     'tmp/cli/init/non-existing',
   )
   await prepare()
 
+  fs.chmodSync('tmp/cli/init', '0444')
```
tbergquist-godaddy: This worked to make the mkdirSync fail locally, but it doesn't seem to work on the CI. 

Is there another/better way to test this?

cli/init.js
```diff
@@ -18,11 +18,18 @@ module.exports = function init(args) {
...
 
-  invariant(
-    dirExists,
-    'Failed to create a Service Worker at "%s": directory does not exist.\nMake sure to include a relative path to the root directory of your server.',
-    absolutePublicDir,
-  )
+  if (!dirExists) {
+    // Try to create the directory if it doesn't exist
+    try {
+      fs.mkdirSync(absolutePublicDir, { recursive: true })
```
kettanaito: I suggest we use its Promise API and wrap it in `until` for better error handling.

```js
const { until } = require('@open-draft/until')

module.exports = async function init(args) {
  const [createDirectoryError] = await until(() => fs.promises.mkdir(...))

  invariant(createDirectoryError, '...')
}
```

test/msw-api/cli/init.test.ts
kettanaito: We can step one level higher in the execution stack and mock `fs.promises.mkdir` to always throw in this particular test. That will reject the promise, which would happen in case creating a directory fails (no matter the reason).

```js
import * as fs from 'fs'

afterEach(() => {
  jest.restoreAllMocks()
})

it('throws an exception when creating a worker directory failed', async () => {
  jest.spyOn(fs.promises, 'mkdir').mockRejectedValue(new Error('Original error message')

  // the rest of the test...
})
```

We should also add the original promise rejection error to our own error message. This provides more useful info to developers to debug possible issues.

cli/init.js
```diff
@@ -18,11 +18,18 @@ module.exports = function init(args) {
...
+    } catch {
+      invariant(
+        false,
+        'Failed to create a Service Worker at "%s": directory does not exist and could not be created.\nMake sure to include a relative path to the root directory of your server.',
```
kettanaito: I think we should include the original error message in our custom one to contain more information when debugging any related issues.

```js
invariant(..., `\
Failed to create ...

See the original error below:
${error}\
`)
```

kettanaito (changes requested):
Hey, @tbergq! Thank you for this improvement, I think it'd lead to a better DX. I've left a few comments to address, please let me know if you have any questions.

test/msw-api/cli/init.test.ts
tbergquist-godaddy: > We can step one level higher in the execution stack and mock fs.promises.mkdir to always throw in this particular test.

That was my first idea as well, but since we are starting the CLI in a child-process the fs that you mock is not the same one as the one running in that child-process. 

> We should also add the original promise rejection error to our own error message. This provides more useful info to developers to debug possible issues.

Yeah, good idea, I will add this 👍

tbergquist-godaddy:
@kettanaito, thanks for your feedback, I will try to address your comments sometime this week!

kettanaito:
Thank you for your work on this, @tbergq! Looking forward to reviewing this again.

tbergquist-godaddy:
I am still having trouble with the test for failing to create the directory. As mentioned above, mocking fs.promises.mkdir doesn't work since it runs in another process, and the chmod doesn't seem to work on the CI, I am guessing since it runs as the root user and is not affected by setting the directory as read-only.

cli/init.js
```diff
@@ -18,11 +20,18 @@ module.exports = function init(args) {
...
-    absolutePublicDir,
-  )
+  if (!dirExists) {
+    // Try to create the directory if it doesn't exist
+    const [createDirectoryError] = await until(() =>
+      fs.promises.mkdir(absolutePublicDir, { recursive: true }),
+    )
+    invariant(
+      createDirectoryError == null,
+      `Failed to create a Service Worker at "%s": directory does not exist and could not be created.${EOL}Make sure to include a relative path to the root directory of your server.${EOL}${EOL}See the original error below:${EOL}%s`,
```
kettanaito: Why prefer `EOL` over `\n`?

tbergquist-godaddy: Because it might differ on windows vs linux/mac
Screenshot 2021-04-27 at 14 29 15

kettanaito: I think it's a good change, but it doesn't belong to this pull request's scope. 

What do you think if we use `\n` here, as we do across the codebase, and fix them _all_ in another pull request? That way changes are scoped, clear, and revertable.

test/msw-api/cli/init.test.ts
```diff
@@ -1,5 +1,6 @@
 import * as fs from 'fs-extra'
 import { exec } from 'child_process'
+
```
kettanaito: Redundant new line?

cli/init.js
tbergquist-godaddy: Sure, I'll put it back

test/msw-api/cli/init.test.ts
```diff
@@ -44,18 +45,45 @@ test('copies the worker script into an existing directory without errors', async
...
 
-test('throws an exception and does not copy the worker script given a non-existing directory', async () => {
+test('creates the directory if it does not exist', async () => {
+  const { prepare, getPath, cleanup } = createTeardown(
+    'tmp/cli/init/non-existing',
+  )
+  await prepare()
+
+  const { stderr } = await promisifyChildProcess(
+    exec(`node cli/index.js init ${getPath('non-existing-public')} --no-save`),
```
kettanaito: I suggest we provide a nested non-existing directory path so we also assert that nested directories are created.
```suggestion
    exec(`node cli/index.js init ${getPath('public/nested/path')} --no-save`),
```

kettanaito (changes requested):
Thank you for addressing the feedback, @tbergq! I've left a comment on a few things, could you please let me know what you think about them?

test/msw-api/cli/init.test.ts
tbergquist-godaddy: Yeah, sorry, it was leftover from trying to mock fs I guess, I'll remove it

test/msw-api/cli/init.test.ts
tbergquist-godaddy: Changed to `public/nested/path`

test/msw-api/cli/init.test.ts
tbergquist-godaddy: Removed

tbergquist-godaddy:
@kettanaito, all comments should be attended to, and I rewrote the failing test to a unit test instead. 

Please have another look.

kettanaito (approved):
Thank you for the great work on this, @tbergq! Welcome to contributors! 🎉

PR #703, opened 2021-04-20
Sets "warn" as the default "onUnhandledRequest" strategy

### LINKED ISSUE #453: Add a debugging request handler

One of the most common issues developers experience is unmatched requests.

I suggest to add a debugger request handler (i.e. `log`) that would act as a transparent handler, meaning it would log out any requests that didn't match preceding handlers.

We usually suggest to do that manually, but neither of the existing handlers can be fully used as a transparent request handler, as they have a predicate built-in which limits requests captured to a certain degree.

kettanaito:
## GitHub

- Closes #453

## Changes

- The default value for `onUnhandledRequest` is now `"warn"`.
- Adds explicit `default` and `bypass` integration tests so there's a clear test suite for each respective behavior.
- Console calls in the tests are stubbed to be silent with `.mockImplementation()`. Removes noise from the test runner output.

## Motivation

Around 80% of issues are caused by request URL mismatch. With "bypass" being the default option for `onUnhandledRequest`, those mismatches never surface to the end-developer. While this option may be slightly intrusive, there are two strong arguments to adopt it:

1. You can always set `onUnhandledRequest: "bypass"` to disable warnings. This is useful for mock-first development. I also like the explicitness of this action, as you're being aware that now all unhandled requests going to go silently.
2. The value you get from unhandled requests warnings is far greater than the console noise it may create.

### DIFF
--- a/src/utils/request/onUnhandledRequest.ts
+++ b/src/utils/request/onUnhandledRequest.ts
@@ -119,7 +119,7 @@ ${handlers.map((handler) => `  • ${handler.info.header}`).join('\n')}`
 export function onUnhandledRequest(
   request: MockedRequest,
   handlers: RequestHandler[],
-  strategy: UnhandledRequestStrategy = 'bypass',
+  strategy: UnhandledRequestStrategy = 'warn',
 ): void {
   if (typeof strategy === 'function') {
     strategy(request)
@@ -172,7 +172,12 @@ Read more: https://mswjs.io/docs/getting-started/mocks\
       break
     }
 
+    case 'bypass':
+      break
+
     default:
-      return
+      throw new Error(
+        `[MSW] Failed to react to an unhandled request: unknown strategy "${strategy}". Please provide one of the supported strategies ("bypass", "warn", "error") or a custom callback function.`,
+      )
   }
 }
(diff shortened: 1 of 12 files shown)


src/utils/request/onUnhandledRequest.ts
```diff
@@ -172,7 +172,10 @@ Read more: https://mswjs.io/docs/getting-started/mocks\
       break
     }
 
+    case 'bypass':
+      break
+
     default:
-      return
+      throw new Error(``)
```
timdeschryver: I think providing an error message here can be useful for future use.

```suggestion
      throw new Error(``)
```

msutkowski (dismissed):

src/utils/request/onUnhandledRequest.ts
kettanaito: Great catch. Will provide a meaningful error message when given an unknown strategy.

msutkowski (dismissed):

timdeschryver (approved):

LeBenLeBen:
FYI The docs still states that "bypass" is the default: https://mswjs.io/docs/api/setup-worker/start#onunhandledrequest

kettanaito:
Thanks for pointing that out, @LeBenLeBen. The changes to the docs have been committed.

PR #706, opened 2021-04-22
Bump `@mswjs/interceptors` to 0.9.0

### LINKED ISSUE #705: `TypeError [ERR_INVALID_URL]: Invalid URL: [object Object]` for requests not made with an `URL` object
## Environment

| Name | Version |
| ---- | ------- |
| msw  | 0.28.1   |
| node | 15.14.0     |
| OS   | Debian     |

## Request handlers

```js
import { rest } from 'msw'

const server = setupServer(
  rest.get('https://github.com', (req, res, ctx) => {
        return res(ctx.status(200));
    })
)

server.listen()
```

## Actual request

```js
import Wreck from '@hapi/wreck';

const request = async () => {
    return await Wreck.get('https://github.com');
}
```

## Current behavior
Requests made in application code that don't use the `URL` object can't be mocked, and cause an `Invalid URL` error:
```
    TypeError [ERR_INVALID_URL]: Invalid URL: [object Object]
        at new NodeError (node:internal/errors:329:5)
        at onParseError (node:internal/url:537:9)
        at new URL (node:internal/url:613:5)
        at CookieStore.get (/srv/node_modules/msw/node/lib/index.js:7404:28)
        at setRequestCookies (/srv/node_modules/msw/node/lib/index.js:7550:109)
        at /srv/node_modules/msw/node/lib/index.js:7609:21
        at Generator.next (<anonymous>)
        at /srv/node_modules/msw/node/lib/index.js:53:71
        at new Promise (<anonymous>)
        at __awaiter (/srv/node_modules/msw/node/lib/index.js:49:12) {
      input: '[object Object]',
      code: 'ERR_INVALID_URL',
      trace: [ { method: 'GET', url: 'https://github.com' } ]
    }
```

## Expected behavior

Requests somewhat comforming to the structure of an `URL` object can be mocked as well.

## Details

In my example, I'm using the `Wreck` module provided by Hapi to make HTTP requests. This causes the follow chain of events:
- Wreck [uses a simple object](https://github.com/hapijs/wreck/blob/master/lib/index.js#L96) to construct the `url` of the `request` (instead of the native `URL`)
- since msw calls `.toString` when [setting request cookies](https://github.com/mswjs/msw/blob/1562bfc901c930bffc8c5eb150bb855b54020676/src/utils/request/setRequestCookies.ts#L10) and [reading response cookies](https://github.com/mswjs/msw/blob/1562bfc901c930bffc8c5eb150bb855b54020676/src/utils/request/readResponseCookies.ts#L9), this causes the above mentioned error

This could be fixed by using `request.url.href` instead of `request.url.toString()`, which also seems to be the behaviour of calling `toString` on a standard `URL` object.

I'll be happy to provide a PR for this, but there might be a better way to support this (or it might not be desired to do so).

lauraseidler:
Fixes #705 (for real this time)

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   "sideEffects": false,
   "dependencies": {
     "@mswjs/cookies": "^0.1.4",
-    "@mswjs/interceptors": "^0.8.0",
+    "@mswjs/interceptors": "^0.9.0",
     "@open-draft/until": "^1.0.3",
     "@types/cookie": "^0.4.0",
     "@types/inquirer": "^7.3.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1171,10 +1171,10 @@
     "@types/set-cookie-parser" "^2.4.0"
     set-cookie-parser "^2.4.6"
 
-"@mswjs/interceptors@^0.8.0":
-  version "0.8.0"
-  resolved "https://registry.yarnpkg.com/@mswjs/interceptors/-/interceptors-0.8.0.tgz#3cf4ff1a8925c999133e8c077b94df5150c54ab1"
-  integrity sha512-Z+ocUNjEYNbTtyKiSAvoX0koN1mXvUzrR1ZqcgzF0jvt1Uqdj8zFmUDB0Lg75O63F53CnngWAF5PN8ueSQnesA==
+"@mswjs/interceptors@^0.9.0":
+  version "0.9.0"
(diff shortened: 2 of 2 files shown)


kettanaito (approved):
Thank you for carrying this update through!

kettanaito:
@lauraseidler, please let me know if you find some other issues interesting. We'd be happy to have you as a regular contributor, and we want to support your process if you ever want to become one. Just let me know. Thank you!

lauraseidler:
@kettanaito thanks! I'm just starting to use msw on a wider scale, so I might need to gain some experience with using it first, but I'll get back to you on that!

PR #720, opened 2021-05-01
Allow passing generator to RequestHandler

### LINKED ISSUE #718: Feature request: better support for stateful response logic
(This might well be a duplicate. If so I'm sorry, it was a tough one to search for.)

I have a few mocks where I'm simulating a polling endpoint, where I want the endpoint to return a 'processing' response continually for some number of times (a hundred, let's say), then finally return either 'success' or 'error'.

At the moment, as I understand it, this is a bit unergonomic with MSW. I have to maintain some state in my test code and reference it in the response handlers like this:

```ts
function() {
    let count = 0;
    server.use(
        rest.get('/polling-endpoint', (_req, res, ctx) => {
            if (count < 100) {
                count += 1;
                return res(ctx.json(pendingResponse));
            }
            return res(ctx.json(successResponse));
        })
    )
}
```
This is fine, but it makes pulling out the resolver into a utility function that I can use elsewhere quite difficult. Alternatively I can do this, which works but is definitely not ideal:

```ts
function() {
    const pending = (req, res, ctx) => res.once(ctx.json(pendingResponse));
    const ok = (req, res, ctx) => res.once(ctx.json(okResponse));
    server.use(
        rest.get('/polling-endpoint', pending),
        rest.get('/polling-endpoint', pending),
        rest.get('/polling-endpoint', pending),
        // ...repeat 97 more times
        rest.get('/polling-endpoint', ok),
    )
}
```

The simplest solution to this specific problem I can think of is to add a `.times()` method to `res`, which would work like `.once()` but allow you to specify the number of times the handler works before its internal `shouldSkip` state goes to true.

```ts
function() {
    const pending = (req, res, ctx) => res.times(100, ctx.json(pendingResponse));
    const ok = (req, res, ctx) => res.once(ctx.json(okResponse));
    server.use(
        rest.get('/polling-endpoint', pending),
        rest.get('/polling-endpoint', ok)
    )
}
```
Then `.once()` would just be an alias for `.times(1)`.

A more complicated but more powerful solution might be to allow you to pass a generator to `rest.get`. Then my `count` variable from the first example can live inside the handler:

```ts
function() {
    server.use(
        rest.get('/polling-endpoint', function* (_req, res, ctx) {
            let count = 0;
            if (count < 100) {
                count += 1;
                yield res(ctx.json(pendingResponse));
            }
            return res.once(ctx.json(successResponse));
        })
    )
}
```

I might just be missing something and this isn't a problem. If so, great!

**Additional context**

drivasperez:
This PR changes `RequestHandler` to allow passing a generator function for stateful handlers, fixes #718. For example:

```ts
const handler = rest.get('/endpoint', function*(res, req, ctx) {
  let count = 0;
  while (count < 5) {
    count += 1;
    yield res(ctx.json({ count }))
  }
  return res(ctx.json({ done: true }))
})
```

The request generator will yield responses until it finishes with `return`. Finishing a generator sets the `RequestHandler`'s `shouldSkip` property to true. It might be that this behaviour isn't what we want: an alternative would be to "reset" the handler by calling the generator function again and getting a new generator.


test/rest-api/generator.mocks.ts
```diff
@@ -0,0 +1,37 @@
...
+        count += 1
+        yield res(
+          ctx.json({
+            status: 'pending',
+            count,
+          }),
+        )
+      }
+
+      return res(
```
kettanaito: Would it be a common practice to keep returning this "complete" response once the generator is done?
It's not semantic to mark the handler as skipped, making it a one-time handler, without using the `res.once`. 

I suggest we achieve this use case by using `res.once`, and drop the `if (this.resolverGeneratorIsDone)` part from the request handling altogether. This way if you wish to respond once with the "complete" state, you'd express your intention explicitly:

```js
rest.get('/:count', function* (req, res, ctx) {
  while (predicate) {
    yield res(ctx.text('pending'))
  }

  // Once the generator is done, respond with the "complete"
  // state _once_. After this point this handler will self-destruct,
  // respecting the existing one-time handlers behavior.
  return res.once(ctx.text('complete'))
})
```

What do you think about this?

drivasperez: Yeah, I think that's probably better. It means you can have the move-on-once-completed behaviour if you want it (with `res.once`), but allows you to also just have the "complete" response repeat if you wish.

src/utils/internal/isGenerator.test.ts
```diff
@@ -0,0 +1,19 @@
+import { isGenerator } from './isGenerator'
+
+test('returns true given a generator function', () => {
+  expect(
+    isGenerator(function* () {
+      yield 2
+    }),
```
drivasperez: This is currently failing because the function itself isn't a generator, but rather it produces a generator once called.

src/utils/internal/isGenerator.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+export function isGenerator<IteratorType>(
+  fn: any,
+): fn is Generator<IteratorType, IteratorType, IteratorType> {
+  if (!fn) {
+    return false
+  }
+
+  return typeof (fn as Generator<unknown>)[Symbol.iterator] == 'function'
+}
```
drivasperez: This might be better named/typed as `isIterable`. The test will return true for anything that has a Symbol.iterator function method; I don't think there's a good way of distinguishing them. I'm also not sure if there's a good reason to disallow iterables that aren't generators, as they should work in the same way without errors.

drivasperez (commented):
Added a couple of thoughts about generators.

src/utils/internal/isGenerator.ts
kettanaito: Great suggestion! I'm preparing some polishing to the logic and will push it later today.

src/utils/internal/isGenerator.test.ts
kettanaito: An oversight, will provide the fix.

kettanaito (approved):

kettanaito:
Thank you for your superb work on this, @drivasperez! Welcome to contributors 🎉 
We'd be happy to collaborate with you in the future, so feel free to choose some of the existing issues, we'd highly appreciate your involvement. Take care!

PR #735, opened 2021-05-14
Allow returning a narrowed BodyType in ResponseTransformer

Idonoghue:
It's cumbersome only being able to send the exact BodyType generic to `ctx.json`

```typescript
rest.get<any, { result: boolean }>('/user', (req, res, ctx) =>
  res(ctx.json({ result: true })), // This will error: Type '{ result: boolean; }' is not assignable to type '{ result: true; }'
)
```

Being able to return a narrowed type is also helpful when dealing with union response types

```typescript
rest.get<any, string | string[]>('/user', (req, res, ctx) =>
  res(ctx.json('hello')), // This will error: Type 'string | string[]' is not assignable to type 'string'
)
```

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -14,8 +14,11 @@ export interface MockedResponse<BodyType = any> {
   delay?: number
 }
 
-export type ResponseTransformer<BodyType = any> = (
-  res: MockedResponse<BodyType>,
+export type ResponseTransformer<
+  BodyType extends TransformerBodyType = any,
+  TransformerBodyType = any
+> = (
+  res: MockedResponse<TransformerBodyType>,
 ) => MockedResponse<BodyType> | Promise<MockedResponse<BodyType>>
 
 export type ResponseFunction<BodyType = any> = (
--- a/test/typings/rest.test-d.ts
+++ b/test/typings/rest.test-d.ts
@@ -37,3 +37,13 @@ rest.get<
   // @ts-expect-error `null` is not a valid response body type.
   null
 >('/user', () => null)
+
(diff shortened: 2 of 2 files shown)


timdeschryver (commented):
That's weird, were you able to reproduce this in a "real" project?
I'm asking because I only seem to be able to reproduce this withing the types test of this repository 🤔

Idonoghue:
Yes we ran into this on an enterprise app. It is strange that I couldn't reproduce this in one of the existing unit or integration tests for msw. If you want I can set up a minimal repo that replicates the issue.

Idonoghue:
@timdeschryver [Here is a sandbox based on the rest-react example.](https://codesandbox.io/s/stupefied-tdd-jn1n5?file=/src/mocks/handlers.ts) After doing a lot of digging it seems the reason this error only appears sometimes is that it is dependent on the tsconfig setting `strictFunctionTypes: true`. To confirm, remove that line from the codesandbox and the error disappears. I believe this is still worth fixing because any projects with `strict: true` in the tsconfig [will also have this problem](https://www.typescriptlang.org/tsconfig#strict).

timdeschryver (dismissed):
That makes sense, thanks for taking a closer look @Idonoghue !

itsfarseen:
Workaround: 
```ts
rest.post<any, "Foo"|"Bar">("/foo", (req, res, ctx) => {
  return res(ctx.json<"Foo"|"Bar">("Foo"));
  //                 ^^^^^^^^^^^^^
}
```

kettanaito (dismissed):
Thank you for the investigation and the solution to this issue, @Idonoghue. You're amazing!

kettanaito (approved):

PR #736, opened 2021-05-16
chore: add .husky folder required by husky 5

marcosvega91:
Using husky 5 all hooks have been moved to `.husky` folder. 

I have also added a `prepare` script that will add a new bash script inside .husky that is used for running hooks

https://typicode.github.io/husky/#/?id=features

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ tmp
 stats.html
 .vscode
 msw-*.tgz
+.husky/_
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx --no-install lint-staged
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -12,3 +12,4 @@ tmp
 stats.html
 .vscode
 msw-*.tgz
+.husky/_
```
kettanaito, line 15: Would it make sense to migrate to [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks) instead of using husky?

My reasoning would be:

- easier setup.
- the same functionality (in the context of our current usage).

kettanaito (approved):
Thank you for improving the setup, @marcosvega91!

.gitignore
kettanaito: We'll update to simple-git-hooks once MSW is migrated to the internal library management solution in the future.

PR #740, opened 2021-05-17
onUnhandledRequest: Suppress exceptions when parsing potential GraphQL requests

### LINKED ISSUE #732: REST query is mistaken for GraphQL query and can't be intercepted (throws an error)
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.28.2   |
| browser | Chrome     |
| OS      | Mac OS X Catalina    |

## Request handlers

no request handler for the following rest request 

```js
https://api.dev/artists?query=Petit+Biscuit&page=0&limit=5&workspaceId=ce54f030-2191-40b1-8b2a-cd5167e9ec5f
```

## Actual request

```js
const results = await Axios.get('/artists', {
    params: {
      query,
      page,
      limit,
      workspaceId,
      onlyBelieveArtists,
    },
  });
```

## Current behavior

Browser console throws an error : 

> [MSW] Failed to intercept a GraphQL request to "GET https://api.dev/artists": cannot parse query. See the error message from the parser below.

## Expected behavior

no error shoud be thrown as it is not a GraphQL request

this happens in 

`parseGraphQLRequest` method

seemingly because it relies on the presence of a _query_ param to start and parse the query

should this method even be called? maybe it should as no rest handler is provided it also tries to find a graphQL handler

thanks for any input on this

thomasmery:
- Fixes https://github.com/mswjs/msw/issues/732

Tries and apply what @kettanaito suggested to avoid logging an error when an unhandled rest request containing a _query_ param would result in a confusing error message in the console

@kettanaito  please let me know if this is what you meant and/or how this can be improved

thanks

### DIFF
--- a/src/handlers/GraphQLHandler.ts
+++ b/src/handlers/GraphQLHandler.ts
@@ -23,6 +23,7 @@ import {
   parseGraphQLRequest,
 } from '../utils/internal/parseGraphQLRequest'
 import { getPublicUrlFromRequest } from '../utils/request/getPublicUrlFromRequest'
+import { tryCatch } from '../utils/internal/tryCatch'
 
 export type ExpectedOperationTypeNode = OperationTypeNode | 'all'
 export type GraphQLHandlerNameSelector = RegExp | string
@@ -108,7 +109,10 @@ export class GraphQLHandler<
   }
 
   parse(request: MockedRequest) {
-    return parseGraphQLRequest(request)
+    return tryCatch(
+      () => parseGraphQLRequest(request),
+      (error) => console.error(error.message),
+    )
   }
 
   protected getPublicRequest(
--- a/src/utils/internal/parseGraphQLRequest.ts
+++ b/src/utils/internal/parseGraphQLRequest.ts
@@ -157,11 +157,9 @@ export function parseGraphQLRequest(
 
     // Encountered a matching GraphQL request that is syntactically invalid.
     // We may consider getting the parsing error and propagating it to the user.
-    console.error(
+    throw new Error(
       `[MSW] Failed to intercept a GraphQL request to "${request.method} ${requestPublicUrl}": cannot parse query. See the error message from the parser below.`,
     )
-    console.error(parsedResult)
-    return undefined
   }
 
   return {
--- a/src/utils/internal/parseGraphQLRequest.ts
+++ b/src/utils/internal/parseGraphQLRequest.ts
@@ -155,10 +155,8 @@ export function parseGraphQLRequest(
   if (parsedResult instanceof Error) {
     const requestPublicUrl = getPublicUrlFromRequest(request)
 
-    // Encountered a matching GraphQL request that is syntactically invalid.
-    // We may consider getting the parsing error and propagating it to the user.
     throw new Error(
-      `[MSW] Failed to intercept a GraphQL request to "${request.method} ${requestPublicUrl}": cannot parse query. See the error message from the parser below.`,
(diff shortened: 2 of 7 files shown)


src/utils/request/onUnhandledRequest.ts
```diff
@@ -116,6 +116,17 @@ ${handlers.map((handler) => `  • ${handler.info.header}`).join('\n')}`
   return `Did you mean to request "${handlers[0].info.header}" instead?`
 }
 
+export function tryAndSilenceIfThrows<Payload>(
```
kettanaito: What do you think about:

```ts
export function withoutExceptions
```

kettanaito: Could you please move this utility function to `src/utils/internal`? That's the directory where we keep such internal helpers.

src/utils/request/onUnhandledRequest.ts
```diff
@@ -126,7 +137,10 @@ export function onUnhandledRequest(
     return
   }
 
-  const parsedGraphQLQuery = parseGraphQLRequest(request)
+  // TODO: comment and exlain why we don't want to ignore the error if parsing fails here
```
kettanaito: We actually _want_ to ignore any errors originating during this parsing. I'll provide some explanation below.

```ts
/**
 * @note Ignore any parsing exceptions.
 * At this step, the library cannot guarantee that the unhandled request
 * is a valid GraphQL requests. Since the point of this function is to
 * provide smart suggestions and not ensure a request's type validity,
 * the exception during GraphQL parsing only means that this request
 * is not a valid GraphQL request and shouldn't be suggested as such.
 */
```

src/utils/internal/parseGraphQLRequest.ts
```diff
@@ -157,11 +157,9 @@ export function parseGraphQLRequest(
 
     // Encountered a matching GraphQL request that is syntactically invalid.
     // We may consider getting the parsing error and propagating it to the user.
-    console.error(
+    throw new Error(
```
kettanaito, line 158: Could you please adjust the tests for `src/utils/internal/parseGraphQLRequest.test.ts` to reflect this change?

src/utils/request/onUnhandledRequest.ts
kettanaito: Please, would you be comfortable writing a few unit tests for this helper function? 
Let me know if you need any help, I'd be glad to help you with the tests.

kettanaito (changes requested):
Thank you for working on these improvements, @thomasmery!
They look great. I've left a few comments, could you please let me know what you think about them?

Looking forward to seeing this fix merged.

thomasmery:
@kettanaito will get to it as soon as I am able :)

thanks for the guidance!

kettanaito:
Hey, @thomasmery. I've updated your feature branch against the latest state of the repo and polished the implementation a little, mainly adding missing tests. I hope you don't mind. Looks good to get it merged, waiting for the CI to pass.

kettanaito (dismissed):

src/handlers/GraphQLHandler.ts
```diff
@@ -108,7 +109,10 @@ export class GraphQLHandler<
   }
 
   parse(request: MockedRequest) {
-    return parseGraphQLRequest(request)
+    return tryCatch(
+      () => parseGraphQLRequest(request),
+      (error) => console.error(error.message),
```
kettanaito, line 114: Handling parsing exceptions as console errors to preserve the previous behavior. At the moment, when parsing a request in a _graphql handler_ fails, it doesn't produce a 500 response, but a console error including the parsing error.

kettanaito (approved):

thomasmery:
@kettanaito ah so sorry - it was just not possible to follow up - thanks for your kindness and this great project

Issue #545, opened 2021-01-18
Codesandbox preview never updates the worker


kettanaito:
## The issue

Our pull requests use a Codesandbox integration to generate [a preview sandbox](https://codesandbox.io/s/msw-react-xx1c8) for the changes. While it correctly pulls the state of the `msw` package, it doesn't update the `public/mockServiceWorker.js` file as a part of the preview. This can lead to a broken sandbox once we introduce a breaking change to the client-worker communication. 

## Expected behavior

I'd expect Codesandbox config to support some bootstrap event where we could pull the `mockServiceWorker.js` file _relevant to the built pull request_ (it doesn't have to be the latest, as pull requests may also change the worker file).

kettanaito:
@msutkowski, hey, Matt 👋 Do you know if it's possible to execute some logic before Sandbox builds its preview?

kettanaito:
It should be possible to use the `startScript` option of the [`codesandbox.config.json`](https://codesandbox.io/docs/configuration) to execute a custom script whenever the sandbox starts. I cannot edit our MSW React template as the entire site gets broken once I unfreeze the template.

kettanaito:
I've added the `startScript` option to the `condesandbox.config.json`, but it still doesn't execute that command before preparing the sandbox. I can clearly see the latest published worker instead of the latest commit worker per pull request it tries to preview. 

Anytime I try to edit Codesandbox config live the entire page freezes and goes out of memory. I'm suspecting a Codesandbox issue.

kettanaito:
Closing this. We should migrate to CodeSandbox Projects and utilize its capabilities to use the present worker based on the branch you're on.

PR #758, opened 2021-05-29
remove-me: test worker change

kettanaito:
A test pull request for #545.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -17,6 +17,7 @@ self.addEventListener('install', function () {
 })
 
 self.addEventListener('activate', async function (event) {
+  console.log('worker activated!')
   return self.clients.claim()
 })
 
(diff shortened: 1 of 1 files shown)


kettanaito:
## Codesandbox configuration change

Sets an explicit `node` version in `ci.json` for Codesandbox to prevent this error during the sandbox installation:

```
error yargs@17.0.1: The engine "node" is incompatible with this module. Expected version ">=12". Got "10.24.1"
error Found incompatible module.
```

It appears that Codesandbox uses node@10 by default, which renders certain dependencies incompatible.

kettanaito:
The worker script is still not updated at `packageJson.msw.workerDirectory` (points to `public/mockServiceWorker.js`). The `postinstall` script executes correctly when using npm/yarn locally and on CI. This seems to be a Codesandbox issue. I've raised it with them on Twitter.

kettanaito:
The `postinstall` script doesn't get called because that's an in-browser sandbox. 

It was suggested by Ives to turn the sandbox into a container sandbox, so it gets executed in Docker, where the `postinstall` script would run. That should be possible by using the "Node" template in the `sandbox.config.json` file, but there appears to be no such option.

PR #774, opened 2021-06-08
feat(gql): support DocumentNode and TypedDocumentNode

### LINKED ISSUE #773: Support TypedDocumentNode in graphql contracts
I would like consistent usage of `TypedDocumentNode` approach for graphql operations and mocks (msw) to minimize cognitive load, coupling, and explicit types needing to be added.

I would like to be able to do this
```js
import { GetRecentActivitiesDocument } from "graphql-operations";
import { graphql } from "msw";

export const getRecentActivitiesHandler = graphql.query(
  GetRecentActivitiesDocument, // generated type that is of type `TypedDocumentNode<,>` from '@graphql-typed-document-node/core'
  (req, res, ctx) => {
    // removed for brevity
  }
);

```
instead of this
```js
import {
  GetRecentActivitiesQuery,
  GetRecentActivitiesQueryVariables,
} from "graphql-operations";
import { graphql } from "msw";

export const getRecentActivitiesHandler = graphql.query<
  GetRecentActivitiesQuery,
  GetRecentActivitiesQueryVariables
>("GetRecentActivities", (req, res, ctx) => {
  // removed for brevity
});

```

I did not consider other alternatives as it seems straightforward.

**Additional context**
If you haven't heard of `TypedDocumentNode`, here are some references
- https://the-guild.dev/blog/typed-document-node
- https://the-guild.dev/blog/graphql-with-typescript-done-right

```js
import { GetRecentActivitiesDocument } from "graphql-operations";
// ...
const { data, refetch } = useQuery(GetRecentActivitiesDocument, {
    variables: { limit: 10 },
  }); // apollo client usage of `TypedDocumentNode`
```

**Questions**
1. Is this something y'all would be supporting in the codebase?
2. If so, is this easy to do where y'all can do in a short amount of time or possibly guide me on how y'all want to approach so I can do and submit a PR?

longility:
Allows consistent usage of `TypedDocumentNode` between graphQL use and mocking.

References to `TypedDocumentNode` 
- https://the-guild.dev/blog/typed-document-node
- https://the-guild.dev/blog/graphql-with-typescript-done-right

closes #773

### DIFF
--- a/src/handlers/GraphQLHandler.test.ts
+++ b/src/handlers/GraphQLHandler.test.ts
@@ -1,6 +1,7 @@
 /**
  * @jest-environment jsdom
  */
+import { parse } from 'graphql'
 import { Headers } from 'headers-utils/lib'
 import { context } from '..'
 import { createMockedRequest } from '../../test/support/utils'
@@ -10,6 +11,7 @@ import {
   GraphQLHandler,
   GraphQLRequest,
   GraphQLRequestBody,
+  isDocumentNode,
 } from './GraphQLHandler'
 import { MockedRequest, ResponseResolver } from './RequestHandler'
 
@@ -81,6 +83,51 @@ describe('info', () => {
     expect(handler.info.header).toEqual('mutation Login (origin: *)')
     expect(handler.info.operationType).toEqual('mutation')
     expect(handler.info.operationName).toEqual('Login')
+  })
+
+  test('parses a query operation name from a given DocumentNode', () => {
+    const node = parse(`
+      query GetUser {
+        user {
+          firstName
+        }
+      }
+    `)
+
+    const handler = new GraphQLHandler('query', node, '*', resolver)
+
+    expect(handler.info).toHaveProperty('header', 'query GetUser (origin: *)')
+    expect(handler.info).toHaveProperty('operationType', 'query')
+    expect(handler.info).toHaveProperty('operationName', 'GetUser')
+  })
+
+  test('parses a mutation operation name from a given DocumentNode', () => {
+    const node = parse(`
+      mutation Login {
+        user {
+          id
+        }
+      }
+    `)
+    const handler = new GraphQLHandler('mutation', node, '*', resolver)
+
+    expect(handler.info).toHaveProperty('header', 'mutation Login (origin: *)')
+    expect(handler.info).toHaveProperty('operationType', 'mutation')
+    expect(handler.info).toHaveProperty('operationName', 'Login')
+  })
+
+  test('throws an exception given a DocumentNode with a mismatched operation type', () => {
+    const node = parse(`
+      mutation CreateUser {
+        user {
+          firstName
+        }
+      }
+    `)
+
+    expect(() => new GraphQLHandler('query', node, '*', resolver)).toThrow(
+      'Failed to create a GraphQL handler: provided a DocumentNode with a mismatched operation type (expected "query", but got "mutation").',
+    )
   })
 })
 
@@ -372,5 +419,27 @@ describe('run', () => {
     const result = await handler.run(request)
 
     expect(result).toBeNull()
+  })
+})
+
+describe('isDocumentNode', () => {
+  it('returns true given a valid DocumentNode', () => {
+    const node = parse(`
+      query GetUser {
+        user {
+          login
+        }
+      }
+    `)
+
+    expect(isDocumentNode(node)).toEqual(true)
+  })
+
+  it('returns false given an arbitrary input', () => {
+    expect(isDocumentNode(null)).toEqual(false)
+    expect(isDocumentNode(undefined)).toEqual(false)
+    expect(isDocumentNode('')).toEqual(false)
+    expect(isDocumentNode('value')).toEqual(false)
+    expect(isDocumentNode(/value/)).toEqual(false)
   })
 })
--- a/test/typings/graphql.test-d.ts
+++ b/test/typings/graphql.test-d.ts
@@ -1,3 +1,4 @@
+import { parse } from 'graphql'
 import { graphql } from 'msw'
 
 graphql.query<{ key: string }>('', (req, res, ctx) => {
@@ -50,3 +51,61 @@ graphql.operation<
 >((req, res, ctx) => {
   return res(ctx.data({ key: 'pass' }))
 })
+
+/**
+ * Supports `DocumentNode` as the GraphQL operation name.
+ */
+const getUser = parse(`
+  query GetUser {
+    user {
+      firstName
+    }
+  }
+`)
+graphql.query(getUser, (req, res, ctx) =>
+  res(
+    ctx.data({
+      // Cannot extract query type from the runtime `DocumentNode`.
+      arbitrary: true,
+    }),
+  ),
+)
+
+const getUserById = parse(`
+  query GetUserById($userId: String!) {
+    user(id: $userId) {
+      firstName
+    }
+  }
+`)
+graphql.query(getUserById, (req, res, ctx) => {
+  req.variables.userId
+
+  // Extracting variables from the native "DocumentNode" is impossible.
+  req.variables.foo
+
+  return res(
+    ctx.data({
+      user: {
+        firstName: 'John',
+        // Extracting a query body type from the "DocumentNode" is impossible.
+        lastName: 'Maverick',
+      },
+    }),
+  )
+})
+
+const createUser = parse(`
+  mutation CreateUser {
+    user {
+      id
+    }
+  }
+`)
+graphql.mutation(createUser, (req, res, ctx) =>
+  res(
+    ctx.data({
+      arbitrary: true,
+    }),
+  ),
+)
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -1,3 +1,4 @@
+import { DocumentNode } from 'graphql'
 import { Path } from 'node-match-path'
 import { ResponseResolver } from './handlers/RequestHandler'
 import {
@@ -9,6 +10,14 @@ import {
   GraphQLHandlerNameSelector,
 } from './handlers/GraphQLHandler'
 
+export interface TypedDocumentNode<
+  Result = { [key: string]: any },
+  Variables = { [key: string]: any },
+> extends DocumentNode {
+  __resultType?: Result
+  __variablesType?: Variables
+}
+
 function createScopedGraphQLHandler(
   operationType: ExpectedOperationTypeNode,
   url: Path,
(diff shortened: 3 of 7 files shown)


test/graphql-api/typed-document-node.node.test.ts
```diff
@@ -0,0 +1,109 @@
...
+const apolloClient = createGraphQLClient({
+  uri: 'http://localhost:3000',
+  fetch,
+})
+
+const server = setupServer(
+  graphql.query(GetUserDetailDocument, (req, res, ctx) => {
+    // MANUAL VERIFICATION: intellisense on `userId` should show that `userId` is a `string`.
+    // Not sure how to unit test this.
+    const { userId } = req.variables
```
longility, line 38: image

test/graphql-api/typed-document-node.node.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    // MANUAL VERIFICATION: intellisense on `ctx.data` should show that `payload` is a `GetUserDetailQuery`.
+    // Not sure how to unit test this.
+    return res(
+      ctx.data({
+        user: {
+          id: userId,
+          firstName: 'John',
+          age: 32,
+        },
+      }),
```
longility, line 49: image

src/graphql.test.ts
```diff
@@ -9,3 +15,25 @@ test('exports supported GraphQL operation types', () => {
...
 })
+
+test('pass proper type to request variable and data payload parameter for query', () => {
+  const op = graphql.query(GetUserDetailDocument, (req, res, ctx) => {
+    // If type doesn't match, this should fail compilation
+    const variables: GetUserDetailQueryVariables = req.variables
+
+    // MANUAL VERIFICATION: intellisense on `ctx.data` should be of type `DataContext<GetUserDetailQuery>`.
+    // Not sure how to test this.
+    ctx.data
```
longility: Manual verification:
image

src/graphql.test.ts
```diff
@@ -9,3 +15,25 @@ test('exports supported GraphQL operation types', () => {
...
+})
+
+test('pass proper type to request variable and data payload parameter for mutation', () => {
+  const op = graphql.mutation(LoginDocument, (req, res, ctx) => {
+    // If type doesn't match, this should fail compilation
+    const variables: LoginMutationVariables = req.variables
+
+    // MANUAL VERIFICATION: intellisense on `ctx.data` should be of type `DataContext<LoginMutation>`.
+    // Not sure how to test this.
+    ctx.data
```
longility: Manual verification: image

longility:
@kettanaito If there are preferred ways or better ways of doing something, let's discuss. I was not completely comfortable with the changes. I used judgment or took a stab at it.

kettanaito:
Hey, @longility. Your changes look great! Please, give me some time to go through them and provide feedback.

test/typings/graphql.test-d.ts
```diff
@@ -50,3 +51,22 @@ graphql.operation<
...
+
+// Supports "DocumentNode" as the GraphQL operation name.
+const getUser = parse(`
+  query GetUser {
+    user {
+      firstName
+    }
+  }
+`)
+graphql.query(getUser, (req, res, ctx) => res(ctx.data({})))
```
kettanaito: @longility, we use this TypeScript module to test types. It's going to be compiled during the test run and if there are any type of violations, the compilation will fail. We can expect failures to test negative scenarios via `@ts-expect-error`. This way we make sure there are no type regressions throughout the releases.

src/handlers/GraphQLHandler.test.ts
```diff
@@ -82,132 +82,49 @@ describe('info', () => {
...
+    expect(handler.info).toHaveProperty('header', 'query GetUser (origin: *)')
     expect(handler.info).toHaveProperty('operationType', 'query')
-    expect(handler.info).toHaveProperty('operationName', 'GetUserDetail')
+    expect(handler.info).toHaveProperty('operationName', 'GetUser')
   })
 
-  test('parses operation name out of DocumentNode for mutation', () => {
-    const handler = new GraphQLHandler('mutation', LoginDocument, '*', resolver)
+  test('parses a mutation operation name from a given DocumentNode', () => {
+    const node = parse(`
```
kettanaito, line 105: I'm replacing the manual `DocumentNode` instances with the invocations of the `parse` function from GraphQL. This way we retain the test scenario while keeping it much more concise.

package.json
```diff
@@ -64,6 +64,7 @@
   ],
   "sideEffects": false,
   "dependencies": {
+    "@graphql-typed-document-node/core": "^3.1.0",
```
kettanaito: @longility, do you think there's a way to support typed document nodes without introducing this as a dependency?

Not all the users will use GraphQL, and even smaller subset of those who do will use `TypedDocumentNode`. Adding this as a fixed dependency is quite a price _everybody_ will pay.

From the type definitions, `TypedDocumentNode` and `DocumentNode` seem compatible. The only issue I see is that one cannot extract the query fields from `DocumentNode` straight away, while `TypedDocumentNode` has generics to infer them in a handler. 

Do you think we could swap `TypedDocumentNode` with the plain `DocumentNode` (users already install `graphql`)? We can write a custom extract type to get the query fields from the plain document node to achieve type safety, I think.

kettanaito (commented):
I've updated your feature branch with a few changes, let me know if they make sense. 
Overall, this looks great. There's only one concern about adding `@graphql-typed-document-node/core` as a dependency, I don't think that's what we should do (see my reasoning in the comments).

src/graphql.test.ts
kettanaito: We should move this into the `test/typings/graphql.test-d.ts`, eventually.

package.json
kettanaito: We can still use this package as a dev dependency, so we could generate some typed document nodes and use them in type tests, for example. But I don't think this is something our users should install.

longility: Yes, I had concerns with this but I did not know how to approach. One thought I had was that we can duplicate the `TypedDocumentNode` contract into the main package code. 

```js
export interface TypedDocumentNode<Result = {
    [key: string]: any;
}, Variables = {
    [key: string]: any;
}> extends DocumentNode {
    __resultType?: Result;
    __variablesType?: Variables;
}
```

Maybe add `@graphql-typed-document-node/core` as dev dependency for testing. What do you think of that approach? I assume `TypedDocumentNode` contract shouldn't change too often and an acceptable risk, but I'm just guessing.

test/typings/graphql.test-d.ts
longility: I did not follow too much on this. Sorry for lack of experience and/or context. Can you guide me or elaborate and is there a specific task or work you were thinking?

longility: Is that the commit you did?

package.json
longility: @kettanaito I updated the code to where there isn't a dependency on `@graphql-typed-document-node/core`, and duplicated the contract. I figure easier to see the code. I'm not sure how to test. You would have to guide me as I'm weak in that area. I decided not to add `@graphql-typed-document-node/core` as a dev dependency unless we are going to test.

longility: @kettanaito I was able to figure out and test following the pattern in place with `graphql.test-d.ts` and having `graphql-typed-document-node/core` as a dev dependency. Let me know what you think.

kettanaito: Hey, @longility. Those changes look fantastic. 
Could you please tell me how do you generate the `graphql.test-data.ts`?

longility: Of course.

Here is an example. I actually used this to generate the `graphql.test-data.ts`. https://github.com/longility/msw/commit/7dc85f9a0d7f0669b469884f97b96b231f997b68

I used this for reference: https://github.com/dotansimha/graphql-typed-document-node#how-to-use

longility: Of course.

Here is an example. I actually used this to generate the `graphql.test-data.ts`. https://github.com/longility/msw/commit/b96da417cbab88d9c6724dae2abb25cede72b36b

I used this for reference: https://github.com/dotansimha/graphql-typed-document-node#how-to-use

Know that you can run against like `gql` tag function. In example, I'm using `*.graphql` file instead.

I have not done enough to know how I feel about all this yet as mentioned before. I'm experimenting. :)

kettanaito: The more I think about this feature, the more it sounds like a standalone usage example. I'm thinking maybe we can move this into a `with-graphql-code-generator` example in our examples repo, while leaving the support implementation in MSW. What are your thoughts on this? Let's discuss this before proceeding if you don't mind.

longility: I don't mind @kettanaito . I want to consider your input more heavily as you are more familiar with the repo and ecosystem of msw than I am.

If you are talking about the codegen example commit I linked (which I used to generate `graphql.test-data.ts`), then yes, I think it makes sense for `with-graphql-code-generator` in https://github.com/mswjs/examples/tree/master/examples is what I believe you are talking about. Also, maybe an addition to the docs like somewhere here: https://mswjs.io/docs/api/graphql/query#usage-with-typescript.

For transparency, this PR does not have that codegen example commit I shared, that was another branch to help you get an idea and I didn't want to add that extra maintenance on this PR.

kettanaito:
I will prepare an example and add it to the examples repo. Then we can clean up this pull request not to contain `@graphql-typed-document-node` at all (previously needed only for testing, the newly added example will act as a test).

kettanaito:
Hey, @longility. I've added a [GraphQL Code Generator usage example](https://github.com/mswjs/examples/pull/59) to the examples repo and it looks good. We'd have to merge this change first for its tests to pass, and there's also a dependency update necessary on the `@graphql-typed-document-node/core` side to make the types compatible.

Please, would you have a chance to rebase this branch and remove everything `@graphql-typed-document-node`-related from it (as it now belongs in the example)? We should use `DocumentNode` type from the `graphql` package to annotate the expected selector type of the GraphQL handlers (`TypedDocumentNode` extends `DocumentNode`).

kettanaito:
I've rebased the branch, squashing the initial implementation. Removed any external dependencies, and added an integration (interception) test to assert that MSW successfully intercepts and mocks GraphQL operations given their DocumentNode. Everything seems to be in order.

Regarding the `TypedDocumentNode` validation, I've prepared the [usage example](https://github.com/mswjs/examples/pull/59), but it requires the GraphQL Code Generator maintainers to update their `graphql` dependency (https://github.com/dotansimha/graphql-typed-document-node/pull/53). Once that's done, the smoke test for the `TypedDocumentNode` will be merged and ready.

kettanaito (approved):
Thank you for making this happen, @longility!

kettanaito:
Welcome to contributors! 🎉

longility:
Thanks @kettanaito for picking up where I left off as I was not as available.

PR #781, opened 2021-06-15
Injects the built package version into the worker script

### LINKED ISSUE #779: Add a version comment to the mockServiceWorker.js
When the `mockServiceWorker.js` in the public folder is updated it is not clear if this is by mistake or an upgrade to a new version.

Add a comment near the top of the `mockServiceWorker.js` file with the npm package version that produced the file in the `public` directory. This would make it more discoverable which version of the file you have and it will be easier when updating.

kettanaito:
## Changes

- The built package version is now injected into the worker script (`mockServiceWorker.js`).

## Motivation

- Makes it easier to review code that commits changes to the worker script (i.e. library updates).
- Makes it easier to support the library users if the package version is directly included in the script, as opposed to figuring out the package version from the script's hash sum.

## GitHub

- Closes #779

### DIFF
--- a/config/plugins/rollup-integrity-check-plugin/index.js
+++ b/config/plugins/rollup-integrity-check-plugin/index.js
@@ -3,15 +3,21 @@ const path = require('path')
 const chalk = require('chalk')
 const replace = require('@rollup/plugin-replace')
 const getChecksum = require('./getChecksum')
+const packageJson = require('../../../package.json')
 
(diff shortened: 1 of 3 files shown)


kettanaito:
Hey, @ibratoev. Do you think these changes would work? I'd like to hear your opinion on this approach.

ibratoev:
It looks great and it would improve our workflow using the library!
Thanks.

PR #787, opened 2021-06-18
Warns on pages that are outside of the worker's scope

kettanaito:
## Changes

- Calling `worker.start()` will now validate the registered worker's scope against the current page's URL. 

## Motivation

Requests issued by a page that lies outside of the worker's scope will not be intercepted. This is an expected behavior according to the Service Worker specification. However, such scenarios are currently difficult to debug, as the library will active the worker just fine, but all the requests will be bypassed.

### DIFF
--- a/src/setupWorker/start/utils/validateWorkerScope.ts
+++ b/src/setupWorker/start/utils/validateWorkerScope.ts
@@ -1,0 +1,17 @@
+import { StartOptions } from '../../glossary'
+
+export function validateWorkerScope(
+  registration: ServiceWorkerRegistration,
+  options?: StartOptions,
+): void {
+  if (!options?.quiet && !location.href.startsWith(registration.scope)) {
+    console.warn(
+      `\
+[MSW] Cannot intercept requests on this page because it's outside of the worker's scope ("${registration.scope}"). If you wish to mock API requests on this page, you must resolve this scope issue.
+
+- (Recommended) Register the worker at the root level ("/") of your application.
+- Set the "Service-Worker-Allowed" response header to allow out-of-scope workers.\
+`,
+    )
+  }
+}
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "jest": "^26.6.3",
     "json-bigint": "^1.0.0",
     "lint-staged": "^11.0.0",
-    "page-with": "^0.3.6",
+    "page-with": "^0.4.0",
     "prettier": "^2.2.1",
     "regenerator-runtime": "^0.13.7",
     "rimraf": "^3.0.2",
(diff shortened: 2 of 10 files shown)


src/setupWorker/start/utils/validateWorkerScope.ts
```diff
@@ -0,0 +1,17 @@
+import { StartOptions } from '../../glossary'
+
+export function validateWorkerScope(
+  registration: ServiceWorkerRegistration,
+  options?: StartOptions,
+): void {
+  if (!options?.quiet && !location.href.startsWith(registration.scope)) {
```
kettanaito, line 7: I'm not sure if such basic string check is enough to guarantee the correctness of the worker's scope.

kettanaito: According to the spec, the worker's directory and any subdirectories will be controlled by the worker. It looks safe to assume that if the current page's URL doesn't start with the worker's scope then it is outside of it.

Aprillion: I would hope this is enough, but I am not familiar how different browsers would handle possible edge cases (is `http` prefix always present, could it happen that one is `localhost` and the other `127.0.0.1`, implicit vs explicit port `80`, trailing `/`) ...

IMHO might be easier to release and fix false positive cases based on reported issues than to figure out more tests (unless other reviewers know more of course)

sudokoi (dismissed):

Aprillion (dismissed):

test/msw-api/setup-worker/start/wait-until-ready.error.test.ts
```diff
@@ -5,6 +5,8 @@ test('restores deferred requests if the worker registration fails', async () =>
   const { page } = await pageWith({
     example: path.resolve(__dirname, 'wait-until-ready.error.mocks.ts'),
     routes(app) {
+      app.set('etag', false)
```
kettanaito, line 8: This prevents express from responding with 304 to the cached responses. Always force 200.

src/setupWorker/start/utils/validateWorkerScope.ts
kettanaito: That's a good point. Well, we can make this validation more granular and compare protocols/hostnames/paths separately. Would that make more sense?

Aprillion: that would make sense, but only if we knew which differences are acceptable (the worker would still work despite differences) and which are not (plus maybe inconsistent between browsers)  😅

I think that having a warning even if everything works will be easy to fix in the future, but if there is no warning and the worker doesn't work will be harder for people to report repro steps

kettanaito: I think going with the current implementation and adjusting it based on developers' feedback is the right approach here. There may be false positives, which would be considered bugs, and could be fixed and released in patch versions. The amount of issues this warning may solve is way higher than the inconvenience it may cause. 

Basically, the incremental strategy you've suggested above. Love that!

Aprillion (approved):

Issue #559, opened 2021-01-22
Having `query` field in XHR json request's body breaks mocks


pa3:
## Describe the bug

Sending REST-request with a payload that includes `query` param in a mixed mock setup consisting of both REST and GraphQL mocks results in a `NetworkError` produced by mocks. That only happens when REST-request is being sent via XMLHttpRequest, sending one via fetch API work fine, as far as I was able to observe.

The reason why that fails is that `parse` until from `grapqhl` package throws when being provided with a non-string query to parse (which is the base in the described setup). 

## Environment

- `msw: 0.25.0`
- `nodejs: 14.11.0`
- `npm: 6.14.8`

## To Reproduce

I wrote a reproducing unit-test: https://github.com/pa3/msw/blob/master/test/rest-api/conflict.test.ts

## Expected behavior

The unit test from the previous section runs fine.

## Possible fixes

1. Check `typeof query` somewhere at the beginning of graphql RequestHandler's `parse` handler.
2. Filter out resolvers based on their URL-mask before invoking their `parse` handler, so that requests sent to non-graphql URLs will not mess with graphql handlers logic. 

Let me know what to do you think about these two and I will happily create a PR with one of them.

kettanaito:
Hey, @pa3. Thanks for reporting this!

Yes, this is a known issue described in #489. I've also opened a discussion on how to efficiently distinguish between an actual GraphQL request and a compatible counterpart (see https://github.com/mswjs/msw/discussions/535).

> Filter out resolvers based on their URL-mask before invoking their parse handler, so that requests sent to non-graphql URLs will not mess with graphql handlers logic.

You can do this without any changes from the library's side:

```js
import { setupWorker, graphql } from 'msw'

const link = graphql.link('https://api.backend.com/graphql')

setupWorker(
  // All request will first be tested against the endpoint of the link
  // before attempting to parse their query/body.
  link.query('GetUser', resolver)
)
```

> Read more about [GraphQL link](https://mswjs.io/docs/api/graphql/link).

I'll close this issue in favor of the aforementioned one. It's recommended you use `graphql.link` in case of mixed API types setup, as there's no reliable way to tell for sure if a request is a GraphQL request or not.

pa3:
Thanks for such a quick response and sorry for rising a duplicate issue (I was trying to find exiting one, I swear :sweat_smile:). I will   drop some thoughts in  https://github.com/mswjs/msw/issues/489

kettanaito:
[...]

PR #793, opened 2021-06-24
fix: prevent GraphQLHandler from parsing queries sent to different enpoint

pa3:
## What's wrong?

At the moment `GraphQLHandler` tries to parse any query which leads to very misleading `console.errors` being show if query parsing fails. One scenario for that is having REST request with an `req.body.query` being a JSON. (see more here: https://github.com/mswjs/msw/issues/559) The proposed solution from this to happen is to use `graphql.link`, but that does not prevent errors in console, as `msw` tries to figure out if it deals with the matching endpoint after parsing the query.

## Expected behavior

`GraphQLHandler` checks if request being handled is actually matching the endpoint specified by `graphql.link` _before_ trying to parse the request, so that non-matching requests would be rejected nicely and silently.

### DIFF
--- a/src/handlers/GraphQLHandler.ts
+++ b/src/handlers/GraphQLHandler.ts
@@ -125,6 +125,11 @@ export class GraphQLHandler<
     }
   }
 
+  public test(request: MockedRequest): boolean {
+    const hasMatchingUrl = matchRequestUrl(request.url, this.endpoint)
+    return hasMatchingUrl.matches && super.test(request)
+  }
+
   predicate(request: MockedRequest, parsedResult: ParsedGraphQLRequest) {
     if (!parsedResult) {
       return false
--- a/src/handlers/GraphQLHandler.test.ts
+++ b/src/handlers/GraphQLHandler.test.ts
@@ -325,6 +325,29 @@ describe('test', () => {
     expect(handler.test(request)).toBe(true)
     expect(handler.test(alienRequest)).toBe(false)
   })
+
+  test('does not parse requests sent to the different endpoints', () => {
+    jest.spyOn(console, 'error')
+
+    const handler = new GraphQLHandler(
+      'query',
+      'GetUser',
+      'https://api.github.com/graphql',
+      resolver,
+    )
+
+    const request = createMockedRequest({
+      method: 'POST',
+      url: new URL('/some-other-endpoint', location.href),
+      body: {
+        query: { message: 'I gonna cause troubles!' },
+      },
+    })
+
+    handler.test(request)
+
+    expect(console.error).not.toHaveBeenCalled()
+  })
 })
 
 describe('run', () => {


kettanaito:
Hey, @pa3. Thank you for issuing this. 

Could you please update to `0.30.0` and let me know if this issue is still there?

The latest release fixes an issue with the GraphQL-compatible REST requests (#740), so I hope it affects the behavior you're experiencing as well. 

By design, `graphql.query`/`graphql.mutation` affect _all_ GraphQL operations, regardless of the endpoint. You've pointed out correctly, that in case you wish to scope the interception to a specific endpoint, you must use `graphql.operation` instead. 

MSW should properly tell apart a GraphQL-compatible REST request and a regular (irrelevant) REST request at the moment.

src/handlers/GraphQLHandler.ts
```diff
@@ -125,6 +125,11 @@ export class GraphQLHandler<
     }
   }
 
+  public test(request: MockedRequest): boolean {
+    const hasMatchingUrl = matchRequestUrl(request.url, this.endpoint)
+    return hasMatchingUrl.matches && super.test(request)
```
kettanaito, line 130: The endpoint URL match is already implemented in the `predicate` method. The parent class calls `this.predicate` as a part of the `test` method:

https://github.com/mswjs/msw/blob/1da34e4fef13e278043d65ad656875186ac84e90/src/handlers/GraphQLHandler.ts#L128-L157

https://github.com/mswjs/msw/blob/1da34e4fef13e278043d65ad656875186ac84e90/src/handlers/RequestHandler.ts#L158-L160

Internally, parsing always precedes the predicate, because the request handler's predicate may depend on some request data available only after the request has been parsed (even if it's an irrelevant request—that's how the relevance is determined). This also allows for a request handler to parse any given request without performing any predicates (direct usage of `handler.parse()` of the handler's public API).

kettanaito:
Even if there's a valid GraphQL request issued to an unrelated endpoint, there should be no parsing errors as long as that request describes a valid GraphQL operation. The only scenario when I see the issue occurring potentially is when a seemingly compatible REST API request is being parsed by the GraphQL handler. We've taken multiple measures to ensure that's not the issue in the past releases, but I don't deny there may be more that we could do to improve the experience. 

At the moment, a seemingly compatible REST request will only be parsed if you have any GraphQL handlers.

pa3:
Hey, @kettanaito, thanks for your fast feedback (as usual)!

> The only scenario when I see the issue occurring potentially is when a seemingly compatible REST API request is being parsed by the GraphQL handler.

My case exactly. REST API request has a `body.query` which makes GraphQL handler trying to parse it. Which in turn leads to error being reported to console:

Screenshot from 2021-06-25 10-29-40

I have a test case reproducing unwanted behavior in this PR.

> Could you please update to 0.30.0 and let me know if this issue is still there?

Yup, logged error is still there.

> Internally, parsing always precedes the predicate, because the request handler's predicate may depend on some request data available only after the request has been parsed (even if it's an irrelevant request—that's how the relevance is determined).

Filtering request not matching current `endpoint` before even trying to parse it feels safe, isn't it?

kettanaito:
> Filtering request not matching current endpoint before even trying to parse it feels safe, isn't it?

It does feel safe, but it's against the internal request handling design. Parsing must always precede the predicate because the predicate may rely on the information available only after parsing (i.e. GraphQL operation type/name, REST API path parameters).

This comes down to the inability to distinguish between a GraphQL operation and a seemingly compatible REST API request. GraphQL operations still implement REST, even if partially, and have no discernible properties to be reliably treated as GraphQL operations (discussed in #535).

I understand that moving the endpoint check into the `parse()` method of the GraphQL handler looks like a solution, but it's a workaround at its finest: we are compromising the request handling design to compensate for the GraphQL spec. This isn't a solution I'd like to follow. 

Instead, perhaps we can find a different approach to take?

pa3:
> Parsing must always precede the predicate because the predicate may rely on the information available only after parsing

I mean in case `msw` wants to support an option of, let's say, overruling `endpoint` matching by some other information gained from parsing, then, indeed, my proposed solution breaks this design decision. In any other case I don't see any issue with early exit  from route handling (before attempting any parsing) when it's clear that `endpoint` is not matching for a given handler. Advantages only, like saving couple of CPU cycles by avoiding wasteful parsing. :smile: 

> Instead, perhaps we can find a different approach to take?

The only other option I see at the moment is to somehow pass a configuration option to `msw` to prevent it from logging error here: https://github.com/mswjs/msw/blob/2213b4ebd78c8459d8d9f39d8dd1d268337b79d8/src/handlers/GraphQLHandler.ts#L114

That being said, I don't want/have a right to question design decisions of `msw` and already silenced annoying `console.error` from my side. So, feel free to close this draft if my reasoning still doesn't make any sense. 

And thanks for the awesome library! :heart:

Issue #622, opened 2021-03-01
React Native support is broken


wwdrew:
This code was taken directly from the React Native GraphQL branch in the examples repository (https://github.com/mswjs/examples/tree/examples/gql-react-native/examples/graphql-react-native-apollo). I've been unable to get that to work at all for other reasons, so I thought I'd create a new repository using the latest versions of everything to demonstrate the error I'm receiving.

https://github.com/wwdrew/mswjs-react-native-failing

## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.27.0   |
| browser | n/a     |
| react-native | 0.63.4 |
| OS      | MacOS 11.2.1     |

## Current behavior

Metro crashes with this stack trace:

```
error: Error: Unable to resolve module os from <obscured>/AnotherTest/node_modules/msw/native/lib/index.js: os could not be found within the project.

If you are sure the module exists, try these steps:
 1. Clear watchman watches: watchman watch-del-all
 2. Delete node_modules and run yarn install
 3. Reset Metro's cache: yarn start --reset-cache
 4. Remove the cache: rm -rf /tmp/metro-*
  4 |
  5 | var XMLHttpRequest = require('node-request-interceptor/lib/interceptors/XMLHttpRequest');
> 6 | var os = require('os');
    |                   ^
  7 | var tty = require('tty');
  8 | var events_1 = require('events');
  9 | var nodeRequestInterceptor = require('node-request-interceptor');
```

## Expected behavior

Expecting the API call to be successful and not crash the bundler.

## Screenshots

<img width="472" alt="msw-error" src="https://user-images.githubusercontent.com/49833875/109526671-bedb3c80-7aaa-11eb-9a83-659936818bb6.png">

wwdrew:
This problem is a serious blocker for me, if anyone has any idea what might be causing it I'd be happy to try to track it down. Any kind of guidance would be really helpful.

kettanaito:
Hey, @wwdrew. Thanks for reporting this.

This issue is caused by how [node-request-interceptor](https://github.com/mswjs/node-request-interceptor) library is bundled (and also how MSW bundles library within itself). Looks like it shouldn't include `os` (and possibly other modules) that don't exist in React Native. 

If this issue blocks your work, I suggest cloning MSW and experimenting with its build configuration. Here's how we bundle the library for React Native:

[...]

PR #808, opened 2021-07-08
fix: use external chalk library to stop os/tty import errors on React Native

wwdrew:
Without configuring chalk as an external dependency in the React Native configuration, rollup will try to import os and tty which don't exist in the React Native environment.

In conjunction with #810 and #809, this fixes React Native support.

See https://github.com/mswjs/msw/issues/622 for more details.


kettanaito (approved):
This looks great, thank you, @wwdrew!

kettanaito:
Welcome to contributors!

PR #809, opened 2021-07-08
fix: use WHATWG url for formatting url to fix React Native importing url package

wwdrew:
This PR removes the legacy format function in favour of using the WHATWG implementation, as suggested by the node.js docs (https://nodejs.org/api/url.html#url_url_format_urlobject). This has the side effect for React Native that it no longer imports the non-existent url library.

In conjunction with #810 and #808, this fixes React Native support.

See #622 for more details.


kettanaito (approved):

kettanaito:
Hey, @wwdrew. Please, is there a way to contact you outside of GitHub (email, slack, twitter)? I'd like to ask you a question. Thank you!

PR #810, opened 2021-07-08
chore: update @mswjs/cookies package to fix React Native incompatibility

wwdrew:
The 0.1.6 version of @mswjs/cookies changes the way it checks whether localStorage is available on the host, which is compatible with React Native.

In conjunction with https://github.com/mswjs/msw/pull/809 and https://github.com/mswjs/msw/pull/808, this fixes React Native support.

See #622 for more details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
   ],
   "sideEffects": false,
   "dependencies": {
-    "@mswjs/cookies": "^0.1.5",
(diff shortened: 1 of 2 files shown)


kettanaito (dismissed):
Thank you for bumping this, @wwdrew!

kettanaito (approved):

Issue #661, opened 2021-03-23
Can't install the service worker in Electron renderers with `nodeIntegration` enabled


tessro:
## Describe the bug

Invoking `worker.start()` in an Electron renderer fails if `nodeIntegration` is enabled. This happens because `process` is defined, which causes `isNodeProcess` to return true. As a result, it's not possible to use `msw` in these windows. (`msw`'s node server doesn't work in that environment either, since `fetch` calls use the Chromium network stack, not Node's.)

Electron provides `process.type` with value `"renderer"` in these situations. Would it be satisfactory for `isNodeProcess` to return false in that case? If so, I'm happy to put up a PR!

## Environment

- `msw: 0.27.1`
- `nodejs: 12.18.3`
- `npm: 7.6.3`

> Please also provide your browser version.

Electron 11.3.0 / Chromium 87

## To Reproduce

1. Create Electron `BrowserWindow` with `nodeIntegration: true`
2. Create an `msw` service worker and invoke `worker.start()`
3. Open DevTools, view Console
4. See error:
```
[MSW] Failed to execute `setupWorker` in a non-browser environment. Consider using `setupServer` for Node.js environment instead.
```

## Expected behavior

Service worker is installed

kettanaito:
Hey, @paulrosania. Thank you for discovering this.

So you suggest adding a special case for the Electron renderer process in `isNodeProcess`?

```js
if (process?.type === 'renderer') {
  return true
}
```

> [isNodeProcess.ts](https://github.com/mswjs/msw/blob/655c93f83b7fcebe9a9093cacd7d891939028d29/src/utils/internal/isNodeProcess.ts)

While I think that solution may work, I have a few concerns with this approach. It makes the logic dependent on Electron implementation details, specifically on the way Electron marks the process type. We need to be cautious with escape hatches like this because most likely the renderer process _is_ a Node.js process. The environment check on the library's side is to prevent failure scenarios that are known beforehand. `setupWorker` won't function in a non-browser environment, so we short circuit when such environment is detected. 

> This may be a good reason to re-consider if relying on the `process` being present is the best way to determine a non-browser environment. Perhaps that would be the correct fix. 

Could you please try the suggested fix locally and see how your Electron application behaves?

tessro:
Ah, I see the messiness of this.

[...]

PR #817, opened 2021-07-18
fix: isNodeProcess misidentifies nw.js and electron as node

Losses:


src/utils/internal/isNodeProcess.ts
```diff
@@ -4,7 +4,7 @@
  */
 export function isNodeProcess() {
   // Check browser environment.
-  if (typeof global !== 'object') {
+  if (typeof global !== 'object' || typeof window !== 'undefined') {
```
kettanaito, line 7: This is not correct: `window` may exist in non-browser environments, such as JSDOM. Such environments must _not_ be treated as a browser, because they cannot have the Service Worker API. Such environments must be treated as Node.js because they, in fact, are. 

The intention behind `isNodeProcess` function is to return `true` for _any_ environment that is not a native browser. This includes:

- Environments that polyfill browser-like behaviors (JSOM);
- Server-side environments of otherwise browser-oriented tools (runners in Storybook and Cypress);
- Actual Node.js environment (i.e. when developing Node.js servers).

If you think this function doesn't determine a native browser environment properly, such as Electron, could you share more info about when this happens? A reproduction repository would be the best. We may adjust `isNodeProcess` accordingly, but that adjustment won't rely on the existence of `window`.

kettanaito (changes requested):
Hey, @Losses. Thank you for preparing this. 

Please, could you share some info about when and how this issue is happening? Sharing a reproduction repository would be the best.

Losses:
@kettanaito Hi! thank you for the reply!

Electron as an example, creating a browser window with the following parameters:

```
new BrowserWindow({
    webPreferences: {
      nodeIntegration: true,
    }
});
```

There will be a `global` under global scope, which will make `msw` think it is running on a nodejs environment, but actually, it is a browser (or, Electron calls it "renderer").

Losses:
@kettanaito Hi! thank you for the reply!

Electron as an example, creating a browser window with the following parameters:

```
new BrowserWindow({
    webPreferences: {
      nodeIntegration: true,
    }
});
```

There will be a `global` under global scope, which will make `msw think it is running on a nodejs environment , but actually, it is a browser.

kettanaito:
@Losses, thanks for a quick follow-up.

This is related to https://github.com/mswjs/msw/issues/661 and is only reproducible if the `nodeIntegration` option of Electron is set to `true`. 

I acknowledge this is an issue, but we need to approach it so as not to introduce any Electron-specific logic into `isNodeProcess`, as that would be tiresome to maintain across Electron releases and is frankly unnecessary. 

Perhaps there's a more reliable way to determine an actual browser? Would you be interested in doing research on this?

Losses:
@kettanaito Let me have a look...

kettanaito:
I've found [this package](https://github.com/flexdinesh/browser-or-node) that determines if the code is running in a browser or Node.js. I can't say that it matches our expectations, but we can try out the logic it uses in the `isNodeProcess` function.

> It doesn't take DOM-like environments into account and they are matched as a browser. 

I'm thinking about relying on something like `navigator.userAgent` or `window.name` (given it's consistently set to a fixed value by non-browser environments, which I doubt to be the case).

kettanaito:
We may consider relying on `navigator` as it does have some properties that can reliably determine a browser or a non-browser environment. I was experimenting with this:

```js
function isNodeProcess() {
  const userAgent =
    typeof navigator !== 'undefined' ? navigator.userAgent.toLowerCase() : ''
  const product = typeof navigator !== 'undefined' && navigator.product

  if (!userAgent || userAgent.includes('jsdom') || product === 'ReactNative') {
    return true
  }

  return false
}
```

This function produces the following output:

- Node.js: `true`
- JSDOM: `true`
- React Native: `true`
- Electron: `false`
- Browser: `false`

kettanaito:
Given how many different environments are needed to reliably test this function, I think we should move it to a separate package and have those tests in place. Relying on the `userAgent`/`product` seems reliable and I'd like to have some automated tests to confirm that.

kettanaito:
I've moved the `isNodeProcess` logic into a standalone [`is-node-process`](https://github.com/mswjs/is-node-process) package and covered the usage scenarios with tests. The only issue I had was with the React Native tests as I cannot install XCode on my machine currently. 

We should replace the internal logic with that package. I confirm that it handles Electron correctly by the automated test.

PR #819, opened 2021-07-19
set: Type-validates the forbidden "cookie"/"set-cookie" response headers

kettanaito:
## GitHub

- Originates from #818 

## Changes

- Setting forbidden cookie headers with the `ctx.set` utility now produces a TypeScript violation.
- Forbidden cookie headers include (case-insensitive):
  - `cookie`
  - `cookie2`
  - `set-cookie`
  - `set-cookie2`

We are not including [all forbidden header names](https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name) for now, starting with the `cookie` ones.

## Motivation

We should fail fast in preventing the developer from setting forbidden response cookies. Setting such cookies will raise a runtime exception anyway. 

## Experience

```ts
set({ cookie: 'secret' })

// Argument of type '[{ cookie: string; }]' is not assignable to parameter of type
// '"SafeResponseHeader: the 'cookie' header cannot be set on the response. Please use the 'ctx.cookie()' function instead."'.ts(2345)
```

> Utilizes [this technique](https://github.com/microsoft/TypeScript/pull/40468#issuecomment-808311132) to have custom type violation messages until https://github.com/microsoft/TypeScript/pull/40468 is merged.

### DIFF
--- a/src/context/set.ts
+++ b/src/context/set.ts
@@ -1,12 +1,43 @@
 import { objectToHeaders } from 'headers-utils'
 import { ResponseTransformer } from '../response'
 
+export type HeadersObject<KeyType extends string = string> = Record<
+  KeyType,
+  string | string[]
+>
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
+ */
+export type ForbiddenHeaderNames =
+  | 'cookie'
+  | 'cookie2'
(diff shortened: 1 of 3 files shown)


kettanaito:
The smoke tests are failing due to the old TypeScript version being used in the usage examples.

PR #835, opened 2021-07-27
Uses the "SerializedResponse" type consistently

kettanaito:
- `SerializedResponse` type cannot be inferred from the `MockedResponse` because the latter contains a bunch of internal properties (`once`, `delay`, etc.). Re-declare the type instead.
- Use the `SerializedResponse` type in the `RequestHandler` methods appropriately.

### DIFF
--- a/src/handlers/GraphQLHandler.ts
+++ b/src/handlers/GraphQLHandler.ts
@@ -160,7 +160,7 @@ Consider naming this operation or using "graphql.operation" request handler to i
 
   log(
     request: Request,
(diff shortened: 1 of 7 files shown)


kettanaito:
The input `prepareResponse` accepts is no longer the internal `MockedResponse` instance. In actuality, it never was, as the `transformResponse` of `handleRequest` function can reduce the mocked response to anything. It's then that reduced response that's passed to the `onMockedResponseSent`, and other methods, including `log` of the handlers.

PR #837, opened 2021-07-27
mockServiceWorker: Bypasses the SSE (text/event-stream) requests

### LINKED ISSUE #834: SSE Bypass for HMR
## Describe the bug
When developing on the Next.js with `next dev`, No problem for a while,
but suddenly the server becomes unresponsive.

## Environment

- `msw: 0.29.0`
- `nodejs: 14.16.0`
- `npm: 0.0.0`
- `next: 10.2.3`

Chrome: Version 91.0.4472.164 (Official Build) (x86_64)

## To Reproduce
1. npm run `next dev`
2. Edit some code many times.
3. Suddenly the server becomes unresponsive.

## Expected behavior
Server always responds.

## Screenshots
The below capture is the Network tab when no response is returned.

image

## Workaround

It looks like there is a problem in `webpack-hmr?page=xxxx`.
As a workaround, I added the below SSE Bypass in generated `mockServiceWorker.js`.
Then, I was able to solve the problem that the response did not come back.

```javascript
self.addEventListener("fetch", function (event) {
  const { request } = event;

  // Bypass server-sent events.
  if (request.headers.get("accept") === "text/event-stream") {
    return;
  }

  // Bypass navigation requests.
  if (request.mode === "navigate") {
    return;
  }
  ...
}
```

This Bypass is useful to us, but what do you think?

takefumi-yoshii:
Fixes #834

## Changes

Requests whose `Accept` header contains `text/event-stream` are not explicitly bypassed by the worker script.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -243,6 +243,11 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
 self.addEventListener('fetch', function (event) {
   const { request } = event
 
+  // Bypass server-sent events.
+  if (request.headers.get('accept') === 'text/event-stream') {
+    return
+  }
+
   // Bypass navigation requests.
   if (request.mode === 'navigate') {
     return
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -242,9 +242,10 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
 
 self.addEventListener('fetch', function (event) {
   const { request } = event
+  const accept = request.headers.get('accept') || ''
 
   // Bypass server-sent events.
-  if (request.headers.get('accept') === 'text/event-stream') {
+  if (accept.includes('text/event-stream')) {
     return
   }
 
--- a/test/msw-api/setup-worker/scenarios/text-event-stream.mocks.ts
+++ b/test/msw-api/setup-worker/scenarios/text-event-stream.mocks.ts
@@ -1,0 +1,9 @@
+import { setupWorker, rest } from 'msw'
+
+const worker = setupWorker(
+  rest.get('/user', (req, res, ctx) => {
+    return res()
+  }),
+)
+
+worker.start()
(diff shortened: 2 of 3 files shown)


src/mockServiceWorker.js
```diff
@@ -243,6 +243,11 @@ If you wish to mock an error response, please refer to this guide: https://mswjs
 self.addEventListener('fetch', function (event) {
   const { request } = event
 
+  // Bypass server-sent events.
+  if (request.headers.get('accept') === 'text/event-stream') {
```
kettanaito: Should we account for the cases when the `accept` header includes `text/event-stream` in a list of accepted content types? 

```
Accept: text/plain; text/event-stream;
```

That should still be possible per specification.

Perhaps something like this?

```js
const accept = request.headers.get('accept') || ''

if (accept.includes('text/event-stream')) {}
```

What do you think about this, @takefumi-yoshii?

takefumi-yoshii: No objection, I'll fix it like this.

kettanaito (commented):
Thank you for addressing the header's multi-value support! 

The fix looks great, I'll add an integration test for this and it'll be ready for merge.

kettanaito:
A useful article on issuing/responding to a SSE with Express:

- https://masteringjs.io/tutorials/express/server-sent-events

kettanaito:
I've added the integration test for the `text/event-stream`, yet for some reason, it passes even without the fix. There must be something else to the webpack's HMR request that we need to emulate in the test to reliably assert that it is bypassed. 

@takefumi-yoshii, please, would you have a chance to look at the difference between the setup we have in the test (`test/msw-api/setup-worker/scenarios/text-event-stream.test.ts`) and what happens upon the HMR request?

kettanaito:
I've investigated the difference and updated the main issue. At the time being, I don't see a reliable and cost-efficient way to put this use case into an integration test. Spawning the Next app as a part of tests is extremely costly. The Next example lives in the Next repo, so we don't have control over it to introduce a testing pipeline like we do with our in-house examples (that act as smoke test scenarios).

kettanaito (approved):
Thank you for addressing this!

kettanaito:
Welcome to contributors, @takefumi-yoshii! 🎉

PR #856, opened 2021-08-09
onUnhandledRequest: Throws an exception when using the "error" strategy

### LINKED ISSUE #855: onUnhandledRequest: throw error should fail test
When testing components, most of the API responses are required to be mocked which means that in most cases, the test should not continue without mocking them and should fail them immediately

If an error is thrown in the functional implementation on `onUnhandledRequest` option, the test should fail.

If we specify:
```js
server.listen({
    onUnhandledRequest: (req) => {
        if(NOT_VALID) { //if req is something not expected,
            throw new Error("any error"); // throw error.
        }
    }
})
```

In test
```js
it('calls an endpoint', () => { 
  const res = await post('some/endpoints');  //Should fail here before reaching assertion
  
  expect(res.status).toBe(200);
})
```

The error thrown is going to fail the single test run, but not the whole process.

n/a

**Additional context**
#539

kettanaito:
## GitHub

- Fixes #855 
- Related to #539 

## Changes

- When the "error" strategy of the `onUnhandledRequest` option is used, any request that has no matching request handler will _throw an exception_ to halt request processing and fail any code dependant on it:

```js
await fetch('https://api.com/unhandled')
// FethError: ...
```

- The developer-friendly `console.error` is preserved because the message is too verbose to be used as the `FetchError` reason. Instead, the fetch error reason is a compact summary of why the request failed (cannot be performed as-is).

### DIFF
--- a/src/utils/request/onUnhandledRequest.ts
+++ b/src/utils/request/onUnhandledRequest.ts
@@ -170,8 +170,13 @@ Read more: https://mswjs.io/docs/getting-started/mocks\
 
   switch (strategy) {
     case 'error': {
+      // Print a developer-friendly error.
       devUtils.error('Error: %s', message)
-      break
+
+      // Throw an exception to halt request processing and not perform the original request.
+      throw new Error(
+        'Cannot bypass a request when using the "error" strategy for the "onUnhandledRequest" option.',
+      )
     }
 
     case 'warn': {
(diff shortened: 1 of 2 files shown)


jchi2241:
Thank you for this amazing library. Does throwing an error also stop tests that depend on `setupWorker()`?

jchi2241:
> Thank you for this amazing library. Does throwing an error also stop tests that depend on `setupWorker()`?

Answered my own question - it does.

PR #857, opened 2021-08-09
Supports escaped characters in the URL (i.e. parenthesis)

### LINKED ISSUE #825: Cannot intercept a request with a parentheses in the URL
## Issue 

The request URL having parenthesis eg: `https://made.up/api/usage('id')`,  `https://made.up/api/usage('id')/relObjects`   etc.. not intercepting in my jest react testing library tests using msw.

## Environment

| Name | Version |
| ---- | ------- |
| msw  |^0.30.1   |
| node | v14.16.1     |
| OS   | Windows 10   |

## Request handlers

```js
import { rest } from 'msw'

export const handlers = [
  /**
   * Example of a request handler—function that captures a request
   * and declares which mocked response to return upon match.
   * @see https://mswjs.io/docs/basics/request-handler
   */
  rest.get("https://made.up/api/usage\('id'\)", (req, res, ctx) => {
    return res(
      ctx.status(301),
      ctx.json({
        id: 1,
        firstName: 'John',
      }),
    )
  }),
]
```

## Actual request
```js
// App.js
const { loading, error, data } = useFetch("https://made.up/api/usage('id')");
```

https://codesandbox.io/s/msw-react-forked-nsqrs?file=/src/App.js

## Current behavior

The request is not intercepting 

## Expected behavior

msw mock should work and my test should pass

## Notes 

https://codesandbox.io/s/msw-react-puui0
in the above code sandbox for the URL "https://made.up/api/usage('id')" How should I match the URL in handlers ?
In a similar lines I am unable to intercept many URLs in my project, Tried different way of escaping characters as well
My project s on Jest React Testing library on a CRA

kettanaito:
## GitHub

- Fixes #825

### DIFF
--- a/src/utils/url/getAbsoluteUrl.ts
+++ b/src/utils/url/getAbsoluteUrl.ts
@@ -11,5 +11,8 @@ export function getAbsoluteUrl(path: string, baseUrl?: string): string {
   // or the current location (in the case of browser/browser-like environments).
   const origin = baseUrl || (typeof location !== 'undefined' && location.origin)
 
-  return origin ? new URL(path, origin).href : path
+  return origin
+    ? // Encode and decode the path to preserve escaped characters.
+      decodeURI(new URL(encodeURI(path), origin).href)
+    : path
 }
--- a/test/rest-api/request/matching/uri.mocks.ts
+++ b/test/rest-api/request/matching/uri.mocks.ts
@@ -33,11 +33,11 @@ const worker = setupWorker(
   ),
 
   rest.get(/(.+?)\.google\.com\/path/, (req, res, ctx) => {
-    return res(
-      ctx.json({
-        mocked: true,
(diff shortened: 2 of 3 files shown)


src/utils/url/getAbsoluteUrl.ts
```diff
@@ -11,5 +11,5 @@ export function getAbsoluteUrl(path: string, baseUrl?: string): string {
   // or the current location (in the case of browser/browser-like environments).
   const origin = baseUrl || (typeof location !== 'undefined' && location.origin)
 
-  return origin ? new URL(path, origin).href : path
+  return origin ? decodeURI(new URL(encodeURI(path), origin).href) : path
```
kettanaito: Encoding and decoding the URL preserves the escaped characters present in the `path`.

PR #860, opened 2021-08-11
Cleanup emitter listeners

### LINKED ISSUE #858: on('request:match', (req) => {}) Does not get cleaned up
## Describe the bug
I'm using `.on('request:match', (req) => {})` in tests to make sure variables are getting send correctly in the request, however is seems that if I have more than 1 across different tests then they start to fail. I have this in my setupTests file:
```typescript
beforeAll(() => server.listen({ onUnhandledRequest: 'error' }))
afterAll(() => server.close())
afterEach(() => server.resetHandlers())
```
It's like the `resetHandlers()` doesn't reset the `server.on`.

As soon as I comment out the `on()` section then the tests all pass.

## Environment

- `msw: 0.33.0`
- `nodejs: 14.17.3`
- `npm: 6.14.13`

## To Reproduce
1. listen for `.on('request:match')` in multiple tests.
2. watch the tests fail

## Expected behavior
Expect all tests to pass with `.on('request:match')` in multiple tests.

eddiemoore:
## GitHub
- Fixes #858 

## Changes

- Exposes the Life-cycle events under the `.events` property of the worker/server:
```diff
-server.on('request:start', cb)
+server.events.on('request:start', cb)
```

- Exposes the entire `EventEmitter` under `.events` so that developers could interact with it as a regular event emitter (incl. removing listeners).

```js
worker.events.removeAllListeners()
```

### DIFF
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -112,7 +112,6 @@ export function createSetupServer(...interceptors: Interceptor[]) {
           requestHandlers,
           ...nextHandlers,
         )
-        emitter.removeAllListeners()
       },
 
       printHandlers() {
@@ -129,9 +128,7 @@ ${bold(`${pragma} ${header}`)}
         })
       },
 
-      on(eventType, listener) {
-        emitter.addListener(eventType, listener)
-      },
+      events: emitter,
 
       close() {
         emitter.removeAllListeners()
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -112,6 +112,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
           requestHandlers,
           ...nextHandlers,
         )
+        emitter.removeAllListeners()
       },
 
       printHandlers() {
--- a/src/node/glossary.ts
+++ b/src/node/glossary.ts
@@ -2,6 +2,7 @@ import { PartialDeep } from 'type-fest'
 import { IsomorphicResponse } from '@mswjs/interceptors'
 import { MockedRequest, RequestHandler } from '../handlers/RequestHandler'
 import { SharedOptions } from '../sharedOptions'
+import { StrictEventEmitter } from 'strict-event-emitter'
 
 export interface ServerLifecycleEventsMap {
   'request:start': (request: MockedRequest) => void
@@ -49,11 +50,5 @@ export interface SetupServerApi {
    */
   printHandlers(): void
 
-  /**
-   * Attaches a listener to one of the life-cycle events.
-   */
-  on<EventType extends keyof ServerLifecycleEventsMap>(
-    eventType: EventType,
-    listener: ServerLifecycleEventsMap[EventType],
-  ): void
+  events: StrictEventEmitter<ServerLifecycleEventsMap>
 }
--- a/src/setupWorker/glossary.ts
+++ b/src/setupWorker/glossary.ts
@@ -230,11 +230,15 @@ export interface SetupWorkerApi {
    */
   printHandlers: () => void
 
-  /**
-   * Attaches a listener to one of the life-cycle events.
-   */
-  on<EventType extends keyof WorkerLifecycleEventsMap>(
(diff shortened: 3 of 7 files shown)


src/node/createSetupServer.ts
```diff
@@ -112,6 +112,7 @@ export function createSetupServer(...interceptors: Interceptor[]) {
           requestHandlers,
           ...nextHandlers,
         )
+        emitter.removeAllListeners()
```
kettanaito: I wonder what's your thoughts on the expectations here. Imagine this setup:

```js
const server = setupServer()

beforeAll(() => {
  server.listen()
  server.on('request:start', callback)
})

afterEach(() => {
  // Calling this will now remove both "request:start" listeners.
  server.resetHandlers()
})

it('some test', () => {
  server.on('request:start', anotherCallback)
})
```

Would you expect `server.resetHandlers` to _wipe out_ all the listeners, including those you've added in the `beforeAll` hook?

The `.resetHandlers` method removes all runtime handlers (those you've added with `server.use()`) but preserves those you've initially passed to the `setupServer` call. I know it's a different thing, but I was wondering if people would expect it behave the same for event listeners in life-cycle events.

eddiemoore: That's a good point. Maybe need to add a separate method to allow the removal of the emitter's listeners.

kettanaito: I've dwelled on this thought and came to the conclusion that coupling event listeners clean-up and `.resetHandlers()` isn't a good idea because it's not evident. I'm leaning towards having a separate method to remove event listeners from the `server`/`worker` instances. 

I also suggest to group the life-cycle events and the clean-up method under a single `events` property. The reason for that is `server.removeAllListeners()` being too ambiguous. 

```diff
-server.on('request:start', cb)
-worker.on('request:end', cb)
+server.events.on('request:start', cb)
+worker.events.on('request:end', cb)
```

Removing the listeners can then become something like:

```js
server.events.removeAllListeners()
```

I think we could also benefit from `server.events` being a regular `EventEmitter`. This would allow developers to remove certain listeners via `.removeListener(eventName, listener)` as they usually do with a regular event emitter.

kettanaito:
Hey, @eddiemoore. 

I've decided to go with the suggested approach above, that being grouping the entire Life-cycle events API under the `events` property on the worker/server. With this change we'll also expose the entire event emitter and not the `.on()` method, so developers can add/remove listeners and interact with `worker.events` as a regular event emitter.

Let me know if you have any comments about this approach, I'd appreciate your opinion. Thank you!

eddiemoore:
@kettanaito yeah I think that's a good idea.

kettanaito:
There are also internal event listeners added to the same `emitter` to manage responses in Node.js:

https://github.com/mswjs/msw/blob/9373ccf5e699fd46406aadbb55580b4a833e4541/src/node/createSetupServer.ts#L86-L90

If the same instance of `emitter` is exposed to the developer, then when they will call `events.removeAllListeners()` with the intent to clear _their_ listeners, they will also be removing internal listeners, which will break the library. 

The same is true for the worker, as it adds internal listeners as well to control I/O from the worker's channel:

https://github.com/mswjs/msw/blob/9373ccf5e699fd46406aadbb55580b4a833e4541/src/setupWorker/setupWorker.ts#L65-L68

> I also dislike exposing the entire `emitter` because it also has the `.emit()` method, which developers should never call. 

Given this context, it seems that exposing the raw emitter may not be such a good idea. Even exposing the _same_ emitter reference is not a good idea, as developers should never remove the internal listeners added by the library.

eddiemoore:
Well that does throw a bit of a spanner in the works.

kettanaito:
Hey, @eddiemoore. Thank you once more for preparing this!

I'll close this pull request in favor of https://github.com/mswjs/msw/pull/868 where I've implemented the said interface of the life-cycle events. It will be possible to remove listeners from the events by calling either `server.events.removeAllListeners()` or `server.events.removeListener()` after the change is merged.

PR #867, opened 2021-08-17
fix: add prettier-ignore in mockServiceWorker

xyuanbuilds:
add comment, prevent this file be formated

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -1,5 +1,6 @@
 /* eslint-disable */
 /* tslint:disable */
+// prettier-ignore
 
 /**
  * Mock Service Worker (<PACKAGE_VERSION>).


kettanaito (commented):
Hey, @xyuanbuilds. Thank you for suggesting this.

I think you should ignore the entire `public` directory in your project as you never want it to be formatted with Prettier. This shouldn't be addressed on the worker script's side, in my opinion. 

```js
// .prettierignore
public
```

Please share your use-case when you wish for other files to be formatted but only the worker script gets in a way (I don't think that's ever the case).

xyuanbuilds:
> Hey, @xyuanbuilds. Thank you for suggesting this.
> 
> I think you should ignore the entire `public` directory in your project as you never want it to be formatted with Prettier. This shouldn't be addressed on the worker script's side, in my opinion.
> 
> ```js
> // .prettierignore
> public
> ```
> 
> Please share your use-case when you wish for other files to be formatted but only the worker script gets in a way (I don't think that's ever the case).

Thanks for your reply, i agree with just ignore the entire public directory, this was my own negligence

kettanaito:
No worries! Thank you for raising this anyway.

PR #908, opened 2021-09-11
fix(handleRequest): include handler header in the no-response warning

### LINKED ISSUE #907: Improved error message for returning undefined
I'm getting this error/warning:

```
[MSW] Expected a mocking resolver function to return a mocked response Object, but got: undefined. Original response is going to be used instead.
```

For this code in the photo I'm guessing.
1. Would be nice if the error mentioned the URL of the request that triggered the error so I'd be more sure where it's coming from
2. Would be nice if there were some way for me to silence the warning as it's intentional.

image

kettanaito:
- Fixes #907 

## Changes

The no-response warning now contains the relevant request handler header and call frame:

```
[MSW] Expected response resolver to return a mocked response Object, but got undefined. The original response is going to be used instead.
    
  • GET /user
    /path/to/handler:6:18
```

### DIFF
--- a/src/utils/handleRequest.ts
+++ b/src/utils/handleRequest.ts
@@ -86,8 +86,15 @@ export async function handleRequest<
   // as it may be an oversight on their part. Perform the request as-is.
   if (!response) {
     devUtils.warn(
-      'Expected a mocking resolver function to return a mocked response Object, but got: %s. Original response is going to be used instead.',
+      `\
+Expected response resolver to return a mocked response Object, but got %s. The original response is going to be used instead.\
+\n
+  \u2022 %s
+    %s\
+`,
       response,
+      handler.info.header,
+      handler.info.callFrame,
     )
 
     emitter.emit('request:end', request)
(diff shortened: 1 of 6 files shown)


src/utils/handleRequest.ts
```diff
@@ -86,8 +86,15 @@ export async function handleRequest<
   // as it may be an oversight on their part. Perform the request as-is.
   if (!response) {
     devUtils.warn(
-      'Expected a mocking resolver function to return a mocked response Object, but got: %s. Original response is going to be used instead.',
+      `\
+Expected response resolver to return a mocked response Object, but got %s. The original response is going to be used instead.\
+\n
```
kettanaito, line 91: Extra new-line is needed because the current `formatMessage` from the [outvariant](https://github.com/open-draft/outvariant) eats newlines for some reason.

kettanaito: Hmm, this must be the MSW issue somehow. This usage scenario works fine in `outvariant` (confirmed by tests).

Issue #851, opened 2021-08-06
Cannot resolve `debug` on yarn 2.x.x


grapgrap:
## Describe the bug

Yarn cannot resolve 'debug' of `msw`

```
✖ ｢wdm｣: ERROR in ../../.yarn/unplugged/msw-npm-0.33.3-7287fb2021/node_modules/msw/lib/esm/index.js 10:0-31
Module not found: Error: Can't resolve 'debug' in '/Users/a202006041/workspace/toolbox/.yarn/unplugged/msw-npm-0.33.3-7287fb2021/node_modules/msw/lib/esm'
 @ ./src/mock/worker.ts 4:0-40 5:12-21 9:20-37
 @ ./src/mock/index.ts 38:19-37
 @ ./src/index.tsx 10:0-41 12:0-15
```

## Environment

- `msw: 0.29.0`
- `nodejs: 14.16.1`
- `npm: 6.14.12`
- `yarn: 2.4.2`
- `webpack: 5.24.3`

> Please also provide your browser version.

## To Reproduce
1. Install `msw@0.29.0` above.
2. Run msw using webpack on `yarn@2.x.x` above

## Expected behavior

When i tried with `msw@0.28.2` and `yarn@1.x.x` each, It was ok.
I think module resolver cannot resolve **pnp module** that `debug` of `msw` external.

## Additional Information

I tried to install `debug` and config it to peer dependency on `.yarnrc.yml`.
And then it works.

I did not use `rollup` well. I think `msw` should have `debug` to peer dependency.

```
// package.json
{
  ...
  devDependencies: {
    ...
    "debug": "^4.3.2",
    "msw": "^0.33.3",
  }
  ...
}

```

```
// .yarnrc.yml

packageExtensions:
  'msw@*':
    peerDependencies:
      'debug': '*'
```

kettanaito:
Hey, @grapgrap. Thanks for reporting this.

Something seems to be off, but I don't have enough info to say on whose side it is. 

> I think msw should have debug to peer dependency.

MSW doesn't depend on `debug` directly, neither it uses that module. That module is likely installed by `@mswjs/interceptors` on which `msw` depends. The `debug` package is [properly specified as a dependency](https://github.com/mswjs/interceptors/blob/ab7068666337d692a639184e5e1b90114f432707/package.json#L55) in the interceptors package, so it must get installed when installing MSW.

kettanaito:
[...]

Issue #796, opened 2021-06-30
body stream already read when handling NextJS's devPagesManifest request


gCardinal:
## Describe the bug
When using the service worker to mock graphql queries, it seems like msw does something to queries that aren't supposed to be intercepted. I'm using NextJS and as you'll see in the screenshot below, calls to `http://crew.shadowshifts.local.majisti.com/_next/static/development/_devPagesManifest.json` will fail because when Next tries to parse the response, it fails with the error:

```
Failed to fetch devPagesManifest TypeError: Failed to execute 'json' on 'Response': body stream already read
    at eval (page-loader.ts?6f7a:67)
```

The code throwing this error is:

```
  getPageList() {
    if (process.env.NODE_ENV === 'production') {
      return getClientBuildManifest().then((manifest) => manifest.sortedPages)
    } else {
      if ((window as any).__DEV_PAGES_MANIFEST) {
        return (window as any).__DEV_PAGES_MANIFEST.pages
      } else {
        if (!this.promisedDevPagesManifest) {
          this.promisedDevPagesManifest = fetch(
            `${this.assetPrefix}/_next/static/development/_devPagesManifest.json`
          )
            .then((res) => res.json())
            .then((manifest) => {
              ;(window as any).__DEV_PAGES_MANIFEST = manifest
              return manifest.pages
            })
            .catch((err) => {
              console.log(`Failed to fetch devPagesManifest`, err)
            })
        }
        return this.promisedDevPagesManifest
      }
    }
  }
```

You'll notice a `.then((res) => res.json())` in the code above, which indicates that before Next even has a chance to process the response, it's already been parsed somehow.

Removing the service worker fixes the issue.

## Environment

- `msw: 0.30.0`
- `nodejs: 15.14.0`
- `npm: 7.7.6`

Brave, Version 1.26.67 Chromium: 91.0.4472.114 (Official Build) (64-bit)

## To Reproduce
I'm not sure I can get an easy reproduction case, it's happening in a pretty big monorepo and I can't make the code public. Moreover, the [example project here](https://github.com/vercel/next.js/tree/canary/examples/with-msw) does not produce the issue.

## Expected behavior
msw should not interfere at all with Next's dev server requests.

## Screenshots
https://i.imgur.com/9eOcRr6.png

kettanaito:
Hey, @gCardinal. Thank you for reporting this.

> msw should not interfere at all with Next's dev server requests.

[...]

PR #913, opened 2021-09-13
Externalize @mswjs/* packages in certain builds

### LINKED ISSUE #905: Cannot access response body after request is intercepted by MSW
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.35.0   |
| chrome | Version 91.0.4472.114     |
| Ubuntu     | 20.04.2 LTS     |

## Request handlers

I have the following setup for msw in my project. The request that I'm having an error should be bypassed.
```js
const setup = () => {
  if (typeof window === 'undefined') {
    const { server } = require('./server');
    server.listen();
  } else {
    const { worker } = require('./browser');
    worker.start({ onUnhandledRequest: 'bypass' });
  }
};
```

## Actual request

```js
const parseJSON = (text) => JSON.parse(text);

fetch('/url').then((res) => parseJSON(res.text()))
```

## Current behavior

I get the following error in my application

```
Uncaught (in promise) TypeError: Failed to execute 'text' on 'Response': body stream already read
    at parseJSON
```

## Expected behavior

I should be able to access the `response.text` in the promise resolve as the request was bypassed.

### LINKED ISSUE #912: ESM build does not work in the browser ("debug")
## Describe the bug
The es module build ships bare module imports, which can be resolved using a dev server or other tool. But when using the es module build in the browser it fails because it's importing the "debug" module which is actually a commonjs module. 

## Environment

- `msw: 0.35.0`
- `nodejs: 16.4.0`
- `npm: 7.18.0`

## To Reproduce
1. Create a simple HTML page that imports MSW: 
```html
<!DOCTYPE html>
<html>
  <head></head>
  <body>
    <script type="module">
      import * as mswjs from 'msw';
      console.log(mswjs);
    </script>
  </body>
</html>
```
2. Use a dev server that resolved bare imports, for example `@web/dev-server`: `npx web-dev-server --node-resolve --open`
3. Visit the HTML page
4. See this error: `Uncaught SyntaxError: The requested module './../../../debug/src/index.js' does not provide an export named 'default'`

## Expected behavior
MSW imports a browser compatible module.

patrickmcdougle-okta:
PR mentioned in https://github.com/mswjs/msw/pull/660#issuecomment-917238064

Maybe fixes: https://github.com/mswjs/msw/issues/905

~More work to be added possibly: https://github.com/mswjs/msw/issues/851#issuecomment-918450294 (which would likely fix https://github.com/mswjs/msw/issues/912)~ Not doing this here.

### DIFF
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -34,7 +34,7 @@ const buildEsm = {
     'src/graphql.ts',
     'src/context/index.ts',
   ],
-  external: ['debug'],
+  external: ['debug', /@mswjs\/interceptors/],
   output: {
     format: 'esm',
     entryFileNames: '[name].js',
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -28,7 +28,7 @@ const integrityPluginOptions = {
  * (and any relative paths under these packages).
  * @see https://github.com/mswjs/interceptors/issues/52
  */
-const mswjsRegex = /^@mswjs\/(interceptors|cookies)/
+const ecosystemDependencies = /^@mswjs\/(interceptors|cookies)/
 
 /**
  * Configuration for the ESM build.
@@ -41,7 +41,7 @@ const buildEsm = {
     'src/graphql.ts',
     'src/context/index.ts',
   ],
-  external: ['debug', mswjsRegex],
+  external: ['debug', ecosystemDependencies],
   output: {
     format: 'esm',
     entryFileNames: '[name].js',
@@ -111,7 +111,7 @@ const buildNode = {
     'tty',
     'os',
     'timers',
-    mswjsRegex,
+    ecosystemDependencies,
   ],
   output: {
     format: 'cjs',
@@ -142,7 +142,7 @@ const buildNode = {
  */
 const buildNative = {
   input: 'src/native/index.ts',
-  external: ['chalk', 'util', 'events', mswjsRegex],
+  external: ['chalk', 'util', 'events', ecosystemDependencies],
   output: {
     file: 'native/lib/index.js',
     format: 'cjs',
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -24,6 +24,13 @@ const integrityPluginOptions = {
 }
 
 /**
+ * Exclude @mswjs/interceptors and @mswjw/cookies
+ * (and any relative paths under these packages).
+ * @see https://github.com/mswjs/interceptors/issues/52
+ */
+const mswjsRegex = /^@mswjs\/(interceptors|cookies)/
+
+/**
  * Configuration for the ESM build.
  */
 const buildEsm = {
@@ -34,7 +41,7 @@ const buildEsm = {
     'src/graphql.ts',
     'src/context/index.ts',
   ],
-  external: ['debug', /@mswjs\/interceptors/],
+  external: ['debug', mswjsRegex],
   output: {
     format: 'esm',
     entryFileNames: '[name].js',
@@ -104,13 +111,7 @@ const buildNode = {
     'tty',
     'os',
     'timers',
-    '@mswjs/interceptors',
-    /**
-     * Exclude Node.js request interceptors from being compiled.
-     * @see https://github.com/mswjs/interceptors/issues/52
-     */
-    '@mswjs/interceptors/lib/interceptors/ClientRequest',
-    '@mswjs/interceptors/lib/interceptors/XMLHttpRequest',
+    mswjsRegex,
   ],
   output: {
     format: 'cjs',
@@ -141,18 +142,7 @@ const buildNode = {
  */
 const buildNative = {
   input: 'src/native/index.ts',
-  external: [
-    'chalk',
-    'util',
-    'events',
-    '@mswjs/interceptors',
-    /**
-     * Exclude Node.js request interceptors from being compiled.
-     * @see https://github.com/mswjs/interceptors/issues/52
-     */
-    '@mswjs/interceptors/lib/interceptors/ClientRequest',
-    '@mswjs/interceptors/lib/interceptors/XMLHttpRequest',
-  ],
+  external: ['chalk', 'util', 'events', mswjsRegex],
   output: {
     file: 'native/lib/index.js',
     format: 'cjs',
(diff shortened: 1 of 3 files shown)


patrickmcdougle-okta:
This isn't quite ready...I have to figure out how to get my corporate mirrors out of the yarn.lock file. Sorry.

rollup.config.ts
```diff
@@ -34,7 +34,7 @@ const buildEsm = {
     'src/graphql.ts',
     'src/context/index.ts',
   ],
-  external: ['debug'],
+  external: ['debug', /@mswjs\/interceptors/],
```
patrickmcdougle-okta: using a regex handles all relative imports from the package

patrickmcdougle-okta:
Let me know if you need me to clean up the history (squash) and / or let me know if you want me to handle the debug dependency problem in this body of work as well.

Thanks!

rollup.config.ts
kettanaito: I wonder if we should mark all `@mswjs/` packages as external? They are all dependencies, so they should be installed alongside the `msw` package. The issue with updates propagation doesn't stop on `interceptors` alone, we've also got packages like `@mswjs/cookies`, for example.

On the other hand, listing such packages _explicitly_ may be a better idea. That'd save us from any unintended externalization of the dependencies if we ever decide to use a `@mswjs/` package as a dev dependency, for example.

What do you say if we extend this regular expression to include `cookies`?

```js
/@mswjs\/(cookies|interceptors)/
```

kettanaito: I think we should also add this regular expression to the `buildNode` and `buildNative` targets. They currently externalize `@mswjs/interceptors` and some specific interceptors, which will be matched as well since you're introducing a regular expression.

kettanaito (changes requested):
Hey, @patrickmcdougle-okta. Thank you for preparing this, the changes look great!

I've had a few comments and would like to know your opinion. I think this is almost ready to go, just a few things left. Looking forward to releasing this!

Regarding the `debug` issue, let's address it in a separate pull request. I feel it may involve some changes in the biuld process of the interceptors library.

patrickmcdougle-okta:
Sorry for the delay, I was off yesterday. I'll make the proposed changes today. Thanks for your patience and your review!

patrickmcdougle-okta:
I rebased to pull in the current tip of master, but kept a separate new commit for easy reviewing. Let me know if you need me to squash anything.

rollup.config.ts
```diff
@@ -34,7 +41,7 @@ const buildEsm = {
     'src/graphql.ts',
     'src/context/index.ts',
   ],
-  external: ['debug'],
+  external: ['debug', mswjsRegex],
```
kettanaito: What do you think if we name this variable for what it represents intention-wise rather than what data type it holds?

```diff
-mswjsRegex
+ecosystemDependencies
```

kettanaito:
The current CI fails on the `init` integration test:

```
> msw@0.35.0 postinstall /root/release/tmp/auto-update-worker/node_modules/msw
> node -e "try{require('./config/scripts/postinstall')}catch(e){}"

/root/release/tmp/auto-update-worker/node_modules/inquirer/lib/prompts/rawlist.js:126
      this.selected = this.selected ?? -1;
                                     ^

SyntaxError: Unexpected token '?'
    at wrapSafe (internal/modules/cjs/loader.js:1054:16)
    at Module._compile (internal/modules/cjs/loader.js:1102:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Module.require (internal/modules/cjs/loader.js:1026:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Function.promptModule.restoreDefaultPrompts (/root/release/tmp/auto-update-worker/node_modules/inquirer/lib/inquirer.js:65:36)
    at Object.inquirer.createPromptModule (/root/release/tmp/auto-update-worker/node_modules/inquirer/lib/inquirer.js:72:16)
    at Object.<anonymous> (/root/release/tmp/auto-update-worker/node_modules/inquirer/lib/inquirer.js:84:28)
```

Seems related to #916.

**Edit:** You can ignore this failure, it won't happen anymore. The issue was fixed by the underlying dependency that introduced it in the first place.

rollup.config.ts
patrickmcdougle-okta: ~if we're going to go that far, you should just externalize all of your node_modules (which is generally standard practice for libraries). This feels scary and out of scope for the small bug I'm personally trying to fix and get released.~

EDIT: I didn't quite understand that you were only asking for a variable name change, not a full externalization of all dependencies...

patrickmcdougle-okta: but sure, I can update the variable name...

patrickmcdougle-okta:
I rebased again to pull in the current tip of master, but kept another separate commit for easy reviewing. Let me know if you need me to squash anything.

rollup.config.ts
patrickmcdougle-okta: Future work idea: use https://www.npmjs.com/package/rollup-plugin-auto-external to automatically exclude all "dependencies" from the esm/node bundles.

kettanaito (approved):
Thank you for this improvement, @patrickmcdougle-okta!

Collaborating with you on this was great. Let me know if you are interested in helping us with the project, there are always some tasks to work on.

Meanwhile, let’s have this one merged.

patrickmcdougle-okta:
Sorry to bother on a closed PR...Will there be a release with this code included soon? It fixes a bug I'm seeing in a product I'm building and I'd like to avoid maintaining a fork.

kettanaito:
No worries. I'd like to include a few more bug fixes in the next release. There's no ETA, but I was planning on releasing the next version somewhere in October.

misha-erm:
+1 for releasing this. I can't use msw right now because of it.

I described my use-case here https://github.com/mswjs/msw/issues/796#issuecomment-924912728

Thanks in advance

msutkowski:
@patrickmcdougle-okta @MikeYermolayev I'd recommend using the CSB build until the next release if you're in a bind. If you weren't aware, you can install as shown here: https://ci.codesandbox.io/status/mswjs/msw/pr/913/builds/169956

PR #925, opened 2021-09-27
Support a single number as a JSON request/response body

### LINKED ISSUE #920: ctx.json doesn't accept number as the value
## Describe the bug
`rest` methods doesn't accept `number` as type parameter for `RequestBodyType` or `ResponseBody` due do extending `DefaultRequestBody`:
```ts
export declare type DefaultRequestBody = Record<string, any> | DefaultRequestMultipartBody | string | undefined;
```
https://github.com/mswjs/msw/blob/master/src/handlers/RequestHandler.ts#L24
## Environment

- `msw: 0.34.0`
- `tsc: 4.3.5`

## To Reproduce
```ts
import { rest } from 'msw';

rest.post<string, number>('/api', (_, response, context) => {
  response(context.status(200));
});

>>  error TS2344: Type 'number' does not satisfy the constraint 'DefaultRequestBody'.

```

## Expected behavior
no  error TS2344 on reproduction snippet above

Shinyaigeek:
Fix: #920 

## Motivation

Hi team! 👋 
As mentioned in #920, a single number value should be handled as a valid JSON value, but `msw` currently cannot handle number as JSON.

> A JSON text is a sequence of tokens. The set of tokens includes six
structural characters, strings, numbers, and three literal names.

https://www.ietf.org/rfc/rfc7159.txt

## Did

- [x] update `DefaultRequestBody` which seems to be used as JSON or JSON string type in request body and response body.
- [x] update some parse logic in order to enable parser  to parse number value as JSON

### DIFF
--- a/src/context/fetch.ts
+++ b/src/context/fetch.ts
@@ -26,8 +26,11 @@ const createFetchRequestParameters = (input: MockedRequest): RequestInit => {
     return requestParameters
   }
 
-  requestParameters.body =
-    typeof body === 'object' ? JSON.stringify(body) : body
+  if (typeof body === 'object' || typeof body === 'number') {
+    requestParameters.body = JSON.stringify(body)
+  } else {
+    requestParameters.body = body
+  }
 
   return requestParameters
 }
--- a/src/handlers/RequestHandler.ts
+++ b/src/handlers/RequestHandler.ts
@@ -25,6 +25,7 @@ export type DefaultRequestBody =
   | Record<string, any>
   | DefaultRequestMultipartBody
   | string
+  | number
   | undefined
 
 export interface MockedRequest<Body = DefaultRequestBody> {
--- a/src/utils/request/parseBody.ts
+++ b/src/utils/request/parseBody.ts
@@ -3,7 +3,7 @@ import { jsonParse } from '../internal/jsonParse'
 import { parseMultipartData } from '../internal/parseMultipartData'
 
 /**
- * Parses a given request/response body based on the `Content-Type` header.
+ * Parses a given request/response body based on the "Content-Type" header.
  */
 export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
   // Return whatever falsey body value is given.
@@ -11,18 +11,18 @@ export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
     return body
   }
 
-  const contentType = headers?.get('content-type')
+  const contentType = headers?.get('content-type') || ''
 
   // If the body has a Multipart Content-Type
   // parse it into an object.
-  const hasMultipartContent = contentType?.startsWith('multipart/form-data')
+  const hasMultipartContent = contentType.startsWith('multipart/form-data')
   if (hasMultipartContent && typeof body !== 'object') {
     return parseMultipartData(body.toString(), headers) || body
   }
 
   // If the intercepted request's body has a JSON Content-Type
   // parse it into an object.
-  const hasJsonContent = contentType?.includes('json')
+  const hasJsonContent = contentType.includes('json')
 
   if (hasJsonContent && typeof body !== 'object') {
     return jsonParse(body.toString()) || body
--- a/src/utils/request/parseBody.ts
+++ b/src/utils/request/parseBody.ts
@@ -17,7 +17,7 @@ export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
   // parse it into an object.
   const hasMultipartContent = contentType?.startsWith('multipart/form-data')
   if (hasMultipartContent && typeof body !== 'object') {
-    return parseMultipartData(body, headers) || body
+    return parseMultipartData(body.toString(), headers) || body
   }
 
   // If the intercepted request's body has a JSON Content-Type
@@ -25,7 +25,7 @@ export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
   const hasJsonContent = contentType?.includes('json')
 
   if (hasJsonContent && typeof body !== 'object') {
-    return jsonParse(body) || body
+    return jsonParse(body.toString()) || body
   }
 
   // Otherwise leave as-is.
--- a/src/utils/internal/jsonParse.ts
+++ b/src/utils/internal/jsonParse.ts
@@ -3,10 +3,10 @@
  * Does not throw an exception on an invalid JSON string.
  */
 export function jsonParse<T extends Record<string, any>>(
-  str: string,
+  value: any,
 ): T | undefined {
   try {
-    return JSON.parse(str)
+    return JSON.parse(value)
   } catch (error) {
     return undefined
   }
--- a/src/utils/internal/jsonParse.ts
+++ b/src/utils/internal/jsonParse.ts
@@ -1,10 +1,10 @@
 /**
- * Parses a given string into a JSON.
+ * Parses a given value into a JSON.
  * Does not throw an exception on an invalid JSON string.
  */
-export function jsonParse<T extends Record<string, any>>(
+export function jsonParse<ValueType extends Record<string, any>>(
   value: any,
-): T | undefined {
+): ValueType | undefined {
   try {
     return JSON.parse(value)
   } catch (error) {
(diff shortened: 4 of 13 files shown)


src/utils/request/parseBody.ts
```diff
@@ -17,15 +17,15 @@ export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
...
+    return parseMultipartData(body.toString(), headers) || body
   }
 
   // If the intercepted request's body has a JSON Content-Type
   // parse it into an object.
   const hasJsonContent = contentType?.includes('json')
 
   if (hasJsonContent && typeof body !== 'object') {
-    return jsonParse(body) || body
+    return jsonParse(body.toString()) || body
```
Shinyaigeek, line 28: I wondered should I fix also `jsonParse` type to enable `jsonParse` to get also a number value because `JSON.parse` can handle a number value.

 Since `JSON.parse` converts its arguments to strings first in processing, I decided not to update `jsonParse`, feeling that this change is trivial in processing and not worth a complex change to the module interface.

How do you think about this?

FYI: https://tc39.es/ecma262/multipage/structured-data.html#sec-json.parse

src/handlers/RequestHandler.ts
```diff
@@ -25,6 +25,7 @@ export type DefaultRequestBody =
   | Record<string, any>
   | DefaultRequestMultipartBody
   | string
+  | number
```
kettanaito, line 28: Number is also a valid input when composing a `Response` instance, so this change should be fine. 

Just note that this change affects accepted types for _all_ mocked responses (not necessarily those you set via `ctx.json`, but also via `ctx.text`, `ctx.body`, etc.). 

```js
let res = new Response(2)
await res.text() // '2'
```

src/context/fetch.ts
```diff
@@ -26,8 +26,13 @@ const createFetchRequestParameters = (input: MockedRequest): RequestInit => {
     return requestParameters
   }
 
-  requestParameters.body =
-    typeof body === 'object' ? JSON.stringify(body) : body
+  if (typeof body === 'object') {
+    requestParameters.body = JSON.stringify(body)
+  } else if (typeof body === 'number') {
+    requestParameters.body = body.toString()
```
kettanaito: Converting number to string and stringifiying it via `JSON` yields the same result:

```js
number.toString() === JSON.stringify(number) // true
```

I suggest we revert this particular change. What do you think about this?

kettanaito: Can you please elaborate on what usage scenario are you covering with this change?

A `MockedRequest` instance should be well-formed when it's passed to `ctx.fetch`. It's a request representation, which, as we've discussed, cannot indeed have a number body (it'll always be a string). Perhaps I'm missing something. Having a test for this would be great.

src/utils/internal/jsonParse.test.ts
```diff
@@ -11,3 +11,7 @@ test('returns undefined without an error given an invalid JSON string', () => {
   expect(parse).not.toThrow()
   expect(parse()).toBeUndefined()
 })
+
+test('parse a single number as valid JSON string', () => {
```
kettanaito: Nitpick: could you please move this test case to be the second in the file?
We usually structure our tests to list the success scenarios first and end up with exceptional or failure scenarios.

src/utils/internal/jsonParse.test.ts
```diff
@@ -11,3 +11,7 @@ test('returns undefined without an error given an invalid JSON string', () => {
...
+  expect(jsonParse('1')).toEqual(1)
```
kettanaito: Is there a practical benefit of adding this test? `jsonParse` is a try/catch wrapper around native `JSON.parse`. What we're doing here is effectively testing how the latter behaves, which shouldn't be our responsibility (asserting that it indeed parses a valid stringified object in the first test case is sufficient). I think we should remove this test, as it's not connected with any change.

src/utils/request/parseBody.ts
kettanaito, line 28: This is strange. `JSON.parse` accepts numbers just fine but it's TS call signature is limited to accept only `string` input. I feel it's a slight overcomplication to convert everything to a string given `JSON.parse` can handle numbers just fine. I wonder what's your opinion on this?

kettanaito: `JSON.parse` indeed accepts _anything_ and attempts to convert it to a string if it's not. I think this is a TS limitation of `JSON.parse` interface being typed as accepting strings only:

```ts
interface JSON {
    /**
     * Converts a JavaScript Object Notation (JSON) string into an object.
     * @param text A valid JSON string.
     * @param reviver A function that transforms the results. This function is called for each member of the object.
     * If a member contains nested objects, the nested objects are transformed before the parent object is.
     */
    parse(text: string, reviver?: (this: any, key: string, value: any) => any): any;
}
```

I think we should adjust the call signature of `jsonParse` to accept `any` instead of converting every input to a string (which is redundant and is only needed to please TS).

```diff
export function jsonParse<T extends Record<string, any>>(
-  str: string,
+  value: any,
): T | undefined {
```

This is a safe change, as `jsonParse` is added precisely to have a safe JSON parsing, which includes given a malformed input (expected scenario). Once we do this, we can get rid of `.toString()` calls anywhere we wish to parse a non-string data with `jsonParse`. _This_ can now be reflected in unit tests for `jsonParse`.

kettanaito (changes requested):
Hey, @Shinyaigeek 👋 Thank you for following up on this!

There are a few points we should tackle before merging this:

1. Elaborate more on the `ctx.fetch` change and consider adding a test for it.
2. Change the call signature of `jsonParse` to accept `any` so we don't have to convert a non-string supported input value to a string (`JSON.parse` can natively handle numbers despite what its TS signature says).
3. Consider removing some tests that assert underlying standard API (`JSON.parse`, see my comments).

Let me know if you need any help with this. I'd like to hear your opinion on all the comments, so feel free to comment on those. 

Thank you once again! Looking forward to having this released.

src/utils/request/parseBody.ts
Shinyaigeek: I feel your proposal reasonable, so I updated `jsonParse` module in 9b5ad25 👍

src/utils/internal/jsonParse.test.ts
Shinyaigeek: I thought of jsonParse as a JSON (string) parser, not as a wrapper for native JSON.parse, and added this test to test whether it can parse JSON (string) correctly. However, it seems reasonable that it is the responsibility of the native JSON.parse to parse numbers, and that the jsonParser, which is its wrapper, does not need to do so. 
So I removed this test in cadd06a.

src/context/fetch.ts
Shinyaigeek: https://github.com/mswjs/msw/pull/925/files#diff-6202d60879ad8aca93f6a6a85bc330597531e1af8194d7ef3e691bcb8406a012R28

With this change, there is now a possibility that the body of a Request may contain a number type, in TypeScript type inference, and this process was added to deal with this.

I've been reviewing the codebase a bit, and it looks like the body may still be a number when the request is passed to createFetchRequestParameters (please let me know if I'm missing something).

If I were to add a test, should I add a test to see if createFetchRequestParameters will stringify the body of a number, like below?

```javascript
test('', () => {
  const request = createMockedRequest({
    body: 1,
    method: 'POST',
  })
  const { body } = createFetchRequestParameters(request)
  expect(body).toBe('1')
})
```

Shinyaigeek:
Hey @kettanaito 👋  Thank you for your reviewing and sorry for my late reply 🙇 

I added a bit of update and reply according to your indication, please check it when you have time 🙇

kettanaito:
Hey, @Shinyaigeek. Thank you once more for preparing these changes, they are great!

I've updated your feature branch, as well as added a few minor internal changes. I took this as an opportunity to polish the things around request/response body (such as improper types) and also added a few integration tests that ensure the library can handle request/response JSON bodies when there's a single number in them.

src/context/fetch.ts
kettanaito: I've added the integration tests for this change. It does behave correctly:

1. You can respond with a number in a mocked response: `ctx.json(123)`. The number is then properly extracted when you read the response as JSON.
2. You can send a number in a JSON request body (using a regular POST request). The number is then properly presented in `req.body` of your request handler.

kettanaito (approved):
Thank you for making MSW better! Welcome to contributors, @Shinyaigeek! 🎉

PR #931, opened 2021-10-05
feat: support multipart binary request bodies

### LINKED ISSUE #929: Binary body of a request gets corrupted in the handler
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.35.0   |
| browser | Microsoft Edge    |
| OS      | Windows 10    |

## Request handlers

Mocking setup via storybook add-on:

```js
import { rest } from "msw";

component.parameters = {
  msw: [
    rest.post(
      "/api/xyz/42",
      ( req,res, ctx ) => { ... }]];
```

In the req.body I expect a multipart content which owns an jpeg image.

In preview.js I have:

```js
import { initializeWorker, mswDecorator } from "msw-storybook-addon";

initializeWorker();
addDecorator(mswDecorator);
```

## Actual request

The request is a multipart content which contains a jpeg image sent by fetch.

## Current behavior

The data received in the request handler is a damaged jpeg file. After a lot debugging I found out that the image data gets damaged at two locations:

1. In mockServiceWorker.js function getResponse. There the body of the request is extracted by using request.text().
    This damages the binary data because the data is converted to utf-8. 
   
https://github.com/mswjs/msw/blob/11bb2441e102666184188f9f576ac65e59b1be9d/src/mockServiceWorker.js#L175
> The request body is always decoded using UTF-8
    
2. In RequestHandler.ts function parseMultipartData there is a call to the constructor of File with

https://github.com/mswjs/msw/blob/11bb2441e102666184188f9f576ac65e59b1be9d/src/utils/internal/parseMultipartData.ts#L87
    
> This fails because the content body given to the File constructor is encoded as UTF-8 but here the data have to be given as an ArrayBuffer.
    
## Expected behavior

Expected is that the data from the post request will not get damaged.
To prevent a damage of the image for 1. the content should be get as unicode string by using something like this:
   
```js 
const body = String.fromCharCode(...new Uint8Array(await request.arrayBuffer()))
```

For 2. we need transform the extracted data of the multipart data so the image correctly used for the creation of the File.
To transform it back from unicode string to ArrayBuffer we need soething like this:

```js
const encoded = Uint8Array.from([...contentBody].map(ch => ch.charCodeAt())).buffer;
const imageFile = new File(encoded, filename, { type: contentType });
```

kettanaito:
- Fixes #929

## Changes

- Adds a missing test case for the original (bypassed) binary response body.

### DIFF
--- a/src/utils/internal/parseMultipartData.ts
+++ b/src/utils/internal/parseMultipartData.ts
@@ -55,8 +55,8 @@ export function parseMultipartData<T extends DefaultRequestMultipartBody>(
 
   const [, ...directives] = contentType.split(/; */)
   const boundary = directives
-    .filter((d) => d.startsWith('boundary='))
-    .map((s) => s.replace(/^boundary=/, ''))[0]
+    .filter((directive) => directive.startsWith('boundary='))
+    .map((directive) => directive.replace(/^boundary=/, ''))[0]
 
   if (!boundary) {
     return undefined
--- a/test/rest-api/response/body/body-binary.test.ts
+++ b/test/rest-api/response/body/body-binary.test.ts
@@ -1,22 +1,55 @@
 import * as fs from 'fs'
 import * as path from 'path'
 import { pageWith } from 'page-with'
+import { createServer, ServerApi } from '@open-draft/test-server'
 
-test('responds with a given binary body', async () => {
-  const runtime = await pageWith({
+const imageBuffer = fs.readFileSync(
+  path.resolve(__dirname, '../../../fixtures/image.jpg'),
+)
+
+let httpServer: ServerApi
+
+function createRuntime() {
+  return pageWith({
     example: path.resolve(__dirname, 'body-binary.mocks.ts'),
   })
+}
+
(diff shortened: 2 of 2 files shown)


kettanaito:
The issue is related to the _request_ binary body. Respective tests must be added.

Ovaron1994:
Hi @kettanaito,

need help here, Added a test but I don't get the test to run.
Added body-form-binary-data.node.test.ts but it hands on formData.append(...)

Michael

kettanaito:
@Ovaron1994, please push your test to your branch where I could look at it. Thanks.

Ovaron1994:
@kettanaito I put the branch at my public fork: https://github.com/Ovaron1994/msw the branch is named: feature/test_to_show_bug.

Thanks

kettanaito:
I have hopes this will be covered by #1436

kettanaito:
This should be fixed in 2.0. Need to update the branch and still merge the test.

kettanaito:
This is supported in 2.0.

PR #940, opened 2021-10-13
getRequestCookies: check "document" and "location" explicitly

gerbyzation:
Following the conversation on discord this improves the environment detection to work when browser globals are polyfilled. This might happen as part of Server-Side Rendering front-end applications like react.

The tests use JSDOM to test msw in the context of the browser.`isNodeProcess` accurately detects that this is a node process. For the tests to work it would have to run like it's in a browser environment though. To accomplish that I considered mocking, but this would  only fix this for the tests of MSW, not for users using MSW in a JSDOM environment. Are there any other testing environments where this could break that we should account for?

### DIFF
--- a/src/utils/request/getRequestCookies.node.test.ts
+++ b/src/utils/request/getRequestCookies.node.test.ts
@@ -4,23 +4,25 @@
 import { getRequestCookies } from './getRequestCookies'
 import { createMockedRequest } from '../../../test/support/utils'
 
+const prevLocation = global.location
+
 beforeAll(() => {
-  // Node applications performing Server-Side Rendering of front-end applications
-  // might have to polyfill some browser globals like document, location etc.
+  // Node.js applications may polyfill some browser globals (document, location)
+  // when performing Server-Side Rendering of front-end applications.
   global.location = {
-    href: 'https://google.nl',
-    origin: 'https://google.nl',
+    href: 'https://mswjs.io',
+    origin: 'https://mswjs.io',
   } as Location
 })
 
 afterAll(() => {
-  global.location = undefined as unknown as Location
+  global.location = prevLocation
 })
 
 test('returns empty object when in a node environment with polyfilled location object', () => {
   const cookies = getRequestCookies(
     createMockedRequest({
-      url: new URL(`${location.origin}/user`),
+      url: new URL('/user', location.origin),
       credentials: 'include',
     }),
   )
--- a/src/utils/request/getRequestCookies.node.test.ts
+++ b/src/utils/request/getRequestCookies.node.test.ts
@@ -1,0 +1,29 @@
+/**
+ * @jest-environment node
+ */
+import { getRequestCookies } from './getRequestCookies'
+import { createMockedRequest } from '../../../test/support/utils'
+
+beforeAll(() => {
+  // Node applications performing Server-Side Rendering of front-end applications
+  // might have to polyfill some browser globals like document, location etc.
+  global.location = {
+    href: 'https://google.nl',
+    origin: 'https://google.nl',
+  } as Location
+})
+
+afterAll(() => {
+  global.location = undefined as unknown as Location
+})
+
+test('returns empty object when in a node environment with polyfilled location object', () => {
+  const cookies = getRequestCookies(
+    createMockedRequest({
+      url: new URL(`${location.origin}/user`),
+      credentials: 'include',
+    }),
+  )
+
+  expect(cookies).toEqual({})
+})
--- a/src/utils/request/getRequestCookies.ts
+++ b/src/utils/request/getRequestCookies.ts
@@ -1,13 +1,8 @@
 import * as cookieUtils from 'cookie'
-import { isNodeProcess } from 'is-node-process'
 import { MockedRequest } from '../../handlers/RequestHandler'
 
 function getAllCookies() {
   return cookieUtils.parse(document.cookie)
-}
-
-function isJSDOMEnvironment() {
-  return global.navigator?.userAgent?.includes('jsdom')
 }
 
 /**
@@ -17,7 +12,7 @@ export function getRequestCookies(request: MockedRequest) {
   /**
    * @note No cookies persist on the document in Node.js: no document.
    */
-  if (isNodeProcess() && !isJSDOMEnvironment()) {
+  if (typeof document === 'undefined' || typeof location === 'undefined') {
     return {}
   }
 
--- a/src/utils/request/getRequestCookies.ts
+++ b/src/utils/request/getRequestCookies.ts
@@ -1,8 +1,13 @@
 import * as cookieUtils from 'cookie'
+import { isNodeProcess } from 'is-node-process'
 import { MockedRequest } from '../../handlers/RequestHandler'
 
 function getAllCookies() {
   return cookieUtils.parse(document.cookie)
+}
+
+function isJSDOMEnvironment() {
+  return global.navigator?.userAgent?.includes('jsdom')
 }
 
 /**
@@ -12,7 +17,7 @@ export function getRequestCookies(request: MockedRequest) {
   /**
    * @note No cookies persist on the document in Node.js: no document.
    */
-  if (typeof location === 'undefined') {
+  if (isNodeProcess() && !isJSDOMEnvironment()) {
     return {}
   }
 


src/utils/request/getRequestCookies.ts
```diff
@@ -1,18 +1,23 @@
...
+import { isNodeProcess } from 'is-node-process'
 import { MockedRequest } from '../../handlers/RequestHandler'
 
 function getAllCookies() {
   return cookieUtils.parse(document.cookie)
 }
 
+function isJSDOMEnvironment() {
+  return global.navigator?.userAgent?.includes('jsdom')
+}
```
gerbyzation: Should this utility function move somewhere else?

kettanaito:
## Context

This change aims to fix an issue. The issue manifests from the `getRequestCookies` function in those cases when the environment has the `location` polyfilled but is not a JSDOM environment (misses `document` referenced later by the library). 

https://github.com/mswjs/msw/blob/11bb2441e102666184188f9f576ac65e59b1be9d/src/utils/request/getRequestCookies.ts#L15

In a nutshell, relying on the `location`'s existence alone is insufficient to guarantee we're running in a browser-compatible environment (browser/jsdom/etc.). In the @gerbyzation's case the `getRequestCookies` function doesn't short-circuits when it should, resulting in an exception when accessing the `document` later in that function:

https://github.com/mswjs/msw/blob/11bb2441e102666184188f9f576ac65e59b1be9d/src/utils/request/getRequestCookies.ts#L4-L6

src/utils/request/getRequestCookies.ts
```diff
@@ -1,18 +1,23 @@
...
+
 /**
  * Returns relevant document cookies based on the request `credentials` option.
  */
 export function getRequestCookies(request: MockedRequest) {
   /**
    * @note No cookies persist on the document in Node.js: no document.
    */
-  if (typeof location === 'undefined') {
+  if (isNodeProcess() && !isJSDOMEnvironment()) {
```
kettanaito: I think we should rather explicitly check each browser global we're relying on here. 

Relying on the environment alone is insufficient: JSDOM (and other browser-like environments) may polyfill the objects we need here, which means that retrieving the cookies would work. 

I wouldn't check for a specific environment (JSDOM) because there may be other browser-like environments used by other libraries. I do acknowledge JSDOM is rather widespread and is a safe choice but MSW has a strict policy of not introducing anything related to third-party libraries, which is JSDOM in this case. 

Fixing this condition to something like this should be enough:

```js
if (typeof location === 'undefined' || typeof document === 'undefined') {
  return {}
}
```

This way it doesn't matter what environment we're in, it's that environment's responsibility to ensure `location` and `document` are properly polyfilled if they are defined as global objects. 

What do you think about this?

src/utils/request/getRequestCookies.ts
kettanaito: We mustn't rely on specific libraries or agents. I've suggested another way to solve this, let me know your opinion on that suggestion. Thanks!

src/utils/request/getRequestCookies.node.test.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * @jest-environment node
+ */
+import { getRequestCookies } from './getRequestCookies'
+import { createMockedRequest } from '../../../test/support/utils'
+
+beforeAll(() => {
+  // Node applications performing Server-Side Rendering of front-end applications
+  // might have to polyfill some browser globals like document, location etc.
+  global.location = {
```
kettanaito, line 12: I wonder if we can utilize `jest.spyOn(global, 'location')` instead of global value assignment. It'd be easier to clean up afterwards as well by calling `jest.restoreAllMocks()`. 

Could you please give it a try?

kettanaito (changes requested):
Hey, @gerbyzation! This is a great improvement but I'd like to brainstorm with you on the most efficient way to fix this.

src/utils/request/getRequestCookies.node.test.ts
gerbyzation: It doesn't seem to like to spy on an undefined property. If you know of any way around this let me know
```
FAIL  src/utils/request/getRequestCookies.node.test.ts
  ● returns empty object when in a node environment with polyfilled location object

    Cannot spy the location property because it is not a function; undefined given instead

       9 |   // might have to polyfill some browser globals like document, location etc.
      10 |   // @ts-ignore
    > 11 |   jest.spyOn(global, 'location').mockReturnValue({
         |        ^
      12 |     origin: 'https://google.nl',
      13 |   })
      14 | })

      at ModuleMockerClass.spyOn (node_modules/jest-mock/build/index.js:792:15)
      at Object.<anonymous> (src/utils/request/getRequestCookies.node.test.ts:11:8)
```

When changing it to mock the getter: 
```
       FAIL  src/utils/request/getRequestCookies.node.test.ts
  ● returns empty object when in a node environment with polyfilled location object

    location property does not exist

       9 |   // might have to polyfill some browser globals like document, location etc.
      10 |   // @ts-ignore
    > 11 |   jest.spyOn(global, 'location', 'get').mockReturnValue({
         |        ^
      12 |     origin: 'https://google.nl',
      13 |   })
      14 | })

      at ModuleMockerClass._spyOnProperty (node_modules/jest-mock/build/index.js:878:13)
      at Object.<anonymous> (src/utils/request/getRequestCookies.node.test.ts:11:8)
      ```

src/utils/request/getRequestCookies.ts
gerbyzation: Yes, I was thinking about detecting the test environments later as well and I agree it is not a good idea to detect each test environment. 

For a test environment like JSDOM your suggestion would work well, it doesn't need to know specifically what environment it's in - enough of the browser is polyfilled that it can behave like it's in one, which is the purpose of these environments. 

For SSR applications this means though that instead of trying to detect the node environment correctly we'd now have to make it behave as if it was in a browser. A proper polyfill for `document.cookie` would not be appropriate as that would add persistence that is not isolated to the session of a user but globally and leak between different users. If that's just for `document.cookie` that would be manageable to provide a stub for but what is the chance that more browser features would have to be polyfilled/stubbed in the future? It might also create a dissonance, msw was started with `setupServer` which is intended for the node environment, this would make it as if it were in a browser environment in certain circumstances. 

I'll give it some more thought as I'm not sure yet what the best course of action would be. Let me know what your thoughts are

kettanaito: This function is agnostic of the environment and can run in both browser and Node.js, as long as the necessary global objects are present. There will be no issue of running it in a Node.js process if it polyfills `document` and `location`, which it does when emulating a browser environment. If an actual server code declares a global `document` object that's an issue with the code, `document` is a browser global. If you declare it, you are responsible for it (the same goes for location). 

I see no arguments against detecting specific objects this code relies on. It must be said that those objects are browser globals, so, primarily, this is a browser-oriented code. However, users may require to have request cookies persistence in JSDOM as well when testing respective code. They emulate the browser environment and expect the cookies to work.

> For SSR applications this means though that instead of trying to detect the node environment correctly we'd now have to make it behave as if it was in a browser

This function must also work for browser-like environments in Node.js. The environment itself is still Node.js, though. So instead of relying on tool-specific indicators (like Jest's agent), we can check the necessary global objects that are polyfilled in such case. 

> It might also create a dissonance, msw was started with setupServer which is intended for the node environment, this would make it as if it were in a browser environment in certain circumstances.

Not really. A Node.js environment may include browser polyfills if the developer _intends_ to emulate the browser environment (i.e. writing tests in JSDOM). In all other cases, browser globals must not be defined.

gerbyzation: > Not really. A Node.js environment may include browser polyfills if the developer intends to emulate the browser environment (i.e. writing tests in JSDOM). In all other cases, browser globals must not be defined.

I might have failed to illustrate my usecase clearly; I want to emulate browser environment for one thing (SSR) but not for the rest of my node application. Detecting the browser globals would put MSW in 'browser-mode' while what I really want is it to just work like it's in nodejs. The fact that we have SSR in this express app is unrelated to the use of MSW - ideally this wouldn't impact each other - SSR is already messy enough as is.

There are essentially two types of requests:
1. React app -> express backend
2. express backend -> microservices
We are mocking only some requests from the second category. If we'd mock the first category it would still be ideal to have node behaviour as we wouldn't want anything to persist across user requests in polyfills. 

We crossed off detecting the specific environments so the only other suggestion I can offer is to provide an option somewhere to explicitly say according/override to which env it should behave. 

This is my usecase and how I'd expect it to work ideally but I'm aware there are more things to consider here. I hope I've explained it clearly enough so you can make a decision on how you want to tackle this. Appreciate the thought you're putting into the work and happy to help get this over the line.

kettanaito: > Detecting the browser globals would put MSW in 'browser-mode'

That's not the case. It would only attempt to retrieve the request cookies from `localStorage`, that's all. MSW will still understand it's being run in Node.js (because you're using `msw/node`).

src/utils/request/getRequestCookies.node.test.ts
kettanaito: You're right, it's problematic to mock global properties this way.

kettanaito:
Hey, @gerbyzation. I've rewritten the condition to explicitly check `document` and `location` regardless of the environment. The test you've written passes so I think we should be good to go with this fix. 

Please let me know if I've missed anything.

src/utils/request/getRequestCookies.ts
gerbyzation: That's true, I was mainly trying to describe the behaviour but couldn't come up with a better way to phrase it

gerbyzation:
@kettanaito Looks good I think, looking at the code it seems you mean `document` and `localStorage` instead of `location` though?

kettanaito:
You're right, @gerbyzation, that was my mistake checking `localStorage` instead of `location`. I've rebased the feature branch and pushed the correct check.

kettanaito (approved):
Thank you for this improvement, @gerbyzation! Welcome to contributors 🎉

PR #954, opened 2021-10-28
Change references from master client to main client

### LINKED ISSUE #953: Usage of different word than "master" in src/mockServiceWorker.js
First off, I absolutely love msw so far. It's such an elegant solution to so many related problems across all kinds of testing.

One change I'm hoping that the community would consider is removing the word "master" from the msw. A lot of people feel really uncomfortable with usage of the word master in code and many Black developers find it quite offputting. While I would encourage any project community to consider removing the word "master", msw is in bit of a unique situation: the code in `src/mockServiceWorker.js` is included in the repo of any project using msw. For projects who want to be especially inclusive, using msw means [including a term](https://github.com/mswjs/msw/blob/55eb6544ea901b32b78f240e278a3b9122cd9d56/src/mockServiceWorker.js#L90) they're trying to remove from their code base.

Looking at the code, I don't believe that the `resolveMasterClient` function is considered part of the external API of msw. Additionally, it only seems to be [used once](https://github.com/mswjs/msw/blob/55eb6544ea901b32b78f240e278a3b9122cd9d56/src/mockServiceWorker.js#L112), in the `handleRequest` function. So I don't believe a change would cause any technical problems to users. Would the terms "canonical" or "primary" be equivalent for this use-case? If the msw team is agreeable, I am happy to make a (very small) PR with the change.

Thanks for your consideration and for the amazing work you've done with msw; it's truly an amazing tool!

wwahammy:
As discussed in ﻿#953, this PR changes to the name of a function in `mockServiceWorker.js` from `resolveMasterClient` to `resolveMainClient`. The goal of this change is to be more inclusive.

As mentioned in the linked issue, this function name does not appear to be in the public API for msw so it's unlikely to cause compatibility issues.

Closes #953

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -83,11 +83,11 @@ self.addEventListener('message', async function (event) {
   }
 })
 
-// Resolve the "master" client for the given event.
+// Resolve the "main" client for the given event.
 // Client that issues a request doesn't necessarily equal the client
 // that registered the worker. It's with the latter the worker should
 // communicate with during the response resolving phase.
-async function resolveMasterClient(event) {
+async function resolveMainClient(event) {
   const client = await self.clients.get(event.clientId)
 
   if (client.frameType === 'top-level') {
@@ -109,7 +109,7 @@ async function resolveMasterClient(event) {
 }
 
 async function handleRequest(event, requestId) {
-  const client = await resolveMasterClient(event)
+  const client = await resolveMainClient(event)
   const response = await getResponse(event, client, requestId)
 
   // Send back the response clone for the "response:*" life-cycle events.


wwahammy:
I'm unsure why the the integration tests are failing. They seem to work on my computer and I don't know how anything I changed would have caused them to fail.

kettanaito:
@wwahammy, don't worry, your changes cannot cause the tests to fail. It's related to some unreliable behavior of `page-with`/Playwright when running on CI. I've rerun that failed job just now.

kettanaito (approved):
Thank you for this improvement, @wwahammy! Welcome to contributors 🎉

PR #959, opened 2021-10-30
fix: add missing debug dependency

jdanil:
`debug` is required in `msw` via the rollup config, but it not declared as a dependency. Strict node linkers like pnp and pnpm throw errors when building a project with `msw` as it cannot access the `debug` package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "chalk": "^4.1.1",
     "chokidar": "^3.4.2",
     "cookie": "^0.4.1",
+    "debug": "^4.3.2",
     "graphql": "^15.5.1",
     "headers-utils": "^3.0.2",
     "inquirer": "^8.1.5",


kettanaito:
Hi, @jdanil. Thank you for submitting this change!

There's an extensive discussion about this in #851. MSW itself never requires `debug`, neither in its source code nor in its Rollup config. We mark `debug` as _external_ to prevent it from being bundled if some of our dependencies rely on `debug` (like `@mswjs/interceptors` does!).

https://github.com/mswjs/msw/blob/f6fbd6c94d86d54ee13d20f0128589332cbddcbf/rollup.config.ts#L44

Listing a module in the dependencies while the package does not, in fact, use or depend on it is not semantic. I agree that there may be issues caused by this fourth-party dependency externalization but marking such dependencies as MSW's dependencies is not the solution here. 

As a side note, we're considering adopting `debug` in MSW to be able to run request interception in a debug mode in both browser and Node.js. This issue may get obsolete once we do this.

jdanil:
Hi @kettanaito, thanks for the explanation! Sorry I missed that existing issue, just had a read through it now and the linked `debug` issue.

IMO after reading through that it seems to me that it would make sense for `debug` to be a dependency of `msw` and a peer dependency of `@msw/interceptors` (rather than being declared as a dependency as it is currently).

No worries though, I can work around it locally by adding it to yarn's package extensions.

Will keep an eye out on any future changes in `msw` to adopt `debug` and hopefully pnp/pnpm users will be able to remove the custom package extension in the future.

Thanks for your work on `msw` btw. Its been working fantastically for us.

PR #974, opened 2021-11-07
fix(rest): support ReadableStream as response body

### LINKED ISSUE #581: ReadableStream object could not be cloned when trying to use stream as a mocked response body
This issue originated from the earlier issue: https://github.com/mswjs/msw/issues/156

If we need msw to support SSE, then we need to make the code work:
```js
export const worker = setupWorker(
  rest.get('/sse', (req, res, ctx) => {
    return res(
      ctx.status(200)
    , ctx.set('Content-Type', 'text/event-stream')
    , ctx.body(sse(function* () {
        yield 'message1'
        yield 'message2'
      }))
    )
  })
)

function sse(gfn) {
  let iter
  return new ReadableStream({
    start() {
      iter = gfn()
    }
  , pull(controller) {
      controller.enqueue(`data: ${iter.next().value}\n\n`)
    }
  })
}
```

Browser:
```
[MSW] Request handler function for "GET http://localhost:8080/sse" has thrown the following exception:

DOMException: Failed to execute 'postMessage' on 'MessagePort': ReadableStream object could not be cloned.
```

Node.js:
```
TypeError: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of ReadableStreamTypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of ReadableStream

      at ClientRequestOverride.<anonymous> (node_modules/node-request-interceptor/src/interceptors/ClientRequest/ClientRequestOverride.ts:216:34)
      at step (node_modules/node-request-interceptor/lib/interceptors/ClientRequest/ClientRequestOverride.js:33:23)
      at Object.next (node_modules/node-request-interceptor/lib/interceptors/ClientRequest/ClientRequestOverride.js:14:53)
      at fulfilled (node_modules/node-request-interceptor/lib/interceptors/ClientRequest/ClientRequestOverride.js:5:58)
```

_Originally posted by @BlackGlory in https://github.com/mswjs/msw/issues/156#issuecomment-739362584_

Aprillion:
Fixes #581 .. but in a totally hacky* way and only tested with a stream of string chunks.

If improvements are needed before merge, please feel free to edit the PR or close it and just take inspiration for some better solution 😅  But I'm afraid I won't have more time until next weekend.

*hacky => original code counts that there will be only 1 message from client, and that will resolve a Promise... but when multiple messages come back, only one of them can resolve a promise - but even after the Promise is resolved, the onmessage handler lives inside the closure of a new Promise function, so it's a bit unexpected that we can handle multiple messages there even AFTER the Promise is resolved

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -203,6 +203,10 @@ async function getResponse(event, client, requestId) {
       )
     }
 
+    case 'MOCK_STREAM_START': {
+      return respondWithStream(clientMessage)
+    }
+
     case 'MOCK_NOT_FOUND': {
       return getOriginalResponse()
     }
@@ -305,8 +309,28 @@ function sendToClient(client, message) {
     const channel = new MessageChannel()
 
     channel.port1.onmessage = (event) => {
-      if (event.data && event.data.error) {
-        return reject(event.data.error)
+      if (event.data) {
+        if (event.data.error) {
+          return reject(event.data.error)
+        }
+
+        switch (event.data.type) {
+          // message 'MOCK_STREAM_START' resolves the Promise, but then the client sends
+          // more messages from a ReadableStream (see issue #581)
+          case 'MOCK_STREAM_CHUNK': {
+            if (event.data.payload) {
+              globalThis.currentStreamController.enqueue(
+                event.data.payload.chunk,
+              )
+            }
+            return
+          }
+
+          case 'MOCK_STREAM_END': {
+            globalThis.currentStreamController.close()
+            return
+          }
+        }
       }
 
       resolve(event.data)
@@ -324,6 +348,23 @@ function delayPromise(cb, duration) {
 
 function respondWithMock(clientMessage) {
   return new Response(clientMessage.payload.body, {
+    ...clientMessage.payload,
+    headers: clientMessage.payload.headers,
+  })
+}
+
+function respondWithStream(clientMessage) {
+  const stream = new ReadableStream({
+    start(controller) {
+      globalThis.currentStreamController = controller
+    },
+  }).pipeThrough(
+    // from https://web.dev/fetch-upload-streaming/#streaming-request-bodies
+    // > Each chunk of a [response] body needs to be a Uint8Array
+    new TextEncoderStream(),
+  )
+
+  return new Response(stream, {
     ...clientMessage.payload,
     headers: clientMessage.payload.headers,
   })
--- a/test/rest-api/response/body/body-stream.mocks.ts
+++ b/test/rest-api/response/body/body-stream.mocks.ts
@@ -1,0 +1,24 @@
+import { setupWorker, rest } from 'msw'
+
+const stream = new ReadableStream({
+  start(controller) {
+    controller.enqueue('line1\n')
+
+    setTimeout(() => {
+      controller.enqueue('line2\n')
+      controller.close()
+    }, 0)
+  },
+})
+
+const worker = setupWorker(
+  rest.get('/sse', (req, res, ctx) => {
+    return res(
+      ctx.status(200),
+      ctx.set('Content-Type', 'text/plain'),
+      ctx.body(stream),
+    )
+  }),
+)
+
+worker.start()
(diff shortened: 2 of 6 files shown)


src/mockServiceWorker.js
```diff
@@ -28,6 +28,9 @@ self.addEventListener('message', async function (event) {
   }
 
   const client = await self.clients.get(clientId)
+  if (event.data !== 'KEEPALIVE_REQUEST') {
+    console.log(event.data, client)
+  }
```
Aprillion: aaah, one moment please, I need to hunt down these logs :D

edit: now removed

Aprillion:
`yarn test:smoke` is passing locally, but failing in CI with `Chrome Headless 83.0.4103.61` .. is there some setup I can do to reproduce the failure?

src/mockServiceWorker.js
```diff
@@ -309,6 +313,22 @@ function sendToClient(client, message) {
         return reject(event.data.error)
       }
 
+      switch (event.data?.type) {
```
kettanaito: You can't use optional chaining, as well as many other modern ES features, in the worker script. This is also why the smoke test fails (older Chromium browsers don't support this syntax). I didn't know the newer browsers support this, it's neat.

test/rest-api/response/body/body-stream.node.test.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * @jest-environment node
+ */
+
+test('Node does not support ReadableStream yet', () => {
+  expect(typeof ReadableStream).toBe('undefined')
```
kettanaito, line 6: This test needs to be rewritten, it doesn't test anything as of yet.

test/rest-api/response/body/body-stream.mocks.ts
```diff
@@ -0,0 +1,24 @@
...
+      controller.close()
+    }, 0)
+  },
+})
+
+const worker = setupWorker(
+  rest.get('/sse', (req, res, ctx) => {
+    return res(
+      ctx.status(200),
+      ctx.set('Content-Type', 'text/plain'),
```
kettanaito, line 18: I believe that such a response must also specify that its body is a stream.

test/rest-api/response/body/body-stream.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+test('responds with a ReadableStream response body', async () => {
+  const { request } = await pageWith({
+    example: path.resolve(__dirname, 'body-stream.mocks.ts'),
+  })
+
+  const res = await request('/sse')
+  const status = res.status()
+  const headers = await res.allHeaders()
+  const text = await res.text()
```
kettanaito, line 12: We need to ensure that `res.text()` actually verifies that you can read that response stream once more on the usage surface (assert that we're correctly piping the original response stream into `PassThrough` in the worker).

test/rest-api/response/body/body-stream.node.test.ts
Aprillion: the idea was to fail the test if Node adds ReadableStream ¯\\\_(ツ)_/¯ if you have something in mind that can be actually tested, that would be awesome!

test/rest-api/response/body/body-stream.mocks.ts
Aprillion: apparently not necessary for a stream of `text/plain`, at least in the integration test setup..

src/mockServiceWorker.js
Aprillion: looks like object spread was already used in `{...clientMessage.payload}`, so I validated the updated script in https://eslint.org/demo using version ES2018

kettanaito:
I'm going to close this in favor of #1288. This proposed change has served as an excellent example of how to reconstruct a readable stream within the worker. I've gone with a slightly different approach but this was tremendously helpful.

kettanaito:
## Released: v0.43.0 🎉

This has been released in v0.43.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.43.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.43.0)

---

PR #979, opened 2021-11-11
fix: no-unused-vars lint warnings

chentsulin:
Ignore all variables start with `_`.

Before:

截圖 2021-11-11 下午12 31 19

After:

截圖 2021-11-11 下午12 36 32

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,5 +20,11 @@ module.exports = {
     ],
     '@typescript-eslint/no-namespace': 0,
     '@typescript-eslint/no-var-requires': 0,
+    '@typescript-eslint/no-unused-vars': [
+      1,
+      {
+        argsIgnorePattern: '^_',
+      },
+    ],
   },
 }
--- a/test/msw-api/setup-server/scenarios/custom-transformers.test.ts
+++ b/test/msw-api/setup-server/scenarios/custom-transformers.test.ts
@@ -11,7 +11,7 @@ const jsonBig = (body: Record<string, any>): ResponseTransformer => {
 }
 
 const server = setupServer(
-  rest.get('http://test.mswjs.io/me', (req, res, ctx) => {
+  rest.get('http://test.mswjs.io/me', (req, res) => {
     return res(
       jsonBig({
         username: 'john.maverick',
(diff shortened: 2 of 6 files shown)


test/msw-api/setup-server/scenarios/custom-transformers.test.ts
```diff
@@ -11,7 +11,7 @@ const jsonBig = (body: Record<string, any>): ResponseTransformer => {
 }
 
 const server = setupServer(
-  rest.get('http://test.mswjs.io/me', (req, res, ctx) => {
+  rest.get('http://test.mswjs.io/me', (req, res, _ctx) => {
```
kettanaito: I think we should drop the unused arguments if they are the last in the arguments list. 

```diff
-(req, res, _ctx)  => {
+(req, res) => {
```

.eslintrc.js
```diff
@@ -20,5 +20,11 @@ module.exports = {
     ],
     '@typescript-eslint/no-namespace': 0,
     '@typescript-eslint/no-var-requires': 0,
+    '@typescript-eslint/no-unused-vars': [
+      1,
+      {
+        argsIgnorePattern: '^_',
```
kettanaito, line 26: That's a great suggestion!
What do you think if we scope this rule only to arguments? 

It's still possible to declare a variable starting with an underscore and forget to use it. In that case, we should get warnings from eslint. I think the current use-case is that we wish to handle arguments with this rule.

kettanaito (changes requested):
Hey, @chentsulin!

Thank you for taking initiative with this improvement. I've left two comments: one suggests to scope this eslint rule to arguments only, the second to consider maybe we should drop the unused arguments altogether if they are the last ones in the list. 

I'd like to hear your opinion on those points.

.eslintrc.js
chentsulin: To the best of my knowledge, this [`argsIgnorePattern`](https://eslint.org/docs/rules/no-unused-vars#argsignorepattern) option should only apply to arguments. There is another option called [`varsIgnorePattern `](https://eslint.org/docs/rules/no-unused-vars#varsignorepattern).

test/msw-api/setup-server/scenarios/custom-transformers.test.ts
chentsulin: Sure. I like the dropping approach.

.eslintrc.js
kettanaito: You're completely right! My bad, I've confused the option name 👍 This looks great.

kettanaito (approved):
Thank you for making our repo better, @chentsulin! Welcome to contributors! 🎉

PR #981, opened 2021-11-11
feat: Support for GraphQL extensions to be mocked

OLingard:
**Description**

With regards to the GraphQL spec found [here](http://spec.graphql.org/October2021/#sec-Response-Format), we can see that a GraphQL response with data will also include extensions which should contain meta data. There is a large use case to be able to mock out these extensions to test tracking and any additional fields users include here.

**Methodology**

Building off the back of the work done to allow data and errors to be mocked in the same request (see #403), I added in a new context method and generalised the data type. Also adjusted the folder structure to keep the GraphQL-specific methods together.

**Comments**

I have not updated the docs with a section for extensions so the doc string points to a broken link


src/context/graphql/data.test.ts
```diff
@@ -3,7 +3,7 @@
  */
```
kettanaito: I think we should preserve the flat directory structure of `src/context`. The idea is that directories resemble the API you're using, and you are still using `ctx.data` and `ctx.extensions`. I know it makes sense to group GraphQL-specific utilities respectively, but there's no downside of not doing so.

src/context/graphql/extensions.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        },
+      },
+      data: {
+        hello: 'world',
+      },
+    }),
+  )
+})
+
+test('sets given extensions on the response JSON body in the presence of data and errors', async () => {
```
kettanaito: Thank you for adding these tests!

We'd also have to include a few integration tests (for browser and Node.js) so we're sure `ctx.extensions` behaves as expected in those environments when using a complete MSW setup. 

Would you be interested in adding such tests? 

It's a matter of adding two new test suites:

- `test/graphql-api/extensions.test.ts` (to test the feature's behavior in a browser)
- `test/graphql-api/extensions.node.test.ts` (to test the feature's behavior in Node.js)

You can take a look at the existing [browser](https://github.com/mswjs/msw/blob/master/test/graphql-api/query.test.ts) and [Node.js](https://github.com/mswjs/msw/blob/8bfcf6e39392790bb164c2dbd80cb28e8b57b140/test/graphql-api/cookies.node.test.ts#L50-L68) GraphQL integration tests for inspiration. 

> Note that you don't have to add an actual `createServer` with the GraphQL resolution. `setupServer` + GraphQL request handler using `ctx.extensions()` would be enough.

kettanaito (changes requested):
Hey, @OLingard! Thank you for this improvement!

I've left a few comments and would love to hear your opinion on those. 

- Don't restructure the GraphQL context directory.
- Add some integration tests for browser and Node.js

I've also noticed that your commits are not associated with your GitHub user. This likely means that your local Git config doesn't have the correct user set. Please follow [this guide](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address#setting-your-email-address-for-every-repository-on-your-computer) to update your existing commits. Push the updated commits once ready.

src/context/graphql/data.test.ts
OLingard: Hey Kettanaito, I agree with you here to follow the API layout, the only reason I changed it was due to creating the generalised type `GraphQLPayloadContext`. In this folder structure, should I rename this file or move it to a different folder to be clear that it is used for Graphql data structures?

src/context/graphql/extensions.test.ts
OLingard: Will add these today!

kettanaito: Thank you for adding the tests!

src/context/graphql/data.test.ts
kettanaito: I see, thanks for sharing the insights. Let me think about where's the right place to put that type...

kettanaito: I've moved the type to `src/typeUtils.ts`, I think that's the right place.

kettanaito:
Thank you for addressing the comments, @OLingard! You've done a superb job on this feature. 

I've rebased your feature branch and added a tad of polishing. Let's wait for the CI to pass before making `graphql.extensions` happen.

kettanaito (approved):

kettanaito:
Welcome to contributors, @OLingard! 👏

kettanaito:
Please, if you're interested in tackling any other open issues, don't hesitate to chime in and share your opinion/take the initiative in opening a pull request. I'm certain your expertise would be appreciated! Thank you for being awesome.

PR #982, opened 2021-11-12
fix(getAbsoluteUrl): fix full url check

MarvelSQ:
Fix absolute url check.

## Changes

- Adds a new `isAbsoluteUrl` function that determines if the given URL is absolute. 
- Replaces a simple `.startsWith('/')` logic in `getAbsoluteUrl` to use `isAbsoluteUrl`.
- Supports `api/xyz` paths to convert to `http://domain/api/xyz`

### DIFF
--- a/src/utils/url/getAbsoluteUrl.ts
+++ b/src/utils/url/getAbsoluteUrl.ts
@@ -9,8 +9,8 @@ export function getAbsoluteUrl(path: string, baseUrl?: string): string {
     return path
   }
 
-  // Ignore path with pattern start with :*
-  if (/^[:*]/.test(path)) {
+  // Ignore path with pattern start with *
+  if (path.startsWith('*')) {
     return path
   }
 
--- a/src/utils/url/getAbsoluteUrl.ts
+++ b/src/utils/url/getAbsoluteUrl.ts
@@ -1,9 +1,16 @@
+import { isAbsoluteUrl } from './isAbsoluteUrl'
+
 /**
  * Returns an absolute URL based on the given path.
  */
 export function getAbsoluteUrl(path: string, baseUrl?: string): string {
-  // Ignore absolute URLs.
-  if (!path.startsWith('/')) {
+  // already absolute URL
+  if (isAbsoluteUrl(path)) {
+    return path
+  }
+
+  // Ignore path with pattern start with :*
+  if (/^[:*]/.test(path)) {
     return path
   }
 
--- a/src/utils/matching/normalizePath.test.ts
+++ b/src/utils/matching/normalizePath.test.ts
@@ -39,10 +39,12 @@ test('removes query parameters and hashes from a relative URL', () => {
 })
 
 test('returns a path pattern string as-is', () => {
-  expect(normalizePath(':api/user')).toEqual(':api/user')
+  expect(normalizePath(':api/user')).toEqual('http://localhost/:api/user')
   expect(normalizePath('*/resource/*')).toEqual('*/resource/*')
 })
 
 test('removeß query parameters and hashes from a path pattern string', () => {
-  expect(normalizePath(':api/user?query=123#some')).toEqual(':api/user')
+  expect(normalizePath(':api/user?query=123#some')).toEqual(
+    'http://localhost/:api/user',
+  )
 })
--- a/src/utils/url/getAbsoluteUrl.test.ts
+++ b/src/utils/url/getAbsoluteUrl.test.ts
@@ -24,6 +24,6 @@ test('returns an absolute URL given a relative path without a leading slash', ()
 })
 
 test('returns a path with a pattern as-is', () => {
-  expect(getAbsoluteUrl(':api/user')).toEqual(':api/user')
+  expect(getAbsoluteUrl(':api/user')).toEqual('http://localhost/:api/user')
   expect(getAbsoluteUrl('*/resource/*')).toEqual('*/resource/*')
 })
--- a/src/utils/url/getAbsoluteUrl.test.ts
+++ b/src/utils/url/getAbsoluteUrl.test.ts
@@ -18,3 +18,12 @@ test('returns a given absolute URL as-is', () => {
     'https://api.mswjs.io/users',
   )
 })
+
+test('returns an absolute URL given a relative path without a leading slash', () => {
+  expect(getAbsoluteUrl('users')).toEqual('http://localhost/users')
+})
+
+test('returns a path with a pattern as-is', () => {
+  expect(getAbsoluteUrl(':api/user')).toEqual(':api/user')
+  expect(getAbsoluteUrl('*/resource/*')).toEqual('*/resource/*')
+})
--- a/src/utils/url/isAbsoluteUrl.test.ts
+++ b/src/utils/url/isAbsoluteUrl.test.ts
@@ -1,0 +1,32 @@
+/**
+ * @jest-environment node
+ */
+import { isAbsoluteUrl } from './isAbsoluteUrl'
+
+test('returns true for the "http" scheme', () => {
+  expect(isAbsoluteUrl('http://www.domain.com')).toEqual(true)
+})
+
+test('returns true for the "https" scheme', () => {
+  expect(isAbsoluteUrl('https://www.domain.com')).toEqual(true)
+})
+
+test('returns true for the "ws" scheme', () => {
+  expect(isAbsoluteUrl('ws://www.domain.com')).toEqual(true)
+})
+
+test('returns true for the "ftp" scheme', () => {
+  expect(isAbsoluteUrl('ftp://www.domain.com')).toEqual(true)
+})
+
+test('returns true for the custom scheme', () => {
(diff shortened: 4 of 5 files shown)


src/utils/url/getAbsoluteUrl.ts
```diff
@@ -2,8 +2,8 @@
  * Returns an absolute URL based on the given path.
  */
 export function getAbsoluteUrl(path: string, baseUrl?: string): string {
-  // Ignore absolute URLs.
-  if (!path.startsWith('/')) {
+  // Ignore Full URL
+  if (path.startsWith('http') || path.startsWith('https')) {
```
kettanaito: Is there an issue you're encountering related to this change?

So far this introduces a degression of logic: absolute URLs do not necessarily start with HTTP/HTTPS protocol. There are also `ftp://`, `data://` and [many other URL schemes](https://en.wikipedia.org/wiki/List_of_URI_schemes). Instead of accounting for each individual scheme, we reverse the predicate, checking if the URL is relative. 

I'd like to learn more about what you're trying to solve by this.

src/utils/url/getAbsoluteUrl.ts
```diff
@@ -2,8 +2,8 @@
  * Returns an absolute URL based on the given path.
  */
 export function getAbsoluteUrl(path: string, baseUrl?: string): string {
-  // Ignore absolute URLs.
-  if (!path.startsWith('/')) {
+  // Ignore Full URL
```
kettanaito: There's no such thing as "full" URL 😉 There are relative and absolute URLs.

kettanaito (changes requested):
Hey, @MarvelSQ. Thank you for opening this. Please take a look at the comments I've left.

src/utils/url/getAbsoluteUrl.ts
MarvelSQ: i just want to use the path without first `/`, like `api/xxx`
for the next, i'd like to set prefix as `basename` to complete the path
```js
// handlers
import { rest } from 'msw';
export const handlers = [
  rest.get('api/xxx', (req, res, ctx) => {
       // handle request
  })
]

// entry
import { worker } from './mock/browser';
worker.start({
  resolutionContext: {
     baseUrl: `${location.origin}/`
  }
});
```
i notice there is `resolutionContext` inside handleRequest

----------

for absolute URL check.
`protocol RegExp` could be better solution

src/utils/url/getAbsoluteUrl.ts
MarvelSQ: sorry for my bad comment

src/utils/url/getAbsoluteUrl.ts
kettanaito: I wonder what benefits do you gain from the setup above? The behavior you get is the default behavior in MSW: all relative requests are resolved against the current origin. A request is considered relative if its URL starts from a leading slash "/". Do you have any reasons not to write request URLs that way?

src/utils/url/getAbsoluteUrl.ts
kettanaito: No worries!

src/utils/url/getAbsoluteUrl.ts
MarvelSQ: in my case, the site location is dynamic, from `/` to `/subpath/`, to prevent change the path all over the project, i set up a `<base href="/">`, and change api as relative.
```js
// no slash start
export const API_USER_INFO = 'api/user_info';
```
in MSW, i need change every handler path
```js
import { rest } from 'msw';
import { API_USER_INFO } from 'api/user_info';

export const handlers = [
  rest.get(`/${API_USER_INFO}`, ....)
]
```
so, i think if msw provide a default prefix would be convenient. then i found change `getAbsoluteUrl` is easier way to fulfill my request.
```diff
import { rest } from 'msw';
import { API_USER_INFO } from 'api/user_info';

export const handlers = [
-  rest.get(`/${API_USER_INFO}`, ....)
+  rest.get(API_USER_INFO, ...)
]
```

MarvelSQ: @kettanaito should this request continue？

src/utils/url/getAbsoluteUrl.ts
```diff
@@ -2,8 +2,8 @@
  * Returns an absolute URL based on the given path.
  */
 export function getAbsoluteUrl(path: string, baseUrl?: string): string {
-  // Ignore absolute URLs.
-  if (!path.startsWith('/')) {
+  // Ignore url with any protocol or starts with *、:
+  if (/^[a-z][a-z0-9+.-]*:/.test(path) || /^[:*]/.test(path)) {
```
kettanaito: Do you think we can achieve the same result by including the `://` in the regular expression?

```suggestion
  if (/^[a-z][a-z0-9+.-]*:\/\//.test(path)) {
```

kettanaito: As the complexity of determining the absolute URL increases, I suggest we isolate this logic into a utility function called `isAbsoluteUrl`. We can then cover that function with unit tests to ensure it behaves as expected. Would you be interested in helping me with this?

src/utils/url/getAbsoluteUrl.ts
kettanaito: @MarvelSQ, I think your use case makes sense and we should not treat `one/two` as an absolute URL. You can proceed with the fix, I've left a few comments.

src/utils/url/getAbsoluteUrl.ts
MarvelSQ: @kettanaito in current test case, the result have no difference. but pattern url like `:api/user`, `*` failed

MarvelSQ: for test, i think `getAbsoluteUrl` test case can be used for `isAbsoluteUrl`

kettanaito: This expression seems to cover the use cases we need:

```js
/^[a-zA-Z-]+:\/\//
```

> I don't think we should account for numbers to be a part of the protocol (is that even permitted?)

Could you please confirm that?

The issue with the current approach is the permissive `*` token, which allows _zero_ or more matches. It's the "zero" part that we should replace with `+` (one or more matches). For the URL to be absolute it must have `something://`.

kettanaito: Not really. `getAbsoluteUrl` accepts a path and ensures that the absolute path is always returned. `isAbsoluteUrl` accepts a path and returns a boolean indicating if that path _is already absolute_. Those are different requirements.

kettanaito: Or did you perhaps mean just to collocate the tests in the single `getAbsoluteUrl.test` file? In that case, I'd still prefer to have a `[module].test.ts` file convention, even if it's a small utility function (absolutely no harm in that).

MarvelSQ: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler#permitted_schemes
custom protocol allow `web+`, so numbers are possible

MarvelSQ: ok, set a `isAbsoluteUrl.test.ts`

MarvelSQ: https://github.com/axios/axios/blob/master/lib/helpers/isAbsoluteURL.js
axios regexp allow numbers

src/utils/url/getAbsoluteUrl.ts
```diff
@@ -1,9 +1,16 @@
...
 export function getAbsoluteUrl(path: string, baseUrl?: string): string {
-  // Ignore absolute URLs.
-  if (!path.startsWith('/')) {
+  // already absolute URL
+  if (isAbsoluteUrl(path)) {
+    return path
+  }
+
+  // Ignore path with pattern start with :*
+  if (/^[:*]/.test(path)) {
```
kettanaito: Need to revisit if this change is necessary. Without it, relative paths with patterns will resolve to `:user` -> `http://localhost/:user`, which would then transform to a regular expression. That is the correct behavior.

MarvelSQ: the pattern match effect multiple top-level test case, it's easy to keep the same behavior.
i  agree `:user` should convert to `http://localhost/:user`
```suggestion
  if (/^\*/.test(path)) {
```
`*` as previous discuss, should be left

kettanaito: Hm. But `*` won't match `isAbsoluteUrl` and the given path will be returned as-is. Am I missing something?

MarvelSQ: sorry, i mean `*` will catch by regexp and ignored.

I changed the conditions and modified the related unit test cases. 

But code failed in integration test. I see some type declaration errors

kettanaito (approved):
Thank you for working on this and addressing all the comments. Great job.

MarvelSQ:
happy to contribute 🎉

Issue #975, opened 2021-11-09
Compilation error when running `tsc` after installation


rvenugopal:
## Environment

- `msw: 0.35.0`
- `nodejs: 12.2.1`
- `npm: 0.0.0`
- `tsc: 3.9.7`

> Please also provide your browser version.
NA.  This is on Node.

## To Reproduce
1. Install msw 
2. Run `tsc -p .`
3. See error below
```
node_modules/msw/lib/types/context/set.d.ts:7:71 - error TS1110: Type expected.

7 export declare type ForbiddenHeaderError<HeaderName extends string> = `SafeResponseHeader: the '${HeaderName}' header cannot be set on the response. Please use the 'ctx.cookie()' function instead.`;
                                                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

[1:21:56 PM] Found 1 error. Watching for file changes.

```

My tsconfig looks like the following

```
{
  "compilerOptions": {
    "module": "commonjs",
    "strict": true,
    "target": "ES6",
    "noEmit": false,
    "incremental": true,
    "inlineSourceMap": true,
    "removeComments": true,
    "outDir": "dist",
    "lib": ["es6", "es2016.array.include", "es2017.object"],
    "alwaysStrict": true,
    "traceResolution": false,
    "baseUrl": "./",
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "resolveJsonModule": true,
    "esModuleInterop": true,
    "strictPropertyInitialization": false
  },
  "include": [
    "app.ts",
    "config",
    "migration",
    "src",
    "test"
  ]  
}

```

## Expected behavior
Compilation succeeds

kettanaito:
Hey, @rvenugopal.

TypeScript 3.9 is not supported by the MSW ecosystem. Please use the newer version of TypeScript (the latest is 4.4, the minimum required should be 4.2).

rvenugopal:
Is this information about minimum typescript version documented?  Thanks @kettanaito

willbuckingham:
Yup, just ran into this one myself, thank fully found this, I think ideally this would have a peer dependency for typescript if that is the case.

kettanaito:
[...]

PR #985, opened 2021-11-16
Add optional peer dependency for typescript

willbuckingham:
Based on conversation in https://github.com/mswjs/msw/issues/975

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,6 +141,14 @@
     "webpack": "^5.68.0",
     "webpack-dev-server": "^3.11.2"
   },
+  "peerDependencies": {
+    "typescript": "4.2.x"
+  },
+  "peerDependenciesMeta": {
+    "typescript": {
+      "optional": true
+    }
+  },
   "resolutions": {
     "chokidar": "3.4.1"
   },


package.json
```diff
@@ -130,6 +130,14 @@
     "webpack": "^5.44.0",
     "webpack-dev-server": "^3.11.2"
   },
+  "peerDependencies": {
+    "typescript": ">=4.2.0"
+  },
+  "peerDependenciesMeta": {
+    "typescript": {
+      "optional": true
```
kettanaito, line 149: Does this have anything to do with the pre-installation of peer dependencies in npm 7?

> UPDATE: npm versions 1, 2, and 7 will automatically install peerDependencies if they are not explicitly depended upon higher in the dependency tree. For npm versions 3 through 6, you will receive a warning that the peerDependency is not installed instead.

package.json
```diff
@@ -130,6 +130,14 @@
     "webpack": "^5.44.0",
     "webpack-dev-server": "^3.11.2"
   },
+  "peerDependencies": {
+    "typescript": ">=4.2.0"
```
kettanaito: This range will target anything above 4.2.0. Perhaps let's provide an upper version boundary?

```suggestion
    "typescript": ">=4.2.0 < 5"
```

It may be running ahead of ourselves but we shouldn't promise compatibility with _any_ version higher than 4.2.0.

kettanaito (changes requested):
Hey, @willbuckingham. Thank you for making this improvement happen! I've left a few thoughts and would love to hear your opinion on them.

package.json
kettanaito: And that is a **horrible suggestion** on my side. 

I've recently learned that [TypeScript doesn't follow semver](https://github.com/microsoft/TypeScript/issues/14116). What should be the major version is, actually, a "market" version. The next one, `.2.` is indeed the _major_ version. In other words, there are breaking changes between 4.2 and 4.3 and that's intended on the TypeScript's side. 

This means we should lock this peer dependency range to:
```suggestion
    "typescript": "4.2.x"
```

package.json
kettanaito: Per NPM:

> When a user installs your package, npm will emit warnings if packages specified in `peerDependencies` are not already installed. The peerDependenciesMeta field serves to provide npm more information on how your peer dependencies are to be used. Specifically, it allows peer dependencies to be marked as optional.
> — [NPM docs](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta)

So `optional` here controls whether to emit the missing peer dependency warning but does not prevent NPM 7 from always installing `typescript`. Looks like we need to research the way to opt-out of that behavior so that JavaScript developers don't get their `node_modules` polluted by the packages they won't use.

willbuckingham: I can try and run some actual tests on npm 7 in the future, but if I'm reading this correctly and it's still accurate it seems like this should be fine.

> peerOptional deps are deps that are listed in peerDependencies and also have an entry in peerDependenciesMeta containing {"optional": true}. These sorts of dependencies are not installed by default, but if installed, must be resolved to a version that satisfies the optional peer dependency relationship.
https://github.com/npm/rfcs/discussions/289#discussioncomment-131068

package.json
```diff
@@ -130,6 +130,14 @@
     "webpack": "^5.44.0",
     "webpack-dev-server": "^3.11.2"
   },
+  "peerDependencies": {
+    "typescript": ">=4.2.0 < 5"
```
kettanaito: I've recently learned that TS doesn't follow semver. There may, and most likely will be breaking changes between 4.2 and 4.3. We should look the supported version by the minor range:
```suggestion
    "typescript": "4.2.x"
```

Please, would you add this change?

package.json
kettanaito: @willbuckingham, thank you for sharing that excerpt. We can proceed with the `optional` flag and iterate on the developers' feedback once we get some.

kettanaito (changes requested):
Thank you for the great work on this, @willbuckingham. I've left one comment regarding the TypeScript version range given my latest findings of how TS is versioned. Would love to hear your opinion on the suggestions.

package.json
willbuckingham: So you had originally said:

> Please use the newer version of TypeScript (the latest is 4.4, the minimum required should be 4.2).

So I'm not sure how to reconcile that with this.

willbuckingham: perhaps what you want is `>= 4.2.x <= 4.4.x` (I'll have to check if that's valid syntax

kettanaito: I've said that initially before learning that TypeScript [ships breaking changes between minor releases](https://github.com/microsoft/TypeScript/issues/14116#issuecomment-280592571). With that knowledge, I suggest we now lock the peer dependency version to exact major + minor versions, allowing only the latest patches.

willbuckingham: Ok, it's just typescript is up to 4.5.2, but if you are only comfortable saying it works for 4.2.x, it's your lib, up to you of course.

kettanaito: We can bump the supported version to 4.5.2 or anything else once we have the guarantee the library works with that version. That means switching to that version in dev dependencies and using it locally. I've had quite a number of painful minor version updates in TS so I highly recommend we don't take minor versions compatibility as a given. 

Going with `4.2.x` is the correct approach at the moment 👍

kettanaito (dismissed):
Thank you, @willbuckingham! 
Let's see if the smoke tests pass, there have been some issues with them on this branch.

kettanaito:
Some tests fail reliably with this change. I'd have to take a look at what's causing this. There are no installation warnings due to missing `typescript` dependency, all examples (which are used for smoke tests) are written in JavaScript.

amaury-hanser:
Hello, a quick comment to tell you that I'm interested in this PR.  
I don't have the knowledge to fix it, but it's a PR that could be nice for the community. :)

kettanaito:
Hey, @amaury-hanser! Thank you for showing interest in this change. 

The task itself is complete here but we've got some issues with the smoke tests failing. I don't see a direct relation between adding an optional dependency and any failing tests, but there must be something we're missing. 

I've rebased this branch to use our recent GitHub Actions CI. Let's see how the build goes.

kettanaito:
This CI will pass now as I've removed smoke tests from the pull request jobs. It still fails the smoke tests, which would be nice to investigate.

kettanaito:
This is likely to fail smoke tests. My initial guess is that the examples are rather outdated and it may force exceptions during the installation of MSW (although TS is marked as an optional dependency).

package.json
```diff
@@ -141,6 +141,14 @@
     "webpack": "^5.68.0",
     "webpack-dev-server": "^3.11.2"
   },
+  "peerDependencies": {
+    "typescript": "4.2.x"
```
kettanaito, line 145: Since we are about to merge #1073 (path parameter inference), I fail to remember which version of TypeScript added that feature. 

- [4.2 Release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-2.html)
- [4.3 Release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html)

kettanaito: The pull request description mentions 4.1. Great!

kettanaito:
## Released: v0.40.0 🎉

---

piotr-cz:
I'm not able to update to v0.40.0 because of dependency conflict (I have already installed `typescript@^4.6.4`):

npm output:
```
Could not resolve dependency:
peerOptional typescript@"4.2.x" from msw@0.40.0
node_modules/msw
  msw@"*" from the root project
Conflicting peer dependency: typescript@4.2.4
node_modules/typescript
  peerOptional typescript@"4.2.x" from msw@0.40.0
  node_modules/msw
    msw@"*" from the root project
```

kettanaito:
Please see the discussion in https://github.com/mswjs/msw/pull/1241.

PR #994, opened 2021-11-22
feat: export Worker and Server api options type

yannbf:
As part of the work done in the [Storybook addon for MSW](https://github.com/mswjs/msw-storybook-addon/pull/61), we need certain types which are not exported directly from MSW.

This PR exports the options used both in `worker.start` and `server.listen`.

## Roadmap

- [x] Export `SetupServerApi` interface

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,11 @@ import * as context from './context'
 export { context }
 
 export { setupWorker } from './setupWorker/setupWorker'
-export { SetupWorkerApi } from './setupWorker/glossary'
+export type {
+  SetupWorkerApi,
+  PartialStartOptions,
+} from './setupWorker/glossary'
+export type { SharedOptions } from './sharedOptions'
 export {
   response,
   defaultResponse,
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,12 +2,42 @@ import * as context from './context'
 export { context }
 
 export { setupWorker } from './setupWorker/setupWorker'
-export type { SetupWorkerApi, StartOptions } from './setupWorker/glossary'
-export type { SharedOptions } from './sharedOptions'
 export {
   response,
   defaultResponse,
   createResponseComposition,
+} from './response'
+
+/* Request handlers */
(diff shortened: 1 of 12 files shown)


src/index.ts
```diff
@@ -2,7 +2,11 @@ import * as context from './context'
 export { context }
 
 export { setupWorker } from './setupWorker/setupWorker'
-export { SetupWorkerApi } from './setupWorker/glossary'
+export type {
+  SetupWorkerApi,
```
kettanaito: What about the `SetupServerApi`? Isn't that needed as well?

yannbf: Good one! I didn't add it because it can be accessed via:

`import type { SetupServerApi } from 'msw/node`

WDYT?

src/setupWorker/glossary.ts
```diff
@@ -181,10 +181,11 @@ export interface SerializedResponse<BodyType = any> {
   body: BodyType
 }
 
+export type PartialStartOptions = PartialDeep<StartOptions>
```
kettanaito: This may be a good chance to review if we need a non-partial `StartOptions` anywhere in the code. 

I don't think that introducing this custom type is a good idea. Or, to be precise, I don't think having to do `PartialDeep<StartOptions>` everywhere is a good idea. I think we should mark `StartOptions` as optional on a per-property basis, utilizing `Partial` for options that equal objects (like `options.serviceWorker`). 

Please, would you be interested in looking into this? If you're strict on time, I can take it from here and let you know.

src/index.ts
kettanaito: That's exactly how it should work. Nice!

src/setupWorker/glossary.ts
yannbf: I unfortunately do not have time at the moment to dig deeper in the codebase. I'd appreciate if you could take it from here and I'd love to continue watching the PR!

kettanaito: No worries. I've added this suggestion. The only thing that remains is to export the `SetupServerApi`.

kettanaito (approved):

PR #999, opened 2021-11-24
fix(RestHandler): set default "req.params" to ambiguous type

kettanaito:
## Breaking changes

- Move the `Params` generic of `rest` handlers to be the _second_ one:
```diff
- rest.get<RequestBodyType, ResponseBodyType, RequestParameters>
+ rest.get<RequestBodyType, RequestParameters, ResponseBodyType>
```
- Removes the `RequestParams` type. Use `PathParams` type instead. 
- Sets the `ResponseBody` generic of `rest` handlers to `DefaultRequestBody` by default. 

## Motivation

Request path parameters that do not have an explicit generic type must have a default ambiguous type of `string | string[]`. Previously `any`.

```ts
rest.get('/user/:id', (req, res, ctx) => {
  const id = req.params.id // string | string[]
})
```

Request path parameter may also be an array of strings in the case of the path including multiple parameters with the same name:

```ts
rest.get('/user/:id/message/:id', (req, res, ctx) => {
  const id = req.params.id // ["a", "b"] given /user/a/message/b
})
```

Until the consumer provides the explicit parameters type generic to `rest.get` (or any other `rest` handler), the library should treat path parameters as ambiguous.

### DIFF
--- a/test/msw-api/setup-server/scenarios/generator.node.test.ts
+++ b/test/msw-api/setup-server/scenarios/generator.node.test.ts
@@ -2,49 +2,61 @@ import fetch from 'node-fetch'
 import { rest } from 'msw'
 import { setupServer } from 'msw/node'
 
+type RequestParams = {
+  maxCount: string
+}
+
 const server = setupServer(
-  rest.get('/polling/:maxCount', function* (req, res, ctx) {
-    let count = 0
+  rest.get<never, RequestParams>(
+    '/polling/:maxCount',
+    function* (req, res, ctx) {
+      const maxCount = parseInt(req.params.maxCount)
+      let count = 0
 
-    while (count < req.params.maxCount) {
-      count += 1
-      yield res(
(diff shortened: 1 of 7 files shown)


test/msw-api/setup-server/scenarios/generator.node.test.ts
```diff
@@ -2,49 +2,61 @@ import fetch from 'node-fetch'
...
+  maxCount: string
+}
+
 const server = setupServer(
-  rest.get('/polling/:maxCount', function* (req, res, ctx) {
-    let count = 0
+  rest.get<never, RequestParams>(
+    '/polling/:maxCount',
+    function* (req, res, ctx) {
+      const maxCount = parseInt(req.params.maxCount)
```
kettanaito, line 13: Request path parameters must be transformed to the expected type (i.e. string -> number).

PR #1008, opened 2021-11-27
feat: use document.baseURI as the relative request's base URL

### LINKED ISSUE #1007: Use "document.baseURI" as the relative request's base URL
## Current behavior

We use `location.origin` to be the base URL for relative requests.

https://github.com/mswjs/msw/blob/445fc63f9447df2df346c64d1a751d43b284e52a/src/utils/url/getAbsoluteUrl.ts#L10-L12

While this works, certain scenarios won't have a location set (i.e. sandboxed iframes). Relative request URL resolution would fail in those cases.

## Expected behavior

Use [`document.baseURI`](https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI) instead, which is always set on the document.

watife:
- Closes #1007 

use document.baseURI as the relative request's base URL

### DIFF
--- a/src/utils/url/getAbsoluteUrl.ts
+++ b/src/utils/url/getAbsoluteUrl.ts
@@ -8,8 +8,9 @@ export function getAbsoluteUrl(path: string, baseUrl?: string): string {
   }
 
   // Resolve a relative request URL against a given custom "baseUrl"
-  // or the current location (in the case of browser/browser-like environments).
-  const origin = baseUrl || (typeof location !== 'undefined' && location.origin)
+  // or the document baseURI (in the case of browser/browser-like environments).
+  const origin =
+    baseUrl || (typeof document !== 'undefined' && document.baseURI)
 
   return origin
     ? // Encode and decode the path to preserve escaped characters.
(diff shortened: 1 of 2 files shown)


src/utils/url/getAbsoluteUrl.ts
```diff
@@ -8,8 +8,8 @@ export function getAbsoluteUrl(path: string, baseUrl?: string): string {
   }
 
   // Resolve a relative request URL against a given custom "baseUrl"
-  // or the current location (in the case of browser/browser-like environments).
-  const origin = baseUrl || (typeof location !== 'undefined' && location.origin)
+  // or the document baseURI (in the case of browser/browser-like environments).
+  const origin = baseUrl || document.baseURI
```
kettanaito: Let's keep the `typeof` guard but check the existence of the `document` instead:

```suggestion
  const origin = baseUrl || (typeof document !== 'undefined' && document.baseURI)
```

This code also runs in Node.js, where `document` won't exist. The guard is a great way to fall back to using `document.baseURI` whenever there _is_ a document (which may be the case in Node.js if you're using browser polyfills like jsdom).

watife: I will update the PR now. Thanks!

kettanaito:
@watife, I've rebased your feature branch and formatted the changes. Let's wait for the CI.

kettanaito (approved):
Thank you for making this happen, @watife! Welcome to the contributors 🎉

PR #1019, opened 2021-12-07
fix(matchRequestUrl): replace negative lookbehind regexp

### LINKED ISSUE #1018: SyntaxError: Invalid group specifier name in Safari (0.36.0)
## Describe the bug
When loading our app in the Safari browser we see an error in the console which prevents the app from loading.

## Environment

- `msw: 0.36.0`
- `nodejs: 17.2.0`
- `npm: 8.1.4`

Safari v14.1.2

## To Reproduce
Load an app which uses MSW in the Safari browser.

## Expected behavior
App to load normally.

## Screenshots

Screenshot 2021-12-07 at 10 07 30

kettanaito:
- Fixes #1018 

## Changes

- Removes the negative lookbehind regular expression (`?<!`) that's not supported in Safari.
- Replaces the said expression with a positive lookup (optional path parameters are now a capturing group).
- Moves the wildcard coercion _before_ the protocol escaping so it could handle simpler paths.
- Restructures tests around `matchRequestUrl` slightly.

### DIFF
--- a/src/utils/matching/matchRequestUrl.test.ts
+++ b/src/utils/matching/matchRequestUrl.test.ts
@@ -3,66 +3,91 @@
  */
 import { coercePath, matchRequestUrl } from './matchRequestUrl'
 
-test('returns true when matched against a wildcard', () => {
-  const match = matchRequestUrl(new URL('https://test.mswjs.io'), '*')
-  expect(match).toHaveProperty('matches', true)
-  expect(match).toHaveProperty('params', {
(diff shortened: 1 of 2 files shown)


kettanaito:
## Issue: Unexpected MODIFIER at 25, expected END

The error was thrown by `path-to-regexp` because the paths like `/us*` were not coerced properly:

```diff
-/us*
+/us(.*)
```

PR #1021, opened 2021-12-08
fix(onUnhandledRequest): ignore regexp method/operationName

### LINKED ISSUE #1020: TypeError: expected.toLowerCase() is not a function when using "rest.all()"
## Describe the bug
When using `rest.all()` to mock all requests of a given path, it blocks HMR (and other requests).

That's due to an issue with the usage of `isStringEqual` in the file [`src/utils/request/onUnhandledRequest.ts`](https://github.com/mswjs/msw/blob/master/src/utils/request/onUnhandledRequest.ts#L60). 
It assumes that `method` is always a string and with the introduction of `rest.all` it seems to be a `RegExp` now.

## Environment

- `msw: 0.36.0`
- `nodejs: 14.17.6`
- `npm: 6.14.15`
- `yarn: 1.22.15`
- `Chrome: Version 96.0.4664.55 (Official Build) (x86_64)`

## To Reproduce
1. Scaffold a create-react-app
2. Set up MSW with a `rest.all()` handler (mine was like this: `rest.all('/todos', (req, res, ctx) => {...});`
3. Run `yarn start`
4. Make a change on one of the files
5. See error in the console like the ones below:

## Expected behavior
Requests not targeted by the `rest.all()` should still be performed.
By temporarily patching the rest-deps changing the return to:
```js
return actual.toString().toLowerCase() === expected.toString().toLowerCase();
```
solved the issue, but I know that's not the right fix 😄 

## Screenshots
HMR broken:
image

(also messes requests for the chrome extension)
image

kettanaito:
- Fixes #1020 

## Changes

- Does not print any suggestions if the `method` or `operationName` of the handler is a `RegExp`. In that case, it's impossible to suggest anything because the regular expression is a dynamic sort of thing. 
- Adds missing unit tests for the `onUnhandledRequest` function.

### DIFF
--- a/src/utils/request/onUnhandledRequest.ts
+++ b/src/utils/request/onUnhandledRequest.ts
@@ -183,7 +183,9 @@ Read more: https://mswjs.io/docs/getting-started/mocks\
 
       // Throw an exception to halt request processing and not perform the original request.
       throw new Error(
-        'Cannot bypass a request when using the "error" strategy for the "onUnhandledRequest" option.',
+        devUtils.formatMessage(
+          'Cannot bypass a request when using the "error" strategy for the "onUnhandledRequest" option.',
+        ),
       )
     }
 
--- a/src/utils/request/onUnhandledRequest.ts
+++ b/src/utils/request/onUnhandledRequest.ts
@@ -15,7 +15,7 @@ const MAX_MATCH_SCORE = 3
 const MAX_SUGGESTION_COUNT = 4
 const TYPE_MATCH_DELTA = 0.5
 
-type UnhandledRequestCallback = (request: MockedRequest) => void
+export type UnhandledRequestCallback = (request: MockedRequest) => void
 
 export type UnhandledRequestStrategy =
   | 'bypass'
@@ -23,11 +23,13 @@ export type UnhandledRequestStrategy =
   | 'error'
   | UnhandledRequestCallback
 
-function groupHandlersByType(handlers: RequestHandler[]) {
-  return handlers.reduce<{
-    rest: RestHandler[]
-    graphql: GraphQLHandler[]
-  }>(
+interface RequestHandlerGroups {
+  rest: RestHandler[]
+  graphql: GraphQLHandler[]
+}
+
+function groupHandlersByType(handlers: RequestHandler[]): RequestHandlerGroups {
+  return handlers.reduce<RequestHandlerGroups>(
     (groups, handler) => {
       if (handler instanceof RestHandler) {
         groups.rest.push(handler)
@@ -46,18 +48,23 @@ function groupHandlersByType(handlers: RequestHandler[]) {
   )
 }
 
-type RequestHandlerSuggestionList = [number, RequestHandler][]
-type ScoreGetterFn = (request: MockedRequest, handler: RequestHandler) => number
+type RequestHandlerSuggestion = [number, RequestHandler]
 
-function getScoreForRestHandler(): ScoreGetterFn {
+type ScoreGetterFn<RequestHandlerType extends RequestHandler> = (
+  request: MockedRequest,
+  handler: RequestHandlerType,
+) => number
+
+function getRestHandlerScore(): ScoreGetterFn<RestHandler> {
   return (request, handler) => {
     const { path, method } = handler.info
 
-    if (path instanceof RegExp) {
+    if (path instanceof RegExp || method instanceof RegExp) {
       return Infinity
     }
 
     const hasSameMethod = isStringEqual(request.method, method)
+
     // Always treat a handler with the same method as a more similar one.
     const methodScoreDelta = hasSameMethod ? TYPE_MATCH_DELTA : 0
     const requestPublicUrl = getPublicUrlFromRequest(request)
@@ -67,15 +74,20 @@ function getScoreForRestHandler(): ScoreGetterFn {
   }
 }
 
-function getScoreForGraphQLHandler(
+function getGraphQLHandlerScore(
   parsedQuery: ParsedGraphQLQuery,
-): ScoreGetterFn {
+): ScoreGetterFn<GraphQLHandler> {
   return (_, handler) => {
     if (typeof parsedQuery.operationName === 'undefined') {
       return Infinity
     }
 
     const { operationType, operationName } = handler.info
+
+    if (typeof operationName !== 'string') {
+      return Infinity
+    }
+
     const hasSameOperationType = parsedQuery.operationType === operationType
     // Always treat a handler with the same operation type as a more similar one.
     const operationTypeScoreDelta = hasSameOperationType ? TYPE_MATCH_DELTA : 0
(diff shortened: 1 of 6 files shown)


src/utils/request/onUnhandledRequest.ts
```diff
@@ -46,18 +48,23 @@ function groupHandlersByType(handlers: RequestHandler[]) {
...
+  request: MockedRequest,
+  handler: RequestHandlerType,
+) => number
+
+function getRestHandlerScore(): ScoreGetterFn<RestHandler> {
   return (request, handler) => {
     const { path, method } = handler.info
 
-    if (path instanceof RegExp) {
+    if (path instanceof RegExp || method instanceof RegExp) {
```
rmachado-studocu, line 62: I haven't yet understood the concept of _score_ in the project so, sorry for asking this, but assuming that this is attributing a _priority_ for each handler, won't this hurt in cases like:

```js
rest.get('/one/endpoint', () => { /* ... does something ... */ });
rest.all('/one/endpoint', () => { /* ... does something different ... */ });
```

If you are certain that the method's RegExp matches with the current request's method, then you can ignore the method here, no?

src/utils/request/onUnhandledRequest.ts
```diff
@@ -46,18 +48,23 @@ function groupHandlersByType(handlers: RequestHandler[]) {
...
       return Infinity
     }
 
     const hasSameMethod = isStringEqual(request.method, method)
+
```
rmachado-studocu, line 67: Or perhaps change this to something like:
```suggestion
    if (path instanceof RegExp) {
      return Infinity
    }

    /** This assumes that `method` can only be a RegExp in the case of `all` which matches all HTTP verbs */
    const hasSameMethod = method instanceof RegExp || isStringEqual(request.method, method)

```

Assuming that the method cannot be set as RegExp by a user (only `all`, or one of the valid HTTP verbs, this would be ... ok :)

kettanaito: The `method` being a regular expression does not guarantee that it's the _same_ method. Originally, I've done this:

```js
const hasSameMetod = method instanceof RegExp
  ? method.test(request.method)
  : isStringEqual(request.method, method)
```

While this does the method matching, it has an unintended side-effect: this handler being included in the suggestions. You see, when we suggest what handlers you might've meant (i.e. in the case of a typo), we print the _handler's header_. That's a string that represents the handler's resource path, effectively:

- Method + URL for REST handlers;
- Operation Type + Operation Name for GraphQL handlers.

If the handler has `RegExp` as its method, the printed suggestion will look like this:

```
Did you mean to request "/^GE http://localhost/api"?
```

That's not a useful suggestion. In fact, due to the nature of regular expressions being dynamic in nature, any suggestion including RegExp becomes quite useless. Partially, because such request handler's "header" string is not static, but mainly because the users can't really understand what request via are suggesting (RegExp matches a subset of requests). 

This was my main reasoning behind ignoring regular expressions entirely when looking up the handler suggestions. Do you see any potential issues with this that I've overlooked?

src/utils/request/onUnhandledRequest.ts
kettanaito: Yes, I think the `rest.all()` handler must not be suggested. Please see my reasoning around this below. 

## The concept of `score`

We're using [js-levenstein](https://github.com/gustf/js-levenshtein) library for string likelihood matching. The said library can take two strings and return a numeric verdict representing how likely those strings are related/identical. That's precisely what we do with URLs and Operation names when looking up the suggestions for REST and GraphQL handlers respectively. 

The `score` here is that numeric representation. The lower the score, the less difference there is between the two given strings.

> That's also why I manually increment the score in the case of matching REST methods! Handlers with the same methods are more likely to be suggested.

So returning `Infinity` as the score means that the strings are not related at all. In the context of this function, it means that the handlers with the score `Infinity` are skipped.

src/utils/request/onUnhandledRequest.ts
kettanaito: Shortly, my main concern with dynamic methods/operation names is that their string representation is not useful for the user to suggest. It's insufficient to make a good conclusion over what might've been missed, or what should be corrected.

rmachado-studocu: > This was my main reasoning behind ignoring regular expressions entirely when looking up the handler suggestions. Do you see any potential issues with this that I've overlooked?

Not at all. In fact, I would argue that methods are finite and they should be limited to the existing ones.
If a user wants to perform the same handler for multiple verbs/methods, he/she can do one of the following:

```js
const  myHandler = (req, res, ctx) => { /* ... */ };
rest.post('/path', myHandler);
rest.get('/path', myHandler);
rest.put('/path', myHandler);
```
or 
```js
const  myHandler = (req, res, ctx) => {
  const method = req.method.toUpperCase();
  if (method === 'POST') {
    /* ... do POST things ... */
    return res(/* ... */);
  }

  if (method === 'GET') {
    /* ... do GET things ... */
    return res(/* ... */);
  }

  if (method === 'PUT') {
    /* ... do PUT things ... */
    return res(/* ... */);
  }
};
rest.all('/path', myHandler);
```

I'd argue also that `rest.all()` should have lowest precedence when it comes to handlers' execution, since it's less specific (it might already be, I'm just mentioning it 😉 ).

If I do 

```
rest.all('/path', (req, res, ctx) => { /* ... */ });
rest.post('/path', (req, res, ctx) => { /* ... */ }); // this should be executed as it's more specific
```

but again, that's my opinion 😉

rmachado-studocu: Oh and btw, @kettanaito, thanks for your quick reply and action in this 👍

kettanaito: Please, just note that this code does not affect how the handlers are run in any way. This is only the logic to _find suggestions_ from the existing handlers. Request handlers have no priority other than their declaration order (which you should utilize if you wish for fall-through and other patterns to apply). This issue (and the fix) are related to how MSW looked up the handlers to suggest when it catches a request that has no handlers. This is a purely DX thing to save from typos and other logical issues.

rmachado-studocu: Hmmm I see ... so if I set up `onUnhandledRequest: 'bypass'` it would actually solve my issue 🤦  ... So I happened to find that issue because I was not bypassing unhandled requests 😄  ... thanks for that @kettanaito ...

kettanaito: I don't think that using "bypass" would've prevented this issue. The library looks up the suggestions _before_ deciding whether to print them or not (hey, a nice place for improvement!).

If you like my assistance and would want to help me make MSW a better experience, please consider supporting me on [GitHub](https://github.com/sponsors/kettanaito) or our entire team on [Open Collective](https://opencollective.com/mswjs). Thank you!

rmachado-studocu: Done! Keep up the good work! 👍

kettanaito: Thank you so much, @rmachado-studocu ❤️ You are incredible.

PR #1026, opened 2021-12-10
fix: make PathParams type readonly

stefanprobst:
following up on #1017 tihs pr makes the `PathParams` type readonly.


kettanaito (approved):
Thank you for this improvement!

PR #1028, opened 2021-12-13
Fixes escaping port number in absolute urls

### LINKED ISSUE #1027: Cannot match absolute URLs with port number
## Describe the bug

Recently in version `0.36.0` `path-to-regexp` was used over `node-match-path` to have similar matching behaviour as `react-router` and `express`. However, `path-to-regexp` doesn't seem to match absolute URLs as `node-match-path` does. It feels that using `path-to-regexp` is a super valuable step forward but can more easily break mocks than anticipated.

## Environment

- `msw: 0.36.3`
- `nodejs: 17.x.x`

### Example 1

1. Define a handler with an absolute url such as `rest.get('http://localhost:1234/*')`
2. Execute a request against `http://localhost:1234/foo/bar`

### Example 2

1. Define a handler with an absolute url such as `rest.get('/:firstSegment/:secondSegment')`
2. Execute a request against `http://localhost:1234/foo/bar`

### Example 3

1. Define a handler with an absolute url such as `rest.get('http\\://localhost\\:3456/:project/:application')`
3. Execute a request against `http://localhost:1234/foo/bar`

In all examples nothing will be matched or an error will be thrown.

## Expected behavior

The URL to be matched or improved documentation how to deal with absolute URLs. It seems that `path-to-regexp` requires some escaping while not working with "catch-all" handers as well. Given that `path-to-regexp` will continue to be used a few more practical examples for migrating handlers with full urls could help the community if I am not the only one having this issue.

tdeekens:
- Closes #1027

### DIFF
--- a/src/utils/matching/matchRequestUrl.test.ts
+++ b/src/utils/matching/matchRequestUrl.test.ts
@@ -64,6 +64,31 @@ describe('matchRequestUrl', () => {
 })
 
 describe('coercePath', () => {
+  test('escapes the colon in protocol', () => {
+    expect(coercePath('https://example.com')).toEqual('https\\://example.com')
+    expect(coercePath('https://example.com/:userId')).toEqual(
+      'https\\://example.com/:userId',
+    )
+    expect(coercePath('http://localhost:3000')).toEqual(
+      'http\\://localhost\\:3000',
+    )
+  })
+
+  test('escapes the colon before the port number', () => {
+    expect(coercePath('localhost:8080')).toEqual('localhost\\:8080')
+    expect(coercePath('http://127.0.0.1:8080')).toEqual(
+      'http\\://127.0.0.1\\:8080',
+    )
+    expect(coercePath('https://example.com:1234')).toEqual(
+      'https\\://example.com\\:1234',
+    )
+
+    expect(coercePath('localhost:8080/:5678')).toEqual('localhost\\:8080/:5678')
+    expect(coercePath('https://example.com:8080/:5678')).toEqual(
+      'https\\://example.com\\:8080/:5678',
+    )
+  })
+
   test('replaces wildcard with an unnnamed capturing group', () => {
     expect(coercePath('*')).toEqual('(.*)')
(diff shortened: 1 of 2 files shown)


src/utils/matching/matchRequestUrl.ts
```diff
@@ -42,6 +42,11 @@ export function coercePath(path: string): string {
        * @see https://github.com/pillarjs/path-to-regexp/issues/259
        */
       .replace(/^([^\/]+)(:)(?=\/\/)/g, '$1\\$2')
+      /**
+       * Escape the port so that "path-to-regexp" can match
+       * absolute URLs including port numbers.
+       */
+      .replace(/((?::)(?:[0-9]+))/g, '\\$1')
```
tdeekens: Alternatively the following should also work `/(:[0-9]+)/`. @kettanaito you can pick :)

src/utils/matching/matchRequestUrl.test.ts
```diff
@@ -104,5 +104,8 @@ describe('coercePath', () => {
     expect(coercePath('https://example.com:1234/:userId')).toEqual(
       'https\\://example.com\\:1234/:userId',
     )
+    expect(coercePath('https://example.com:1234/:5678')).toEqual(
+      'https\\://example.com\\:1234/:5678',
```
tdeekens: I am not sure how to deal with this case. Suggestions welcome.

tdeekens:
An alternative approach, not using Regular Expressions, could be to use `new URL()` in a `try {} catch() {}`. Then one could attempt to build a url if possible and escape as needed and if a URL can't be constructed default to standard escaping e.g. only wildcard syntax.

kettanaito:
Thanks for opening this, @tdeekens!

I've adjusted the regular expression to match the port number pattern while allowing leading numbers in the path parameter names (as well as parameter names represented by numbers only). I've also changed the order in which the port is escaped, so it happens before escaping the protocol.

kettanaito (approved):
Thank you for helping fix this 👏

PR #1029, opened 2021-12-14
chore: add github actions

kettanaito:
Migrates to GitHub actions from CircleCI.


kettanaito:
- `FAIL test/msw-api/context/delay.test.ts (10.276 s)`
Rewrites the test's assertions to check the `res.request().timings()` directly (previously: checked timestamps from the browser runtime). This rules out the worker registration time, making this test more reliable.

PR #1042, opened 2022-01-02
fix: support boolean request body

azat-io:


src/handlers/RequestHandler.ts
```diff
@@ -26,6 +26,7 @@ export type DefaultRequestBody =
   | DefaultRequestMultipartBody
   | string
   | number
+  | boolean
```
kettanaito, line 29: While this will change the types, the actual body processing will never return a Boolean as the request's body value:

https://github.com/mswjs/msw/blob/527add209585c134fd7702327ded3f4721922ee3/src/utils/request/parseBody.ts#L9-L32

I'm frankly unsure if you can receive a transformed boolean when handling requests in conventional server-side request handling. Afaik, that won't work without an explicit body transformer, and so all request bodies will be plain text by default. 

It'd be great to create a quick Express server and see what `req.body` it receives upon the following request:

```js
fetch('http://localhost:3000', { body: true })
```

At the same time, if you try to construct a `Request` instance in the browser with `true`/`false` as the request's body, it won't preserve the boolean. The request body is always a string. 

@azat-io, could you please give us more context as to when you need this change? I'm particularly interested how you're making/handling a request with the boolean body.

Vavassor:
I ran into this too. My scenario is an API call response body that's `true` or `false` (without quotes).

```
server.use(
  rest.get("https://testurl/api/events/is-unique", (req, res, ctx) => {
    // ...
    const isUnique = eventName !== NON_UNIQUE_EVENT_NAME;
    return res(ctx.status(200), ctx.json(isUnique));
  })
);
```
It gave this type error for `ctx.json(isUnique)`.
```
Argument of type 'ResponseTransformer<boolean, any>' is not assignable to parameter of type 'ResponseTransformer<DefaultRequestBody, any>'.
  Type 'boolean' is not assignable to type 'DefaultRequestBody'.
```
true/false are valid top-level elements in JSON. So I suppose that's how the API got designed that way. I don't know if changing DefaultRequestBody is a solution or if that has larger implications.

src/handlers/RequestHandler.ts
azat-io: image

PR #1064, opened 2022-01-18
fix: parse request body with letter case header

### LINKED ISSUE #171: Request JSON body is not parsed in Node

whyDontI:
Fixes the problem mentioned in this [comment](https://github.com/mswjs/msw/pull/172#discussion_r786449936)

The request body is not parsed if the `Content-type` header has the value with letter cases like `Application/JSON`.

# GitHub
Fixes #171 (Might need to reopen or create a new one)

### DIFF
--- a/src/utils/request/parseBody.ts
+++ b/src/utils/request/parseBody.ts
@@ -11,7 +11,7 @@ export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
     return body
   }
 
-  const contentType = (headers?.get('content-type') || '').toLowerCase()
+  const contentType = headers?.get('content-type')?.toLowerCase() || ''
 
   // If the body has a Multipart Content-Type
   // parse it into an object.
--- a/src/utils/request/parseBody.ts
+++ b/src/utils/request/parseBody.ts
@@ -11,7 +11,7 @@ export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
     return body
   }
 
-  const contentType = headers?.get('content-type') || ''
+  const contentType = (headers?.get('content-type') || '').toLowerCase()
 
   // If the body has a Multipart Content-Type
   // parse it into an object.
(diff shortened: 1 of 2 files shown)


src/utils/request/parseBody.ts
```diff
@@ -11,7 +11,7 @@ export function parseBody(body?: MockedRequest['body'], headers?: Headers) {
     return body
   }
 
-  const contentType = headers?.get('content-type') || ''
+  const contentType = (headers?.get('content-type') || '').toLowerCase()
```
kettanaito: That's a nice catch. What do you think if we utilize optional chaining here since we're already using it?

```suggestion
  const contentType = headers?.get('content-type')?.toLowerCase() || ''
```

whyDontI: Sure 👍🏽

kettanaito (approved):
Thank you for fixing this, @whyDontI! Welcome to contributors 🎉

Issue #429, opened 2020-10-20
TypeScript 4.1 and template literal types to infer Request Params


timdeschryver:
**Is your feature request related to a problem? Please describe.**

Coming from https://github.com/mswjs/msw/pull/421
We could try to infer the request params from the URL by using template literal types (introduced in TS 4.1).
See this [example](https://www.typescriptlang.org/play?ts=4.1.0-pr-40336-88#code/C4TwDgpgBAogHsATgQwMbAEoHsCuwIAKyKAtgM4A8AKlBAhAHYAmZUZSAlgwOYB8UAXgBQUNpx616zVlRFQA-FAwRUWREwrtEXbgBox2nrzkAuKDTr5pUAAYASAN5cAZhERQAysGLAAviccXNygiUl8AekCGV3dldl8bOUUHAG0AayguEOJkEigAHyg0iBAsZ1gEFHRsPEIc8go44F4AXTMtHV9Tc0krFlsomM9vRD8Ap2jg0NyEpKhUjKzpkjaDTu6HXwBuISFQSBDBCqQ0TFx8ZcoAcnCwLHYycJM79gBJJiveHd2f-egAYSO8BO1XOdVI11u92Aj2e0PeT1UJBIjGA70+312QA)(not mine).

**Describe the solution you'd like**

I think this can be done within 2 steps

- [ ] Update to TS 4.1
- [ ] Try to use template literal types, because I'm not 100% sure if it works within this situation

**Describe alternatives you've considered**

The current solution is to provide a generic for the params.

**Additional context**

/

kettanaito:
Hey, @timdeschryver. Thank you for suggesting this. I think path parameters matching in this way would be much more versatile and seamless for the developer. I wonder how TS works when that path string is dynamically constructed?

It looks like bumping to `typescript: 4.1.x` wouldn't even be a breaking change, as we rely on `4.0.x` at the moment. That being said, how should we handle such TypeScript updates for the consumers of the library? Should we specify a TypeScript version in the `peerDependencies`?

timdeschryver:
That's a good question.

I'm not familiar with the template expression but I think dynamically constructed won't get inferred automatically. The developer will have to provide the generic manually, as she has to do today.

About the TypeScript  version, we'll have to try it first. I do think it would be breaking because it's a new feature, it could be that the type will be `any` or `unknown` in earlier version but again, I'm not sure about it.

Sorry if I can't provide answers on this one (yet).
[...]

PR #1073, opened 2022-01-23
feat(rest): infer path parameters from the path string

cmolina:
- Implements #429 

**This is a work in progress**

I am trying to redefine `PathParams` type to only include properties based on the `route`.

This is the main change I aim to introduce:

``` typescript
type PathParams<R extends string, O extends string = R> = /* where R = "route", and the O = "original route" */
  R extends `${infer _Before}:${infer P}/${infer After}` ? (Record<P,StringOrArray<P,O>> & PathParams<After,O>) :
  R extends `${infer _Before}:${infer P}` ? Record<P,StringOrArray<P,O>> :
  R extends `:${infer P}` ? Record<P,StringOrArray<P,O>> : unknown;
```

`StringOrArray` inspects the route and decides if the param `P` appears once or multiple times,
and returns a `string` or `string[]`.

I know I am supposed to write unit tests first; it is just this is my first time writing _unit tests for types_ so I am a bit lost here. I kind of wrote a few "unit tests" while I wrote the type, [you can see them in this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgzgVQHYEsCOBXCAeAClCAD2AiQBM4o5gAnFJAcwBooAlA40iq2+hgPigBeKAFgAUFDYcS5SgAMAJAG96AMwg0oAfQBCENQHsaEAL4AuFblMr1mqAEE1JGqflQA-FAAU8ZOiw8FicXQSJZblosTyg1AEMAGzhocygoiABKKHMJKXZwrgVLZWsAencvdOzYxOSJZVyoAGNDJGo0miwARlS-VEwcAHIMfqxBlkHzEYCIQcERdMaWtuAOrAAmXsRRoemB8ahJvaxSueE1iCXW9vSAZi3-AexhnYPB0qnX+YurlYuAFgeO2ex1mEw+ZAwYFKoIhUI+oLOC06l0kzWuq3SAFYgTMQa9weZIdDYUT4Z8ZqdvotxL92vEkhAerBtnjBhBGW9zABbCCGADuVPODLqaOW9NqEE2LMegXZnImPL5gqRNUZdNWIog9xlwPlyTeH15AtVWo1auSgN1bOJhrJJJ2cOhFP23zN4lMEgkoEgUAAyrxGAB5GgOGg0OIgbAAdRkhR4dEYLD9cbkCb480afTx0aYfrCnDTVS81ETDGqpb4AG0ALoAbnq5rgqQDZZDYYjUZeMztxJhjvtCMdqu7+3NcRbgYY7fDkeetsJfdJfZdJ1VVcGtprEk94lKpTS4GgrEMGBIADVEoF8AU05Wk9Jb9x7wJzqwIC0aGQgv6pzPOz+rD8PwEhEGAxiYkeUC4HEwAABYwRG3JwOs2D5IWz5TiwQapphZZvpmaLoRECi2EgGhaHoBjGGYxR2FoZRkRRjjOJobgxN476ft+uB5n+oazlGvFBsBUAAGTQbBCFxEhKHYCEmhMCJWQ5ERuGkao5H2FRRgmBYTH2NYFRsB+xg8XxbYCQBwmiapeTqVA8h0VpDHsV4XFmUErZ8P+c42YIqQjAA1kgApIA24gNGKGJQGAUnMjB8GIXEyGoe8g7LuSiLfFFUhSMSqQbl0gw1kwjRSKCqSDOs4yNKYEVSOKqxxfB0qJdJslpU6-aUoOq4QEKIi5XlBVQEVJVlWiFU7FVNWTVI9Xmi1cE6u1yWpc8-WqsN00zLNtVoot0V-MtVprTJKVyUcw45eVUCVYcc11Q16InVJOKSUlF0bel2XnDt90zY9B0LRFphAA) but I am still not sure how to write them in the code; maybe under `test/typings/rest.test-d.ts`?

In any case, I need help building the project since this change broke the existing tests. Eventually I will understand what's wrong, but it will probably take me a while since I am not familiar with all the interfaces yet.

This is the tail of the output from `yarn build `, when running `tsc`:

``` shell
$ tsc -p test/typings/tsconfig.json
test/typings/rest.test-d.ts:32:3 - error TS2578: Unused '@ts-expect-error' directive.

32   // @ts-expect-error `null` is not a valid response body type.
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

test/typings/rest.test-d.ts:34:18 - error TS2322: Type 'null' is not assignable to type 'AsyncResponseResolverReturnType<MockedResponse<DefaultRequestBody>>'.

34 >('/user', () => null)
                    ~~~~

  lib/types/handlers/RequestHandler.d.ts:42:122
    42 export declare type ResponseResolver<RequestType = MockedRequest, ContextType = typeof defaultContext, BodyType = any> = (req: RequestType, res: ResponseComposition<BodyType>, context: ContextType) => AsyncResponseResolverReturnType<MockedResponse<BodyType>>;
                                                                                                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The expected type comes from the return type of this signature.

test/typings/rest.test-d.ts:47:11 - error TS2339: Property 'userId' does not exist on type 'unknown'.

47   const { userId } = req.params
             ~~~~~~

test/typings/rest.test-d.ts:51:7 - error TS2578: Unused '@ts-expect-error' directive.

51       // @ts-expect-error "userId" parameter is not annotated
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

test/typings/rest.test-d.ts:60:3 - error TS2578: Unused '@ts-expect-error' directive.

60   // @ts-expect-error Path parameters are always strings.
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

test/typings/rest.test-d.ts:63:23 - error TS2322: Type 'null' is not assignable to type 'AsyncResponseResolverReturnType<MockedResponse<DefaultRequestBody>>'.

63 >('/posts/:id', () => null)
                         ~~~~

  lib/types/handlers/RequestHandler.d.ts:42:122
    42 export declare type ResponseResolver<RequestType = MockedRequest, ContextType = typeof defaultContext, BodyType = any> = (req: RequestType, res: ResponseComposition<BodyType>, context: ContextType) => AsyncResponseResolverReturnType<MockedResponse<BodyType>>;
                                                                                                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The expected type comes from the return type of this signature.


Found 6 errors.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:unit": "cross-env BABEL_ENV=test jest --maxWorkers=3",
     "test:integration": "jest --config=test/jest.config.js --maxWorkers=1",
     "test:smoke": "config/scripts/smoke.sh",
-    "test:ts": "tsc -p test/typings/tsconfig.json",
+    "test:ts": "tsc -p test-d.tsconfig.json",
     "prepare": "yarn simple-git-hooks init",
     "prepack": "yarn build",
     "release": "semantic-release",
--- a/src/handlers/RestHandler.ts
+++ b/src/handlers/RestHandler.ts
@@ -21,23 +21,25 @@ import {
   Match,
   matchRequestUrl,
   Path,
-  PathParams,
+  ExtractPathParams,
+  DefaultParamsType,
 } from '../utils/matching/matchRequestUrl'
 import { getPublicUrlFromRequest } from '../utils/request/getPublicUrlFromRequest'
 import { cleanUrl, getSearchParams } from '../utils/url/cleanUrl'
 import {
-  DefaultRequestBody,
+  DefaultBodyType,
   MockedRequest,
   RequestHandler,
   RequestHandlerDefaultInfo,
   ResponseResolver,
 } from './RequestHandler'
 
-type RestHandlerMethod = string | RegExp
+export type RestMethodType = RESTMethods | RegExp
 
-export interface RestHandlerInfo extends RequestHandlerDefaultInfo {
-  method: RestHandlerMethod
-  path: Path
+export interface RestHandlerInfo<ExactPath extends Path>
+  extends RequestHandlerDefaultInfo {
+  method: RestMethodType
+  path: ExactPath
 }
 
 export enum RESTMethods {
@@ -81,35 +83,54 @@ export type RequestQuery = {
 }
 
 export interface RestRequest<
-  BodyType extends DefaultRequestBody = DefaultRequestBody,
-  ParamsType extends PathParams<Path> = PathParams<RegExp>,
+  BodyType extends DefaultBodyType = DefaultBodyType,
+  ParamsType extends ExtractPathParams<Path> = ExtractPathParams<RegExp>,
 > extends MockedRequest<BodyType> {
   params: ParamsType
 }
 
 export type ParsedRestRequest<P extends Path> = Match<P>
 
+export type RestResponseResolver<
+  MethodType extends RestMethodType,
+  RequestBodyType extends DefaultBodyType,
+  ParamsType extends DefaultParamsType,
+  ResponseBodyType extends DefaultBodyType,
+> = ResponseResolver<
+  RestRequest<
+    MethodType extends RESTMethods.HEAD | RESTMethods.GET
+      ? never
+      : RequestBodyType,
+    ParamsType
+  >,
+  RestContext,
+  ResponseBodyType
+>
+
 /**
  * Request handler for REST API requests.
  * Provides request matching based on method and URL.
  */
 export class RestHandler<
-  RequestType extends MockedRequest<DefaultRequestBody> = MockedRequest<DefaultRequestBody>,
+  MethodType extends RESTMethods | RegExp = RESTMethods | RegExp,
+  PathType extends Path = Path,
+  RequestBodyType extends DefaultBodyType = DefaultBodyType,
+  ResponseBodyType extends DefaultBodyType = DefaultBodyType,
 > extends RequestHandler<
-  RestHandlerInfo,
-  RequestType,
-  ParsedRestRequest<RestHandlerInfo['path']>,
-  RestRequest<
-    RequestType extends MockedRequest<infer RequestBodyType>
-      ? RequestBodyType
-      : any,
-    PathParams<RestHandlerInfo['path']>
-  >
+  RestHandlerInfo<PathType>,
+  MockedRequest<RequestBodyType>,
+  ParsedRestRequest<PathType>,
+  RestRequest<RequestBodyType, ExtractPathParams<PathType>>
 > {
   constructor(
-    method: RestHandlerMethod,
-    path: Path,
-    resolver: ResponseResolver<any, any>,
+    method: MethodType,
+    path: PathType,
+    resolver: RestResponseResolver<
+      MethodType,
+      RequestBodyType,
+      ExtractPathParams<PathType>,
+      ResponseBodyType
+    >,
   ) {
     super({
       info: {
@@ -150,7 +171,10 @@ export class RestHandler<
     )
   }
 
-  parse(request: RequestType, resolutionContext?: ResponseResolutionContext) {
+  parse(
+    request: MockedRequest<RequestBodyType>,
+    resolutionContext?: ResponseResolutionContext,
+  ) {
     return matchRequestUrl(
       request.url,
       this.info.path,
@@ -159,18 +183,18 @@ export class RestHandler<
   }
 
   protected getPublicRequest(
-    request: RequestType,
-    parsedResult: ParsedRestRequest<RestHandlerInfo['path']>,
-  ): RestRequest<any, PathParams<RestHandlerInfo['path']>> {
+    request: MockedRequest<RequestBodyType>,
+    parsedResult: ParsedRestRequest<PathType>,
+  ): RestRequest<any, ExtractPathParams<PathType>> {
     return {
       ...request,
-      params: parsedResult.params || {},
+      params: parsedResult.params,
     }
   }
 
   predicate(
-    request: RequestType,
-    parsedResult: ParsedRestRequest<RestHandlerInfo['path']>,
+    request: MockedRequest<RequestBodyType>,
+    parsedResult: ParsedRestRequest<PathType>,
   ) {
     const matchesMethod =
       this.info.method instanceof RegExp
@@ -180,7 +204,7 @@ export class RestHandler<
     return matchesMethod && parsedResult.matches
   }
 
-  log(request: RequestType, response: SerializedResponse) {
+  log(request: MockedRequest<RequestBodyType>, response: SerializedResponse) {
     const publicUrl = getPublicUrlFromRequest(request)
     const loggedRequest = prepareRequest(request)
     const loggedResponse = prepareResponse(response)
--- a/src/handlers/RestHandler.ts
+++ b/src/handlers/RestHandler.ts
@@ -82,12 +82,12 @@ export type RequestQuery = {
 
 export interface RestRequest<
   BodyType extends DefaultRequestBody = DefaultRequestBody,
-  ParamsType extends PathParams = PathParams,
+  ParamsType extends PathParams<Path> = PathParams<RegExp>,
 > extends MockedRequest<BodyType> {
   params: ParamsType
 }
 
-export type ParsedRestRequest = Match
+export type ParsedRestRequest<P extends Path> = Match<P>
 
 /**
  * Request handler for REST API requests.
@@ -98,12 +98,12 @@ export class RestHandler<
 > extends RequestHandler<
   RestHandlerInfo,
   RequestType,
-  ParsedRestRequest,
+  ParsedRestRequest<RestHandlerInfo['path']>,
   RestRequest<
     RequestType extends MockedRequest<infer RequestBodyType>
       ? RequestBodyType
       : any,
-    PathParams
+    PathParams<RestHandlerInfo['path']>
   >
 > {
   constructor(
@@ -160,15 +160,18 @@ export class RestHandler<
 
   protected getPublicRequest(
     request: RequestType,
-    parsedResult: ParsedRestRequest,
-  ): RestRequest<any, PathParams> {
+    parsedResult: ParsedRestRequest<RestHandlerInfo['path']>,
+  ): RestRequest<any, PathParams<RestHandlerInfo['path']>> {
     return {
       ...request,
       params: parsedResult.params || {},
     }
   }
 
-  predicate(request: RequestType, parsedResult: ParsedRestRequest) {
+  predicate(
+    request: RequestType,
+    parsedResult: ParsedRestRequest<RestHandlerInfo['path']>,
+  ) {
     const matchesMethod =
       this.info.method instanceof RegExp
         ? this.info.method.test(request.method)
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -4,18 +4,13 @@ import {
   RestHandler,
   RestResponseResolver,
 } from './handlers/RestHandler'
-import {
-  Path,
-  ExtractPathParams,
-  DefaultParamsType,
-} from './utils/matching/matchRequestUrl'
+import { Path, ExtractPathParams } from './utils/matching/matchRequestUrl'
 
 function createRestHandler<MethodType extends RESTMethods | RegExp>(
   method: MethodType,
 ) {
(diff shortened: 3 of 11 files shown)


kettanaito:
Hey, @cmolina. Thank you for preparing this!

Those `tsc` errors mean that we're breaking some type annotations with this change. Let me dive into the changes and see what may be causing this.

src/handlers/RestHandler.ts
```diff
@@ -98,12 +98,12 @@ export class RestHandler<
 > extends RequestHandler<
   RestHandlerInfo,
   RequestType,
-  ParsedRestRequest,
+  ParsedRestRequest<RestHandlerInfo['path']>,
```
kettanaito: `RestHandlerInfo['path']` always refers to a _type_ but we need the value here. I've changed that in the follow-up commit.

kettanaito:
Hey, @cmolina. Thank you for kicking off this improvement! It's fantastic. 

I've pushed a few additional changes to properly infer path parameters in `RestHandler` and `rest`. All tests should be passing now. 

There's one thing I couldn't solve, though. At the moment I've removed the support for explicit `ParamsType` generic, so path parameters _are always inferred_. For some reason I had trouble annotating the value of the parameters like this:

```diff
  return <
    RequestBodyType extends DefaultBodyType = DefaultBodyType,
+   ParamsType extends DefaultParamsType = never,
    ResponseBodyType extends DefaultBodyType = DefaultBodyType,
    PathType extends Path = Path,
  >(
    path: PathType,
    resolver: RestResponseResolver<
      MethodType,
      RequestBodyType,
+     ParamsType extends never ? ExtractPathParams<PathType> : ParamsType,
      ResponseBodyType
    >,
  ) => {
```

This should allow overriding the parameters type upon usage like so:

```ts
rest.get<never, { id: string }>('/user', (req) => {
  req.params.id.trim()
})
```

However, it doesn't compile. It yields a `RestResponseResolver` type incompatible with the one created by the `RestHandler` generic. 

I'd like to get your opinion on two things:

1. Do you think an explicit `ParamsType` generic is viable? When do you think developers would use it?
2. Do you see any potential issues with the explicit `ParamsType` generic support I've shared above? I'd love to get your expertise on this. 

Thank you!

package.json
```diff
@@ -17,7 +17,7 @@
     "test:unit": "cross-env BABEL_ENV=test jest --maxWorkers=3",
     "test:integration": "jest --config=test/jest.config.js --maxWorkers=3",
     "test:smoke": "config/scripts/smoke.sh",
-    "test:ts": "tsc -p test/typings/tsconfig.json",
+    "test:ts": "tsc -p test-d.tsconfig.json",
```
kettanaito, line 25: Moving this config to the root so that we can write `*.test-d.ts` files next to the source code. We still keep the `test/typings` for the feature-based typings that can't be placed near a single module.

kettanaito:
# Other changes

## Breaking changes

- The `DefaultRequestBody` type is now `DefaultBodyType`.

## Features

- Exports the following types:
  - `ExtractPathParams`
  - `DefaultParamsType`
  - `RestResopnseResolver`

kettanaito:
Sharing a Twitter discussion about the explicit `Params` type: https://twitter.com/TkDodo/status/1487798191286374403

Aprillion:
> ```js
> rest.get<never, { id: string }>('/user', (req) => {
>   req.params.id.trim()
> })
> ```
> However, it doesn't compile. It yields a RestResponseResolver type incompatible with the one created by the RestHandler generic.

That error sounds legit to me, isn't it? What else would be desired if someone used that syntax? IMHO prompting the developer to fix the path from `'/user'` to `'/user/:id'` makes sense here.

> 1. Do you think an explicit ParamsType generic is viable? When do you think developers would use it?

If I got my types generated from BE, I would be tempted to use explicit `ParamsType` to get an error if my path stops matching a changed API in the future. But I can't imagine wanting to write inline explicit type manually if it can be inferred from the path :D

kettanaito:
> If I got my types generated from BE, I would be tempted to use explicit ParamsType to get an error if my path stops matching

Afaik that's not possible to achieve in TS. You can extract params from a string but I haven't seen an example that'd validate a string against a certain pattern. 

One good point Dominik makes on Twitter is that developers may want to narrow down the parameter type to specific values:

```js
req.params.locale // "en-US" | "de-DE"
```

It'd be great to see this pattern as an extension of the generated params record where you can't introduce new keys but can override values as long as they extend `string | string[]`.

```ts
rest.get<never, { locale: 'en-US' }>('/user/:id/:locale', (req) => {
  req.params.id // string, still exists (inferred)
  req.params.locale // "en-US" (narrowed string)
})
```

I think this may also align well with backend specs that annotate path params as enums.

Aprillion:
> align well with backend specs that annotate path params as enums

hmm, OpenApi spec like `/user/{id}/{locale}` where id is a number and locale is some enum could be theoretically converted to 3 TS types:
* ` '/user/:id/:locale' ` as a string literal type to validate the MSW handler path parameter
* `{id: number, locale: 'en-US' | 'de-DE'}` as the `ParamsType` generic in MSW handler function body
* `` `/user/${number}/${'en-US' | 'de-DE'}` `` as a template literal type to validate `fetch` calls in client code

no idea how to pass the first 2 types to MSW in some easy way, let alone how the 3rd type fits into the MSW typing picture, but I believe `rest.get(...)` is not the best place, it would be much more convenient to pass those generated types globally to `setupWorker`/`setupServer`

=> I believe I won't miss explicit `ParamsType` support at this place for individual `RequestHandler`s if removed 🤔

kettanaito:
The changes are not released candidate yet. We still need to support the params generics for people to override inferred params. I think I will start with the only behavior that whenever you give an explicit generic, it completely overrides what's inferred from the path string. We can iterate on this later and consider extending the inferred object type instead.

kettanaito:
`832463f`

cmolina:
@kettanaito thanks for all the progress with the integration and tests!

I will spend a few minutes today to work on the override of params.

---

I haven't figured out it yet, but I learned something interesting when I was writing  a new test for the override functionality.

In `test/typings/rest.test-d.ts:28`, my IDE shows me the right types when hovering over `rest.get`:

``` typescript
rest.get('/user/:id/message/:id', (req) => { /*
     ^ (property) get: <DefaultBodyType, never, DefaultBodyType, "/user/:id/message/:id">(path: "/user/:id/message/:id", resolver: RestResponseResolver<RESTMethods.GET, DefaultBodyType, Record<...>, DefaultBodyType>) => RestHandler<...> */
```

However, when I overrode one of the types, I lost the information of the route being handled. Notice how adding `<never>` renders a type way shorter than the original, replacing the route string with a generic `Path`:

``` typescript
rest.get<never>('/user/:id/message/:id', (req) => { /*
     ^ (property) get: <never, never, DefaultBodyType, Path>(path: Path, resolver: RestResponseResolver<RESTMethods.GET, never, DefaultParamsType, DefaultBodyType>) => RestHandler<...> */
```

I will focus on how I can define types without losing the non-overridden types.

kettanaito:
Hey, @cmolina! Thank you for following up on that.

Yeah, it's tricky to get the override generic to work properly. I think right now the intention is that whenever an explicit path generic is provided, it must be used instead of the inferred path parameters. 

There's but one concern here: we should make it easy for people to annotate request and response bodies without having to explicitly annotate path parameters. This is relevant to the generics order right now:

```js
rest.get<RequestBody, PathParams, ResponseBody>()
```

I do think this order makes sense, I'm just mentioning that we should probably have a special value for the `PathParams` that'd mean "I don't want to override them, use the inferred values". I was thinking about using `unknown` for this purpose and checking that internally:

```ts
PathParams extends unknown ? USE_INFERRED : PathParams
```

Frankly, I'm not sure if conditional types based on `unknown` are possible. We may try looking into that.

cmolina:
Yeah, we are on the same page.

What I am saying is, right now when people annotate a type, _any type_, the path information is lost, and because of that we are  unable to infer the parameters.

(I already wrote the conditional :) but I'm order to work, we need to find a way for TypeScript to infer the type of `PathParams` as a specific string literal, instead of `string`).

cmolina:
I found out why we are loosing the `PathParams` information: currently TypeScript does not support skipping generics and performing inference 😬 

> Regarding inferring the trailing type parameters automatically - I've noticed a comment from @RyanCavanaugh in https://github.com/microsoft/TypeScript/issues/39008#issuecomment-642074983 that:

> > When any type arguments are specified, the remaining arguments are set to their defaults rather than going through inference. IMHO this was probably a mistake, but it was written that way on purpose and would be too large of a breaking change to modify at this point

You can read more about it in [this open issue](https://github.com/microsoft/TypeScript/issues/10571) and [this open PR](https://github.com/microsoft/TypeScript/pull/26349).

---

Apparently I delved too greedily and too deep into TypeScript 😂 

I will explore if there are workarounds that accommodate to our use case.

kettanaito:
Thanks for investigating this, @cmolina! Yeah, it looks like it's impossible to skip a generic even if it doesn't have an explicit value. What a bummer. I wonder if we can still have explicit/inferred path parameters somehow 🤔

kettanaito:
One thing I've noticed is that if a path parameter has modifiers (i.e. `:name+` or `:name*`) the modifier gets inferred as a part of the path parameter name. I will push the fix for this.

kettanaito:
I slowly lean towards merging this without explicit generic support (conditional generics such as this one are not supported in TypeScript). Your path string is always the source of truth. Even if you generate type definitions out of external spec, like OpenAPI spec, path parameter type will be synced with the path string there anyway. 

I still acknowledge the need to narrow down the path parameter type from a generic `string` to an enum or an explicit value but I'd recommend using type-casting for this:

```ts
type Roles = 'admin' | 'editor' | 'viewer'

rest.get('/roles/:type', (req, res, ctx) => {
  const type = req.params.type as Roles
})
```

What do you think, @cmolina? Do you have any concerns about merging this in the current state?

cmolina:
These are my thoughts:
- I think this change is as good as we can deliver, given the current state of TypeScript
- It may be confusing for users to have params inferred only sometimes
- I wonder how useful this change would be in real life codebases. For instance, in my current project, in won't be useful since I override both `RequestBody` and `ResponseBody`

I guess I feel a bit disappointed with the current solution; before you wrote the last message I was going to propose to close this PR. But maybe this could be a datapoint for the TypeScript team to encourage them to add support to skipping generics.

kettanaito:
Thanks for sharing, @cmolina! 

To be fair, the current solution works and is written fantastically! It achieves its goal, that being inferred path parameter types. It became more difficult because we tried to bring more variable use cases to the table. I do believe those use cases are A) rare; B) can be achieved with type-casting. Even those who want to limit path parameter types will still benefit from those types being inferred!

Perhaps one thing that's left to do is to remove the `PathParams` generic to `rest.*` and the entire chain down the line. 

This is definitely a breaking change. But I think it's a breaking change for a better developer experience.

src/rest.ts
```diff
@@ -4,18 +4,13 @@ import {
...
-  DefaultParamsType,
-} from './utils/matching/matchRequestUrl'
+import { Path, ExtractPathParams } from './utils/matching/matchRequestUrl'
 
 function createRestHandler<MethodType extends RESTMethods | RegExp>(
   method: MethodType,
 ) {
   return <
     RequestBodyType extends DefaultBodyType = DefaultBodyType,
-    ParamsType extends DefaultParamsType = never,
```
kettanaito: It's no longer possible to specify the path parameters generic.

kettanaito (dismissed):
This is fantastic 🎉 Thank you for making it happen, @cmolina!

We should merge #1025 alongside this since string inference is available only after specific versions of TypeScript.

PR #1076, opened 2022-01-25
chore: schedule daily releases

kettanaito:
Runs the `release` workflow once per day at 23:00 only if the `ci` workflow has successded.


kettanaito:
I have no confirmation from the GH Actions docs on how `schedule` and `workflow_run` will combine. Without `schedule`, it runs _whenever_ the dependent job succeeds. That's not the expected result.

PR #1096, opened 2022-02-11
feat(onUnhandledRequest): allow default behaviour in a custom callback

### LINKED ISSUE #1080: Enable a way to run the default behaviour from inside the onUnhandledRequest callback
When I want to create a custom `onUnhandledRequest` callback, I want to block a certain url from being flagged, but then I have to re-write the default behaviour in my own code.

Allow me to hand control back to `msw` for the `onUnhandledRequest` callback.

Re-writing the code that already exists in https://github.com/mswjs/msw/blob/ddbc5a7ce30eabc56390eba79e0d260f3a8f598d/src/utils/request/onUnhandledRequest.ts#L180 into my own test code, which would mean it may diverge from `msw` code in future and may not work exactly the same if I don't re-write it correctly.

**Additional context**
We could solve this by adding an extra parameter that is passed to `onUnhandledRequest` called `next`, which will yield control back to `msw`. This would be similar to the `next()` functionality from express.

JamesZoft:
Allow users to fall back to the default behaviour for `onUnhandledRequest` if they just want to cover a small number of scenarios, then hand back control to MSW

- Closes #1080

### DIFF
--- a/src/utils/handleRequest.test.ts
+++ b/src/utils/handleRequest.test.ts
@@ -92,7 +92,11 @@ test('reports request as unhandled when it has no matching request handlers', as
     ['request:unhandled', request],
     ['request:end', request],
   ])
-  expect(options.onUnhandledRequest).toHaveBeenNthCalledWith(1, request)
+  expect(options.onUnhandledRequest).toHaveBeenNthCalledWith(
+    1,
+    request,
+    expect.anything(),
+  )
   expect(callbacks.onBypassResponse).toHaveBeenNthCalledWith(1, request)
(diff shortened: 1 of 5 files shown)


kettanaito:
Thanks for making this improvement happen, @JamesZoft! 🎉 

I've pushed to your feature branch a few minor changes: renamed `next` to `print` and made it an object, so the usage looks like this:

```js
onUnhandledRequest(request, print) {
  print.error() // or `print.warning()`
}
```

Also added a few integration and E2E tests. Everything looks in order!

kettanaito (approved):

PR #1097, opened 2022-02-12
chore: use "release-please"

kettanaito:
- Uses [release-please](https://github.com/googleapis/release-please) instead of Semantic release.

Mainly so that multiple merges to the default branch would accumulate in the release pull request and then be released under a single version bump.

## Pre-requisites

- release-please must support GitHub Release generation (https://github.com/google-github-actions/release-please-action/issues/427). We do not maintain a `CHANGELOG.md` file.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,30 +1,36 @@
 name: release
 
 on:
-  schedule:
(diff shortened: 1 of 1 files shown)


kettanaito:
Keeping it as a template but closing.

PR #1110, opened 2022-02-19
fix(handleRequest): bypass request only if "x-msw-bypass" header equals "true"

### LINKED ISSUE #1085: "x-msw-bypass" header is treated differently in the worker and on the client
Here's how the worker treats the `x-msw-bypass` header: 

https://github.com/mswjs/msw/blob/6e7ec2871f5f82fbf228b803df565286b1d4dbfc/src/mockServiceWorker.js#L159-L161

While on the client-side we only check if that said header is set (regardless of its value):

https://github.com/mswjs/msw/blob/6e7ec2871f5f82fbf228b803df565286b1d4dbfc/src/utils/handleRequest.ts#L60-L65

This may lead to mismatching behavior when using the `x-msw-bypass: "false"` request header.

## Expected behavior

Requests are bypassed if the `x-msw-bypass` header has the `"true"` value. All the other values are ignored. 

We should also add an integration test for this:

1. Create a handler for a request.
1. Dispatch a matching request with `x-msw-bypass: "false"`.
2. Expect the mocked response. (Actual: get the original response, no mocking applied).

kettanaito:
- Fixes #1085

### DIFF
--- a/src/utils/handleRequest.test.ts
+++ b/src/utils/handleRequest.test.ts
@@ -44,7 +44,7 @@ afterEach(() => {
   emitter.removeAllListeners()
 })
 
-test('returns undefined for a request with the "x-msw-bypass" header', async () => {
+test('returns undefined for a request with the "x-msw-bypass" header equal to "true"', async () => {
   const request = createMockedRequest({
     url: new URL('http://localhost/user'),
     headers: new Headers({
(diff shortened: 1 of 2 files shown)


src/utils/handleRequest.test.ts
```diff
@@ -72,6 +72,33 @@ test('returns undefined for a request with the "x-msw-bypass" header', async ()
   expect(callbacks.onMockedResponseSent).not.toHaveBeenCalled()
 })
 
+test('does not bypass a request with "x-msw-bypass" header set to arbitrary value', async () => {
+  const request = createMockedRequest({
+    url: new URL('http://localhost/user'),
+    headers: new Headers({
+      'x-msw-bypass': 'anything',
```
kettanaito, line 79: Arbitrary values of the `x-msw-bypass` request header do not bypass the request.

PR #1138, opened 2022-03-02
chore: use node 14 in codesandbox

jbergstroem:
It seems like engine requirements for dependencies are requiring the node 14 runtime. Since `devDependencies` has types for node 14 I thought it might make sense to just bump it.

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,4 +1,4 @@
 {
-  "node": "12",
+  "node": "14",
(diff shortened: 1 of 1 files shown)


kettanaito:
Hey, @jbergstroem. Thanks for proposing this! 

I think this change is now obsolete since I've just merged the Node.js 16 upgrade (#1136). The entire MSW toolchain will require Node.js version 16 or higher.

Issue #1090, opened 2022-02-09
Early request not intercepted in Firefox for most recent msw version


celcius112:
## Environment

| Name    | Version |
| ------- | ------- |
| msw     | 0.36.8   |
| browser | Firefox  96  |
| OS      | MacOS 12.1     |

First of all, thank you for this very handy library :)

## Current behavior

Updating the msw version to 0.36.8 in the Angular example project (https://github.com/mswjs/examples/tree/master/examples/rest-angular) breaks the request interception in Firefox for early requests.

Here's a zip containing : 
- An upgrade to msw 0.36.8
- A modification in AppComponent to perform an early request to `login`
[examples.zip](https://github.com/mswjs/msw/files/8031346/examples.zip)

Launching the Angular application for the first time after a `yarn start` will result with a 404 being thrown. Reloading the page or clicking the submit button will make it work.
<img width="964" alt="Screenshot 2022-02-09 at 10 42 23" src="https://user-images.githubusercontent.com/1503748/153169861-e220040c-933c-4785-ae2b-0c86cf743db4.png">

kettanaito:
Hey, @celcius112. Thanks for reporting this. 

I confirm this is still an issue on Firefox only even with `0.39.1`. The recent worker activation Promise fixes we've published do not fix this. Needs investigation.

kettanaito:
This is related to #220. Firefox does not expose `XMLHttpRequest` to Service Worker, so the worker doesn't trigger its "fetch" event when such requests occur. This is a known limitation of Firefox and there's nothing we can do about it. 

I've ensured that the fetch/XMLHttpRequest requests are deferred correctly and they are. Requests are performed only once the worker is active. It still cannot handle the XMLHttpRequest because Firefox doesn't notify the worker about such requests. It does work as expected when using `window.fetch` upon initial load (early request). 

```ts
  ngOnInit(): void {
    fetch('/login', {
      method: 'POST',
      headers: { 'content-type': 'application/json' },
      body: JSON.stringify({ username: 'test' }),
    })
  }
```

I understand this may be not the way to request resources in Angular but overall you must consider using `window.fetch`. I suspect that `this.http.post` used in `this.userService.login` still creates an XHR instead. May be a dependency issue, may be a framework consensus.

Either way, MSW cannot (and should not) circumvent browser limitations. Being a developer tool, you can use MSW in other browsers instead.

celcius112:
Hello @kettanaito, thank you for your answer. 

[...]

Issue #1051, opened 2022-01-10
Initial load request interception


kettanaito:
As reported in #1049 and a few other issues, the library is having trouble intercepting initial load requests. That often happens because of the race condition between the worker registration and application making requests. We do attempt to eliminate that race condition by capturing and replaying all requests that happened in that timeframe but that clearly doesn't work for everybody. 

https://github.com/mswjs/msw/blob/527add209585c134fd7702327ded3f4721922ee3/src/utils/deferNetworkRequestsUntil.ts#L8

## Expected behavior

The library is capable of intercepting requests issued upon the initial load/mount of an application. 

## Suggested solutions

1. For once, we may investigate why `deferNetworkRequestsUntil` doesn't work in those problematic cases. Perhaps there's an issue in the implementation that's the culprit.
2. Alternatively, think of a different approach to the problem.

marcosvega91:
Hellooo, it could be bacause the `deferNetworkRequestsUntil` function has been called ad the end of the start process? There are some await before and this could make the application ready before MSW is ready. Maybe it could be better to move the function at the beginning of the function using a different promise ?

kettanaito:
This has a high chance of being fixed by #1146. The `deferNetworkUntil` hijacked `window.fetch` and inserted an intermediary Promise to await—the worker registration Promise. Since the latter used to resolve improperly on the worker's registration but not activation, it was possible that the hijacked fetch would restore itself earlier than the worker was indeed activated, becoming ready to handle requests. 

This needs testing and verification.

kettanaito:
One way to test this is to use the NextJS example and start the worker without explicitly awaiting it activation before rendering the app (without the deferred mounting approach). This must always work.

kettanaito:
I confirm that #1146 fixes the initial load issue some of our users have been experiencing. Here's a video of the `GET /reviews` request being delayed until the worker has been activated (for 1.5s):

https://user-images.githubusercontent.com/14984911/157251896-5a6eff51-4c3c-475f-958c-aafb60b57d08.mov

The request happens on the `App` component mount. There are no changes made to how the NextJS app is rendered, the mocking is enabled as per the [NextJS usage example](https://github.com/vercel/next.js/tree/canary/examples/with-msw).

PR #1146, opened 2022-03-06
fix(setupWorker): resolve the "start" promise after the worker has activated

### LINKED ISSUE #1145: "setupWorker" Promise may resolve before the worker is activated
Service Worker activation is an async process which undergoes multiple stages:

- installing
- installed
- activating
- activated

Currently, the Promise returned from `setupWorker` returns the _registration_ Promise:

https://github.com/mswjs/msw/blob/d7bdeeca6a9f0bb3f7abd6179af2babeb68e1cf7/src/setupWorker/start/createStartHandler.ts#L120

The worker registration may be in the aforementioned states when this Promise resolves. 

## Proposed behavior

The library should resolve the `setupWorker` Promise when, and only when, the worker has been registered and activated (ready to handle requests):

```ts
return workerRegistration.then((registration) => {
  const pending = registration.installing || registration.waiting
  return new Proimse((resolve) => {
    pending.addEventListener('statechange', () => {
      if (pending.state === 'activated') {
        return resolve()
      }

      // Can we assume failed activation by state alone?
    })
  })
})
```

## Context

This issue is currently failing the tests in the Karma usage example. Implementing the said fix works.

kettanaito:
- Fixes #1145 
- May be related to #1051, #1090

### DIFF
--- a/src/setupWorker/start/createStartHandler.ts
+++ b/src/setupWorker/start/createStartHandler.ts
@@ -92,10 +92,6 @@ If this message still persists after updating, please report an issue: https://g
       `)
       }
 
-      await enableMocking(context, options).catch((err) => {
-        throw new Error(`Failed to enable mocking: ${err?.message}`)
-      })
-
       context.keepAliveInterval = window.setInterval(
         () => context.workerChannel.send('KEEPALIVE_REQUEST'),
         5000,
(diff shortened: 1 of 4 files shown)


src/setupWorker/start/createStartHandler.ts
```diff
@@ -108,7 +104,31 @@ If this message still persists after updating, please report an issue: https://g
...
+    const workerRegistration = startWorkerInstance().then(
+      async (registration) => {
+        const pendingInstance = registration.installing || registration.waiting
+
+        // Wait until the worker is activated.
+        // Assume the worker is already activated if there's no pending registration
+        // (i.e. when reloading the page after a successful activation).
+        if (pendingInstance) {
+          await new Promise<void>((resolve) => {
+            pendingInstance.addEventListener('statechange', () => {
```
Aprillion, line 116: both `registration.installing.addEventListener` and `registration.waiting.addEventListener` can listen to the same event `'statechange'`?

kettanaito: @Aprillion yes. Those properties return a worker instance in a different state. But it's the same worker instance as the activated worker, having the same event listeners. This is how you know if a state transitions into another state.

PR #1155, opened 2022-03-09
fix(parseIsomorphicRequest): bypassing cookies properly

### LINKED ISSUE #1154: `parseIsomorphicRequest` doesn't handle `credentials` and `cookies`.
## Describe the bug

I met a problem that the `ctx.fetch(req)` doesn't respect the `credentials` and `cookies` fields.
So we can't bypass the request properly, I think.

Then, I found the cause that was caused by `parseIsomorphicRequest` function.
The function creates a new mocked request but it always has
 `credentials: 'same-origin'` even if the actual request was specified `credential: 'include'`.

## Environment

- `msw: 0.39.1`
- `nodejs: 16.13.0`
- `npm: 8.4.1`

## To Reproduce

I've create the PR and it contains the test case for this problem.

## Expected behavior

Bypass the cookies and credentials.

hrsh7th:
Fix #1154

This PR aims to bypass correctly APIs that require cookies.

I've created the [msw-snapshot](https://github.com/hrsh7th/msw-snapshot) package for testing that will response from actual server or disk-cache.
I met this problem in the test with the API that requires the login session.

### DIFF
--- a/src/utils/request/parseIsomorphicRequest.ts
+++ b/src/utils/request/parseIsomorphicRequest.ts
@@ -17,6 +17,9 @@ export function parseIsomorphicRequest(
     method: request.method,
     body: parseBody(request.body, request.headers),
     credentials: request.credentials || 'same-origin',
+
+    // The `setRequestCookies` will modify the `Headers` object but we don't want to modify the original header field.
+    // So we create new `Headers` instance here.
     headers: new Headers(Array.from(request.headers.entries())),
     cookies: {},
     redirect: 'manual',
--- a/src/utils/request/parseIsomorphicRequest.ts
+++ b/src/utils/request/parseIsomorphicRequest.ts
@@ -1,4 +1,5 @@
 import * as cookieUtils from 'cookie'
+import { Headers } from 'headers-polyfill/lib'
 import { IsomorphicRequest } from '@mswjs/interceptors'
 import { MockedRequest } from '../../handlers/RequestHandler'
 import { parseBody } from './parseBody'
@@ -15,7 +16,8 @@ export function parseIsomorphicRequest(
     url: request.url,
     method: request.method,
     body: parseBody(request.body, request.headers),
-    headers: request.headers,
+    credentials: request.credentials,
+    headers: new Headers(Array.from(request.headers.entries())),
     cookies: {},
     redirect: 'manual',
     referrer: '',
@@ -26,7 +28,6 @@ export function parseIsomorphicRequest(
     integrity: '',
     destination: 'document',
     bodyUsed: false,
-    credentials: 'same-origin',
   }
 
   // Set mocked request cookies from the `cookie` header of the original request.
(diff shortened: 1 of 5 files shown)


src/utils/request/parseIsomorphicRequest.ts
```diff
@@ -15,7 +16,8 @@ export function parseIsomorphicRequest(
     url: request.url,
     method: request.method,
     body: parseBody(request.body, request.headers),
-    headers: request.headers,
+    credentials: request.credentials,
```
kettanaito: We should still have the `same-origin` fallback to preserve the previous behavior. 

```suggestion
    credentials: request.credentials || 'same-origin',
```

src/utils/request/parseIsomorphicRequest.ts
```diff
@@ -15,7 +16,8 @@ export function parseIsomorphicRequest(
...
+    headers: new Headers(Array.from(request.headers.entries())),
```
kettanaito: I think `request.headers` is already an instance of the `Headers` polyfill. Could you share more info as to why we need to construct it again? Are some headers being omitted? If so, have you understand why?

kettanaito (changes requested):
Hey, @hrsh7th. Thanks for opening this pull request.

I've left a few suggestions/clarifications. Let me know about those when you have time.

src/utils/request/parseIsomorphicRequest.ts
hrsh7th: To be honest, I don't understand perfectly the reason for this but it's needed to keep the `headers` field. Even if we call parseIsomorphicRequest multiple times.

I'll check how to disappear the headers field when calling parseIsomorphicRequest multiple times.

hrsh7th: ~~Oh. I understood the reason. That the reason why is the `setRequestCookies` will overwrite the `headers` field.~~

No. Apparently it should be confirmed a little more firmly. I'll consider it. (_google translated_)

kettanaito: No worries!
The `headers` must not disappear—that's for certain. I just think that we don't need to deconstruct an existing `Headers` instance to construct it again 🤔 It'd be great to catch the issue you're experiencing into a unit test.

hrsh7th: I've considered it.

### The `parseIsomorphicRequest` always attaches the all cookies

This fact is commented in https://github.com/mswjs/msw/pull/1155/files#diff-738c9fc8b1eee13315e077f1aed118cc90af2e9cfb3965436290ff67a6d64284R33 .

I think it's correct so I don't modify it.

### The `setRequestCookie` consumes the `cookie` headers

So we should re-create the new `headers` field. (because it's overwritten).


Finally, I think current PR is correct.

kettanaito (approved):
Thanks for improving the library, @hrsh7th! Welcome to contributors 🎉

kettanaito:
## Released: v0.40.0 🎉

---

PR #1172, opened 2022-03-17
feat: add service worker scope and worker script URL to "Mocking enabled" message

### LINKED ISSUE #1103: Print the worker's scope in the "Mocking enabled" message
It may help in some cases of debugging to see what is the registered worker's scope. I don't think there's a way to see the worker's scope in the DevTools, it only lists "clients".

```
[MSW] Mocking enabled.
- Documentation: https://mswjs.io/docs
- Find an issue? <GITHUB_LINK>
- Scope: /
```

We may also consider printing additional information:
- The worker script location;
- The list of initial request handlers.

Poivey:
close #1103

As discussed, printing service worker scope & location in "Mocking enabled" message.
Scope and location are found in context to get effective service worker values and not just user raw options.

## Question
Do we need to cover cases where these fields would not be present in context (Eg. worker or registration is null) ? I think it should not happen because mocking would not be enabled in this case anyway so we would not print this message, I would like your opinion on that as this is my first PR here.

## Screenshot
Screenshot 2022-03-17 at 18 18 32

### DIFF
--- a/src/setupWorker/start/utils/printStartMessage.ts
+++ b/src/setupWorker/start/utils/printStartMessage.ts
@@ -1,20 +1,16 @@
-import { SetupWorkerInternalContext } from '../../glossary'
 import { devUtils } from '../../../utils/internal/devUtils'
 
-interface PrintStartMessageArgs {
+export interface PrintStartMessageArgs {
   quiet?: boolean
   message?: string
-  scope?: string
-  workerLocation?: string
+  workerUrl?: string
+  workerScope?: string
 }
 
 /**
  * Prints a worker activation message in the browser's console.
  */
-export function printStartMessage(
-  context: SetupWorkerInternalContext,
-  args: PrintStartMessageArgs = {},
-) {
+export function printStartMessage(args: PrintStartMessageArgs = {}) {
   if (args.quiet) {
     return
   }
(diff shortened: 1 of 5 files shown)


src/setupWorker/start/utils/printStartMessage.ts
```diff
@@ -1,14 +1,20 @@
+import { SetupWorkerInternalContext } from '../../glossary'
 import { devUtils } from '../../../utils/internal/devUtils'
 
 interface PrintStartMessageArgs {
   quiet?: boolean
   message?: string
+  scope?: string
+  workerLocation?: string
```
MartinJaskulla: Why are you adding them to `PrintStartMessageArgs`? 

You don't seem to be using `args.workerLocation` or `args.scope` anywhere.

You are only using `context` e.g. `context.worker?.scriptURL`.

kettanaito:
Hey, @Poivey. Thank you for your work on this!

I'd like to look closer at the changes and see if exposing a `context` to the print message function is the best course of action we could take. I lean towards `printStartMessage` remaining pure and letting the callee provide it with the data like worker script location upon invocation.

kettanaito:
Okay, so what I've tried is I removed the `context` from the `printStartMessage`, making the function contextless. It also reduced the setup we need to test it actually printing the worker info upon the activation message. 

Renamed the commit to `fix:` since this is not a breaking change technically.

kettanaito (approved):

src/setupWorker/start/utils/printStartMessage.ts
Poivey: You are right, I'm sorry I missed that. Thank you @kettanaito for fixing it.

kettanaito:
Welcome to contributors, @Poivey! 🎉

kettanaito:
Thank you for reviewing this, @MartinJaskulla!

kettanaito:
## Released: v0.40.0 🎉

---

PR #1180, opened 2022-03-24
feat: add support for ctx.forward

### LINKED ISSUE #676: Add "ctx.forward" for forwarding (bypassing) requests
## What

A new `ctx.forward` utility may be added to forward the currently captured request to its original destination (bypass).

## Why

- See #655 

Currently bypassing is done via implicit/explicit return in the resolver. This is not intuitive and deterministic action, which also makes it hard to distinguish when the user _forgot_ to return a mocked response, and when they _meant_ to bypass intentionally. 

With the introduction of this feature, the library could reliably print a warning when you have a handler, but return no response from the resolver—it becomes safe to assume that you forgot to do that. 

## API

```js
  if (anything) {
    return ctx.forward()
  }

  return res(ctx.text('mocked'))
})
```

rtalvarez:
## What
Adds support for `ctx.forward()`

## Why

This allows developers to explicitly forward the request to its original destination conditionally inside a handler, without relying on the existing `return undefined` mechanism (which is in place for a different reason).

Using `ctx.forward()` is interpreted as a direct command to forward the current request, and not a mistake by the developer

```
  // Skip this handler and perform the request as-is
  // if the request URL contains the "skip" query parameter.
  if (req.url.searchParams.get('skip')) {
    return ctx.forward()
  }

  return res(ctx.json({ id: 1 }))
})
```

Closes: https://github.com/mswjs/msw/issues/676
Additional context: https://github.com/mswjs/msw/issues/655

## Testing done
I've used a dummy app to verify these scenarios manually:

- Normal use-case: Mock a route, and return mocked data -> Verified mock data is returned 
- Mistake usecase: Declare a handler that does nothing -> Verified request is bypassed, and a warning printed to the console
- Explicit forward: Declare a handler that returns `ctx.forward()` -> Verified request is bypassed, NO warning printed

@kettanaito, The unit test included is pretty dumb, let me know if you think this warrants testing at an additional layer


kettanaito:
Hey, @rtalvarez. Thank you so much for your work on this feature. I've decided to go with https://github.com/mswjs/msw/pull/1204.

kettanaito:
## Released: v0.40.0 🎉

---

PR #1188, opened 2022-04-03
fix(multipart/form-data): Include form data in the request payload when making a proxy request with ctx.fetch(req)

### LINKED ISSUE #1158: Response patching is misbehave with FormData
## Describe the bug

FormData was not sent alone with `ctx.fetch(req)`, an empty object instead.

## Environment

- `msw: 0.39.1`
- `nodejs: 16.13.2`
- `npm: 8.5.0`
- `firefox: 97.0.1`

## To Reproduce

handler:
```js
export const handlers = [
  rest.post("/post", async (req, res, ctx) => {
    console.log("msw body:", req.body);
    const proxy = await ctx.fetch(req);

    return res(
      ctx.status(proxy.status),
      ctx.body(await proxy.text())
    );
  }),
];
```

invoke:
```js
const formData = new FormData();
formData.append("key", "value");
fetch("/post", { method: "POST", body: formData });
```

The original request has form data:
image

But request by msw has no form data:
image

## Expected behavior

I'm expecting that `ctx.fetch(req)` behaves exact the same way as it without thourgh msw.

I'm not sure why this happens. Maybe fetch with `content-type: multipart/form-data` can't deal with plain object?
Sorry, I can not reproduce a precise fetch request to confirm my guessing.

> FormData is converted to an plain object ([see also issue-599](https://github.com/mswjs/msw/issues/599#issuecomment-779728207)):
https://github.com/mswjs/msw/blob/0f0827c00d185533af258da8c0bb1285ab1f9b4c/src/context/fetch.ts#L29-L34

MartinJaskulla:
Fixes https://github.com/mswjs/msw/issues/1158

When making a FormData request, the browser adds the content-type header automatically and adds a boundary e.g.
`"content-type": multipart/form-data; boundary=----WebKitFormBoundarygyGKnRF8C9LT0BhB"`.
```ts
const formData = new FormData();
formData.append("key", "value");
fetch("/post", { method: "POST", body: formData });
```

When this request is intercepted and proxied

```ts
export const handlers = [
  rest.post("/post", async (req, res, ctx) => {
    console.log("msw body:", req.body);
    const proxy = await ctx.fetch(req);

    return res(
      ctx.status(proxy.status),
      ctx.body(await proxy.text())
    );
  }),
];
```
and we are creating a new request in `mockServiceWorker.js`,

```ts
const originalRequest = new Request(requestClone, {
      headers: new Headers(cleanRequestHeaders),
})
```
then `cleanRequestHeaders` will have the property `"content-type": multipart/form-data; boundary=----WebKitFormBoundarygyGKnRF8C9LT0BhB"`. The browser treats this as a user-provided header and does not add a new boundary value. This will break the body and no payload will be send. If we delete the content-type header, the browser will automatically re-add it with a new boundary value and the form data is included in the payload.

Solution found here:
https://community.cloudflare.com/t/cannot-seem-to-send-multipart-form-data/163491/2

Notice the new boundary value:
Screenshot 2022-04-04 at 08 45 07
Screenshot 2022-04-04 at 08 46 03


kettanaito:
Hey, @MartinJaskulla. Thanks for opening this!

This is interesting. As I understand, the issue happens only when proxying requests with `ctx.fetch`. What do you think if we add this headers logic to that very function instead of the worker script?

Changes to the worker script are always breaking changes, as developers would have to update those on their machines to guarantee MSW works as expected (even if they're using an automatic update route). I'd much prefer for this to be tackled in the underlying API like `ctx.fetch`, especially if that's where the problem occurs. 

Do you see any implications of moving this change to `ctx.fetch`?

MartinJaskulla:
Makes sense. `ctx.fetch` seems like a much better place for this check. Also easier to unit test.

Will take a look tomorrow evening.

kettanaito:
Sounds fantastic! Thank you so much.

src/context/fetch.ts
```diff
@@ -26,6 +28,23 @@ const createFetchRequestParameters = (input: MockedRequest): RequestInit => {
...
 
+  // When the user made their original request, the browser added a boundary e.g.
+  // "content-type": multipart/form-data; boundary=----WebKitFormBoundarygyGKnRF8C9LT0BhB"
+  // If we now reuse this string as a user-provided header, the form-data will break and no payload will be send
+  // https://community.cloudflare.com/t/cannot-seem-to-send-multipart-form-data/163491/2
+  if (
+    input.headers.get('content-type')?.includes('multipart/form-data') &&
+    typeof body === 'object'
+  ) {
+    input.headers.delete('content-type')
```
MartinJaskulla: Is it a problem that I am mutating the input?

It might be better to make a copy of the headers and mutate the copy:

```ts
const requestParameters: RequestInit = {
    ...input,
    headers: new Headers(input.headers),
    body: undefined,
  }
```

Would I use `Headers` or `HeadersPolyfill` for that?

src/context/fetch.ts
```diff
@@ -26,6 +28,23 @@ const createFetchRequestParameters = (input: MockedRequest): RequestInit => {
...
+    const formData = new FormData()
```
MartinJaskulla, line 40: If we make the fix in `fetch.ts`, we need to create fresh form data, which was not necessary in `mockServiceWorker.js`. Previous commit: https://github.com/mswjs/msw/pull/1188/commits/e00de1e4b8bafb2dbbe29c0ffc500daf7b3c3246.
 
I would prefer doing the fix in` mockServiceWorker.js` as seen in previous commit, but I understand the concern about breaking changes

src/context/fetch.ts
kettanaito: You're absolutely correct: let's make a copy of the original input headers and mutate that copy:

```js
const headers = new Headers(input.headers)
headers.delete('content-type')
```

And yes, you'd use `headers-polyfill` for this so it works in any environment 👍

src/context/fetch.test.ts
```diff
@@ -34,3 +36,29 @@ test('aguments RequestInit with the Record<string, string> headers', () => {
...
+    // User decides to proxy the intercepted request to real endpoint in their handler.js:
+    // const proxy = await ctx.fetch(req);
+    // Now fetch() from fetch.ts will receive the body as a plain object even though the user send FormData:
+    body: {
+      key: 'value',
+    },
+    headers: new HeadersPolyfill({
+      'content-type': CONTENT_TYPE,
+    }),
+  } as unknown as MockedRequest
```
kettanaito: Please consider using `createMockRequest` utility function. You can find its usage examples if you search the project for its name. It constructs a full mocked response for you so you don't have to deal with partial representations and type casting.

src/context/fetch.ts
kettanaito: What would happen if instead of removing the entire `Content-Type` header we:

1. Parsed it.
2. Removed only the `boundary` part?

I have concerns that removing the header means we're entrusting the browser to guess the user's intentions. This may work in most cases but I think it may fail in other. Specifying a `Content-Type` is a user's indention, so if we can preserve the header but remove any automatically appended boundaries that'd be the least invasive solution. 

What do you think?

MartinJaskulla: I intercepted the requests made by Chrome with [HTTP Toolkit](https://httptoolkit.tech/).

There are two different issues:
- Chrome dev tools has a visual bug. The payload is always included, but displayed in three different ways in the dev tools (see screenshots below). There is no difference between not sending the `Content-Type` header, sending the header without a `boundary` and sending the header with a `boundary` (at least in Chrome and Firefox). The data send over the network is always the same. **So we don't need need to delete or modify the header at all**. Btw: Chrome and Firefox are adding their own `boundary` value whatever the user does. 
- However we still need to create a new `FormData` in `fetch.ts`

After the fix the correct data will be send over the network, but developers using Chrome will not realise this if they rely only on Chrome's dev tools, but don't inspect the network traffic.

# Chrome
### Sending no header
Screenshot 2022-04-04 at 18 32 07
Screenshot 2022-04-04 at 18 32 14

### Sending `'content-type': 'multipart/form-data'`
Screenshot 2022-04-04 at 18 33 07
Screenshot 2022-04-04 at 18 33 12

### Sending ` 'content-type': 'multipart/form-data; boundary=----WebKitFormBoundarygyGKnRF8C9LT0BhB'`
Screenshot 2022-04-04 at 18 33 28
Screenshot 2022-04-04 at 18 33 40

# HTTP Toolkit
### Before this PR
no-new-form-data

### Creating new form data, but not modifying the header
new-form-data-only-sends-payload

src/context/fetch.ts
```diff
@@ -26,14 +28,24 @@ const createFetchRequestParameters = (input: MockedRequest): RequestInit => {
...
-    typeof body === 'number' ||
-    typeof body === 'boolean'
-  ) {
-    requestParameters.body = JSON.stringify(body)
-  } else {
-    requestParameters.body = body
+  switch (typeof body) {
+    case 'object':
+      if (input.headers?.get('content-type')?.includes('multipart/form-data')) {
+        const formData = new FormData()
```
kettanaito, line 34: So the issue was that `req.body` was a string? Because all intercepted requests have a string body. 

And here you're reconstructing a `FormData` instance to assign to the proxied request's body without even touching the `content-type` header. Do I understand this correctly?

MartinJaskulla: Correct

src/context/fetch.ts
kettanaito: Can I just take a moment to appreciate your effort on this? You've done a fantastic job getting to the core of the issue 👏 

I see now. It's a matter of re-constructing the `FormData`. I still think that's because the intercepter `req.body` will be text received from the worker/Node.js interceptor. I'm not comfortable to assess if we should extend upon the `IsomorphicRequest` to allow deriving textual request data to other formats similar to have `window.Request` does. This may be a good idea for the future discussions. 

For now, I agree with the point to address the issue at its immediate context: `ctx.fetch`.

src/context/fetch.test.ts
```diff
@@ -1,7 +1,9 @@
 /**
  * @jest-environment jsdom
  */
-import { augmentRequestInit } from './fetch'
+import { augmentRequestInit, createFetchRequestParameters } from './fetch'
+import { createMockedRequest } from '../../test/support/utils'
+import HeadersPolyfill from 'headers-polyfill/lib/Headers'
```
kettanaito:
```suggestion
import { Headers } from 'headers-polyfill'
```

src/context/fetch.ts
kettanaito, line 34: There are two issues with our approach here. 

1. It will only work in the browser. `FormData` doesn't exit in Node.js, so this line will throw. 
2. I think we're trying to handle a scenario much later than we should be doing that. 

It looks like we should re-create a `FormData` when parsing intercepted request body right here:

https://github.com/mswjs/msw/blob/6fc89f762681aa7f6f697ea78fa5f7c736b35c5d/src/utils/request/parseBody.ts#L16-L21

I think the fix should be somewhere in `parseBody` or `parseMultipartData`. 

I may be wrong on this though. Do you find this request body processing chain to be correct?

1. Request is intercepted, its body is text because we only transfer text via WebSocket channel or Node.js interceptor.
2. Request body is parsed in `parseBody`. We see it's multipart form data, we convert the string to a JSON representation of the form data.
3. We pass the converted JSON request body to `req.body`, it arrives to `ctx.fetch`.
4. We reconstruct `FormData` from the JSON request body and dispatch the native fetch.

It looks like if we reconstruct `FormData` in `parseBody` then we can drop any extra body processing in `ctx.fetch`. This only comes down to one question:

- What do you expect `req.body` to be when intercepted a `multipart/form-data` request? Do you expect it to be `FormData`? Or a JSON representation of the form data?

src/context/fetch.test.ts
```diff
@@ -34,3 +36,26 @@ test('aguments RequestInit with the Record<string, string> headers', () => {
...
+  // const proxy = await ctx.fetch(req);
+  // Now fetch() from fetch.ts will receive the body as a plain object even though the user send FormData:
+  const input = createMockedRequest({
+    method: 'POST',
+    body: {
+      key: 'value',
+    },
+    headers: new HeadersPolyfill({
+      'content-type':
+        'multipart/form-data; boundary=----WebKitFormBoundaryostedjdEBW2xxin5',
```
kettanaito, line 55: A useful warning from MDN:

> Warning: When using FormData to submit POST requests using [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) or the [Fetch_API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) with the multipart/form-data [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) (e.g. when uploading Files and Blobs to the server), do not explicitly set the Content-Type header on the request. Doing so will prevent the browser from being able to set the Content-Type header with the boundary expression it will use to delimit form fields in the request body.
> —[Source](https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects)

So we do have to remove the `Content-Type` request header for the `FormData` to be handled correctly.

src/context/fetch.ts
MartinJaskulla: > What do you expect req.body to be when intercepted a multipart/form-data request? Do you expect it to be FormData? Or a JSON representation of the form data?

Do you mean inside a user defined `RequestHandler`?  That's an interesting question.

Without `msw`:
1. User calls native `fetch`: ` fetch('/post', { method: 'POST', body: formData })`
2. The browser modifies the request e.g. adds `Content-Type` header, changes body to a string: `"------WebKitFormBoundaryxOsEsNP9Pphx5RC6\r\nContent-Disposition: form-data; name=\"key\"\r\n\r\nvalue\r\n------WebKitFormBoundaryxOsEsNP9Pphx5RC6--\r\n"` etc.
3. Browser sends request to the backend

With `msw`:

4. The service worker intercepts the request in step 3 and passes it in a 4th step to the user's `RequestHandler`.

If one is not an expert in service workers or `msw`, it would be reasonable to expect `RequestHandler` to receive either
- the original request from step 1 without any modifications (has no headers, has `FormData` body)
- whatever is send to the backend (result of step 2)

However what is actually happening is that `msw` modifies the request again between step 3 and 4 (to make usage in `RequestHandler` more convenient?)  e.g. by changing `.headers` from a plain object to `new Headers` in `parseWorkerRequest.ts`.

So I guess it would be valid to change the body back to `FormData` for convenience as well. 

On the other hand `msw` tries to keep `MockedRequest` in the browser and NodeJS equivalent (Why does `parseWorkerRequest` return `RestRequest` with an extra `params` property?), so users can work with requests in the same way, which speaks for a plain object body in both environments.

But then again how useful is that equivalence in the case of NodeJS anyways? Without `FormData` you cannot write your unit tests in the same way the browser makes the fetch request anyways.

Maybe going for changing the body back to `FormData` is more in line with the current approach? What about also adding a NodeJS polyfill for FormData?

Very curious about what you would expect (JSON or FormData)!

MartinJaskulla: If we wanted the original body from step 1, but the modifications from step 2:

Probably a bad idea with a lot of breaking changes, but would it be possible to wrap the native `fetch` (as it is done while waiting for the serviceworker to initialise in `deferNetworkRequestsUntil.ts` ) in a function `keepOriginalThenFetch`, which stores the original request from step 1 and then calls the native fetch? 
```ts
const originalFetch = window.fetch
window.fetch = function keepOriginalThenFetch(...nativeFetchArgs) {
  storeOriginalRequestForParseWorkerRequest(...nativeFetchArgs)
  originalFetch(...nativeFetchArgs)
}
```

Then in `parseWorkerRequest` we get the request modified by the browser from the service worker and create the `RestRequest` the same way as before (keeping the modifications made by the browser in step 2), but afterwards we retrieve the original request from step 1 stored by `keepOriginalThenFetch` and overwrite the properties of `RestRequest`. Then `RestRequest` would have the original body from step 1 and all other original user specified properties such as `headers` (if given).

https://github.com/mswjs/msw/issues/1158 was about `FormData` only, but what about all the other types supported by `BodyInit`, which are missing in`DefaultBodyType`?
```ts
type BodyInit = Blob | BufferSource | FormData | URLSearchParams | ReadableStream<Uint8Array> | string;
```
With `keepOriginalThenFetch` all types would work and there would not be any need to parse the body or call `parseMultipartData` any more.

I haven't used `msw` enough or even made a POST request with `BufferSource` in my life, so I don't really know what I am talking about here 😅

PR #1189, opened 2022-04-04
fix(graphql): prevents REST API requests with a "query" from being treated as GraphQL

### LINKED ISSUE #1184: Rest Api from Cubejs gets mistaken as GraphQL when using both in one handler
### Prerequisites

### Environment check

### Node.js version

14.17.5

### Reproduction repository

https://github.com/brantleyenglish/msw-test

### Reproduction steps

Run `yarn` then `yarn test`

### Current behavior

Currently I am getting this error, and the mocked rest endpoint does not get populated with mock data. 

```
console.error
[MSW] Failed to intercept a GraphQL request to "GET https://azure-bonobo.aws-us-west-2.cubecloudapp.dev/cubejs-api/v1/load": cannot parse query. 
See the error message from the parser below.
Syntax Error: Expected Name, found String "measures".
```
This url should be parsed as a rest endpoint. This is not a graphql query.

### Expected behavior

I expect the test to pass the `resultSet` to return some form of data.

kettanaito:
Fixes #1184

### DIFF
--- a/src/utils/internal/parseGraphQLRequest.ts
+++ b/src/utils/internal/parseGraphQLRequest.ts
@@ -11,7 +11,7 @@ import { devUtils } from './devUtils'
 import { jsonParse } from './jsonParse'
 
 interface GraphQLInput {
-  query: string | null
+  query: string | undefined
   variables?: GraphQLVariables
 }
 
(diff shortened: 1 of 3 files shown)


src/utils/internal/parseGraphQLRequest.ts
```diff
@@ -86,24 +86,43 @@ function extractMultipartVariables<VariablesType extends GraphQLVariables>(
...
+ * Check if the given GraphQL query string is potentially valid
+ * without parsing it. This returns false fast when analyzing
+ * GET/POST requests with the "query" parameter/body property
+ * which are not a GraphQL request.
+ */
+function maybeGraphQLQuery(query?: string | null): query is string {
+  const normalizedQuery = query?.toLowerCase().trim() || ''
+
+  return (
+    normalizedQuery.startsWith('query') ||
```
kettanaito, line 99: This will return `false` fast for GraphQL queries with comments. But I don't think it's allowed for the query string to contain comments, I'd expect GraphQL clients should strip them. 

That'd be nice to be researched.

kettanaito:
Not a good idea.

A better idea: https://github.com/mswjs/msw/discussions/535#discussioncomment-2504052

PR #1199, opened 2022-04-11
feat: add "unhandledException" life-cycle event

### LINKED ISSUE #754: Add way to detect and log exceptions in request handlers.

I'm currently using `msw` for mocking network requests in `jest` tests. I haven't found an easy way to detect when exceptions are thrown in my request handlers. For example, the code below will produce an exception due to bug and `msw` will silently convert the error to a 500 response.

```js
rest.post("/authorize", (req, res, ctx) => {
	// ... some authorization logic ...
	
	// getCurrentUser throws because there is a bug in it
	const user = getCurrentUser();

	// ... more authorization logic ...

	return res(
		ctx.status(200),
		ctx.json({
			redirectUri,
		}),
	);
});
```

None of the [lifecycle events](https://mswjs.io/docs/extensions/life-cycle-events#events-list) are emitted when the exception is thrown. A new lifecycle event like `response:error` would allow me to write something like
```js
worker.on('response:error', (res, reqId, error) => {
	console.error("Request handler threw error.", error);
});
```

This would help developers quickly ascertain the source of some test failures without having to follow a chain of symptoms back to a failed `msw` request handler.

### Alternative 1
I could use a higher order function wrapper like
```js
rest.post("/authorize", logAndRethrowErrors((req, res, ctx) => {
	// ...
}));
```
```js
function logAndRethrowErrors(fn) {
	return (...args) => {
		try {
			return fn(...args);
		} catch (e) {
			console.error(e);
			throw e;
		}
	};
}
```
Pros:
This code logs an error which likely includes helpful stack trace.

Cons:
This requires all developers on my project to add `logAndRethrowErrors` which may be easy to forget. I'd instead prefer to fail loudly globally.

### Alternative 2
I could do something like this during setup
```js
// Setup
const server = setupServer(...handlers);
const requestsWithoutResponses = new Map();
server.on("request:start", (req) => {
	requestsWithoutResponses.set(req.id, req);
});
server.on("request:end", (req) => {
	requestsWithoutResponses.delete(req.id);
});
```
and this during teardown
```js
// Teardown
server.close();
for (const [, req] of requestsWithoutResponses) {
	console.error(
		"Request received no response. This likely means the request handler threw an error, the error was swallowed by msw, and silently converted to a 500 response.\nRequest:",
		req,
	);
}
```
and assuming any request that did not emit `"request:end"` resulted in an exception.

Pros:
This gives me the global ability fail loudly.

Cons:
Does not have the error message to emit.

### Alternative 3
Combine alternative 1 and 2 above.

Pros:
Allows us to emit error trace with `logAndRethrowErrors` and use the fallback teardown code to hint that a developer forgot `logAndRethrowErrors`.

Cons:
This is quite a few lines of non-standard tooling for something that could be done succinctly if `msw` offered a simple API for it.

MartinJaskulla:
- Closes https://github.com/mswjs/msw/issues/754

Users can now react to errors thrown in their RequestHandlers. Before this PR all errors were silently converted into a 500 response (only in the browser). Now users can subscribe to a new event:

```js
worker.events.on('unhandledException', (error, request) => {})
```

### DIFF
--- a/src/sharedOptions.ts
+++ b/src/sharedOptions.ts
@@ -21,6 +21,7 @@ export interface LifeCycleEventsMap<ResponseType> {
   'request:end': (request: MockedRequest) => void
   'response:mocked': (response: ResponseType, requestId: string) => void
   'response:bypass': (response: ResponseType, requestId: string) => void
+  unhandledException: (error: Error, request: MockedRequest) => void
 }
 
 export type LifeCycleEventEmitter<ResponseType> = Pick<
--- a/src/utils/handleRequest.ts
+++ b/src/utils/handleRequest.ts
@@ -1,3 +1,4 @@
+import { until } from '@open-draft/until'
 import { StrictEventEmitter } from 'strict-event-emitter'
 import { MockedRequest, RequestHandler } from '../handlers/RequestHandler'
 import { ServerLifecycleEventsMap } from '../node/glossary'
@@ -65,11 +66,20 @@ export async function handleRequest<
   }
 
   // Resolve a mocked response from the list of request handlers.
-  const lookupResult = await getResponse(
-    request,
-    handlers,
-    handleRequestOptions?.resolutionContext,
-  )
+  const [lookupError, lookupResult] = await until(() => {
+    return getResponse(
+      request,
+      handlers,
+      handleRequestOptions?.resolutionContext,
+    )
+  })
+
+  if (lookupError) {
+    // Allow developers to react to unhandled exceptions in request handlers.
+    emitter.emit('unhandledException', lookupError, request)
+    throw lookupError
+  }
+
   const { handler, response } = lookupResult
 
   // When there's no handler for the request, consider it unhandled.
--- a/test/msw-api/setup-server/life-cycle-events/on.test.ts
+++ b/test/msw-api/setup-server/life-cycle-events/on.test.ts
@@ -35,6 +35,9 @@ beforeAll(async () => {
     rest.post(httpServer.http.makeUrl('/no-response'), () => {
       return
     }),
+    rest.get(httpServer.http.makeUrl('/unhandled-exception'), () => {
+      throw new Error('Unhandled resolver error')
+    }),
   )
   server.listen()
 
@@ -60,6 +63,12 @@ beforeAll(async () => {
 
   server.events.on('response:bypass', (res, requestId) => {
     listener(`[response:bypass] ${res.body} ${requestId}`)
+  })
+
+  server.events.on('unhandledException', (error, req) => {
+    listener(
+      `[unhandledException] ${req.method} ${req.url.href} ${req.id} ${error.message}`,
+    )
   })
 
   // Supress "Expected a mocking resolver function to return a mocked response"
@@ -153,6 +162,16 @@ test('emits events for an unhandled request', async () => {
   expect(listener).toHaveBeenNthCalledWith(
     4,
     `[response:bypass] majestic-unknown ${requestId}`,
+  )
+})
+
+test('emits unhandled exceptions in the request handler', async () => {
+  const url = httpServer.http.makeUrl('/unhandled-exception')
+  await fetch(url).catch(() => undefined)
+  const requestId = getRequestId(listener)
+
+  expect(listener).toHaveBeenCalledWith(
+    `[unhandledException] GET ${url} ${requestId} Unhandled resolver error`,
   )
 })
 
--- a/test/msw-api/setup-worker/life-cycle-events/on.mocks.ts
+++ b/test/msw-api/setup-worker/life-cycle-events/on.mocks.ts
@@ -8,6 +8,9 @@ const worker = setupWorker(
   rest.post('/no-response', () => {
     return
   }),
+  rest.get('/unhandled-exception', () => {
+    throw new Error('Unhandled resolver error')
+  }),
 )
 
 worker.events.on('request:start', (req) => {
@@ -35,6 +38,12 @@ worker.events.on('response:mocked', async (res, requestId) => {
 worker.events.on('response:bypass', async (res, requestId) => {
   const body = await res.text()
   console.warn(`[response:bypass] ${body} ${requestId}`)
+})
+
+worker.events.on('unhandledException', (error, req) => {
+  console.warn(
+    `[unhandledException] ${req.method} ${req.url.href} ${req.id} ${error.message}`,
+  )
 })
 
 worker.start({
--- a/test/msw-api/setup-worker/life-cycle-events/on.test.ts
+++ b/test/msw-api/setup-worker/life-cycle-events/on.test.ts
@@ -92,6 +92,17 @@ test('emits events for an unhandled request', async () => {
   ])
 })
 
+test('emits unhandled exceptions in the request handler', async () => {
+  const runtime = await createRuntime()
+  const url = runtime.makeUrl('/unhandled-exception')
+  await runtime.request(url)
+  const requestId = getRequestId(runtime.consoleSpy)
+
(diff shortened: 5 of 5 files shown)


src/sharedOptions.ts
```diff
@@ -18,6 +18,11 @@ export interface LifeCycleEventsMap<ResponseType> {
   'request:start': (request: MockedRequest) => void
   'request:match': (request: MockedRequest) => void
   'request:unhandled': (request: MockedRequest) => void
+  'request:unhandledException': (
+    request: MockedRequest,
```
kettanaito: What do you think if me model the order of arguments based on the developer's expectations of them?

I think when handling exceptions, developers are most interested in the error thrown, so I'd recommend it go first. 

```ts
'request:unhandledException': (error: Error, request: MockedRequest, handler: RequestHandler) => void
```

src/utils/getResponse.ts
```diff
@@ -47,7 +51,14 @@ export const getResponse = async <
       return executionResult
     }
 
-    const result = await handler.run(request, resolutionContext)
+    const [error, result] = await until(() =>
+      handler.run(request, resolutionContext),
```
kettanaito: I do like checking for exceptions immediately around `handler.run` but I don't think there are many things that may throw in `getResponse` on our part. I don't think that making this function accept an `emitter` is a good idea, it should be a function of handlers to mocked response, with as little side-effects as possible. 

I'd probably try moving this logic around `getResponse` and gather feedback from developers. We are unlikely to experience exceptions out of handler's context, but if we do we can always revisit—the surface at which we check for exception is an implementation detail, so we can change it at any time without introducing a breaking change.

test/msw-api/setup-server/life-cycle-events/on.test.ts
```diff
@@ -156,6 +165,23 @@ test('emits events for an unhandled request', async () => {
   )
 })
 
+test("emits errors from the user's RequestHandler", async () => {
+  const url = httpServer.http.makeUrl('/error')
+  // In NodeJS errors are not gracefully transformed into 500 responses
+  await fetch(url).catch(() => undefined)
+  const requestId = getRequestId(listener)
+
+  expect(listener).toHaveBeenNthCalledWith(
```
kettanaito: I believe there's a different test that ensures that `request:start` is called. We can remove this assertion from this particular test, keeping it focused on the new event we introduce.

test/msw-api/setup-worker/life-cycle-events/on.mocks.ts
```diff
@@ -22,6 +25,12 @@ worker.events.on('request:unhandled', (req) => {
   console.warn(`[request:unhandled] ${req.method} ${req.url.href} ${req.id}`)
 })
 
+worker.events.on('request:unhandledException', (req, handler, error) => {
+  console.warn(
+    `[request:unhandledException] ${req.method} ${req.url.href} ${req.id} ${handler.info.header} ${error.message}`,
```
kettanaito: `handler.info.header` is synonymous to `req.method req.url.href` at the beginning of this string. I think we can drop `header.info.header` from this warning.

test/msw-api/setup-server/life-cycle-events/on.test.ts
```diff
@@ -50,6 +53,12 @@ beforeAll(async () => {
     listener(`[request:unhandled] ${req.method} ${req.url.href} ${req.id}`)
   })
 
+  server.events.on('request:unhandledException', (req, handler, error) => {
```
kettanaito: I don't like the way I've modeled these tests in the past. Instead of invoking a listener in this callback function's body, I should've passed the listener as the callback itself. This would allow us to assert _all_ arguments passed along a particular event.

```js
server.events.on('foo', listener)

expect(listener).toHaveBeenCalledWith([mockedRequest, handler, error])
```

Let's refactor this after this feature 👍

kettanaito (commented):
Hey, @MartinJaskulla! Thank you for your work on this, it looks superb!

I had only one substantial comment about the order of arguments in the listener. I'd like to hear your opinion on that.

src/utils/getResponse.ts
MartinJaskulla: How to pass the responsible `handler` back to `handleRequest` in an elegant way (in order to emit it there)?

`handleRequest` passes down `handlers`, so it does not know which `handler` threw the error.

I still need to wrap `hander.run` in `getResponse`.
```ts
    const [error, result] = await until(() =>
      handler.run(request, resolutionContext),
    )

    if (error) {
      // Don't emit anymore (emitter is not passed down to keep function as pure as possible)
      // emitter.emit('request:unhandledException', error, request, handler)
   
      // Somehow pass the handler
      error.handler = handler // Property 'handler' does not exist on type 'Error'.
      throw error
    }
```

Adding random properties on errors is probably not great. You also don't seem to make use of custom errors classes extending the default error anywhere yet.

test/msw-api/setup-worker/life-cycle-events/on.mocks.ts
MartinJaskulla: It is true that it makes no sense to add `handler.info.header` to the warning in a real scenario.

However I want to test that `handler` was actually passed and its properties can be used. `handler.info.header` makes the most sense to me as it is the only property that has a value related to the test setup (e.g. uses correct url). `handler.markAsSkipped` has no meaning here.

src/utils/getResponse.ts
kettanaito: Let's implement this without `handler` in the arguments and see what we can do. If anything, we can drop the handler in the end until our users actually need its reference. For discovery purposes they can always follow the stack trace of the thrown error to arrive at the relevant handler.

test/msw-api/setup-worker/life-cycle-events/on.mocks.ts
MartinJaskulla: Since we will remove `handler` from the arguments of the event, this problem does not exist anymore

MartinJaskulla:
> Hey, @MartinJaskulla! Thank you for your work on this, it looks superb!
> 
> I had only one substantial comment about the order of arguments in the listener. I'd like to hear your opinion on that.

Thank you for your support! Learning a lot 🥳

And yes, errors as the first argument make sense!

test/msw-api/setup-server/life-cycle-events/on.test.ts
MartinJaskulla: I am on vacation until 25.04. Will refactor it afterwards 👍

kettanaito:
Hey, @MartinJaskulla. I've force-pushed to your feature branch because we've reverted a significant chunk of one change, which made rebasing utterly impossible. I've tried to replicate your changes, so it should be okay now!

src/utils/handleRequest.ts
```diff
@@ -65,11 +66,20 @@ export async function handleRequest<
...
+    return getResponse(
+      request,
+      handlers,
+      handleRequestOptions?.resolutionContext,
+    )
+  })
+
+  if (lookupError) {
+    // Allow developers to react to unhandled exceptions in request handlers.
+    emitter.emit('unhandledException', lookupError, request)
```
kettanaito, line 79: I'm using the `unhandledException` event name (dropping the "request:" part) because exceptions are not related to requests being handled. They are related to request handlers or resolvers. I think a context-free `unhandledException` makes more sense in that regard.

kettanaito (approved):

kettanaito:
Welcome to the contributors, @MartinJaskulla! 🎉 
I can't wait to get this to our users.

kettanaito:
## Released: v0.40.0 🎉

---

PR #962, opened 2021-10-31
feat: allow forwarding of requests with ctx.forward

mcky:
As discussed in #234 and #676 `(Add "ctx.forward" for forwarding (bypassing) requests)` this adds a `ctx.forward()` method.

The new API looks like
```ts
rest.get('/user', (req, res, ctx) => {
  if (anything) {
    return res(ctx.forward())
  }

  return res(ctx.text('mocked'))
})
```
It differs slightly to the suggestion in #676 as it wraps the forward call in `res()` - if that's a deal breaker I can try to figure out how to implement it as per the issue

The behaviour is exactly the same as a `x-msw-bypass` header or returning undefined, just without the warnings.

Happy with any and all feedback - for now I've hacked in a `x-msw-bypass` into my own test-suite just to silence the undefined warnings 😅 

Todo:
- [ ] Update documentation

PR #923, opened 2021-09-22
feat(ctx): Add "ctx.forward" for forwarding requests

lsroa:
This tries to solve the discussed here #676 @kentcdodds @kettanaito 
Probably the test needs some improvements :thinking:

Issue #1175, opened 2022-03-20
Inferred path parameters fallback to "DefaultParamsType" when specifying "RequestBodyType"


kettanaito:
- Related to #1073 

## Current behavior

```ts
rest.get<never>('/user/:id', (req) => {
  req.params.id // is not an inferred type
  req.params // DefaultParamsType
})
```

## Expected behavior

```ts
rest.get<never>('/user/:id', (req) => {
  req.params.id // is an inferred "string"
  req.params // { id: string } (inferred)
})
```

kettanaito:
TypeScript does not support partial generics default values, so the behavior above is, sadly, expected. I can give it a try if we can somehow split the generic internally but I don't see that happening: the end developer must be able to provide both `ReqBodyType` generic and the string literal value of the path on the same function.

kettanaito:
I think this is a deal breaker of the path parameter inference because it renders the inference null when the user annotates their request/response body, which I expect most of the users to do. 

Do you have any ideas on how we can fix this, @cmolina? 🤔

cmolina:
I don't know. 

I just asked in the TypeScript's open PR, in case somebody can help us with this.

TylorS:
[...]

Issue #413, opened 2020-10-03
Provide a way (or document an existing way) to return a NetworkError once


noahbrenner:
## Is your feature request related to a problem? Please describe.

I would like to have a documented way of defining a handler that returns a NetworkError only once, then falls back to an existing handler. This can be helpful in testing an app's request-retrying logic.

---

I know how to return a valid response or a failing status code once:
```javascript
server.use(
  rest.get('/foo', (_req, res, ctx) => {
    return res.once(
      ctx.status(404, 'Not found')
    );
  })
);
```

And I know how to return a network error:
```javascript
server.use(
  rest.get('/foo', (_req, res, ctx) => {
    return res.networkError('Failed to conect');
  })
);
```

**But since these both depend on returning the result of different methods on `res`, I'm not sure how to use the current API to return a network error once.**

---

I tried a few options, including the three below, but they all *continued* to respond with errors after the first request. They also resulted in TypeScript errors, which makes sense:
```javascript
server.use(
  rest.get('/foo', (_req, res, ctx) => {
    return res.once(res.networkError('Failed to conect'));
  })
);
```

```javascript
server.use(
  rest.get('/foo', (_req, res, ctx) => {
    return res.once(() => res.networkError('Failed to conect'));
  })
);
```

```javascript
server.use(
  rest.get('/foo', (_req, res, ctx) => {
    return res.once((innerRes) => innerRes.networkError('Failed to conect'));
  })
);
```

## Describe the solution you'd like

I'm not familiar with `msw`'s internals, so there may be a better fit for the API, but an intuitive solution for me would be:

- **Provide `networkError` on the `ctx` object.**
  This would be consistent with the API for changing just about every other aspect of a response (`networkError` strikes me as out of place on the `res` object).

It could then be used something like this:
```javascript
server.use(
  rest.get('/foo', (_req, res, ctx) => {
    return res.once(
      ctx.networkError('Failed to conect')
    );
  })
);
```

## Describe alternatives you've considered

Another good option would be to provide `networkError` as a method on `once`:
```javascript
server.use(
  rest.get('/foo', (_req, res, ctx) => {
    return res.once.networkError('Failed to conect');
  })
);
```

Or to provide `once` on `rest` and `graphql` and all request methods on `once`:
```javascript
server.use(
  rest.once.get('/foo', (_req, res, ctx) => {
    return res.networkError('Failed to conect');
  })
);
```

Or to provide `once` as a method of all request methods and have it inherit the preceding method's behavior:
[...]

PR #1204, opened 2022-04-13
feat: add "req.passthrough"

### LINKED ISSUE #676: Add "ctx.forward" for forwarding (bypassing) requests

kettanaito:
> This is a breaking change.

- Originates from #655
- Closes #676
- Closes #1180 
- Closes #962 
- Closes #923

## Changes

- Adds `req.passthrough` to explicitly instruct MSW to perform an intercepted request as-is.
- Returning `undefined`/`null` from handlers now treats the request as unhandled.

### DIFF
--- a/src/handlers/RequestHandler.ts
+++ b/src/handlers/RequestHandler.ts
@@ -1,5 +1,10 @@
 import { Headers } from 'headers-polyfill'
-import { MockedResponse, response, ResponseComposition } from '../response'
+import {
+  MaybePromise,
+  MockedResponse,
+  response,
+  ResponseComposition,
+} from '../response'
 import { getCallFrame } from '../utils/internal/getCallFrame'
 import { isIterable } from '../utils/internal/isIterable'
 import { status } from '../context/status'
@@ -54,6 +59,7 @@ export interface MockedRequest<Body extends DefaultBodyType = DefaultBodyType> {
   referrerPolicy: Request['referrerPolicy']
   body: Body
   bodyUsed: Request['bodyUsed']
+  passthrough: typeof passthrough
 }
 
 export interface RequestHandlerDefaultInfo {
@@ -71,9 +77,9 @@ export type ResponseResolverReturnType<ReturnType> =
   | undefined
   | void
 
-export type MaybeAsyncResponseResolverReturnType<ReturnType> =
-  | ResponseResolverReturnType<ReturnType>
-  | Promise<ResponseResolverReturnType<ReturnType>>
+export type MaybeAsyncResponseResolverReturnType<ReturnType> = MaybePromise<
+  ResponseResolverReturnType<ReturnType>
+>
 
 export type AsyncResponseResolverReturnType<ReturnType> =
   | MaybeAsyncResponseResolverReturnType<ReturnType>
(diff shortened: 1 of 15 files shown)


src/handlers/RequestHandler.ts
```diff
@@ -272,3 +285,22 @@ export abstract class RequestHandler<
...
 }
+
+/**
+ * Bypass this intercepted request.
+ * This will make a call to the actual endpoint requested.
+ */
+export function passthrough(): MockedResponse<null> {
+  // Constructing a dummy "101 Continue" mocked response
+  // to keep the return type of the resolver consistent.
+  return {
```
kettanaito, line 296: I'm returning a `MockedResponse` from `passthrough` to keep the same return type of the response resolver. I don't like this approach but I'll use it for now. 

We need to refactor the return type of the resolver to be some sort of `Instruction`, which could be:

- Passthrough
- Fallthrough
- Use a mocked response
- Return a network error
- Anything else in the future

It should be a strictly defined object representing the library's next steps. Resolvers don't necessarily return mocked responses, as in the Passthrough/Fallthrough scenarios. Each intention can also be permanent or one-time (like `res.once()`). This is a great opportunity to solve #413. 

The source for these instructions can also differ:

- `res()` instructs to use a mocked response.
- `req.passthrough()` instructs to passthrough this request.
- `res.networkError()` (or a different API) instructs to return a network error.

Decoupling the source from the instruction will allow us to design semantic APIs without locking ourselves in a `MockedResponse` structure.

kettanaito: Proposing this in https://github.com/mswjs/msw/issues/1207.

kentcdodds:
Sweet! Where are the docs for this?

kettanaito:
@kentcdodds, I've added the documentation in https://github.com/mswjs/mswjs.io/pull/192.

abhishekarun:
When is this coming in formal release?

kettanaito:
@abhishekarun, as soon as #1175 is addressed. If you have some spare time I'd certainly appreciate some help with that issue. 

Also, keep in mind that this is not a new feature but rather an improvement over the existing bypass/passthrough behavior.

kettanaito:
## Released: v0.40.0 🎉

---

yshrsmz:
This PR seems to break the compatibility with `@mswjs/http-middleware`. I've created a PR to fix the issue https://github.com/mswjs/http-middleware/pull/10

taylorkline:
> Returning undefined/null from handlers now treats the request as unhandled.

I would politely suggest @kettanaito that the release notes for this release mention this as a breaking change.

kettanaito:
@taylorkline, [updated](https://github.com/mswjs/msw/releases/tag/v0.40.0). Breaking changes between minor releases in pre-1.0 packages are implied but I agree this should've been communicated better.

taylorkline:
> @taylorkline, [updated](https://github.com/mswjs/msw/releases/tag/v0.40.0). Breaking changes between minor releases in pre-1.0 packages are implied but I agree this should've been communicated better.

Thank you!

PR #1219, opened 2022-04-26
feat: improve PathParams type to support interface

kotarella1110:
Passing an interface to the generics of PathParams causes a type error because interface doesn't have an [implicit index signature](https://github.com/microsoft/TypeScript/pull/7029).

```ts
interface Params {
  userId: string
}

rest.get<never, Params>('/user', (req) => { // Type 'Params' is not assignable to type 'PathParams'. Index signature for type 'string' is missing in type 'Params'.(2322)
  req.params
})
```

I would like to resolve this, but for type safety reasons if you cannot approve this PR please close it 🙏🏻

### DIFF
--- a/src/rest.ts
+++ b/src/rest.ts
@@ -12,7 +12,7 @@ function createRestHandler<Method extends RESTMethods | RegExp>(
 ) {
   return <
     RequestBodyType extends DefaultBodyType = DefaultBodyType,
-    Params extends PathParams = PathParams,
+    Params extends PathParams<Params> = PathParams,
     ResponseBody extends DefaultBodyType = DefaultBodyType,
   >(
     path: Path,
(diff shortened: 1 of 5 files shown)


kettanaito (approved):
Awesome changes, @kotarella1110! 🎉 
I've added some polishing with the latest commit. Now the tests should pass without issues. Let's wait for that CI before we merge.

src/rest.ts
```diff
@@ -12,7 +12,7 @@ function createRestHandler<Method extends RESTMethods | RegExp>(
 ) {
   return <
     RequestBodyType extends DefaultBodyType = DefaultBodyType,
-    Params extends PathParams = PathParams,
+    Params extends PathParams<keyof Params> = PathParams,
```
kettanaito, line 15: This is an interesting pattern. 
So, if I read this correctly, by accepting key type as a generic, we are able to limit the known keys to the immediate keys of the current `Params` interface. This way the index signature is not ambiguous and instead is limited to the known keys.

kettanaito:
Welcome to contributors! 🎉

kettanaito:
## Released: v0.40.0 🎉

---

PR #1241, opened 2022-05-18
fix(deps): make TypeScript peerDep more forgiving

kentcdodds:
Without this I can't upgrade my project beyond 4.2.x and TS is now at 4.6.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "webpack-dev-server": "^3.11.2"
   },
   "peerDependencies": {
-    "typescript": "4.2.x"
+    "typescript": ">=4.2.x"
   },
   "peerDependenciesMeta": {
     "typescript": {


package.json
```diff
@@ -141,7 +141,7 @@
     "webpack-dev-server": "^3.11.2"
   },
   "peerDependencies": {
-    "typescript": "4.2.x"
+    "typescript": ">=4.2.x"
```
kettanaito, line 144: I understand the concern here but I don't think we can safely bump this. TypeScript is not versioned according to SemVer, so there may be (and often are, as practice shows) breaking changes between minor TypeScript releases (migrating from 4.2 to 4.3 may break stuff). We've had users reporting issues when using MSW on certain newer TypeScript versions because TS often likes to change how it handles small things between minor releases. 

We've added this exact peer dependency version to give the users a clear understanding what's the officially supported TS version (the one we're developing the library on). If we make this range more forgiving, we are claiming MSW will work on any version past 4.2 but there's actually no guarantee of that. 

The right approach here is to bump the TS version for the development environment and also in the officially supported version incrementally. This also poses a few challenges as migrating from 4.2 to 4.7, let's say, would lock all the library users who are using any TS version between those ranges. I'm frankly unsure how to properly tackle this, maybe you have some suggestions?

chrisguttandin: Hi, I hope it's okay to add my two cents here.

There is a tool to downlevel type definitions. It's called [downlevel-dts](https://github.com/sandersn/downlevel-dts). It's meant to support multiple TS versions at the same time. I once discovered it since users of my open source libraries had similar problems. I gave it a try and it worked as expected. But I never got around to actually use it to publish different type definitions.

kettanaito: Hey, @chrisguttandin! A great suggestion. The only concern I have is how does one publish type definitions oriented at different TS versions? Is there a way to declare something like:

```json
{
  "typings": {
    "4.2.x": "./typings/index.4.2.d.ts",
    "4.7": "./typings/index.d.ts"
  }
}
```

I think there's no such support in NPM. I wouldn't want to introduce downlevel as a regular dependency that'd run on users' machines—type distribution is the library's responsibility, they shouldn't be concerned with that. 

The issue itself comes from TS not being versioned per SemVer. If you think of the actual TS versions like 4.x and 7.x, there's nothing unusual that you can't use MSW with version 7 when it supports version 4. The only applicable strategy I see is to adopt newer versions and drop older versions. But it has a huge risk for our users, as not everybody stays on the bleeding edge of TS, and I myself is reluctant to upgrade TS often because of the breaking changes it constantly introduces that are challenging to debug.

kettanaito: Alternatively, you can ignore the peer dependency error (or suppress it) because MSW may not rely on anything that breaks between 4.2 and 4.7, as an example.

chrisguttandin: If I recall correctly discovering different type definitions based on an entry in the `package.json` file is a feature of TypeScript itself. It's just rarely used in the wild.

https://github.com/sandersn/downlevel-dts#usage

I also found this in some old release notes.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions

Just to be clear, I don't want to push you into that direction. I have no experience with it. However I would be more than happy to copy your solution in case you figure out how to integrate that into the build pipeline. :-)

Methuselah96: If you really care about advertising a correct TS version I would suggest just running your build against each minor version of TypeScript you plan to support. Redux Toolkit does this [here](https://github.com/reduxjs/redux-toolkit/blob/master/.github/workflows/tests.yml#L90-L99) which should provide a good starting point of how to implement this using GH Actions.

kettanaito: @Methuselah96, that's interesting, thanks for the suggestion!

The main reason we've added the peer dependency is to reduce the number of issues from the older TS versions. We do use certain TS features that were added in 4.2, so that was set as the supported version. 

But I like the idea of having multiple jobs to test against different TS versions to have observable regressions. I can try that.

Methuselah96: @kettanaito If your goal is just to reduce the number of issues from older TS versions, allowing `>=4.2.x` as suggested in this PR seems reasonable to me. While it's true that technically breaking changes happen between minor versions, from my experience it's rarely a practical issue, especially if you're building with the oldest version in the range. However, if you really care about correctness, testing against each minor version of TypeScript is the way to go. I'm not aware of any other package that restricts the TypeScript version it works with to a single minor version. Looking forward to this getting resolved.

Methuselah96: If you still decide to just allow a finite range of minor versions, then that means that users of your library have to wait for you to update the peer dependency range to include the new minor version, which is not ideal.

kettanaito: I'd combine the best of two solutions: have a permissive TS peer dependency range and also a matrix of CI jobs that actually verify the library is compatible with each supported version. That'd also allow for catching of type-related regressions. 

That's what I'm pursuing with #1242 and a few follow-up pull requests. I hope there aren't many things we should fix to become compatible with 4.4+ (that's the breaking point currently).

kettanaito:
## Released: v0.40.2 🎉

This has been released in v0.40.2!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.40.2)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.40.2)

---

PR #1257, opened 2022-05-27
fix: add "ctx.field()" for GraphQL responses

### LINKED ISSUE #1200: Support custom fields on a mocked GraphQL response
### Scope

Improves an existing behavior

### Compatibility

### Feature description

I'm using msw for mocking grapghql api calls in react testing library tests of my application.
I have a requirement to pass following object as the response context, to simulate an error response behavior
```
{
            "networkError": {
                "statusCode": "404"
            },
            "message": "Not Found",
            "metadata": {
                "errorCode": "Repository Not Accessible",
                "additionalData": "http://github.com/testUser/repo-to-be-deleted"
            }
        }
```
However when setting the response context it only allows me to set `data`, `headers`, `delay`, `extensions` and some other predefined fields for the context.
I need to be able to set custom values, for an example as per the above object I need to set `metadata` and `message` fields

Poivey:
- Closes #1200

As discussed in the issues, this PR adds a `field(fieldName, fieldValue)` method to graphQL response context.

`fieldName` is typed in a way to prevent using `errors`, `data`, `extensions` and `<empty string>` as a value. Is then any other value we should forbid ?
`fieldName` is also validated at runtime level.

Screenshot 2022-06-02 at 15 44 22

~@kentcdodds In the issue you mentioned [`invariant`](https://www.npmjs.com/package/outvariant) as a solution to make sure this field is valid, but I am not sure how it could be use here as we need to check for specific values, not if the value is undefined. Could you tell me how you thought this coule be used here ?~

## TODO

- [ ] Update the documentation, new page : `https://mswjs.io/docs/api/context/field`

### DIFF
--- a/src/context/field.test.ts
+++ b/src/context/field.test.ts
@@ -62,11 +62,22 @@ test('throws when trying to set non-serializable values', async () => {
   )
 })
 
-test.each(['', 'data', 'errors', 'exceptions'])(
-  'throws when passing "%s" as field value',
+test('throws when passing an empty string as field name', async () => {
+  await expect(response(field('' as string, 'value'))).rejects.toThrow(
+    `[MSW] Failed to set a custom field on a GraphQL response: field name cannot be empty.`,
+  )
+})
+test('throws when passing an empty string (when trimmed) as field name', async () => {
+  await expect(response(field('   ' as string, 'value'))).rejects.toThrow(
+    `[MSW] Failed to set a custom field on a GraphQL response: field name cannot be empty.`,
+  )
+})
+
+test.each(['data', 'errors', 'extensions'])(
+  'throws when passing "%s" as field name',
   async (fieldName) => {
     await expect(response(field(fieldName, 'value'))).rejects.toThrow(
-      'ctx.field() first argument must not be an element of ["","data","errors","exceptions"]',
+      `[MSW] Failed to set a custom "${fieldName}" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.${fieldName}()" instead?`,
     )
   },
 )
--- a/src/context/field.test.ts
+++ b/src/context/field.test.ts
@@ -67,17 +67,51 @@ test('throws when passing an empty string as field name', async () => {
     `[MSW] Failed to set a custom field on a GraphQL response: field name cannot be empty.`,
   )
 })
+
 test('throws when passing an empty string (when trimmed) as field name', async () => {
   await expect(response(field('   ' as string, 'value'))).rejects.toThrow(
     `[MSW] Failed to set a custom field on a GraphQL response: field name cannot be empty.`,
   )
 })
 
-test.each(['data', 'errors', 'extensions'])(
-  'throws when passing "%s" as field name',
-  async (fieldName) => {
-    await expect(response(field(fieldName, 'value'))).rejects.toThrow(
-      `[MSW] Failed to set a custom "${fieldName}" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.${fieldName}()" instead?`,
-    )
-  },
-)
+test('throws when using "data" as the field name', async () => {
+  await expect(
+    response(
+      field(
+        // @ts-expect-error Test runtime value.
+        'data',
+        'value',
+      ),
+    ),
+  ).rejects.toThrow(
+    '[MSW] Failed to set a custom "data" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.data()" instead?',
+  )
+})
+
+test('throws when using "errors" as the field name', async () => {
+  await expect(
+    response(
+      field(
+        // @ts-expect-error Test runtime value.
+        'errors',
+        'value',
+      ),
+    ),
+  ).rejects.toThrow(
+    '[MSW] Failed to set a custom "errors" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.errors()" instead?',
+  )
+})
+
+test('throws when using "extensions" as the field name', async () => {
+  await expect(
+    response(
+      field(
+        // @ts-expect-error Test runtime value.
+        'extensions',
+        'value',
+      ),
+    ),
+  ).rejects.toThrow(
+    '[MSW] Failed to set a custom "extensions" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.extensions()" instead?',
+  )
+})
--- a/src/context/field.test.ts
+++ b/src/context/field.test.ts
@@ -1,0 +1,72 @@
+/**
+ * @jest-environment jsdom
+ */
+import { field } from './field'
+import { response } from '../response'
+import { data } from './data'
+import { errors } from './errors'
+
+test('sets a given field value string on the response JSON body', async () => {
+  const result = await response(field('field', 'value'))
+
+  expect(result.headers.get('content-type')).toBe('application/json')
+  expect(result).toHaveProperty('body', JSON.stringify({ field: 'value' }))
+})
+
+test('sets a given field value object on the response JSON body', async () => {
+  const result = await response(
+    field('metadata', {
+      date: new Date('2022-05-27'),
+      comment: 'nice metadata',
+    }),
+  )
+
+  expect(result.headers.get('content-type')).toBe('application/json')
+  expect(result).toHaveProperty(
+    'body',
+    JSON.stringify({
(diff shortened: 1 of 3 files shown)


src/context/field.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Set a custom field on the GraphQL mocked response.
+ * @example res(ctx.fields('customField', value))
+ * @see {@link https://mswjs.io/docs/api/context/field}
+ */
+export const field = <FieldNameType extends string, FieldValueType>(
+  fieldName: FieldNameType extends ForbiddenFieldNames ? never : FieldNameType,
+  fieldValue: FieldValueType,
+): ResponseTransformer<string> => {
+  return (res) => {
+    assertFieldNameIsValid(fieldName)
```
kettanaito: What do you think about `validateFieldName`? Sounds like the same meaning but much shorter.

src/context/field.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const prevBody = jsonParse(res.body) || {}
+    const nextBody = mergeRight(prevBody, { [fieldName]: fieldValue })
+
+    return json(nextBody)(res as any) as any
+  }
+}
+
+function assertFieldNameIsValid(fieldName: string) {
+  if (forbiddenFieldNamesList.includes(fieldName as ForbiddenFieldNames)) {
```
kettanaito: Consider using `invariant` here:

```ts
import { devUtils } from '../utils/internals/devUtils'

invariant(
  fieldName !== '',
  devUtils.formatMessage('Failed to set a custom field on a GraphQL response: field name cannot be empty.')
)

invariant(
  fieldName !== 'data',
  devUtils.formatMessage('Failed to set a custom "%s" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.data()" instead?', fieldName)
)

invariant(
  fieldName !== 'errors',
  devUtils.formatMessage('Failed to set a custom "%s" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.errors()" instead?', fieldName)
)
```

I'd also drop the forbidden names array and tackle each scenario individually. It may seem repetitive but we gain more contextual error messages for the developer this way. 

What do you think?

src/handlers/GraphQLHandler.ts
```diff
@@ -48,6 +50,7 @@ export const graphqlContext: GraphQLContext<any> = {
   extensions,
   errors,
   cookie,
+  field: field,
```
kettanaito: You can use shorthand assignment here!

```suggestion
  field,
```

kettanaito (changes requested):
Hey, @Poivey! Thank you for preparing this, an outstanding work!

I've left a few suggestions, would love to know what you think.

kettanaito:
I've renamed the pull request to be a `fix:` because we're pre-1.0, and this does not need to bump the minor version—it's a new feature.

src/context/field.ts
Poivey: Sure I will rename the function !

src/context/field.ts
Poivey: Thank you for the examples, I wasn't sure about using invariant when checking for a specific value. I see the value of more explicit error message, I will update the PR.

kettanaito: I think it's really cool that, being a dev tool, MSW can be as explicit about error messages and error behaviors as it wants. Yeah, I'd generally recommend being detailed when it comes to error messages, and if we structure each individual name check separately, we can produce a separate message for each failed state.

src/handlers/GraphQLHandler.ts
Poivey: ✅

src/context/field.ts
Poivey: ✅

src/context/field.ts
Poivey: ✅
I made these changes.

Do you think we could change errors messages generation a bit to reduce duplication ? Or is it better to have the message in the code without formatting to make the search in the code easier ?

It could be something like this
```ts
  invariant(
    fieldName !== 'data' && fieldName !== 'errors' && fieldName !== 'extensions',
    devUtils.formatMessage(
      'Failed to set a custom "%s" field on a mocked GraphQL response: forbidden field name. Did you mean to call "ctx.%s()" instead?',
      fieldName, fieldName,
    ),
  )
```

What bothers me with the current solution is that we use string formatting to create the error message, but we could either:
* use it more to reduce duplication (like in the example above)
* or not use it a all and keep the same number of lines of code (we use formatting in cases where we already know the value of `fieldName`). Instead have string without formatting to make search in the code easier.

Do you think we should change it for one of these propositions or keep it as it is ?

Poivey:
There was a mistake in the forbidden values for field name of the initial commit. It is now corrected (I wrote `exceptions` instead of `extensions` 🤦)

src/context/field.ts
kettanaito: I don't think this is worth investing our time at the moment. Repetition is not inherently bad, and I'd much more prefer to see 

```ts
invariant(a, messageA)
invariant(b, messageB)
invariant(c, messageC)
```

Even if it means all messages are almost the same. Yes, you're right, it improves discoverability of the code but also allows for custom cases without having to extend the abstraction. If we decide to give a better message for the `b` case then we just do that. We know `a` and `c` are not affected.

src/context/field.ts
```diff
@@ -0,0 +1,57 @@
...
+    validateFieldName(fieldName)
+
+    const prevBody = jsonParse(res.body) || {}
+    const nextBody = mergeRight(prevBody, { [fieldName]: fieldValue })
+
+    return json(nextBody)(res as any) as any
+  }
+}
+
+function validateFieldName(fieldName: string) {
```
kettanaito, line 29: We should export this function and unit test it. 

Also, let's include an explicit return type:

```suggestion
function validateFieldName(fieldName: string): void {
```

src/context/field.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    `[MSW] Failed to set a custom field on a GraphQL response: field name cannot be empty.`,
+  )
+})
+test('throws when passing an empty string (when trimmed) as field name', async () => {
+  await expect(response(field('   ' as string, 'value'))).rejects.toThrow(
+    `[MSW] Failed to set a custom field on a GraphQL response: field name cannot be empty.`,
+  )
+})
+
+test.each(['data', 'errors', 'extensions'])(
```
kettanaito: This is absolutely biased but I don't like any logic around test execution. It always makes tests smarter than they should be. If a test is repetitive, then I recommend tackling that on the test setup level but not test orchestration/definition level. 

```ts
it('throws an error when using forbidden field names', async () => {
  const tryWithFieldName = (fieldName: string) => {
    return response(ctx.field(fieldName, 'value'))
  }

  expect(() => tryWithFieldName('data')).toThrow('...')
  expect(() => tryWithFieldName('extensions')).toThrow('...')
})
```

- A single test case because input validation is a single piece of logic (even though we validate against multiple forbidden names—it's fine; each name does not equal new behavior). 
- Clear assertions.
- Repetition is abstracted into `tryWithFieldName`, which is both scoped and readable.

src/context/field.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  expect(result.headers.get('content-type')).toEqual('application/json')
+  expect(result).toHaveProperty(
+    'body',
+    JSON.stringify({
+      field2: 123,
+      field: { errorCode: 'value' },
+      errors: [
+        {
+          message: 'exceeds the limit of awesomeness',
```
kettanaito, line 49: ❤️

src/context/field.ts
Poivey: Ok ! So we keep it as it is ?

kettanaito (approved):

kettanaito:
I've pushed a few polish items, the feature should be ready to go! 🎉

kettanaito:
@Poivey, I'm so thankful for your contributions. You're the best 🥳

Poivey:
> @Poivey, I'm so thankful for your contributions. You're the best 🥳

Happy to contribute ! I planned to do the last changes later this week but it's great if you found the time to update it yourself :)

kettanaito:
## Released: v0.42.1 🎉

This has been released in v0.42.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.42.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.42.1)

---

PR #1288, opened 2022-06-11
feat: send mocked response body as ReadableStream to the worker

### LINKED ISSUE #581: ReadableStream object could not be cloned when trying to use stream as a mocked response body

### LINKED ISSUE #929: Binary body of a request gets corrupted in the handler

### LINKED ISSUE #966: No response when using `jest.useFakeTimers()` with axios and msw
## Describe the bug
My app uses Svelte with Axios to fetch data from a REST API endpoint. I noticed that Axios didn't sometimes wait long enough for the request and it timed out. Now I increased the Axios' timeout to 12000 ms and wanted to verify it with tests that it actually waits for long enough for the request.

So I mocked the API with msw for the basic test that it actually works. No problems there. And when I add the delay to my msw mocked API, it works too, I just need to increase the test's timeout value so it actually waits long enough.

Which brings me to my point: I don't really want to wait for that long to verify that the Axios' 12 second timeout actually works. So here's where I wanted to use the jest's `useFaketimers` method but I just can't get it to work. Even when trying to advance the mocked timer with every command I found in jest, the response never comes.

## Environment

- `msw: 0.35.0`
- `nodejs: 14.17.3`
- `npm: 7.24.0`
<details>
  <summary>packages.json deps</summary>
  
  ```javascript
  "devDependencies": {
    "@jest/types": "^27.2.5",
    "@testing-library/dom": "^8.10.1",
    "@testing-library/jest-dom": "^5.14.1",
    "@testing-library/svelte": "^3.0.3",
    "@tsconfig/svelte": "^2.0.1",
    "@types/jest": "^27.0.2",
    "@types/node": "^16.11.6",
    "@types/testing-library__jest-dom": "^5.14.1",
    "@typescript-eslint/eslint-plugin": "^4.33.0",
    "@typescript-eslint/parser": "^4.33.0",
    "eslint": "^7.32.0",
    "eslint-import-resolver-typescript": "^2.5.0",
    "eslint-plugin-import": "^2.25.2",
    "eslint-plugin-svelte3": "^3.2.1",
    "eslint-plugin-testing-library": "^4.12.4",
    "jest": "^27.3.1",
    "msw": "^0.35.0",
    "svelte": "^3.44.0",
    "svelte-check": "^2.2.7",
    "svelte-jester": "^2.1.5",
    "svelte-preprocess": "^4.9.8",
    "ts-jest": "^27.0.7",
    "ts-node": "^10.4.0",
    "tslib": "^2.3.1",
    "typescript": "^4.4.4"
  },
  "dependencies": {
    "axios": "^0.24.0"
  }
  ```
</details>

Not really browser related but sure:

* Chrome 95.0.4638.69 (Official Build) (64-bit)
* Firefox 94.0.1 (64-bittinen)

## To Reproduce
1. use axios to fetch data from a REST API endpoint
2. mock the API with msw
3. add a delay to the msw mocked API
4. use `jest.useFakeTimers()` to mock timers
5. use `jest.advanceTimersByTime()` to pass the time forward until there should be an answer from the delayed API request
6. the test times out with an error like `Unable to find an element with the text` or similar

## Expected behavior
jest successfully mocks timers and skips the time so the test doesn't timeout.

## Reproduction repo
https://github.com/asode/msw-usefaketimers

Failing test code:
https://github.com/asode/msw-usefaketimers/blob/main/src/App.test.ts#L43

Failing test run:
https://github.com/asode/msw-usefaketimers/runs/4106445180#step:4:27

kettanaito:
- Fixes #966 (does not wrap all responses in `setTimeout` anymore). For Node, awaits timeout conditionally only if `response.delay` is set. 
- Closes #974 
- Fixes #581
- Fixes #929
- Related to #931
- Answers #1272

## Changes

- Transfers mocked response data as a `ReadableStream` between the client and the worker. This way we support both text and non-text data like `Uint8Array` to be used as a mocked response body. 

## Roadmap

- [x] Add an integration test for `ctx.delay()` in Node. Removing `setTimeout` in `handleRequest` breaks response delays for Node but no tests catch that. 
- [ ] Add an integration test for fake timers that ensures that no timer advance is needed if the mocked response uses no delay.


src/utils/handleRequest.ts
```diff
@@ -125,27 +125,24 @@ Expected response resolver to return a mocked response Object, but got %s. The o
...
-      emitter.emit('request:end', request)
+  handleRequestOptions?.onMockedResponseSent?.(
+    transformedResponse,
+    requiredLookupResult,
+  )
 
-      resolve(transformedResponse as ResponseType)
-    }, response.delay ?? 0)
-  })
+  emitter.emit('request:end', request)
```
kettanaito: Delay is performed in the worker in order to affect the response timings.

src/mockServiceWorker.js
```diff
@@ -316,23 +320,48 @@ function sendToClient(client, message) {
...
+
+        case 'MOCK_RESPONSE_END': {
+          streamCtrl.close()
+          operationChannel.close()
+        }
+      }
+    }
+
+    await sleep(mockResponse.delay)
+    return resolve(new Response(stream, mockResponse))
```
kettanaito, line 365: ## Gotcha

A _really_ important nuance here is to respond with a stream response _straight away_. This way the client will receive the response immediately and will be able to read the stream as the operation channel keeps pushing the chunks from the response resolver.

src/node/createSetupServer.ts
```diff
@@ -67,24 +67,28 @@ export function createSetupServer(
...
+            body: response.body,
+            delay: response.delay,
+          }
         },
-      )
+      })
 
       if (response) {
+        await new Promise<void>((resolve) => {
+          if (!response.delay) {
```
kettanaito: Since I've removed a hard-coded `setTimeout` around responses in `handleRequest`, the Node part now needs to delay mocked responses manually. The worker has been always delaying the responses in the worker thread so that the mocked delay counts for server computing time.

kettanaito:
## No `ctx.stream` support possible (yet)

There's no `ReadableStream` in Node.js prior to version 17. We can't support using readable streams as mocked response bodies until version 17 is the minimal version of Node we support. I've tried writing a polyfill for a readable stream but it also depends on the writable stream for `pipeTo`/`pipeThrough` methods; the writable stream needs to know about `AbortController`, and here you are polyfilling half of the browser standard API. 

We can still use a readable stream _internally_ to stream mocked responses from the client to the worker. In Node, we'd have to convert mocked responses to `Uint8Array` (to preserve binary/buffers in mocks) and feed it to the respective patched modules.

kettanaito:
For Node, we can convert mocked response bodies to `Buffer` or similar, achieving the unified response body processing as with the `ReadableStream` in the browser.

kettanaito:
## Released: v0.43.0 🎉

This has been released in v0.43.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.43.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.43.0)

---

PR #1321, opened 2022-07-12
feat: transfer response body from worker as arraybuffer (zero copy)

95th:
## Changes

- Reads request/response bodies in the worker as `ArrayBuffer`. This unifies the body handling as we use `IsomorphicRequest` and `MockedRequest` that extends it, and they operate with bodies as `ArrayBuffer`.

### DIFF
--- a/test/rest-api/custom-request-handler.mocks.ts
+++ b/test/rest-api/custom-request-handler.mocks.ts
@@ -10,7 +10,10 @@ import {
 
 // This is an example of a custom request handler that matches requests
 // based on the presence of a given header.
-class HeaderHandler extends RequestHandler<{ headerName: string }> {
+class HeaderHandler extends RequestHandler<{
+  header: string
+  headerName: string
+}> {
   constructor(headerName: string, resolver: ResponseResolver) {
     super({
       info: {
@@ -36,7 +39,7 @@ interface CustomContext {
   halJson: (body: Record<string, any>) => ResponseTransformer
 }
 
-class UrlHandler extends RequestHandler<{ url: string }> {
+class UrlHandler extends RequestHandler<{ header: string; url: string }> {
   constructor(
     url: string,
     resolver: ResponseResolver<MockedRequest, CustomContext>,


src/mockServiceWorker.js
```diff
@@ -254,11 +257,12 @@ async function getResponse(event, client, requestId) {
...
       referrer: request.referrer,
       referrerPolicy: request.referrerPolicy,
-      body: await request.text(),
+      body,
       bodyUsed: request.bodyUsed,
       keepalive: request.keepalive,
     },
-  })
+  }
+  const clientMessage = await sendToClient(client, message, body)
```
kettanaito: Why should we pass body as another argument if it's in `message.body` already?

src/mockServiceWorker.js
```diff
@@ -316,7 +320,11 @@ function sendToClient(client, message) {
       resolve(event.data)
     }
 
-    client.postMessage(message, [channel.port2])
+    const transfer = [channel.port2]
+    if (body) {
+      transfer.push(body)
```
kettanaito: Can array buffer be transmitted through a message channel without pushing it to the transport?

src/utils/request/pruneGetRequestBody.ts
```diff
@@ -1,21 +0,0 @@
-import { ServiceWorkerIncomingRequest } from '../../setupWorker/glossary'
-import { isStringEqual } from '../internal/isStringEqual'
-
-type Input = Pick<ServiceWorkerIncomingRequest, 'method' | 'body'>
-
-/**
- * Ensures that an empty GET request body is always represented as `undefined`.
- */
-export function pruneGetRequestBody(
```
kettanaito: I think we still perform this logic in MockedRequest. May I please ask you to double-check?

kettanaito (commented):
Like these changes. I've left a couple of questions to make sure I understand them fully.

src/mockServiceWorker.js
```diff
@@ -147,20 +147,22 @@ async function handleRequest(event, requestId) {
   if (client && activeClientIds.has(client.id)) {
     ;(async function () {
       const clonedResponse = response.clone()
-      sendToClient(client, {
+      const body =
+        clonedResponse.body === null ? null : await clonedResponse.arrayBuffer()
```
kettanaito: Do we have to make this `null` check anymore? I think it was needed because we didn't want to send an empty string if the request didn't have a body, so we could distinguish between explicit "no body" scenario and explicit "empty string" body scenarios on the client-side.

Now, I think empty request body will be read as `ArrayBuffer(0)`, which is totally what we expect on the client, as it will be the developer's responsibility to read that body into something (like `req.text()`, getting an empty string on such an empty buffer). 

The only downside of this is that you won't be able to read explicit `null` with the deprecated `req.body`, as for the library `''` and `null` will both be received as empty buffers, and we wouldn't be able to distinguish when `req.body` reference should return an empty string or a `null`. 

What do you think about this use case?

src/mockServiceWorker.js
95th: It's bit easy to check if body is there since `payload` may not contain `body`, or have it `boolean`...

src/mockServiceWorker.js
95th: no it has to be transferred explicitly.

src/utils/request/pruneGetRequestBody.ts
95th: We removed it in the Request PR

src/mockServiceWorker.js
95th: Response constructor throws error if we provide body with 204.

src/mockServiceWorker.js
kettanaito: Could you please share some context as to why?

src/mockServiceWorker.js
kettanaito: But the `body` property is not set conditionally on `payload`. So it's always there. Then, we can check its value as `typeof payload.body !== 'undefined'` (or `== null` since we're coercing it to null if it's empty) in the `sendToClient` to achieve the same result, can't we?

src/mockServiceWorker.js
95th: it is converted to an opaque object if not transferred.

src/mockServiceWorker.js
95th: We can. Let me try that

kettanaito: Go ahead.

It's be great not to duplicate things, as two arguments imply two different values being passed. Since `body` is the same as `message.body` and really shouldn't differ intention-wise, I think an extra argument creates a confusion.

95th: Done

src/mockServiceWorker.js
```diff
@@ -321,8 +319,8 @@ function sendToClient(client, message, body) {
     }
 
     const transfer = [channel.port2]
-    if (body) {
-      transfer.push(body)
+    if (typeof message.payload !== 'undefined' && message.payload.body) {
```
kettanaito: What do you think if we flip the responsibility around? Instead of `sendToClient` checking a potential body input, we'd expose a callback that allows the consumer to push additional data to the transport?

```js
const clientMessage = await sendToClient(
  client,
  requestObj,
  (transport) => transport.push(request.body)
)

// ...

function sendToClient(client, message, goodNameHere) {
  if (goodNameHere) {
    goodNameHere(transport)
  }
}
```

The only question now is whether we can call `goodNameHere` at a certain point of the transport being constructed. I don't see why we wouldn't be able to, but I feel you have better understanding of whether this suggestion makes sense.

95th: I think it makes it more or less the same as my original approach.

kettanaito: Then I'd move the responsibility of pushing to the `transport` to the consumer of `sentToClient`. Imo, it makes the function more predictable, as it doesn't try to leverage the complexity of input, delegating that complexity to the consumer (closer to the context).

95th: SendToClient is the last one on this side of the channel. I think it make the most sense at callers to provide transferrable content. We can make it similar to post message - an array of transferrable content. Others will just pass empty array

src/mockServiceWorker.js
```diff
@@ -147,20 +147,22 @@ async function handleRequest(event, requestId) {
   if (client && activeClientIds.has(client.id)) {
     ;(async function () {
       const clonedResponse = response.clone()
-      sendToClient(client, {
+      const body =
+        clonedResponse.body == null ? null : await clonedResponse.arrayBuffer()
```
kettanaito: Something we may consider for future improvement is reading request/response body as streams. This way we don't have to wait for the entire body to be read in order to be sent from the worker, we can start sending as soon as we receive the first byte. 

This will require some larger refactor but may be worth it.

src/utils/request/parseWorkerRequest.ts
```diff
@@ -13,9 +12,5 @@ export function parseWorkerRequest(
   const url = new URL(rawRequest.url)
   const headers = new Headers(rawRequest.headers)
 
-  return new MockedRequest(url, {
-    ...rawRequest,
-    body: encodeBuffer(rawRequest.body || ''),
-    headers,
-  })
+  return new MockedRequest(url, { ...rawRequest, headers })
```
kettanaito: _This_ is great.

PR #1328, opened 2022-07-14
fix: prevent "MockedRequest" constructor loop

### LINKED ISSUE #1326: msw fails to run on 0.44.0
### Prerequisites

### Environment check

### Browsers

Chromium (Chrome, Brave, etc.), Firefox, Safari

### Reproduction repository

https://github.com/netcoding87/msw-safari

### Reproduction steps

- `npm install` (will install `msw` in v. `0.43.0`) 
- `npm run start`
=> Worker is enabled and working
---

- update `msw` to latest version (`0.44.0`)
- `npm run start`
=> Worker is enabled but throwing errors

### Current behavior

image

### Expected behavior

Working `msw` 😄

95th:
- Fixes https://github.com/mswjs/msw/issues/1326

### DIFF
--- a/src/setupWorker/start/createFallbackRequestListener.ts
+++ b/src/setupWorker/start/createFallbackRequestListener.ts
@@ -25,7 +25,10 @@ export function createFallbackRequestListener(
   })
 
   interceptor.on('request', async (request) => {
-    const mockedRequest = new MockedRequest(request)
+    const mockedRequest = new MockedRequest(request.url, {
+      ...request,
+      body: await request.arrayBuffer(),
+    })
 
     const response = await handleRequest<SerializedResponse>(
       mockedRequest,
--- a/src/utils/request/MockedRequest.ts
+++ b/src/utils/request/MockedRequest.ts
@@ -84,14 +84,8 @@ export class MockedRequest<
   public readonly referrer: string
   public readonly referrerPolicy: RequestReferrerPolicy
 
-  constructor(request: IsomorphicRequest)
-  constructor(url: URL, init?: MockedRequestInit)
-  constructor(input: URL | IsomorphicRequest, init: MockedRequestInit = {}) {
-    if (input instanceof IsomorphicRequest) {
-      super(input)
-    } else {
-      super(input, init)
-    }
+  constructor(url: URL, init: MockedRequestInit = {}) {
+    super(url, init)
     if (init.id) {
       this.id = init.id
     }
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -66,7 +66,10 @@ export function createSetupServer(
     })
 
     interceptor.on('request', async function setupServerListener(request) {
-      const mockedRequest = new MockedRequest(request)
+      const mockedRequest = new MockedRequest(request.url, {
+        ...request,
+        body: await request.arrayBuffer(),
+      })
 
       const response = await handleRequest<
         MockedInterceptedResponse & { delay?: number }


src/utils/request/MockedRequest.ts
```diff
@@ -84,14 +84,8 @@ export class MockedRequest<
   public readonly referrer: string
   public readonly referrerPolicy: RequestReferrerPolicy
 
-  constructor(request: IsomorphicRequest)
-  constructor(url: URL, init?: MockedRequestInit)
-  constructor(input: URL | IsomorphicRequest, init: MockedRequestInit = {}) {
-    if (input instanceof IsomorphicRequest) {
-      super(input)
-    } else {
-      super(input, init)
```
95th, line 93: This call goes to IR constructor in tests but MR constructor in repro.

src/setupWorker/start/createFallbackRequestListener.ts
```diff
@@ -25,7 +25,10 @@ export function createFallbackRequestListener(
   })
 
   interceptor.on('request', async (request) => {
-    const mockedRequest = new MockedRequest(request)
+    const mockedRequest = new MockedRequest(request.url, {
+      ...request,
+      body: await request.arrayBuffer(),
```
kettanaito, line 30: Why can't we pass `request` as `init` anymore?

`request` is IR, so it should be a valid init to another IR instance, I think. 

I wonder if us reading the request body here doesn't lock it to be read by the user. I suppose if tests pass then we are safe.

95th: because the field is called _body now in IR. it will return empty body as if it is missing

95th: > I wonder if us reading the request body here doesn't lock it to be read by the user. 

It shouldn't. We're constructing a new MR (hence IR) here which will have bodyUsed false

kettanaito: Oh, you're right. The consumer never interacts with the original intercepted IR, we always expose a newly constructed MR, which is effectively a cloned request instance.

kettanaito (commented):
This looks great. Thank you once more for making this library better, @95th!

The only thing remaining is to depict this issue in a test case.

kettanaito:
I can't reproduce this locally at all. I agree that this is a transpilation issue, as the `super()` call transpiled by TypeScript turns into something like this:

```js
  constructor(input) {
    var _supercall = function () {
      return _supercall(...arguments);
    };

    let init = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};

    var __super = function () {
      _supercall(...arguments);
    };

    if (input instanceof import_interceptors.IsomorphicRequest) {
      __super(input);
    } else {
      __super(input, init);
    }

    if (init.id) {
      this.id = init.id;
    }

    this.cache = init.cache || "default";
    this.destination = init.destination || "";
    this.integrity = init.integrity || "";
    this.keepalive = init.keepalive || false;
    this.mode = init.mode || "cors";
    this.priority = init.priority || "auto";
    this.redirect = init.redirect || "follow";
    this.referrer = init.referrer || "";
    this.referrerPolicy = init.referrerPolicy || "no-referrer";
    this.cookies = init.cookies || this.getCookies();
  }
```

How is this ever a valid transpilation result?

```js
var _supercall = function () {
  return _supercall(...arguments);
};
```

kettanaito (approved):

kettanaito:
## Released: v0.44.1 🎉

This has been released in v0.44.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.44.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.44.1)

---

PR #1332, opened 2022-07-16
feat(worker): read request/response body from as ArrayBuffer

kettanaito:
- Clone of 1321

## Changes

- Reads request/response bodies in the worker as `ArrayBuffer`. This unifies the body handling as we use `IsomorphicRequest` and `MockedRequest` that extends it, and they operate with bodies as `ArrayBuffer`.


src/setupWorker/glossary.ts
```diff
@@ -38,9 +38,9 @@ export interface ServiceWorkerIncomingRequest extends RequestWithoutMethods {
   id: string
 
   /**
-   * Text response body.
+   * Response body.
    */
-  body?: string
+  body?: ArrayBuffer
```
kettanaito, line 43: Can this ever be undefined? 🤔

src/mockServiceWorker.js
```diff
@@ -316,7 +321,7 @@ function sendToClient(client, message) {
       resolve(event.data)
     }
 
-    client.postMessage(message, [channel.port2])
+    client.postMessage(message, [channel.port2].concat(transfer || []))
```
kettanaito, line 326: @95th, one thing I've changed is added this `.concat()` instead of spreading. I know spreading is quite supported right now but I wanted to make sure we ship worker code that most of the browsers can understand.

src/setupWorker/glossary.ts
95th: We pass null intentionally sometimes .. I think this for handling that case.

kettanaito: Then it should be 
```suggestion
  body: ArrayBuffer | null
```

In TypeScript null and undefined are treated differently. Could you please make this change to the type definition?

kettanaito:
This may be partially/fully obsolete with #1404.

kettanaito:
Closing in favor of #1436

kettanaito:
## Released: v2.0.0 🎉

This has been released in v2.0.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.0)

---

PR #1353, opened 2022-07-26
Fix/spelling errors and syntax warnings

kmfb:
Fix/spelling errors and syntax warnings

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -200,7 +200,7 @@ async function getResponse(event, client, requestId) {
 
   function passthrough() {
     // Clone the request because it might've been already used
-    // (i.e. its body has been read and sent to the cilent).
+    // (i.e. its body has been read and sent to the client).
     const headers = Object.fromEntries(clonedRequest.headers.entries())
 
     // Remove MSW-specific request headers so the bypassed requests
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -11,8 +11,8 @@
 const INTEGRITY_CHECKSUM = '<INTEGRITY_CHECKSUM>'
 const activeClientIds = new Set()
 
-self.addEventListener('install', function () {
-  self.skipWaiting()
+self.addEventListener('install', async function () {
+  await self.skipWaiting()
 })
 
 self.addEventListener('activate', function (event) {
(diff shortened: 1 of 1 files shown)


src/mockServiceWorker.js
```diff
@@ -38,14 +38,14 @@ self.addEventListener('message', async function (event) {
 
   switch (event.data) {
     case 'KEEPALIVE_REQUEST': {
-      sendToClient(client, {
+      await sendToClient(client, {
```
kettanaito, line 41: Do you find any particular benefit of awaiting this? The "message" event is not tied to the request handling pipeline in any way, so what we introduce here are all side effects. We may await them if we depend on async actions execution order but in this case we don't.

src/mockServiceWorker.js
```diff
@@ -145,9 +145,9 @@ async function handleRequest(event, requestId) {
   // Ensure MSW is active and ready to handle the message, otherwise
   // this message will pend indefinitely.
   if (client && activeClientIds.has(client.id)) {
-    ;(async function () {
+    await (async function () {
       const clonedResponse = response.clone()
-      sendToClient(client, {
+      await sendToClient(client, {
```
kettanaito, line 150: This is another example of a side effect which we don't need to await. In fact, I think it'd make request resolution slower if we do.

This line clones and sends back a response to the main thread. At this point, the request is finished, and we should really "release" it as soon as possible so the client reacts to its response. Anything that happens past this point is a side effect.

src/mockServiceWorker.js
```diff
@@ -11,16 +11,16 @@
 const INTEGRITY_CHECKSUM = '<INTEGRITY_CHECKSUM>'
 const activeClientIds = new Set()
 
-self.addEventListener('install', function () {
-  self.skipWaiting()
+self.addEventListener('install', async function () {
+  await self.skipWaiting()
```
kettanaito, line 15: Take a look at the [.skipWaiting()](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/skipWaiting) page on MDN. Although some service worker scope methods return promises, those are implicitly tied to the worker internals and you don't have to await them. In fact, some event listeners, like "fetch", must be entirely synchronous otherwise they will throw. This means they can still execute async things sequentially but rid you of the need to await them explicitly.

kettanaito (changes requested):
Hi! Thanks for opening this. 

At this point, I don't think these changed would actually change anything, if not make things worse. 

Can you tell me more about why you think they are necessary? Is there some issue they help you fix?

kettanaito:
I'm going to close this because I don't find this improving anything per se. Feel free to follow up with more argumentation in this pull request. Thanks.

PR #1356, opened 2022-08-01
feat: make GraphQL a peer dependency, support GraphQL v15.0

JohnDaly:
The `graphql` package throws errors if it's not treated as a [singleton dependency](https://github.com/graphql/graphql-js/blob/94234c8ce0a33cd8d884d0fced3d7d107db8f8aa/src/jsutils/instanceOf.js#L32). It's recommended for libraries to install it as a peer dependency (see: https://github.com/graphql/graphql-js/issues/594#issuecomment-1189850052).

To allow MSW to work with projects that are using `graphql` v15, I've made a few small changes to the code to avoid using `OperationTypeNode` as an enum. The updated code is compatible with both v15 and v16 of the `graphql` package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "chalk": "4.1.1",
     "chokidar": "^3.4.2",
     "cookie": "^0.4.2",
-    "graphql": "^16.3.0",
     "headers-polyfill": "^3.0.4",
     "inquirer": "^8.2.0",
     "is-node-process": "^1.0.1",
@@ -116,6 +115,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "fs-extra": "^10.0.0",
     "fs-teardown": "^0.3.0",
+    "graphql": "^16.3.0",
     "jest": "26",
     "json-bigint": "^1.0.0",
     "lint-staged": "^11.0.1",
@@ -134,6 +134,7 @@
     "webpack-dev-server": "^3.11.2"
   },
   "peerDependencies": {
+    "graphql": "^15.0.0 || ^16.0.0",
     "typescript": ">= 4.2.x <= 4.7.x"
   },
   "peerDependenciesMeta": {
--- a/package.json
+++ b/package.json
@@ -138,6 +138,9 @@
     "typescript": ">= 4.2.x <= 4.7.x"
   },
   "peerDependenciesMeta": {
+    "graphql": {
+      "optional": true
+    },
     "typescript": {
       "optional": true
     }
--- a/src/context/errors.ts
+++ b/src/context/errors.ts
@@ -1,4 +1,4 @@
-import { GraphQLError } from 'graphql'
+import type { GraphQLError } from 'graphql'
 import { ResponseTransformer } from '../response'
 import { jsonParse } from '../utils/internal/jsonParse'
 import { mergeRight } from '../utils/internal/mergeRight'
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -83,7 +83,7 @@ const standardGraphQLHandlers = {
    * })
    * @see {@link https://mswjs.io/docs/api/graphql/query `graphql.query()`}
    */
-  query: createScopedGraphQLHandler(OperationTypeNode.QUERY, '*'),
+  query: createScopedGraphQLHandler('query' as OperationTypeNode, '*'),
 
   /**
    * Captures a GraphQL mutation by a given name.
@@ -93,14 +93,14 @@ const standardGraphQLHandlers = {
    * })
    * @see {@link https://mswjs.io/docs/api/graphql/mutation `graphql.mutation()`}
    */
-  mutation: createScopedGraphQLHandler(OperationTypeNode.MUTATION, '*'),
+  mutation: createScopedGraphQLHandler('mutation' as OperationTypeNode, '*'),
 }
 
 function createGraphQLLink(url: Path): typeof standardGraphQLHandlers {
   return {
     operation: createGraphQLOperationHandler(url),
-    query: createScopedGraphQLHandler(OperationTypeNode.QUERY, url),
-    mutation: createScopedGraphQLHandler(OperationTypeNode.MUTATION, url),
+    query: createScopedGraphQLHandler('query' as OperationTypeNode, url),
+    mutation: createScopedGraphQLHandler('mutation' as OperationTypeNode, url),
   }
 }
 
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -1,4 +1,4 @@
-import { DocumentNode, OperationTypeNode } from 'graphql'
+import type { DocumentNode, OperationTypeNode } from 'graphql'
 import { ResponseResolver } from './handlers/RequestHandler'
 import {
   GraphQLHandler,
--- a/src/handlers/GraphQLHandler.ts
+++ b/src/handlers/GraphQLHandler.ts
@@ -1,4 +1,4 @@
-import { DocumentNode, OperationTypeNode } from 'graphql'
+import type { DocumentNode, OperationTypeNode } from 'graphql'
 import { SerializedResponse } from '../setupWorker/glossary'
 import { data } from '../context/data'
 import { extensions } from '../context/extensions'
--- a/src/utils/internal/parseGraphQLRequest.ts
+++ b/src/utils/internal/parseGraphQLRequest.ts
@@ -1,8 +1,7 @@
-import {
+import type {
   DocumentNode,
   OperationDefinitionNode,
   OperationTypeNode,
-  parse,
 } from 'graphql'
 import { GraphQLVariables } from '../../handlers/GraphQLHandler'
 import { getPublicUrlFromRequest } from '../request/getPublicUrlFromRequest'
@@ -41,6 +40,9 @@ export function parseDocumentNode(node: DocumentNode): ParsedGraphQLQuery {
 
 function parseQuery(query: string): ParsedGraphQLQuery | Error {
   try {
+    // Since 'graphql' is an optional peer dependency,
+    // we'll attempt to import it dynamically
+    const { parse } = require('graphql')
     const ast = parse(query)
     return parseDocumentNode(ast)
   } catch (error) {


package.json
```diff
@@ -134,6 +134,7 @@
     "webpack-dev-server": "^3.11.2"
   },
   "peerDependencies": {
+    "graphql": "^15.0.0 || ^16.0.0",
```
mattcosta7, line 137: can graphql be an optional peer?  I'm curious if it's required for users who only define `rest` methods?

JohnDaly: That would be ideal, however it would require additional changes. Currently, the `graphql` package is included in the main entrypoint, so it's a required peer dependency of `msw`.

JohnDaly:
> This pull request is automatically built and testable in [CodeSandbox](https://codesandbox.io).
> 
> To see build info of the built libraries, click [here](https://ci.codesandbox.io/status/mswjs/msw/pr/1356/builds/277254) or the icon next to each commit SHA.
> 
> Latest deployment of this branch, based on commit [ebe786c](https://github.com/mswjs/msw/commit/ebe786c6317bfa6ff03e91609ee55cd0ecb697c5):
> 
> Sandbox	Source
> [MSW React](https://codesandbox.io/s/msw-react-dzkgyt)	Configuration

The Sandbox template will need to be updated to include `graphql` in its list of dependencies

package.json
```diff
@@ -77,7 +77,6 @@
     "chalk": "4.1.1",
     "chokidar": "^3.4.2",
     "cookie": "^0.4.2",
-    "graphql": "^16.3.0",
```
kettanaito, line 80: We should still keep `graphql` as a dev dependency so we could develop the library. I suppose newer versions of NPM install peer dependencies out of the box, but let's keep it explicit.

package.json
kettanaito: I believe `graphql` should come in as an optional dependency. Otherwise, we'd be demanding our users who are not using GraphQL in their apps to install `graphql` alongside MSW. 

By making `graphql` an optional peer dependency (similar to how we list `typescript`), we are not making any breaking changes to our consumers, actually:

- Devs who are using GraphQL in their apps, already have GraphQL installed—no change here for them;
- Devs who are not using GraphQL are not affected at all.

kettanaito, line 137: But one concern I do have is to double-check MSW's tree shaking. 

1. List this as an optional peer dependency.
2. Create an empty project and install MSW there.
3. Don't use `graphql`. Would your app compile?

JohnDaly: @kettanaito do you have an idea of how we might be able to make `graphql` optional? 

As far as I can tell, the only code being imported from `graphql` that is needed at runtime by `msw` is the `parse` function. It is used [here](https://github.com/mswjs/msw/blob/fc7f00c96e79ad0be83cbaf7c82ceff23f0c4355/src/utils/internal/parseGraphQLRequest.ts#L44).

My thought was to do something like this, to treat the package as optional:

```ts
// Since 'graphql' is an optional peer dependency,
// we'll attempt to import it dynamically
let parse: typeof import('graphql').parse
try {
  parse = require('graphql').parse
} catch (e) {}
```

What I'm unsure of is if the ESM output of `msw` will work with this approach.

The other option I considered is to move the code that depends on `graphql` to a different entrypoint (e.g. `msw/graphql`). This would be a breaking change though.

What do you think?

kettanaito: If we have proper code splitting in place, if you're not using `graphql` from `msw`, your app shouldn't get the line of code where we do `import { parse } from 'graphql'`. This is something I'd kindly ask you to test out (steps described in some other comment on this pull request). 

And in terms of package.json, I believe you can take a look how we specify `"typescript"`, as it's an optional peer dependency we already have. 

> Side note: I think NPM also force-installs optional peer dependencies, which is sort of okay. As long as we don't break users' apps by forcing them to manually install a dependency they don't otherwise need.

kettanaito:
I will update the sandbox once this is merged.

package.json
JohnDaly: I'm not sure we can solve this with code splitting, without introducing a breaking change.

The problem is that the `graphql` import occurs within the main entrypoint of `msw`. When a consumer imports `msw`, and they don't have `graphql` in their `node_modules`, the `msw` entrypoint will fail to load because there will be a `MODULE_NOT_FOUND` error when trying to resolve `graphql`.

To code split, we would need to move the `graphql` import to a different entrypoint, which would be a breaking change.

I've opted to do the following in https://github.com/mswjs/msw/pull/1356/commits/7a8482ab8a1745533d9b845c0a0e6713eb79c2d1:
1. Mark `graphql` as an optional peer dependency
2. Dynamically load `graphql` within the `parseQuery` function using `require`. This function should only be called when the consumer is attempting to mock GraphQL with `msw` 
3. Update all static imports of `graphql` in the project to be `type imports`

I tested these changes by using my version of `msw` with the [with-msw](https://github.com/vercel/next.js/tree/canary/examples/with-msw) Next.js example. I was able to run the app, and mock rest requests, without needing to install `graphql`.

The CodeSandbox example linked to this PR now compiles. It does warn about the `graphql` module not being found, but this is just a console warning and doesn't affect the application's ability to mock rest requests.

JohnDaly:
@kettanaito have you had a chance to review the latest changes?

package.json
kettanaito: I agree with your approach, I think it's great! Thanks for looking into this 👍

package.json
```diff
@@ -116,6 +115,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "fs-extra": "^10.0.0",
     "fs-teardown": "^0.3.0",
+    "graphql": "^16.3.0",
```
kettanaito, line 118: Is it safe to migrate to v16 given the optional range is 15/16? Are there any features that are dropped in v16 that we may accidentally rely upon?

package.json
kettanaito: This is what you've confirmed with the sandbox, right, @JohnDaly?

kettanaito:
Hey, @JohnDaly. I've finally got to review the changes you've made and they are great!

I've left two questions:

- Whether the compiled sandbox is a confirmation that we are not breaking apps for people now;
- Is it safe to rely on GraphQL v16 internally given we permit v15-16 as optional dependency?

package.json
JohnDaly: Yes, I was able to run the Sandbox and use MSW to mock REST requests

package.json
JohnDaly: Since v16 is the version that MSW has used in the latest release, I think it makes sense to keep it as the devDependency version.

There are two main changes in this PR that affect run-time compatibility:
1. Changing usage of `OperationTypeNode` from enum to string
2. The `parse` function

Changing `OperationTypeNode` should be safe to do, the `enum` in v16 maps to the same strings that were used in v15.

The `parse` function from `graphql` is the only piece of runtime code that is actually used by MSW. It's being used to get the `OperationDefinitionNode` for the GraphQL operation that is intercepted. From what I can tell, the fields that MSW is interested in on `OperationDefinitionNode` are unchanged between v15 and v16.

Based on this, I think it should be safe to list v15 as a dependency. 

It's also worth noting that v15 is, at the time of writing, the most commonly downloaded version of `graphql` on NPM (3+ million downloads a week, compared to ~2 million for v16). It would be good to be backwards compatible with v15, because lots of MSW users probably have that version in their projects.

JohnDaly:
> Hey, @JohnDaly. I've finally got to review the changes you've made and they are great!
> 
> I've left two questions:
> 
> * Whether the compiled sandbox is a confirmation that we are not breaking apps for people now;
> * Is it safe to rely on GraphQL v16 internally given we permit v15-16 as optional dependency?

@kettanaito Let me know if you have any follow-up questions on this

kettanaito (approved):

kettanaito:
Thank you for the superb improvement, @JohnDaly! 🎉 Feel free to tackle any other tasks as long as they look interesting to you. I know I could use your help. Thanks.

kettanaito:
## Released: v0.45.0 🎉

This has been released in v0.45.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.45.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.45.0)

---

limonte:
This is the misuse of peer dependencies. Please read the docs about peer deps and revert this change as the testing helper shouldn't require projects to install `graphql` https://nodejs.org/en/blog/npm/peer-dependencies/#the-solution-peer-dependencies

> What we need is a way of expressing these "dependencies" between plugins and their host package. Some way of saying, "I only work when plugged in to version 1.2.x of my host package, so if you install me, be sure that it's alongside a compatible host." We call this relationship a peer dependency.

`msw` will most certainly work without `graphql` installed, so `graphql` is definitely not a peer dependency.

kleinfreund:
As an example for what now happens in a project that doesn’t use/need graphql during build:

```
Module not found: Error: Can't resolve 'graphql' in 'node_modules/msw/lib'
```

jljouannic:
Another example, here is what `tsc` command gives me now in a project in which I do not need `graphql`:

```
node_modules/msw/lib/index.d.ts:4:63 - error TS2307: Cannot find module 'graphql' or its corresponding type declarations.

4 import { GraphQLError, OperationTypeNode, DocumentNode } from 'graphql';
                                                                ~~~~~~~~~
Found 1 error in node_modules/msw/lib/index.d.ts:4
```

limonte:
To clarify my previous comment. What this PR and `graphql` as peer dependency means is: 

> starting from `0.45.0`, `msw` **will not** work without `graphql` dependency

IMO It doesn't make sense to require `msw` users to have `graphql` as a dependency for their projects.

kettanaito:
Thanks for your concern, @limonte. What is the correct way to express the MSW-graphql dependency then? Because I think NPM, as a package manager, severely lacks in the means of expressing this kind of dependencies. 

To be absolutely clear: `graphql` _is_ an optional peer dependency of MSW:

- Optional, because `graphql` is not required for MSW to execute;
- Peer, because we have no intention of bundling `graphql` as a part of MSW. That is also a bad idea since GraphQL versions mismatches would be horrible.

If anyone has a better way to express this relationship, the pull requests are open and your contribution would be highly valued. 

As to the compilation issues, those happen because `graphql` is not correctly isolated from the rest of the package (just as I've suspected). It's either lacking proper tree-shacking or TypeScript configuration so that it doesn't attempt to load it. For instance, this line of code is entirely incorrect:

> import { GraphQLError, OperationTypeNode, DocumentNode } from 'graphql';

We can no longer import things from `graphql` directly. Moreover, that entire import should be `import type` because we rely exclusively on `parse` from `graphql` and nothing else for runtime.

kettanaito:
> IMO It doesn't make sense to require msw users to have graphql as a dependency for their projects.

This was never the intention. We've spent a lot of time testing this but this still slipped through. I'd appreciate if everyone stopped treating this as a deliberate shortcoming and expressed an ounce of compassion for people trying to make this package better. Thanks.

kettanaito:
I'm not going to revert anything. 0.45.0 only introduces the `graphql` change (and thus was correctly released as a breaking change), so you're not missing on anything else. Let this be an opportunity for people to contribute instead of complain. This needs a fix, not a revert. I will be more than glad to review and merge the fix within a day if anyone actually decides to tackle this.

limonte:
Thank you for the replies @kettanaito 🙏 

I won't be able to contribute, but IMO graphql API should be separated from the core package and imported like this:

```js
import { graphql } from 'msw/graphql'
```

similar to

```js
```

---

In this case the `graphql` dependency can be moved to `devDependencies` so tests will work.

And from the users' perspective graphql types will be imported from whatever version they have installed in their project.

kettanaito:
> And from the users' perspective graphql types will be imported from whatever version they have installed in their project.

That's the issue though. We need to specify the supported range for GraphQL you can use with MSW. We can't guarantee that new features in GraphQL 17 won't break MSW (imagine GraphQL renames the `parse` function). 

Moving the GraphQL to `msw/graphql` is interesting. I don't know if we can bundle things like that straight away. There's `msw/node` because it's, basically, a separate library. `msw/graphql` would be an entire copy of `msw` plus `graphql`. Your bundles would get huge (not really a concern since it's a devtool but still).

kettanaito:
And, once again, the issue is actually with bundling and TypeScript, not NPM. And so the solution should likely be in those areas. But, as I've said, I have no interest in working on this, so it's up to everyone to contribute with ideas and fixes.

PR #1369, opened 2022-08-24
fix: add "listHandlers" method to server and worker

AlexeiDarmin:
@kettanaito is something like this what you had in mind? 

**Change overview**

This PR adds a `listHandlers()` method that returns a list of currently active handlers as suggested in https://github.com/mswjs/msw/issues/474#issuecomment-1225571039.

> You can list currently active handlers via [server.printHandlers()](https://mswjs.io/docs/api/setup-server/print-handlers). Alas, it prints the handlers summary stdout instead of returning a list. If you are interested, it'd be awesome to have a change where printHandlers() returns a list of handlers:

**Additional change proposal**

Now that devs would have access to individual handlers, handlers would benefit from having a `.toString()` or `.stringSummary()` to easily see a string summary of the handler. The logging logic within `printHandlers` could be moved to `handler.toString()`, then `printHandlers` would simply be:

```typescript
 printHandlers() {
        currentHandlers.forEach((handler) => {
          console.log(handler.toString())
        })
      }
```

### DIFF
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -17,6 +17,7 @@ import { devUtils } from '../utils/internal/devUtils'
 import { pipeEvents } from '../utils/internal/pipeEvents'
 import { RequiredDeep } from '../typeUtils'
 import { MockedRequest } from '../utils/request/MockedRequest'
+import { toReadonlyArray } from '../utils/internal/toReadonlyArray'
 
 const DEFAULT_LISTEN_OPTIONS: RequiredDeep<SharedOptions> = {
   onUnhandledRequest: 'warn',
@@ -138,11 +139,13 @@ export function createSetupServer(
       },
 
       listHandlers() {
-        return currentHandlers
+        return toReadonlyArray(currentHandlers)
       },
 
       printHandlers() {
-        currentHandlers.forEach((handler) => {
+        const handlers = this.listHandlers()
+
+        handlers.forEach((handler) => {
           const { header, callFrame } = handler.info
 
           const pragma = handler.info.hasOwnProperty('operationType')
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -137,6 +137,10 @@ export function createSetupServer(
         )
       },
 
+      listHandlers() {
+        return currentHandlers
+      },
+
       printHandlers() {
         currentHandlers.forEach((handler) => {
           const { header, callFrame } = handler.info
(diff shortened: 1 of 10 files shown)


kettanaito:
Hey, @AlexeiDarmin. Thank you for opening this on such a short notice!

I absolutely agree with the summary proposal and `.toString()` on individual handlers. A great idea. Would you have a chance to add it to this pull request?

src/node/createSetupServer.ts
```diff
@@ -137,6 +137,10 @@ export function createSetupServer(
         )
       },
 
+      listHandlers() {
+        return currentHandlers
```
kettanaito: We need to be cautious exposing internals, and `currentHandlers` is internals. I fear that developers will be tempted to meddle with this array directly. That should never happen. They should use explicit APIs like `use` or `resetHandlers` to affect the list of handlers, and `listHandlers()` must be purely readonly. 

Maybe let's make this array readonly before exposing it?

```ts
// First, guard the array on the type level.
listHandlers(): ReadonlyArray<RequestHandler> {
  // Second, prevent mutations on runtime.
  const immutableHandlers = [...currentHandlers]
  Object.freeze(immutableHandlers)
  return immutableHandlers
}
```

src/node/glossary.ts
```diff
@@ -40,11 +42,17 @@ export interface SetupServerApi {
    */
   resetHandlers(...nextHandlers: RequestHandler[]): void
 
+  /**
+ * Returns lists of all active request handlers.
+ * @see {@link https://mswjs.io/docs/api/setup-server/list-handlers `server.list-handlers()`}
+ */
+  listHandlers(): RequestHandler<RequestHandlerDefaultInfo, MockedRequest<DefaultBodyType>, any, MockedRequest<DefaultBodyType>>[]
```
kettanaito: If we go to the readonly list route (which I think we should), let's adjust this return type as well.

kettanaito (changes requested):
I absolutely love the changes! The only suggestion I have is to make sure that we expose a readonly array from `listHandlers()` so that people are not tempted to mutate it.

kettanaito:
Since we're not deprecating `printHandlers()` just yet, I will rename this change to be a `fix`.

kettanaito:
I've covered some of the previous suggestions, also added `listHandlers` to the worker API. Let's see if the tests pass.

kettanaito (approved):

AlexeiDarmin:
@kettanaito thanks for applying the changes! I appreciate your feedback and was going to incorporate when I had a minute.

kettanaito:
## Released: v0.46.0 🎉

This has been released in v0.46.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.46.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.46.0)

---

PR #1370, opened 2022-08-24
test(concurrency): add a failing concurrency test

AlexeiDarmin:
Goal: create a test that reliably fails when msw is used with tests running in parallel.

Execute `yarn test:concurrency` in terminal to run only the concurrency test.

###  Method
- Create a shared server to be consumed by two seperate tests
- Define tests within one describe block with `it.concurrent()`
  - Using serial execution via `it()` syntax will make the tests pass.
  - `beforeAll()` is not officially supported with `.concurrent()` syntax so I use [this method](https://github.com/facebook/jest/issues/7997#issuecomment-796965078) to ensure `beforeAll()` is triggered before tests are run.
- use `async await sleep` to ensure operations upon the server happen in the precise order despite being in different tests.
- Assert that a request made in _test suite A_ is erroneously handled by a handler attached in _test suite B_.
-  Use a request param to distinguish between requests from both test suites `one` and `two`.

## Expected logs, if msw supported concurrency:

- `step 1 - process one attach handler to server`
- `step 2 - process two attach handler to server`
- `step 3 - process one make request`
- `step 4 - process one handler resolver triggered`
- `step 5 - process two make request`
- `step 6 - process two handler resolver triggered`

## Actual Logs

- `step 1 - process one attach handler to server`
- `step 2 - process two attach handler to server`
- `step 3 - process one make request`

  - **this is where it goes wrong**
- `step 6 - process two handler resolver triggered`
- `step 5 - process two make request`
- `step 6 - process two handler resolver triggered`

**Explanation:** 
- Process two was the most recent place where a dynamic handler was added to the server via `server.use()`. 
- Therefor all requests are routed to that implementation since it is at the front of the list of handlers.
- Two handlers are attached at this time, however only the newest one gets all of the API requests.

## Observations

Response to @kettanaito  proposal https://github.com/mswjs/msw/issues/474#issuecomment-1225571039

> Similar to requests identity, I propose to add moduleId/modulePath only to runtime handlers (those defined via server.use()). As long as the request and the runtime handler share the moduleId, it guarantees that they were issued/prepended in the same module and can affect each other.

Supporting parallization of testing within one file, in addition to by `moduleId` could be something worth striving towards.

## Runtime handler identity

I noticed the `callFrame` property from within `handler.info` exposes not only the module, but also the exact line number from which it was called. 

Idea is to add a `lineNumber` to the identity of requests and handlers.

1. Use `moduleId/modulePath` to narrow down the list of possibly valid runtime handlers to this module.
2. When more than one remains:
  a. find the one that has the closest line number **BEFORE** the `lineNumber` within the request.

```typescript 
{
  id: "uuid-of-request",
  modulePath: "/User/octocat/GitHub/test/Profile.test.tsx"
  lineNumber: 337
}
```

### DIFF
--- a/test/msw-api/setup-server/listHandlers.test.ts
+++ b/test/msw-api/setup-server/listHandlers.test.ts
@@ -1,0 +1,43 @@
+import { rest, graphql } from 'msw'
+import { setupServer } from 'msw/node'
+
+const resolver = () => null
+const github = graphql.link('https://api.github.com')
+
+const server = setupServer(
+  rest.get('https://test.mswjs.io/book/:bookId', resolver),
+  graphql.query('GetUser', resolver),
+  graphql.mutation('UpdatePost', resolver),
+  graphql.operation(resolver),
+  github.query('GetRepo', resolver),
+  github.operation(resolver),
+)
+
+beforeAll(() => {
+  server.listen()
+})
+
+afterEach(() => {
+  server.resetHandlers()
+})
+
+afterAll(() => {
+  server.close()
+})
+
+test('lists all current request handlers', () => {
+  const handlers = server.listHandlers()
+  expect(handlers.length).toEqual(6)
+})
+
+test('respects runtime request handlers when listing handlers', () => {
+  server.use(
+    rest.get('https://test.mswjs.io/book/:bookId', resolver),
+    graphql.query('GetRandomNumber', resolver),
+  )
+
+  const handlers = server.listHandlers()
+  // Runtime handlers are prepended to the list of handlers
+  // and they DON'T remove the handlers they may override.
+  expect(handlers.length).toEqual(8)
+})
--- a/test/msw-api/setup-server/scenarios/concurrency/api.ts
+++ b/test/msw-api/setup-server/scenarios/concurrency/api.ts
@@ -1,0 +1,12 @@
+import fetch from 'node-fetch'
+
+export const makeRequest = async (endpoint: string) => {
+  const res = await fetch(endpoint, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  })
+
+  return res
+}
--- a/test/msw-api/setup-server/scenarios/concurrency/process-one.test.ts
+++ b/test/msw-api/setup-server/scenarios/concurrency/process-one.test.ts
@@ -1,0 +1,89 @@
+import { rest, graphql } from 'msw'
+import { setupServer } from 'msw/node'
+import { makeRequest } from './api'
+import {
+  sharedServer,
+  sleep,
+  endpoint,
+  getNow,
+  beforeAllResolve,
+  beforeAllPromise,
+  sleepDelay,
+} from './shared-server'
+
+describe('concurrency test', () => {
+  beforeAll(async () => {
+    sharedServer.listen()
+    beforeAllResolve()
+  })
+
+  afterAll(async () => {
+    sharedServer.close()
+  })
+
+  it.concurrent('process one tests', async () => {
+    await beforeAllPromise
+
+    const mockFn = jest.fn()
+
+    console.log('step 1 - process one attach handler to server: '),
+      sharedServer.use(
+        rest.get(`${endpoint}/:processId`, (req, res, ctx) => {
+          console.log(
+            'step 4 - process one handler resolver triggered: ',
+            getNow(),
+          )
+
+          mockFn(req.params.processId)
+
+          return res(
+            ctx.json({
+              text: 'process one handler resolver response',
+            }),
+          )
+        }),
+      )
+
+    const handlers = sharedServer.listHandlers()
+
+    await sleep(sleepDelay)
+
+    console.log('step 3 - process one make request: ')
+
+    await makeRequest(`${endpoint}/one`)
+
+    expect(mockFn).toHaveBeenCalledWith('one')
+
+    console.log('process one end: ', handlers)
+  })
+
+  it.concurrent('process two tests', async () => {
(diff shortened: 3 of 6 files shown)


test/msw-api/setup-server/listHandlers.test.ts
```diff
@@ -0,0 +1,43 @@
+import { rest, graphql } from 'msw'
```
kettanaito, line 1: I believe this is a test leaking from #1369. Consider removing?

kettanaito:
Hey, @AlexeiDarmin. This is a fantastic effort, I'm beyond happy to read through your thoughts and implementation of concurrency tests. 

I think what you describe in the test setup and the expected/actual behavior makes total sense and this is the direction we should pursue when writing this test. 

As to the `moduleId` and using the number of `callFrame`, I feel that it's a workaround rather than a solution. I keep comparing MSW to Nock in terms of parallel runs, but I guess the main distinction here is that Nock doesn't have a concept of a "server" (global singleton that keeps the list of handlers/interceptors). So whatever Nock attaches within a single test doesn't get exposed outside of that test. Perhaps that's what gives parallel support to Nock without much effort. 

In our case, we keep the list of `currentHandlers` internally, and methods like `use()` prepend handlers to that list. The issue is that those runtime handlers are meant to be scoped per the test that adds them, but since the `server` is global, they are never scoped. Maybe this is a chance to re-think how we keep and modify the list of handlers.

test/msw-api/setup-server/scenarios/concurrency/api.ts
```diff
@@ -0,0 +1,12 @@
+import fetch from 'node-fetch'
+
+export const makeRequest = async (endpoint: string) => {
+  const res = await fetch(endpoint, {
+    method: 'GET',
```
kettanaito, line 5: `GET` is the default method, we may drop it.

test/msw-api/setup-server/scenarios/concurrency/api.ts
```diff
@@ -0,0 +1,12 @@
...
+    headers: {
+      'Content-Type': 'application/json',
```
kettanaito, line 7: Specifying `Content-Type` for `GET` requests does nothing as GET requests cannot have body (content).

test/msw-api/setup-server/scenarios/concurrency/process-one.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+          return res(
+            ctx.json({
+              text: 'process one handler resolver response',
+            }),
+          )
+        }),
+      )
+
+    const handlers = sharedServer.listHandlers()
```
kettanaito, line 47: Oh, so you want to assert the list of active handlers as well? I think that's smart. It makes this change dependent on #1369, which is okay. Now I see why you have the `listHandlers` test.

test/msw-api/setup-server/scenarios/concurrency/process-one.test.ts
```diff
@@ -0,0 +1,89 @@
...
+          mockFn(req.params.processId)
+
+          return res(
+            ctx.json({
+              text: 'process two handler resolver response',
+            }),
+          )
+        }),
+      )
+    await sleep(sleepDelay)
```
kettanaito, line 82: Can you please add a brief comment right here above the `sleep` call about why it's necessary? How does it contribute to asserting the behavior?

test/msw-api/setup-server/scenarios/concurrency/process-one.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  it.concurrent('process two tests', async () => {
+    await beforeAllPromise
+
+    const mockFn = jest.fn()
+
+    console.log('step 2 - process two attach handler to server: '),
+      sharedServer.use(
+        rest.get(`${endpoint}/:processId`, (req, res, ctx) => {
+          console.log(
```
kettanaito, line 68: I think we should turn these logs from purely observational to the data we actually assert in the test. We're doing a similar pattern with the life-cycle events, you can take a look at it here:

https://github.com/mswjs/msw/blob/b93aca9f165294d064517990e6c55aee7557d9b1/test/msw-api/setup-server/life-cycle-events/on.test.ts#L94

Checking that the mock function was called with the correct arguments still doesn't guarantee that the entire request processing pipeline was correct.

test/msw-api/setup-server/scenarios/concurrency/shared-server.ts
```diff
@@ -0,0 +1,43 @@
+import { setupServer } from 'msw/node'
+
+export const sharedServer = setupServer()
+
+export const sleepDelay = 3000
+export const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms))
```
kettanaito, line 6: We already have the `sleep` utility, you can reuse it:

https://github.com/mswjs/msw/blob/b93aca9f165294d064517990e6c55aee7557d9b1/test/support/utils.ts#L4

kettanaito (changes requested):
This is a great start! Let's iterate a little on how/what we assert in these tests and think about the implementation approach that could support concurrent runs meanwhile.

AlexeiDarmin:
@kettanaito thank you again for the feedback, I'll apply what changes I can this week!

svarlet:
I'm suffering from these concurrency issues too. 
Has the 'one http port per test' solution been considered to allow a single service worker to support per-test config?
Alternatively, perhaps a header based solution could work too.
Could we create many service workers? for example: one per test?

kettanaito:
## Released: v2.2.0 🎉

---

Issue #1373, opened 2022-08-26
GraphQl in client bundle although it's unused


ashtonlance:
Latest version of msw and @mswjs/data. I'm only using rest in my app but I'm getting huge chunks of GraphQL in my client bundle.

<img width="794" alt="Screen Shot 2022-08-25 at 8 28 22 PM" src="https://user-images.githubusercontent.com/13652692/186792794-08996449-9938-49f2-93e4-62bfbce610fa.png">

nathanredblur:
Even when I'm only using rest, I can start the project because this module requires GraphQL

```ERROR in ../node_modules/msw/lib/index.js 696:30-48
Module not found: Error: Can't resolve 'graphql' in '/Users/pepito/app/app-web/node_modules/msw/lib'
resolve 'graphql' in '/Users/pepito/app/app-web/node_modules/msw/lib'
  Parsed request is a module
  using description file: /Users/pepito/app/app-web/node_modules/msw/package.json (relative path: ./lib)
    Field 'browser' doesn't contain a valid alias configuration
    resolve as module
      /Users/pepito/app/app-web/node_modules/msw/lib/src doesn't exist or is not a directory
      looking for modules in /Users/pepito/app/app-web/node_modules/msw/node_modules
        single file module
          using description file: /Users/pepito/app/app-web/node_modules/msw/package.json (relative path: ./node_modules/graphql)
            no extension
              Field 'browser' doesn't contain a valid alias configuration
              /Users/pepito/app/app-web/node_modules/msw/node_modules/graphql doesn't exist
            .tsx
              Field 'browser' doesn't contain a valid alias configuration
              /Users/pepito/app/app-web/node_modules/msw/node_modules/graphql.tsx doesn't exist
            .ts
              Field 'browser' doesn't contain a valid alias configuration
              /Users/pepito/app/app-web/node_modules/msw/node_modules/graphql.ts doesn't exist
            .jsx
              Field 'browser' doesn't contain a valid alias configuration
              /Users/pepito/app/app-web/node_modules/msw/node_modules/graphql.jsx doesn't exist
            .js
              Field 'browser' doesn't contain a valid alias configuration
              /Users/pepito/app/app-web/node_modules/msw/node_modules/graphql.js doesn't exist
        /Users/pepito/app/app-web/node_modules/msw/node_modules/graphql doesn't exist
      /Users/pepito/app/app-web/node_modules/msw/lib/node_modules doesn't exist or is not a directory```

[...]

PR #1381, opened 2022-08-30
fix(deps): include "graphql" as a dependency

### LINKED ISSUE #1371: tsc error: Cannot find module 'graphql' or its corresponding type declarations
### Prerequisites

### Environment check

### Node.js version

v16.11.0

### Reproduction repository

https://github.com/takanokana/msw-for-issue

### Reproduction steps

1．npm ci
2. npm run typecheck

### Current behavior

node_modules/msw/lib/index.d.ts:4:63 - error TS2307: Cannot find module 'graphql' or its corresponding type declarations.

4 import { GraphQLError, OperationTypeNode, DocumentNode } from 'graphql';

### Expected behavior

No errors when executing tsc.

mattcosta7:
this makes the graphql peer dependency required. While the hope is to eventually avoid the need to require the graphql library for users who don't declare graphql requests directly, the current architecture means that many tools fails to properly compile without error/warning without this.

Since NPM 7, npm will automatically install peer dependencies, which should avoid consumers of msw having to manually install graphql.

While ideally we'd avoid this now - this PR may be a reasonable stopgap that avoids frustration/confusion in the interim

I also reverted the change to `require('graphql')`.  The semantics of lazily requiring are a bit off imo, and given that it doesn't provide the desired effect, there's a bit more clarity maintaining import syntax. 

A follow-up that separates the package/bundle in a way that allows to properly treeshake the graphql imports (or provide them from a different import specifier), so they aren't required dependencies would be ideal.

Totally get it, if the goal is to try a fail forward approach instead of trying to smooth out the current usage, until a more holistic fix can be merged - but figured this might avoid some annoyance in the meantime (both on consumers and avoid issues like #1371 and the handful of duplicates that have popped up over the last week)

- Fixes #1371

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,9 +138,6 @@
     "typescript": ">= 4.2.x <= 4.8.x"
   },
   "peerDependenciesMeta": {
-    "graphql": {
-      "optional": true
-    },
     "typescript": {
       "optional": true
     }
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "chalk": "4.1.1",
     "chokidar": "^3.4.2",
     "cookie": "^0.4.2",
+    "graphql": "^15.0.0 || ^16.0.0",
     "headers-polyfill": "^3.0.4",
     "inquirer": "^8.2.0",
     "is-node-process": "^1.0.1",
@@ -115,7 +116,6 @@
     "eslint-plugin-prettier": "^3.4.0",
     "fs-extra": "^10.0.0",
     "fs-teardown": "^0.3.0",
-    "graphql": "^16.3.0",
     "jest": "26",
     "json-bigint": "^1.0.0",
     "lint-staged": "^11.0.1",
@@ -134,7 +134,6 @@
     "webpack-dev-server": "^3.11.2"
   },
   "peerDependencies": {
-    "graphql": "^15.0.0 || ^16.0.0",
     "typescript": ">= 4.2.x <= 4.8.x"
   },
   "peerDependenciesMeta": {
--- a/src/utils/internal/parseGraphQLRequest.ts
+++ b/src/utils/internal/parseGraphQLRequest.ts
@@ -3,6 +3,7 @@ import type {
   OperationDefinitionNode,
   OperationTypeNode,
 } from 'graphql'
+import { parse } from 'graphql'
 import { GraphQLVariables } from '../../handlers/GraphQLHandler'
 import { getPublicUrlFromRequest } from '../request/getPublicUrlFromRequest'
 import { MockedRequest } from '../request/MockedRequest'
@@ -40,9 +41,6 @@ export function parseDocumentNode(node: DocumentNode): ParsedGraphQLQuery {
 
 function parseQuery(query: string): ParsedGraphQLQuery | Error {
   try {
-    // Since 'graphql' is an optional peer dependency,
-    // we'll attempt to import it dynamically
-    const { parse } = require('graphql')
     const ast = parse(query)
     return parseDocumentNode(ast)
   } catch (error) {
--- a/test/graphql-api/compatibility.node.test.ts
+++ b/test/graphql-api/compatibility.node.test.ts
@@ -1,6 +1,5 @@
 import fetch from 'cross-fetch'
-import { graphql as executeGraphql } from 'graphql'
-import { buildSchema } from 'graphql/utilities'
+import { graphql as executeGraphql, buildSchema } from 'graphql'
 import { graphql } from 'msw'
 import { setupServer } from 'msw/node'
 import { createGraphQLClient, gql } from '../support/graphql'
--- a/test/graphql-api/extensions.node.test.ts
+++ b/test/graphql-api/extensions.node.test.ts
@@ -1,14 +1,11 @@
 /**
  * @jest-environment node
  */
-import fetch from 'node-fetch'
-import {
-  graphql as executeGraphql,
-  buildSchema,
-  ExecutionResult,
-} from 'graphql'
+import type { ExecutionResult } from 'graphql'
+import { buildSchema, graphql as executeGraphql } from 'graphql'
 import { graphql } from 'msw'
 import { setupServer } from 'msw/node'
+import fetch from 'node-fetch'
 import { gql } from '../support/graphql'
 
 const schema = gql`
--- a/test/graphql-api/extensions.test.ts
+++ b/test/graphql-api/extensions.test.ts
@@ -1,6 +1,6 @@
 import * as path from 'path'
 import { pageWith } from 'page-with'
-import { ExecutionResult } from 'graphql'
+import type { ExecutionResult } from 'graphql'
 import { executeGraphQLQuery } from './utils/executeGraphQLQuery'
 import { gql } from '../support/graphql'
 
--- a/test/graphql-api/response-patching.node.test.ts
+++ b/test/graphql-api/response-patching.node.test.ts
@@ -4,8 +4,7 @@
 import { graphql } from 'msw'
 import { setupServer } from 'msw/node'
 import fetch from 'cross-fetch'
-import { graphql as executeGraphql } from 'graphql'
-import { buildSchema } from 'graphql/utilities'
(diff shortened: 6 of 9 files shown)


package.json
```diff
@@ -138,9 +138,6 @@
     "typescript": ">= 4.2.x <= 4.8.x"
   },
   "peerDependenciesMeta": {
-    "graphql": {
-      "optional": true
```
kettanaito, line 142: I find this a rather harsh strategy. If we wish to solve compilations for people who don't use GraphQL, then I'd propose to add back `graphql` as a dependency. Respect the version range as we specify in `peerDependencies`, and install it alongside MSW. Here's what going to happen for the two groups of users:

- Those who use GraphQL will get it deduped, given that they comply with the peer dependency version MSW requires (which they should).
- Those who don't use GraphQL won't have to make any changes to their projects. GraphQL will be installed as MSW's dependency but won't get into their bundle unless they A) use it; B) we code-split correctly. 

What do you think about this? Returning `graphql` as a regular dependency.

kettanaito: Since MSW is a dev tool, the price of adding dependencies is relatively low. People won't ship MSW, so even if GraphQL ends up in their dev bundles, there's zero harm in that.

src/utils/internal/parseGraphQLRequest.ts
```diff
@@ -3,6 +3,7 @@ import type {
   OperationDefinitionNode,
   OperationTypeNode,
 } from 'graphql'
+import { parse } from 'graphql'
```
kettanaito, line 6: While we're at it, I believe there's a place or two where we import types from `graphql` but do not mark those imports as `import type`. Could you please look into those?

kettanaito (changes requested):
Thanks for tackling this, @mattcosta7! 

I've left one proposal about how to move forward with this. I'm curious what you think about this.

package.json
ashtonlance: I'd really prefer to not have msw in my dev bundle. There's potential that I could be using as a stopgap in prod until some micro-services are completed by another team.

kettanaito: > I'd really prefer to not have msw in my dev bundle

How do you see it getting into your code then? Did you mean "prod" bundle?

MSW is and will always be a part of your dev bundle. Well, as long as you tell it to (use conditional require as suggested in the docs). My point here was that MSW won't ever get to a prod bundle unless that's your intention. And since it's not a user-facing KBs of code, it doesn't matter how many dependencies it installs, within the reason, of course.

ashtonlance: Sorry yes, my prod bundle. I had an issue that was closed (https://github.com/mswjs/msw/issues/1373) where graphql was in my prod bundle even though I'm not using it anywhere.

mattcosta7: I'd be fine with returning it as a dependency.  I think the peer strategy here should install a version at the top of the semver range as well, which _shouldn't_ require users import it directly in this case either, but a plain dependency is fine I think

> Those who don't use GraphQL won't have to make any changes to their projects. GraphQL will be installed as MSW's dependency but won't get into their bundle unless they A) use it; B) we code-split correctly.

Re this - I don't believe that's true currently - msw isn't shipping esm, and ships a single file bundle, so I doubt any tools could correctly remove/shake graphql from any consumer bundles (maybe possible, but offhand I don't think so). 

It might take a bit to dig into that - but emitting an esm build that's transpiled rather than bundled might help unlock that

mattcosta7: I can look at fixing up this pr later today, or - if someone has time before I get to it (likely 8+ hours from now) - they're welcome to take it over

mattcosta7: a1564c37aed185febc48381b3476806d6e69dd3e makes graphql a dependency again using the same range previously in the peer requirement

src/utils/internal/parseGraphQLRequest.ts
mattcosta7: Audited those here (mostly in tests) - 1dbc669ce1b9d2bfb229bc655542c336098804f9

test/graphql-api/compatibility.node.test.ts
```diff
@@ -1,6 +1,5 @@
 import fetch from 'cross-fetch'
-import { graphql as executeGraphql } from 'graphql'
-import { buildSchema } from 'graphql/utilities'
```
mattcosta7, line 3: we use this from `/utilities` and from `graphql` in different files, so I moved to just using the base package export

src/utils/internal/parseGraphQLRequest.ts
mattcosta7: https://github.com/mswjs/msw/blob/main/src/handlers/GraphQLHandler.test.ts#L5

Just calling this out separately.  the import for `OperationTypeNode` is used as an enum in this test, so we can't safely just import types there

mattcosta7:
Failing on typescript 4.5 looks like it was a network blip and I expect it to succeed if we re-run

src/utils/internal/parseGraphQLRequest.ts
kettanaito: Oh, great catch! We need to be careful with enums. Luckily, those won't compile if we import them incorrectly.

test/graphql-api/compatibility.node.test.ts
kettanaito: 💯 agree with this. We should rely on the base package all the time.

kettanaito:
@mattcosta7, yeap, looks like a hiccup when installing dependencies on CI. Re-running...

package.json
```diff
@@ -77,6 +77,7 @@
     "chalk": "4.1.1",
     "chokidar": "^3.4.2",
     "cookie": "^0.4.2",
+    "graphql": "^15.0.0 || ^16.0.0",
```
kettanaito, line 80: The only thing I'd like to verify is whether this won't get multiple `graphql` packages in an app that uses a different GraphQL version.

mattcosta7: is there a possibility to make a prerelease/canary version, which might be the easiest way to validate that?

I'm also not positive, if it does we can drop back to declaring this as a peer and dealing with the automatically installed peers behavior for projects that don't declare a local version

mattcosta7: @kettanaito 

To test, I published a version of this package to `@mattcosta7/msw`

I installed this package instead of `msw` in a project that consumes msw currently (without graphql included as a dependency - it was a transient dep by virtue of msw being in this pacakge already)

Screen Shot 2022-08-30 at 10 28 37 PM


I then decided that project should use `graphql@15.8`

Screen Shot 2022-08-30 at 10 30 42 PM

I think that should give a good indication that this de-duplication should happen on the package manager side

previously graphql _was_ in the pacakge-lock due to msw@0.44 

Screen Shot 2022-08-30 at 10 33 24 PM

mattcosta7: Let me know if there are any other steps you'd like me to take to help verify this?

kettanaito: That sounds great! Thank you for validating this, @mattcosta7. 

FIY, you can also install repositories at a specific PR state as

```json
{
  "msw": "mswjs/msw#1381/head"
}
```

kettanaito (approved):
This looks good!

This should solve the dependency issue for those who're not using `graphql`. This should also not introduce any conflicts to those who are using `graphql`.

Let's test it out.

kettanaito:
Thank you so much for your work on this, @mattcosta7! 👏 You're our dependency wizard.

ashtonlance:
Thanks @mattcosta7!
@kettanaito When will this version be versioned and released?

kettanaito:
We need to write this somewhere.

@ashtonlance, MSW is released automatically at regular intervals. I believe it's every 3 days:

https://github.com/mswjs/msw/blob/8436515fd93e3f372205815b52ffc3a572da2690/.github/workflows/release.yml#L5

But since this is a blocking issue for our users I will trigger the release manually right now.

kettanaito:
## Released: v0.46.1 🎉

This has been released in v0.46.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.46.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.46.1)

---

PR #1383, opened 2022-09-01
feat: add explicit `exports` map in package.json

ivanhofer:
This PR adds the missing `exports` field to the `package.json`.
This makes `msw` compatible with `TypeScript` projects using the `moduleResolution` set to `node16`.

https://www.typescriptlang.org/docs/handbook/esm-node.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,11 +5,13 @@
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
   "exports": {
-    ".": "./lib/index.js",
+    ".": {
+      "default": "./lib/index.js"
+    },
     "./native": "./lib/native/index.js",
     "./node": {
-      "import": "./lib/node/index.mjs",
-      "require": "./lib/node/index.js"
+      "require": "./lib/node/index.js",
+      "default": "./lib/node/index.mjs"
     }
   },
   "bin": {
--- a/package.json
+++ b/package.json
@@ -4,6 +4,15 @@
   "description": "Seamless REST/GraphQL API mocking library for browser and Node.js.",
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
+  "exports": {
+    ".": "./lib/index.js",
+    "./iife": "./lib/iife/index.js",
+    "./native": "./lib/native/index.js",
+    "./node": {
+      "import": "./lib/node/index.mjs",
+      "require": "./lib/node/index.js"
+    }
+  },
   "bin": {
     "msw": "cli/index.js"
   },
--- a/package.json
+++ b/package.json
@@ -8,7 +8,9 @@
     ".": {
       "default": "./lib/index.js"
     },
-    "./native": "./lib/native/index.js",
+    "./native": {
+      "default": "./lib/native/index.js"
+    },
     "./node": {
       "require": "./lib/node/index.js",
       "default": "./lib/node/index.mjs"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -4,6 +4,15 @@
   "description": "Seamless REST/GraphQL API mocking library for browser and Node.js.",
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
+  "exports": {
+    ".": "./lib/index.js",
+    "./iife": "./lib/iife/index.js",
```
kettanaito: I don't think we promise `msw/iife` import. It's true in the case of `msw/node` and `msw/native` but as far as I reacal IIFE should be imported directly from the build directory like so:

```html
<script src="https://unpkg.com/msw/lib/iife/index.js"></script>
```

We don't have a placeholder directory `./iife` so I believe the `iife` export rule is misleading.

package.json
```diff
@@ -4,6 +4,15 @@
...
+    "./native": "./lib/native/index.js",
+    "./node": {
+      "import": "./lib/node/index.mjs",
+      "require": "./lib/node/index.js"
```
mattcosta7: is it worth moving these to 

`require` and `default` instead of explicitly using import?

package.json
```diff
@@ -4,6 +4,15 @@
   "description": "Seamless REST/GraphQL API mocking library for browser and Node.js.",
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
+  "exports": {
+    ".": "./lib/index.js",
```
mattcosta7: is it worth being a little forward thinking here and using the 

```
{ "default": "./lib/index.js" } 
```
configuration, which might allow for separately supporting an esm and cjs build at some point, with a smaller diff here at that point

not really necessary this pass though, since it would be equivalent in use

package.json
ivanhofer: You are right. I'll remove this.

package.json
ivanhofer: sure I can change this

package.json
ivanhofer: I did it. Any reasons using `default` over `import`?

mattcosta7: not really should be fine either way.  I tend to think of it as ideally everything should go to esm, and `require` is the special case, so it kind of bakes that thinking into the configuration - but that's only semantic and not functional

package.json
```diff
@@ -4,6 +4,18 @@
...
+  "exports": {
+    ".": {
+      "default": "./lib/index.js"
+    },
+    "./native": {
+      "default": "./lib/native/index.js"
+    },
+    "./node": {
+      "require": "./lib/node/index.js",
+      "default": "./lib/node/index.mjs"
```
kettanaito, line 16: Does the "default" behavior mean ESM?

mattcosta7: Default would be anything not explicitly matched above - so `import` usage would fall into that export

kettanaito: Got it. Thanks for the explanation!

kettanaito (approved):
Looks good from my side. Waiting for the tests to pass.

kettanaito:
Thank you, @ivanhofer @mattcosta7 for your contribution! You folks are simply fantastic. Let's ship this.

kettanaito:
## Released: v0.47.0 🎉

This has been released in v0.47.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.47.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.47.0)

---

PR #1392, opened 2022-09-07
fix: print response log only when response arrives

### LINKED ISSUE #1343: Display request log once the response is signaled by the worker
### Scope

Improves an existing behavior

### Compatibility

## Current behavior

Right now MSW prints the `[MSW] /resource 200 OK` request log _immediately_ once the mocked response has been sent to the worker. 

## Proposed behavior

Instead, print the request log once the worker reports the response event. This way if the response has some delay, or took some time to process for other reasons, the developer would know that, translating the moment when the request log has appeared to the moment when the response has been truly handled by the worker. 

## Links

https://github.com/mswjs/msw/blob/6c6e1193327ba6c7d55826fd7997ddc9f1fb4fd5/src/setupWorker/start/createRequestListener.ts#L66-L80

This is the wrong place to call `handler.log()`. We should migrate from this and reconsider if `onMockedResponseSent` is used for any other purpose. 

We may also consider rewriting `onMockedResponseSent` to actually await the response event from the worker and _then_ dispatching its callback. For example, `handleRequest()` accepts `context.emitter` as an input, so it can create a listener on the response events:

```ts
// handleRequest.ts
// We have two distinct events for different response resolution type,
// so listen to both. One request will always have only one of the response
// events emitter by the worker. 
context.emitter.once('response:mocked', handleResponse)
context.emitter.once('response:bypassed', handleResponse)

const handleResponse = (response, requestId) => {
  // Ignore response events from irrelevant requests.
  if (request.id !== requestId) {
    return
  }

  onMockedResponseSent?.(info)
}
```

kettanaito:
- Fixes #1343 

## Changes

- The response log message in the console will now be printed only once the response actually arrives. This will respect the response timing in visual reporting.
- Fallback mode will now emit `response:mocked` and `response:bypass` life-cycle events (previously, never emitted those, that's a bug). 
- Removes `onMockedResponseSent` internal hook in `handleRequest`.

### DIFF
--- a/src/handlers/GraphQLHandler.ts
+++ b/src/handlers/GraphQLHandler.ts
@@ -197,8 +197,7 @@ Consider naming this operation or using "graphql.operation" request handler to i
 
   log(
     request: Request,
-    response: SerializedResponse,
-    handler: this,
+    response: SerializedResponse<any>,
     parsedRequest: ParsedGraphQLRequest,
   ) {
     const loggedRequest = prepareRequest(request)
--- a/src/handlers/RequestHandler.ts
+++ b/src/handlers/RequestHandler.ts
@@ -140,7 +140,6 @@ export abstract class RequestHandler<
   abstract log(
     request: Request,
     response: SerializedResponse<any>,
-    handler: this,
     parsedResult: ParsedResult,
   ): void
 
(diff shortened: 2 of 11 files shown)


kettanaito:
## Released: v0.47.1 🎉

This has been released in v0.47.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.47.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.47.1)

---

PR #1394, opened 2022-09-07
chore: use playwright/test for integration tests

### LINKED ISSUE #1342: Migrate to "@playwright/test"
### Scope

Improves an existing behavior

### Compatibility

### Feature description

I'd like to use [@playwright/test](https://www.npmjs.com/package/@playwright/test) library for our integration tests. It comes with great debugging experience and should overall be faster.

## Precautions

We currently use `page-with`, which also compiles the examples, including runtime compilation. This may be a task for `page-with`, or for MSW to integrate such examples' compilation with Playwright Test.

kettanaito:
- Closes #1342 

## Changes

- Adds and configures `@playwright/test` for integration tests instead of `page-with`.
- Improves the stability of the integration tests. 

## Roadmap

- [x] Fix failing `src/utils/handleRequest.test.ts:251:30`
- [x] Migrate the rest of the tests
   - [x] `rest-api`
   - [x] `graphql-api`
   - [x] `msw-api`
- [x] Add `.use()` support to `@open-draft/test-server` for runtime middleware
- [x] Rewrite existing tests to always use `HttpServer` from `test-server`
- [x] **Consider if spawning API server per test would be a better idea**
- [x] Adjust the `test:integration` script to use `playwright`
- [x] Fix the failing 9 tests
- [x] Split the browser and node integration tests (incl. CI)
- [x] Account for new tests from #1418
- [x] Resolve leaking state in tests
- [x] Update `CONTRIBUTING.md` guidelines on how to run `--debug` mode in PW
- [x] **Fix all `test.skip()` and `test.fixme()`**
- [x] Configure CI to upload Playwright test output artifacts for better debugging (although videos and screenshots are of little use, given most of the things happen in Network/Console). It'd be much better to export each failed test's Network into an HAR file and attach that to the failed run. 

## Resources

- [webpack-http-server](https://github.com/kettanaito/webpack-http-server). As a part of this change, I'm decoupling the runtime compilation capabilities of `page-with` into a standalone package.


test/playwright.config.ts
```diff
@@ -0,0 +1,18 @@
+import { PlaywrightTestConfig, devices } from '@playwright/test'
+
+const config: PlaywrightTestConfig = {
+  workers: 1,
+  globalSetup: require.resolve('./playwright.setup.ts'),
+  forbidOnly: !!process.env.CI,
+  use: {
+    serviceWorkers: 'allow',
```
kettanaito: This is redundant, remove.

test/playwright.extend.ts
```diff
@@ -0,0 +1,28 @@
...
+      )
+      const result = await res.json()
+
+      await page.goto(result.previewUrl, { waitUntil: 'networkidle' })
+    })
+  },
+  async fetch({ page }, use) {
+    await use(async (url) => {
+      page.evaluate((url) => fetch(url), url)
+      return page.waitForResponse(url)
```
kettanaito: ## Discovery: `page.request` vs `page.evaluate(fetch)`

PW comes with a built-in `page.request` utility but requests it makes bypass the network. I believe that's intentional in order for PW to spy on requests you make. Since we need requests to actually happen to hit the service worker, I'm adding a custom `fetch` fixture.

test/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
-    "target": "ES2015",
+    "target": "esnext",
+    "module": "commonjs",
```
kettanaito: May also be irrelevant, reconsider.

src/utils/url/getAbsoluteWorkerUrl.ts
```diff
@@ -3,5 +3,5 @@
  * relative URL (known during the registration).
  */
 export function getAbsoluteWorkerUrl(relativeUrl: string): string {
-  return new URL(relativeUrl, location.origin).href
+  return new URL(relativeUrl, location.href).href
```
kettanaito: ## TODO

I need to move this into a standalone fix. The worker script URL is not resolved correctly when using relative paths.

kettanaito: Moved it to #1456.

kettanaito:
Working on splitting tests into browser and node directories as it's less overhead when configuring different testing frameworks to run there.

shwarcu:
Hi @kettanaito , I am a happy user of MSW and I also work with Playwright recently. I subscribed to this issue thinking that maybe you will have some tasks that can be delegated to contributors. If you need some help with test automation and you have some bits of work around Playwright that can be delegated to other people let me know. I will try to organize some time on weekends to help you with the project :) 

Big thanks for making mswjs <3

kettanaito:
Hey, @shwarcu! Teaming up on this would be amazing, actually. I struggle to get Playwright to work reliably as I suspect some leaking state in my testing setup (not PW's fault). Most of the time a few random e2e tests fail with something like

```
Test timeout of 30000ms exceeded while tearing down "context".
```

- [Example](https://github.com/mswjs/msw/actions/runs/4089213767/jobs/7051680036)

The usefulness of that error message leaves to be desired. However, I've already hunted down and resolved a few leaking states, such as `workerConsole` that persisted between different tests, leading to false negative matches. 

## Context

Basically, I'm migrating to `@playwright/test` with this pull request. It's roughly 90% done, save for the flakiness issue I've described above. I wish I could share more details but that's as far as I got to myself. If I can get local tests to pass sometimes, on CI they fail absolutely every run. I also notice that GitHub Actions give us only 2 core CPUs, which means Playwright is running in a single-worker mode, making tests slow. Individual tests are also slower (~1.5s local, ~5s remote), which I also contribute to the sheer power difference between machines. I wouldn't say that's an issue right now. 

## How can you help?

First of all, I'd be tremendously thankful if we could team up and ship this. These changes have been floating in the air for months (as most of my open source endeavors because life) but I find them crucial for the library's future. The current E2E setup we have on `main` is reliable but extremely slow.

Here's a couple of things where you could help:

1. Look into that `context` timeout issue. Can we get a more specific error message/stack trace out of it?
2. Inspect the Playwright setup (`test/browser/playwright.config.ts` and `test/browser/playwright.extend.ts`) for potential leaked state/other reasons of flakiness. It should be rather straightforward but I can answer any questions you have any time. 
3. Suggest any Playwright best practices you find applicable in MSW tests. I'd deem this as less priority and we can dive into those even after this pull request is merged. I think that'd make more sense to iterate on this change, given how long I've been dragging it already. 

Please let me know if you'd like to address some of those points. We can also switch to Discord for more efficient communication if you find that to your liking. Thank you!

shwarcu:
@kettanaito thanks a lot for sharing the context of current work and issues. I will start with getting the repo locally, running tests on my machine,  getting familiar with the setup and how tests look like. I will get back to you once I'm able to sit down and tackle one of these points :telephone_receiver: I will keep in mind what you said while running things on my machine, maybe I will have some early thoughts

shwarcu:
@kettanaito just very first thing that came to my mind when I took a look at config and few files - when it comes to tests that run in the browser, have you tried using https://playwright.dev/docs/api/class-testproject#test-project-fully-parallel configuration option? I think having it enabled from the start can be beneficial in few ways:
- you are following good practices from the very start - tests should not depend on each other and shouldn't be affected by side effects of any other tests
- easier to manage setup and teardown - browser context is not shared between tests
- if there are some memory issues in the browser, they won't affect entire test file (suite)

So I would propose following changes https://github.com/mswjs/msw/pull/1532 (PR might not be complete at the moment, I haven't checked all tests yet)

Also, even though in this case we're not testing a web application, it might be helpful in some cases to have traces recorded for failed tests https://playwright.dev/docs/test-configuration#record-test-trace
MSW outputs information in the console, depending on how much console logs are helpful when debugging - it might be worth enabling traces with `'on-first-retry'` option (because I see you already have retries enabled and set to 2)


Edit: btw I just found out you can enable GPU acceleration when running tests in headless mode which may speed up test run 
https://github.com/michel-kraemer/sudocle/blob/main/playwright.config.js
https://michelkraemer.com/enable-gpu-for-slow-playwright-tests-in-headless-mode/

Edit: I noticed that when I put 100% load on my CPU by running all tests in 50% workers, then my OS is barely handling any other tasks. So when OS is under 100% load and can't take any more, I can imagine that tests may become unstable - it did happen to me multiple times locally that tests kept failing until I reduced the load.
Enabling GPU acceleration seemed to improve the situation. I still wonder how to apply these findings into making Github Actions test runs more stable. It might be a generic Playwright issue that it can't handle big loads well. My colleagues faced similar issues when they run Playwright tests on a macbook with Intel CPU while Playwright config was created and tested only on M1 prior to them running on Intel (tl;dr - there were more workers spawned than Intel based macbook could handle well)


Example failures I get when running tests on my PC:

The one you mentioned already
```sh

  4) [chromium] › msw-api\setup-worker\start\wait-until-ready.error.test.ts:3:5 › restores deferred requests if the worker registration fails 

    Test timeout of 30000ms exceeded while tearing down "context".
```

```sh
3) [chromium] › msw-api\setup-worker\start\start.test.ts:25:5 › resolves the "start" Promise when the worker has been activated

    Test timeout of 30000ms exceeded.

    page.evaluate: Target closed

      34 |
      35 |   const untilWorkerActivated = page
    > 36 |     .evaluate(() => {
         |      ^
      37 |       return new Promise((resolve) => {
      38 |         navigator.serviceWorker.addEventListener('controllerchange', resolve)
      39 |       })

```
⬆️ This one may be caused by **missing** `await` before `page.evaluate`. There is eslint plugin for Playwright code https://www.npmjs.com/package/eslint-plugin-playwright which will warn you about missing awaits and some bad patterns.
Also the first error  ` Test timeout of 30000ms exceeded while tearing down "context".` could be caused by this too but I'm not too sure. At the moment I don't see other orange/red flags.
```sh

  5) [chromium] › msw-api\setup-worker\start\wait-until-ready.test.ts:25:5 › defers network requests until the worker is ready

    Test timeout of 30000ms exceeded.

    page.waitForResponse: Page closed

      40 |
      41 |   const [numbersResponse, lettersResponse] = await Promise.all([
    > 42 |     page.waitForResponse((res) => {
         |          ^
      43 |       return res.url().includes('/numbers')
      44 |     }),
      45 |     page.waitForResponse((res) => {
```

⬆️ Will investigate a bit on this one too

kettanaito:
Hey, @shwarcu! Thanks for the investigation here!

I've merged the `fullyParallel` proposal to this branch, I think it makes perfect sense in the light of what we're trying to achieve here. The CPU shenanigans were interesting to learn about, we may consider enabling the GPU acceleration and compare the results. 

> This one may be caused by missing await before page.evaluate

Maybe. I need to emphasize that sometimes I intentionally omit awaiting `page.evaluate` to trigger an action but await it later on (e.g. with `page.waitForResponse()`). If you await evaluation, you make `.waitForResponse()` flaky as it will not resolve if the response has already occurred before this method was called.

kettanaito:
I was testing the `fullyParallel `changes and it seemed to have improved the stability of tests quite a lot! Alas, some tests are still flaky, mainly:

- `scenarios/errors/network-error.test.ts`

```
  1) [chromium] › msw-api/setup-worker/scenarios/errors/network-error.test.ts:58:5 › propagates a CORS violation error from a non-matching request 

    Error: expect(received).toEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 1

      Array [
    +   "[MSW] Caught an exception from the \"POST http://localhost:8080/graphql\" request (TypeError: Failed to fetch). This is probably not a problem with Mock Service Worker. There is likely an additional logging output above.",
        "[MSW] Caught an exception from the \"GET http://127.0.0.1:50438/resource\" request (TypeError: Failed to fetch). This is probably not a problem with Mock Service Worker. There is likely an additional logging output above.",
```

Which may suggest some shared state as well... Need to look into this one. 

Yeah, it's a leaking state. The network error test doesn't even do any requests to `/graphql` at all! That console array is not properly reset between tests, and so an error from one test leaks to an unrelated test, failing it. 

So far this has been happening only with the `workerConsole`.

shwarcu:
@kettanaito I'm super glad I could already help a bit!

> Maybe. I need to emphasize that sometimes I intentionally omit awaiting page.evaluate to trigger an action but await it later on (e.g. with page.waitForResponse()). If you await evaluation, you make .waitForResponse() flaky as it will not resolve if the response has already occurred before this method was called.

Yeah I was a bit confused about missing await and wasn't sure if it was intentional, thanks a lot for explaining 😄 

I will need more time to dig deeper in this branch code and to debug failing tests too. At the moment I'm thinking about another **hypothetical** low hanging fruits when it comes to tests performance (maybe also stability?):
#### Using `esbuild-loader` instead of `ts-loader` in webpack http server 
Out of pure curiosity I have run browser tests with `ts-loader` and with `esbuild-loader` and based on my very rough comparison results were slightly better for esbuild


| Loader         | Run 1           | Run 2          | Run 3           | Run 4           | Run 5          | Run 6 |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |------------- |
| ts-loader      | 82.16s         | 81.36s          | 80.84s          | 82.13s          | 78.92s         | 81.54s         |
| esbuild-loader | 78.99s     | 77.51s          | 80.63s          | 77.72s          | 78.05s         | 79.39s         |

Results could be within margin of error. Probably more reliable results can be gathered with [hyperfine](https://github.com/sharkdp/hyperfine). I will do it maybe tomorrow. However even if change is small it will start making a difference in a bigger scale - when tests are run frequently or when tests suite grows.

I allowed myself a little personal trip here as I really enjoy  ⚡ _b l a z i n g l y   f a s t_ ⚡ tools like esbuild and I was very curious how it would do it this unusual testing setup 😄 
[Changes required to use esbuild-loader are smol](https://github.com/mswjs/msw/pull/1534)

#### Running webpack http server once for entire run

Starting and stopping separate servers for each test case **might be** costly (there are many tests that use it). So maybe it would be worth a shot to see how test run would perform if we start http server once at the beginning, and serve all test-files under corresponding paths. And then close the server after entire test run ends.
There is configuration option in Playwright to run web server https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests

If there's some flakiness is coming from custom setup/teardown in fixtures then maybe running one global web server is one of solutions.
In this case I'm wondering if tests would still remain self-contained and not leak state between each other. I think that if we run them in separate browser context we're probably safe (?). I mean, is there a chance that `WebpackHttpServer` will influence tests?

Let me know what do you think and if you want me to verify these 2 hypothetical improvements 👋

kettanaito:
We can consider `esbuild-loader`, no hard opinion on that. The performance difference is not that impressive but I found ts-loader to behave strangely in some corner cases. 

> So maybe it would be worth a shot to see how test run would perform if we start http server once

That's how it's intended to behave. I definitely misconfigured things if it spawns the Webpack server on each test. The whole point of that server is that it's spawned once, and it supports runtime compilations, returning unique compilation URLs so compilations are scoped. Let's take a look if it's spawned on every test...

kettanaito:
I'm spawning the Webpack server with `{ scope: 'worker' }`, which, I believe, means that this is performed only _once_. But now when I'm thinking about it... Does it mean once _per worker_? So in a multi-threaded test run that'd mean multiple spawns of the Webpack server.

kettanaito:
I'm rewriting the Worker Console setup to be a router on the compilation route so that it's scoped to the compilation vs being a global thing. That does seem to help with the flakiness, just takes time to finish...

kettanaito:
@shwarcu, so I've tried spawning the webpack server once but the issue is that the tests need a _reference_ to that server. It's used to perform compilations (`server.compile()`) and also to modify compilation's behavior on the test's runtime, such as appending new routes, overriding what gets served as the worker script, etc. 

There doesn't seem to be a way to create server _once_ and then expose it to tests. Playwright's configs run in a separate process, and I've come to believe that even tests are isolated as I cannot use a singleton pattern—the value of persisted server is always undefined across tests. 

I think I can technically split the compilation server and the preview server so that we use a single compilation server to save up resources but spawn a new preview server for each test but that's an overkill given that the compilation server already does a great job at scoping compilations and serving them like a preview (its entire point). 

Do you happen to know if we can achieve what I've described above?

kettanaito:
What I decided to do in the end is to spawn 1 compilation server per 1 worker. I approached it this way: tests are completely isolated within the worker process so there's no way to have a shared state between workers. Within a single worker, however, I will establish _a single compilation server_ to save resources. 

I've given this a quick test and can confirm that this setup works. Now testing the flakiness...

kettanaito:
Sadly, even after rewriting the Worker console logic and ensuring the webpack server is scoped per worker, the tests are still flaky, timing out on the "context" error. Huh. 

@shwarcu, what do you think we could look at as the next step?

shwarcu:
Hi @kettanaito 👋 I need more time for 'decompression' after work and I won't be looking at the code today and tomorrow. **But I will continue my engagement in this topic and I will get back to you once I have something valuable to say (probably closer to Saturday/Sunday)**

kettanaito:
Absolutely no worries! Please make sure to reset properly. Work can be daunting but a good chunk of rest does wonders. See you around!

kettanaito:
I went and improved a bit on how we handle runtime compilations. I've added the ability to dispose of the compilation once the test settles so they don't take memory anymore:

- Kill the underlying webpack compiler;
- Remove any in-memory records of the compilation that `webpack-http-server` keeps internally.

I don't know if that helped or not but I've got the first CI passing since forever. Will retry it to confirm... 


The tests do seem more reliable now. I had 4 runs, 3 succeeded, but the 4th one failed with:

```js
  1) [chromium] › msw-api/setup-worker/scenarios/iframe/iframe.test.ts:40:5 › intercepts a request from a deeply nested iframe 
error Command failed with exit code 1.

    Test timeout of 10000ms exceeded.

    frame.waitForSelector: Target closed
    =========================== logs ===========================
    waiting for locator('#first-name') to be visible
    ============================================================

      57 |
      58 |   await deepFrame.evaluate(() => window.request())
    > 59 |   const firstNameElement = await deepFrame.waitForSelector('#first-name')
         |                                            ^
      60 |   const firstName = await firstNameElement.evaluate((node) => node.textContent)
      61 |
      62 |   expect(firstName).toBe('John')

        at /Users/runner/work/msw/msw/test/browser/msw-api/setup-worker/scenarios/iframe/iframe.test.ts:59:44
```

These iframe tests are dependent on file system, so I suspect there's an issue where two tests try to write to the same directory (although it should be impossible now given the dir name is based on the worker id + directory name). 

Well, at least it seems that disposing of webpack instances helped with the other tests.

kettanaito:
As the next step, I've skipped the iframe tests to verify whether they are the only ones left failing. Will re-run the CI multiple times to confirm that.

kettanaito:
Alas, the tests are still flaky so I moved on with adding some additional debugging to them. Now each failed test will flush its runtime console state to a log file and upload it as an artifact to the CI. This should help us debug flaky tests since screenshots and videos are pretty much useless given the way we write things in the console primarily.

kettanaito:
So based on https://github.com/mswjs/msw/actions/runs/4148868358, which has failing tests with no logs whatsoever, I suspect the issue is in asynchronicity/servers/etc. Quite a pickle then, I was hoping for some browser-side reporting to help but it looks like I need to dive into why PW/webpack server has trouble doing what it's designed to do.

kettanaito:
## Released: v1.1.0 🎉

This has been released in v1.1.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v1.1.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/1.1.0)

---

Issue #1267, opened 2022-06-01
Can't resolve '@mswjs/interceptors/lib/interceptors/ClientRequest' 


ntucker:
### Prerequisites

- [X] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [X] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [X] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [X] I'm using Node.js version 14 or higher

### Browsers

NODE 16/18

### Reproduction repository

https://github.com/coinbase/rest-hooks/tree/a3963b2618981992bab115a8e8543abdb9cdc655/website

Gonna revert MSW upgrade to make this work so master will no longer have this problem.

### Reproduction steps

1. clone repo above
2. git checkout a3963b2618981992bab115a8e8543abdb9cdc655
3. navigate to website folder
4. yarn install
5. yarn start
6. observe error in cli as well as opened browser tab

### Current behavior

```bash
ERROR in ./node_modules/msw/lib/node/index.mjs 45:0-94

Module not found: Error: Can't resolve '@mswjs/interceptors/lib/interceptors/ClientRequest' in '/home/ntucker/src/rest-hooks/website/node_modules/msw/lib/node'
Did you mean 'index.js'?
BREAKING CHANGE: The request '@mswjs/interceptors/lib/interceptors/ClientRequest' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

https://github.com/mswjs/msw/pull/1247 breaks in node versions that support ESM

### Expected behavior

To fix this - the ESM version of MSW must be explicit about file extensions. Otherwise it is by definition incorrectly built. If this cannot be done you should not ship a custom ESM version as this is incorrect. The only reason it would work is if it is used by a system that doesn't enforce fully (webpack). However this will break in fully-spec compliant things like node.

More simply:

`import '@mswjs/interceptors/lib/interceptors/ClientRequest'` => `import '@mswjs/interceptors/lib/interceptors/ClientRequest.mjs'`

kettanaito:
Hey, @ntucker. Thanks for reporting this. 

[...]

PR #1395, opened 2022-09-07
fix: correctly link types in `package.json` `exports` field

ivanhofer:
Fixes an issue where type annotations would be broken on Node 16 (TS 4.7-4.8).


package.json
```diff
@@ -9,9 +9,11 @@
       "default": "./lib/index.js"
     },
     "./native": {
+      "types": "./lib/native/index.d.ts",
       "default": "./lib/native/index.js"
```
kettanaito, line 13: Do you encounter an issue that these types aren't being picked up?

We have the typings mapped explicitly in respective `package.json` files:

https://github.com/mswjs/msw/blob/6931c0ecadca5d4b156c1299bb402f074a3d1a9a/node/package.json#L4

https://github.com/mswjs/msw/blob/6931c0ecadca5d4b156c1299bb402f074a3d1a9a/native/package.json#L4

ivanhofer: It doesn't look like typescript uses the nested `package.json` file. I quickly tested the solution above with a `.mjs` file and typescript correctly resolves everything.

kettanaito: That's interesting. Perhaps #1383 affected how TS sees things. This would be great to turn into an integration test so we could catch broken typings. 

We do have some typings tests but, apparently, they didn't catch this issue?

https://github.com/mswjs/msw/blob/6931c0ecadca5d4b156c1299bb402f074a3d1a9a/test/typings/index.test-d.ts#L2-L3

I'd expect these imports to yield `any` if the typings map is broken, failing the typings tests.

ivanhofer: Well, those tests can pass or fail depending on the settings inside `tsconfig.json`. 
I can try a few things over the weekend to see in what scenarios it is woorking and where not.

ivanhofer: I have added some tests that test the moduleResultion of node 16.
I was wrong before. This PR does not fix #1267, but it fixes using `msw/node` in a `.mts` file

ivanhofer:
I'm not sure what to do about the failing tests.. Older versions don't have "Node16" as an option

kettanaito:
Looks like we need a way to run `test:ts` with specific parameters based on which TS version is currently being run. 

## Proposal

Let's add support for `tsconfig.[TS_VERSION].json` files under `test/typings`. If a config file for the current TS version is present, it will be used. If it's not present, the default `test/typings/tsconfig.json` will be used as a fallback. 

To achieve this, we can move the typings tests logic into a custom script.

```ts
// test/typings/run.ts
import * as fs from 'fs'
import * as path from 'path'
import { spawnSync } from 'child_process'
import { invariant } from 'outvariant'
import tsPackageJson from 'typescript/package.json'

const tsVersion = tsPackageJson.version
invariant(tsVersion, 'Failed to run typings tests: unable to determine TypeScript version')

const tsConfigPaths = [
  path.resolve(__dirname, `tsconfig.${tsVersion}.json`),
  path.resolve(__dirname, 'tsconfig.json')
]
const tsConfigPath = tsConfigPaths.find((path) => fs.existsSync(path))

spawnSync('yarn', ['test:ts', tsConfigPath], {
  cwd: path.resolve(__dirname, '../..')
})
```

Then, you'd be able to create custom TS configurations for targets 4.2 - 4.6. There'd be no need to repeat yourself as we can write a single "old" config and use it for other compatible targets:

```json
// tsconfig.4.4.json
{
  "extends": "tsconfig.4.2.json"
}
```

@ivanhofer, does this sound interesting to you? Or do you want me to add this once I have time?

ivanhofer:
I have some free time next week. I can take a look at this

kettanaito:
Thank you! Don't hesitate to ping me in this thread for any help. Excited for this.

ivanhofer:
@kettanaito I have fixed the tests like you have suggested. I had to add `tsx` as a `devDependency` to be able to execute the `run.ts` script.
I could not make it work with `ts-node`, that is already included in the repo. But it looks like that dependency is not getting used anywhere..

package.json
```diff
@@ -143,6 +145,7 @@
     "ts-loader": "^9.2.6",
     "ts-node": "^10.1.0",
     "tsup": "^5.12.8",
+    "tsx": "^3.9.0",
```
kettanaito: Can you please share some more details about why `tsx` was needed over `ts-node`?

ivanhofer: I don't exactly remember. But I'm tired of debugging `ts-node` stuff, so whenever I encounter a problem with it, I just replace the command with `tsx`. Mostly it just works :)
If you want to investigate, you can run `ts-node test/typings/run.ts`.

kettanaito: Yeah, I'll try using `ts-node` locally and see if there are any issues. I'd not introduce new dependencies unless it's really necessary. I haven't got any issues with `ts-node` in the past, would like to learn about this one.

kettanaito: I've tried `ts-node`. There were a few issues that indicated issues with our `test/typings/tsconfig.json`. I've fixed those in 702147f and now the typings tests are passing without issues. I've reverted it to use `ts-node` and deleted `tsx`.

kettanaito (approved):
I simply love these automations! 😍 
Thank you so much for fixing the issue _and_ improving our testing setup, @ivanhofer. 

Let's wait for the build status and merge this.

package.json
ivanhofer: Thanks for investigating and fixing it!

kettanaito:
## Released: v0.47.4 🎉

---

PR #1399, opened 2022-09-11
feat: makes the library esm-compatible

### LINKED ISSUE #1267: Can't resolve '@mswjs/interceptors/lib/interceptors/ClientRequest'

ivanhofer:
fixes #1267

I have investigated #1267 a bit. There are a few road blocks to make `msw/node` work in a `.mjs` file:
 - [x] reference `lib/index.js` instead of `'lib'` in a few files (easy to achive)
 - [x] ~`tsup` does not support import assertions (needed here https://github.com/mswjs/msw/blob/main/src/context/status.ts#L1) https://github.com/egoist/tsup/issues/714~
    workaround: https://github.com/mswjs/msw/pull/1399#discussion_r968360168
 - [x] ~dynamic import of `node-fetch` (https://github.com/mswjs/msw/blob/main/src/context/fetch.ts#L6) is not supported in esm context. https://github.com/evanw/esbuild/issues/1921~
    workaround: https://github.com/mswjs/msw/pull/1399#issuecomment-1243681143

### DIFF
--- a/src/context/status.ts
+++ b/src/context/status.ts
@@ -1,4 +1,4 @@
-import statuses from 'statuses/codes.json' assert { type: 'json' }
+import statuses from 'statuses/codes.json'
 import { ResponseTransformer } from '../response'
 
 /**
--- a/src/context/status.ts
+++ b/src/context/status.ts
@@ -1,4 +1,4 @@
-import statuses from 'statuses/codes.json'
+import statuses from 'statuses/codes.json' assert { type: 'json' }
 import { ResponseTransformer } from '../response'
 
 /**
--- a/src/node/setupServer.ts
+++ b/src/node/setupServer.ts
@@ -1,5 +1,5 @@
-import { ClientRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/ClientRequest'
-import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/XMLHttpRequest'
+import { ClientRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/ClientRequest/index.js'
+import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/XMLHttpRequest/index.js'
 import { createSetupServer } from './createSetupServer'
 
 /**
--- a/package.json
+++ b/package.json
@@ -100,7 +100,6 @@
     "node-fetch": "^2.6.7",
     "outvariant": "^1.3.0",
     "path-to-regexp": "^6.2.0",
-    "statuses": "^2.0.0",
     "strict-event-emitter": "^0.2.0",
     "type-fest": "^2.19.0",
     "yargs": "^17.3.1"
@@ -139,6 +138,7 @@
     "regenerator-runtime": "^0.13.9",
     "rimraf": "^3.0.2",
     "simple-git-hooks": "^2.8.0",
+    "statuses": "^2.0.0",
     "ts-jest": "26",
     "ts-loader": "^9.2.6",
     "ts-node": "^10.1.0",
--- a/src/context/fetch.ts
+++ b/src/context/fetch.ts
@@ -3,7 +3,12 @@ import { Headers } from 'headers-polyfill'
 import { MockedRequest } from '../utils/request/MockedRequest'
 
 const useFetch: (input: RequestInfo, init?: RequestInit) => Promise<Response> =
-  isNodeProcess() ? require('node-fetch') : window.fetch
+  isNodeProcess()
+    ? (input, init) =>
+        import('node-fetch').then(({ default: nodeFetch }) =>
+          (nodeFetch as unknown as typeof window.fetch)(input, init),
+        )
+    : window.fetch
 
 export const augmentRequestInit = (requestInit: RequestInit): RequestInit => {
   const headers = new Headers(requestInit.headers)
--- a/src/node/createSetupServer.ts
+++ b/src/node/createSetupServer.ts
@@ -1,4 +1,5 @@
-import { bold } from 'chalk'
+import chalk from 'chalk'
+const { bold } = chalk
 import { isNodeProcess } from 'is-node-process'
 import { StrictEventEmitter } from 'strict-event-emitter'
 import {
--- a/src/utils/matching/matchRequestUrl.ts
+++ b/src/utils/matching/matchRequestUrl.ts
@@ -1,5 +1,5 @@
 import { match } from 'path-to-regexp'
-import { getCleanUrl } from '@mswjs/interceptors/lib/utils/getCleanUrl'
+import { getCleanUrl } from '@mswjs/interceptors/lib/utils/getCleanUrl.js'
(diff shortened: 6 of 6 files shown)


src/context/status.ts
```diff
@@ -1,4 +1,4 @@
-import statuses from 'statuses/codes.json'
+import statuses from 'statuses/codes.json' assert { type: 'json' }
```
ivanhofer: this will currently be removed on build

mattcosta7: we could also configure `tsup` to inline this library by moving it to `devDependencies` instead of listing in dependencies

src/node/setupServer.ts
```diff
@@ -1,5 +1,5 @@
-import { ClientRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/ClientRequest'
-import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/XMLHttpRequest'
+import { ClientRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/ClientRequest/index.js'
+import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/lib/interceptors/XMLHttpRequest/index.js'
```
mattcosta7, line 2: we should really just update interceptors to expose an esm build.  

I started on that over the weekend, and have some thoughts on how we could do that, but I think we'll need to probably remove `debug` as a dependency there - since it relies on `tty` and `os`, and i'm having issues getting those properly bundled to a browser compat version, that multiple build tools like.

src/handlers/GraphQLHandler.test.ts
```diff
@@ -3,7 +3,7 @@
  */
 import { encodeBuffer } from '@mswjs/interceptors'
 import { OperationTypeNode, parse } from 'graphql'
-import { Headers } from 'headers-polyfill/lib'
+import { Headers } from 'headers-polyfill/lib/index.js'
```
mattcosta7: When I get some time, exposing an esm build of headers-polyfill should be pretty quick, and might be better than importing from lib directly here

mattcosta7:
> dynamic import of node-fetch

I think we could get around this with something like

```
useFetch = isNodeProcess() ? (input, init) => import('node-fetch').then(({default: nodeFetch}) => fetch(input, init)) : window.fetch
```

I'm curious if we could also avoid using node-fetch directly in this library and just stipulate that `globalThis.fetch` be defined already (via a user polyfill) instead?

src/handlers/GraphQLHandler.test.ts
ivanhofer: Yes this would be the cleanest solution. Until then this would be an easy workaround

ivanhofer:
With this branch it is now poosible to run the following `esm` file:

_test.mjs_
```js

setupServer()
```

with the command

`node ./test.mjs`

src/handlers/GraphQLHandler.test.ts
mattcosta7: just opened https://github.com/mswjs/headers-polyfill/pull/38

kettanaito: `headers-polyfill@3.1` ships with ESM support now! Thank you, @mattcosta7! 🎉 

I've updated `@mswjs/interceptors` to use it, now updating MSW as well.

src/node/setupServer.ts
kettanaito: We need to look into some `debug` recipes perhaps. It has browser support, so it should run in the browser. Perhaps we can do some import map for this dependency to ensure that in the browser it always imports the browser dist.

mattcosta7: Yeah, it mostly relates to the fact that `debug` doesn't have an esm build, so loading in a browser directly from `@web/dev-server` seems problematic.  

Maybe that just means users in those environments have to do some extra setup, and that's enough.  Will keep playing with configurations there, because it's also possible we don't need to be as explicit there.

I was testing using tsup to bundle, however using tsc to generate cjs and esm transpilations might help avoid issues related to that (since it won't try to handle debug directly)

src/handlers/GraphQLHandler.test.ts
ivanhofer: I just did a rebase

filipw01:
Nice, work. I was about to open an issue, but it looks like it should be fixed once this PR is merged 😄

ivanhofer:
I think this PR is ready to be merged. Or is something still missing?

mataspetrikas:
@kettanaito @ivanhofer is this coming into the next release?

kettanaito (approved):
Finally got to this improvement. Looks superb 🎉 Updated against the latest main, waiting for the CI before marking it as a release candidate. Expect to land in the next minor release. 

Thank you for improving MSW, @ivanhofer! If you ever feel like exercising your skill again, we could also migrate the [interceptors](https://github.com/mswjs/interceptors) library to ESM.

kettanaito:
The only thing I'd add is an automated test for the distributed ESM format. Let me see if I can dig out the previous test I've written...

kettanaito:
## Released: v0.48.0 🎉

This has been released in v0.48.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.48.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.48.0)

---

PR #1417, opened 2022-09-26
fix: include error stack in log when handling uncaught exceptions

jkieboom:
Fixes #1416.

### DIFF
--- a/src/setupWorker/start/createRequestListener.ts
+++ b/src/setupWorker/start/createRequestListener.ts
@@ -105,7 +105,7 @@ export const createRequestListener = (
 This exception has been gracefully handled as a 500 response, however, it's strongly recommended to resolve this error, as it indicates a mistake in your code. If you wish to mock an error response, please see this guide: https://mswjs.io/docs/recipes/mocking-error-responses`,
           request.method,
           request.url,
-          error,
+          error.stack ?? error,
         )
 
         // Treat all other exceptions in a request handler as unintended,
--- a/test/msw-api/exception-handling.test.ts
+++ b/test/msw-api/exception-handling.test.ts
@@ -28,4 +28,13 @@ test('transforms uncaught exceptions into a 500 response', async () => {
       'ReferenceError: nonExisting is not defined',
     ),
   })
+
+  const errors = runtime.consoleSpy.get('error')
+
+  expect(errors).toEqual(
+    expect.arrayContaining([
+      expect.stringContaining('ReferenceError: nonExisting is not defined'),
+      expect.stringContaining('  at '),
+    ]),
+  )
 })


src/setupWorker/start/createRequestListener.ts
```diff
@@ -105,7 +105,7 @@ export const createRequestListener = (
 This exception has been gracefully handled as a 500 response, however, it's strongly recommended to resolve this error, as it indicates a mistake in your code. If you wish to mock an error response, please see this guide: https://mswjs.io/docs/recipes/mocking-error-responses`,
           request.method,
           request.url,
-          error,
+          error.stack ?? error,
```
kettanaito, line 108: Would this not print the `error` if its stack exists?

I thought we may add a new `console.error(error)` call below this `devUtils.error()` to forward the error to the console. What do you think about this?

test/msw-api/exception-handling.test.ts
```diff
@@ -28,4 +28,8 @@ test('transforms uncaught exceptions into a 500 response', async () => {
       'ReferenceError: nonExisting is not defined',
     ),
   })
+
+  const consoleError = runtime.consoleSpy.get('raw').get('error').join('\n')
```
kettanaito: You can use `consoleSpy.get('error')` straight away to retrieve a list of errors. Then, you can write an assertion that checks if a certain phrase occurs in that list:

```ts
const errors = runtime.consoleSpy.get('error')
expect(errors).toEqual(expect.arrayContaining([
  expect.stringContaining('ReferenceError: ...'),
  // By the way, not sure if this is not a brittle assertion
  // as it relies on implementation details of the "RestHandler" class.
  // Perhaps we can rephrase this so it'd be abstract. 
  expect.stringContaining('at RestHandler.run')
]))
```

src/setupWorker/start/createRequestListener.ts
jkieboom: This prints the error (which will just be name + message) if the stack is undefined. The stack actually also includes the name and message already. Not sure about `console.error` vs `devUtils.error`. You have to tell me what is more appropriate. For me it fits will with displaying the error directly within this message and not as a separate disconnected thing on another log.

test/msw-api/exception-handling.test.ts
jkieboom: Ah thanks, I'm still getting used to `expect`. Wrt the rest handler, yes that's brittle. I was not 100% sure, but I've changed it to just `"  at "` which is a bit generic, but I think works for this particular case.

src/setupWorker/start/createRequestListener.ts
kettanaito: Oh, so the issue was that at some point there are errors that don't have a `stack`? And those were somehow ignored?

jkieboom: No the issue is that `Error.toString()` only includes the `name` and `message`, never the `stack`. Per spec, `stack` on `Error` is optional, so it's best to guard against it being `undefined` (although I don't know when this could happen in practice)

kettanaito: Got it. Thanks for the explanation!

kettanaito (approved):
Thank you for making MSW better, @jkieboom! Welcome to contributors 🎉

kettanaito:
## Released: v0.47.4 🎉

---

PR #1420, opened 2022-09-30
chore: fixes tests on Windows using Unix-style path

kleinfreund:
## Description

I noticed that tests weren't passing on Windows and went ahead and fixed them. All unit and integration tests are now passing on Windows, too.

**Note**: This also contains a controversial change to the commitlint.config.js which **removes** the line length limitations from commit bodies. I've made that change because I wasn't able to commit. Of course this isn't my repository so I'm happily removing this change if it's unwanted. However, I'd like to take this opportunity to argue against such rules: Nowadays, there just isn't a good reason for hard-wrapping code or natural language content (which I believe is the main reason these rules exist). Editors have capabilities to soft-wrap as a user likes per file type/language. Additionally, I've come to really like writing up a commit's changes in ... the commit itself. This saves from needing to do so in two places: the commit and the inadvertent pull/merge request.

In any case, this is not for me to decide, so please let me know whether I should be reverting this.

## Changes

Fixes a test passing a Unix-style path to child_process.exec which failed on Windows by wrapping it in `path.resolve`.

Fixes a test relying on some internal logic which only took Unix-style paths into account.

Changes the commitlint configuration to allow commit bodies of any length.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,6 +1,7 @@
 module.exports = {
   extends: ['@commitlint/config-conventional'],
   rules: {
+    'body-max-line-length': [0, 'always', Infinity],
     'footer-max-line-length': [1, 'always'],
   },
 }
--- a/src/utils/internal/getCallFrame.ts
+++ b/src/utils/internal/getCallFrame.ts
@@ -1,5 +1,5 @@
 // Ignore the source files traces for local testing.
-const SOURCE_FRAME = /\/msw\/src\/(.+)/
+const SOURCE_FRAME = /[\/\\]msw[\/\\]src[\/\\](.+)/
 
 const BUILD_FRAME =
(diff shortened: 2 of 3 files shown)


kleinfreund:
One thing that I noticed is that the following code:

https://github.com/11ty/webc/blob/9ce79dbb3430e80a0db0f289a2b98041b14d08c6/src/ast.js#L221

won't find children of document fragments which is the case for the scenario described in this issue. The `template` node's `childNodes` will be empty. Instead, its children will be on `content.childNodes`.

I've tried to conditionally switch to a document fragment's children to address this like so:

```javascript
let childNodes = rootTagName === 'template' && root.childNodes.length === 0 ? root.content.childNodes : root.childNodes
```

This gets us further, but the slot is still not being replaced.

kettanaito (approved):
Thanks for making the tests better! Welcome to contributors, @kleinfreund! 🎉

kettanaito:
## Released: v0.47.4 🎉

---

PR #1430, opened 2022-10-10
fix: prevent corruption of binary request bodies (#929)

nknapp:
Use `request.arrayBuffer` in worker-message rather than `request.text()`. This prevents re-encoding and the resulting corruption

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -247,7 +247,7 @@ async function getResponse(event, client, requestId) {
       redirect: request.redirect,
       referrer: request.referrer,
       referrerPolicy: request.referrerPolicy,
-      body: await request.text(),
+      body: await request.arrayBuffer(),
       bodyUsed: request.bodyUsed,
       keepalive: request.keepalive,
     },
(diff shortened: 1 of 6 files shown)


nknapp:
The fix will follow in the next commit. But I want to see the pipeline fail first.

kettanaito:
Hi, @nknapp! Thanks for opening this fix. It reminds me largely about a similar change I took on reading request/response bodies as buffer in here https://github.com/mswjs/msw/pull/1332/. May I ask for your fresh eyes to see if our approaches differ? Are we, perhaps, missing something? 🤔

nknapp:
Oh, I didn't see that one. But you are right. It basically contains the same changes, and some more.

That is, except for the integration tests that I wrote.

Feel free to cherry pick those changes, or use it as template for your own test. The bytes that I sent in the test are the first bytes of a JPEG file.

kettanaito:
Thank you! Will do. I recall some difficulty when integrating those changes in the past. Need to revisit that pull request to see what's going on.

kettanaito:
We are transferring the `ArrayBuffer` from the worker now since 2.0:

https://github.com/mswjs/msw/blob/5c5470a0f2204b4164fcc65e210471661208ce4d/src/mockServiceWorker.js#L218

This is fixed in 2.0.

Issue #1442, opened 2022-10-22
Request body with Protobuf binary data is corrupted


DGollings:
### Prerequisites

- [X] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [X] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [X] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [X] I'm using Node.js version 14 or higher

### Browsers

Chromium (Chrome, Brave, etc.), Firefox

### Reproduction repository

not easily possible

### Reproduction steps

Good news, I saw some issues with people having problems(?) mocking binary protobuf messages. I can confirm it works though for Twirp (which is like gRPC but simplified). Simply encode and decode with a properly set header

Well, mostly:
My input
```js
const input = new Uint8Array([138, 1, 6, 10, 4, 10, 2, 32, 1]);
```

I can encode/decode without issues before sending to msw, but in the mock itself it turns into either:
```js
const usingArrayBuffer = new Uint8Array(await req.arrayBuffer());
value usingArrayBuffer = Uint8Array([239, 191, 189, 1, 6, 10, 4, 10, 2, 32, 1]);
  ```

or with a little (unsuccessful) hacking:
```js
const usingText = new Uint8Array((await req.text()).split("").map(x => x.charCodeAt( 0 )));
value usingText = [ 253, 1, 6, 10, 4, 10, 2, 32, 1 ]
```

`usingText[0] = 138` 'fixes it', but unfortunately doesn't work in all cases

Notice the values `[239, 191, 189, ...]`, that's a [unicode replacement character](https://unicodeplus.com/U+FFFD). As in, the arrayBuffer has been at some point converted from raw data to a string and 'back' (?)

A likely important note: I use `msw-storybook-addon`

EDIT: whilst writing this I think I realize the problem. That addon has version 
```
  "peerDependencies": {
    "msw": ">=0.35.0 <1.0.0"
  },
```
set. Which doesn't contain the fixes I think I need.

I'll likely have to open an issue there. But leaving this here for completeness/cross reference

### Current behavior

arrayBuffer is stringified

### Expected behavior

arrayBuffer is completely untouched

DGollings:
Related issue:https://github.com/mswjs/msw/issues/929

DGollings:
[...]

PR #1443, opened 2022-10-22
test: add "application/protobuf" handling test

DGollings:
Added test which mimicks the behaviour I see in issue https://github.com/mswjs/msw/issues/1442

Simply returns the request, which should remain completely unchanged whilst moving through msw


kettanaito:
We need to re-assess this test in the context of the recent MSW 2.0 release. I have a suspicion that this will no longer be an issue. I will rebase the branch and let's see.

test/node/rest-api/request/body/body-protobuf.node.test.ts
```diff
@@ -0,0 +1,38 @@
+// @vitest-environment node
+import { http, HttpResponse } from 'msw'
+import { setupServer } from 'msw/node'
+
+const server = setupServer(
+  http.post('https://example.com/protobuf', async ({ request }) => {
+    const buffer = await request.arrayBuffer()
+
+    return new HttpResponse(new Uint8Array(buffer), {
```
kettanaito, line 9: @DGollings, hi 👋 Do you think it's necessary to preserve this conversion here or should be simply forward the `buffer` as-is to the response? It'd also change how we assert the actual response body sent back. 

In any case, the test is now passing.

kettanaito (approved):

test/node/rest-api/request/body/body-protobuf.node.test.ts
DGollings: I honestly don't remember why that conversion is there, maybe to be slightly more realistic to actual usage? You generally don't respond with the exact request as-is (I think? Its been over a year)

kettanaito:
## Released: v2.0.7 🎉

---

PR #1445, opened 2022-10-26
feat: add "SetupApi" base class

### LINKED ISSUE #1428: Support Edge Runtime (no http module)
### Scope

Improves an existing behavior

### Compatibility

### Feature description

NextJS has introduced an [Edge Runtime](https://nextjs.org/docs/api-reference/edge-runtime) ([more docs here](https://edge-runtime.vercel.app/), which is a stricter node.js runtime based on web standards to run "on the edge". This is used, for example, but their [middleware](https://nextjs.org/docs/advanced-features/middleware). 

I am performing user authentication in this middleware, and would like to mock the `fetch` requests I make in my middleware. This means I need MSW to be compatible with the edge runtime and support intercepting fetch requests made there.

I have cloned the with-msw example and added a `middleware.ts` file which makes a simple `fetch` request. The failure can be seen in my repository: [with-msw-app](https://github.com/kristojorg/with-msw-app)

The error I see when trying to run this is:
```
error - unhandledRejection: Error: The edge runtime does not support Node.js 'http' module.
Learn More: https://nextjs.org/docs/messages/node-module-in-edge-runtime
    at Object.get (webpack-internal:///(middleware)/./node_modules/next/dist/server/web/adapter.js:172:19)
    at __extends (webpack-internal:///(middleware)/./node_modules/@mswjs/interceptors/lib/interceptors/ClientRequest/NodeClientRequest.js:15:73)
    at eval (webpack-internal:///(middleware)/./node_modules/@mswjs/interceptors/lib/interceptors/ClientRequest/NodeClientRequest.js:102:5)
    at eval (webpack-internal:///(middleware)/./node_modules/@mswjs/interceptors/lib/interceptors/ClientRequest/NodeClientRequest.js:441:2)
    at Object.(middleware)/./node_modules/@mswjs/interceptors/lib/interceptors/ClientRequest/NodeClientRequest.js (evalmachine.<anonymous>:161:1)
    at __webpack_require__ (evalmachine.<anonymous>:37:33)
    at fn (evalmachine.<anonymous>:281:21)
    at eval (webpack-internal:///(middleware)/./node_modules/@mswjs/interceptors/lib/interceptors/ClientRequest/http.get.js:25:27)
    at Object.(middleware)/./node_modules/@mswjs/interceptors/lib/interceptors/ClientRequest/http.get.js (evalmachine.<anonymous>:172:1)
    at __webpack_require__ (evalmachine.<anonymous>:37:33)
error - Error [TypeError]: fetch failed
    at Object.processResponse (evalmachine.<anonymous>:3163:27)
    at <unknown> (evalmachine.<anonymous>:3507:42)
    at <unknown> (node:internal/process/task_queues:141:7)
    at AsyncResource.runInAsyncScope (node:async_hooks:199:9)
    at AsyncResource.runMicrotask (node:internal/process/task_queues:138:8)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  middleware: true
}
```

Toxiapo:
- Closes #1428 

Hi @kettanaito, I would like to hear your thoughts about the current setup and if I am on the right track. I am also curious if you mean to write the `SetupApi` as an interface and update the return types on `setupWorker` and `setupServer` or a base class and mimic the current logic into a class as I did here. However, this might be tricky to implement `interceptor.on('request')` for both the server and the worker.

We will have to create a custom setup specifically for Edge in the consuming app to consume the API.
```ts
import { SetupApi } from "msw"
import { FetchInterceptor } from '@mswjs/interceptor'

class SetupEdge extends SetupApi {
  constructor(handlers: RequestHandler[]) {
      super([ClientRequestInterceptor], handlers)
  }
  
   public listen(options: ...): void {
      // ...
      super.apply()
    }
  // ...
}

```


kettanaito:
Hey, @Toxiapo. I'm so excited to see this in the works! Thanks for working on the changes, I will try to get to them today/tomorrow to give some feedback. 

>  I am also curious if you mean to write the SetupApi as an interface and update the return types on setupWorker and setupServer or a base class and mimic the current logic into a class as I did here.

I was thinking of introducing a base class, then extending that class with specific `SetupServerApi` and `SetupWorkerApi` classes, and then returning their instances from the `setupServer()` and `setupWorker()` existing API to preserve the same public interfaces. I think a class is the best syntax to describe fixed life-cycles and behaviors, which is exactly what we're doing here.

src/createSetupApi.ts
```diff
@@ -0,0 +1,104 @@
...
+  public readonly events: LifeCycleEventEmitter<Record<string | symbol, any>>
+
+  constructor(
+    interceptors: {
+      new (): Interceptor<HttpRequestEventMap>
+    }[],
+    initialHandlers: RequestHandler[],
+  ) {
+    this.interceptor = new BatchInterceptor({
+      name: 'setup-api',
```
kettanaito: We would need to somehow parametrize the batch interceptor's `name` so it's different for each child class (`SetupServerApi`, `SetupWorkerApi`). This name affects how the interceptor stores global module overrides, and having multiple batch interceptors with the same name is a no-op.

src/node/setupServer.ts
```diff
@@ -1,15 +1,144 @@
...
+    this.resolvedOptions = {} as RequiredDeep<SharedOptions>
+
+    // TODO: Re-think this
+    this.init()
+  }
+
+  public async init(): Promise<void> {
+    const _self = this
+
+    this.interceptor.on('request', async function setupServerListener(request) {
```
kettanaito: We don't need to keep this function's context, so this can be an arrow function ridding us of the `_self = this` assignment.

kettanaito: I believe we've added this function simply to have a name to make these listeners easier to debug.

src/node/setupServer.ts
```diff
@@ -1,15 +1,144 @@
...
+    })
+  }
+
+  public close(): void {
+    super.dispose()
+  }
+}
+
+export const setupServer = (...handlers: RequestHandler[]) => {
+  handlers.forEach((handler) => {
```
kettanaito: I think this validation needs to live in `SetupApi` so that the input handlers are always validated in the same way (this behavior is the same for any child class).

src/createSetupApi.ts
kristojorg: We could pass the name in to the constructor or use a private property that the sub classes override (eg `this.name`) and 
```ts
class SetupServer extends SetupApi {
  private name = "setup-server"
  ...
}
```

If the name doesn't need to be changed except when defining the concrete classes, using the private property might be simplest?

src/node/setupServer.ts
Toxiapo: I can add this to the base, thanks!

src/node/setupServer.ts
Toxiapo: Ok! Thank you.

src/createSetupApi.ts
Toxiapo: I am also thinking of passing the name to the constructor. 
Another option is to just set the `interceptor` in the subclass since the interceptor will not be useful for the `worker` class I think.

kettanaito: I like the private property approach but afaik you cannot force that to be set via TypeScript, can you?

kettanaito:
Let me know when you would like a full review of this. Really exciting changes going on here 👀

src/createSetupApi.ts
Toxiapo: I was having trouble setting that as private property, so I did the constructor approach. I could make the property abstract to force the subclasses to define it. However, I wouldn't be able to use that property in the constructor, so I would have to create a helper function to initialize the interceptor. If you know a better way, please let me now, I would love to change it!

src/setupWorker/setupWorker.ts
```diff
@@ -1,238 +1,226 @@
...
+  }
 
-            if (!message) {
-              return
-            }
+  public initializeWorkerContext(): void {
+    this.context = {
+      // Mocking is not considered enabled until the worker
+      // signals back the successful activation event.
+      isMockingEnabled: false,
```
Toxiapo: This portion of the code is nearly untouched, I moved the full logic to a method for this new class. You will see that I had to recreate most of the base methods for the worker class because the logic for the worker was around the `this.context` object.

src/node/setupServer.ts
```diff
@@ -1,15 +1,137 @@
...
  * Sets up a requests interception in Node.js with the given request handlers.
  * @param {RequestHandler[]} requestHandlers List of request handlers.
  * @see {@link https://mswjs.io/docs/api/setup-server `setupServer`}
  */
-export const setupServer = createSetupServer(
-  // List each interceptor separately instead of using the "node" preset
-  // so that MSW wouldn't bundle the unnecessary classes (i.e. "SocketPolyfill").
-  ClientRequestInterceptor,
-  XMLHttpRequestInterceptor,
-)
```
Toxiapo, line 15: The `createSetupServer` function is not being used for the node server anymore, but it's still being used for `src/native` so I didn't delete the file.

test/msw-api/setup-server/printHandlers.test.ts
```diff
@@ -36,6 +36,7 @@ afterAll(() => {
 test('lists all current request handlers', () => {
   server.printHandlers()
 
+  // Test failed here, commenting so it shows up in the PR
   expect(console.log).toBeCalledTimes(6)
 
   expect(console.log).toBeCalledWith(`\
```
Toxiapo, line 42: I am struggling with these tests and can't figure out why!! I wonder if you have had this problem before and can give me some pointers on this.  😔

(It does seem like it's outputting the correct number of calls, however, the test results were printed in a batch instead of in sequence to the terminal)

src/node/setupServer.ts
```diff
@@ -1,15 +1,137 @@
...
+export class SetupServerApi extends SetupApi<ServerLifecycleEventsMap> {
+  private resolvedOptions: RequiredDeep<SharedOptions>
+
+  constructor(handlers: RequestHandler[]) {
+    super(
+      [ClientRequestInterceptor, XMLHttpRequestInterceptor],
+      'server-setup',
+      handlers,
+    )
+
```
Toxiapo: I am contemplating if I should make the interceptors in the concrete class dynamic so that whoever uses the server-like environment setup can pass in their interceptors and reuse this class

```ts 
class SetupServerApi extends SetupApi<ServerLifecycleEventsMap> {
    constructor(handlers: RequestHandler[], interceptors: ...[]) {
       super(interceptors, 'server-setup', handlers)
    }
}

```

Then to use this..

```ts
  export const setupServer = (...handlers: RequestHandler[]) => {
    return new SetupServerApi(handlers, [ClientRequestInterceptor, XMLHttpRequestInterceptor])
  }
```

Toxiapo:
Yes, @kettanaito I think we can start the review! I left some comments that I thought might be useful, but most of the logic was kept the same.

src/node/setupServer.ts
kettanaito: We can leave it for now, good call.

test/msw-api/setup-server/printHandlers.test.ts
kettanaito: I've fixed the tests, the CI is passing now.

src/node/setupServer.ts
kettanaito: I think the entire interceptors concept must be moved away from the base `SetupApi` and implemented as a custom thing on those child classes that need interceptors (Node and React Native ones). This would also make the usage of the base class more streamlined as there are setup APIs that do not need any interceptors at all, like `setupWorker`. 

This way the base class is basically an internal representation of handlers, internal emitters (and even handling), and the public methods that are shared across all setup APIs.

src/setupWorker/setupWorker.ts
kettanaito: No worries. I've removed some of the base class methods from `SetupWorkerApi` and made `this.context.requestHandlers` a getter, so it always returns the latest `this.currentHandlers` value inherited from the base class. I wish simple `this.X` reference would work but I assume since we're nesting it in another object (`this.context)` we lose the object magic. I'm fine with the getters.

kettanaito:
Thanks for moving forward with this, @Toxiapo! This looks great. I've had a chance to try it out locally, and also pushed a few changes regarding tests and child classes. One thing that remains is to distribute the interceptors properly (not all child classes rely on interceptors at all). I can pick this up once I have time next week so we can finish this.

kettanaito:
I've got some time to work on this today. Hope to wrap it up soon and prepare the next release.

src/createSetupApi.ts
kettanaito: Once I've moved the batch interceptor to the `SetupServerApi` child class, this has resolved itself.

kettanaito (dismissed):

kettanaito (approved):

kristojorg:
Just getting back from some time off and super happy to see this get done! Thanks @Toxiapo and @kettanaito : )

kettanaito:
Thanks for making this happen, @Toxiapo! 👏

Toxiapo:
Woohoo, I am excited to try this! Thanks, @kettanaito!

kettanaito:
## Released: v0.48.0 🎉

This has been released in v0.48.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.48.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.48.0)

---

smblee:
Hey guys! wondering what this would enable us to do? Is this for better edge support?

kettanaito:
Hey, @smblee. This would enable you to create custom `setup*` APIs that manage interception. One example of this is indeed creating a subset of `setupServer` that'd use different interceptors to run on the edge. 

```js
// Import the interceptors you want from @mswjs/interceptors
const edgeInterceptors = [fetchInterceptor]

class SetupEdgeServerApi extends SetupServerApi {
  constructor(handlers) {
    super(edgeInterceptors, handlers)
  }
}

export function setupEdgeServer(...handlers) {
  return new SetupEdgeServerApi(handlers)
}
```

Also note this is a rather advanced usage. If this doesn't make sense to you, it's very likely you don't need this API at all.

PR #1446, opened 2022-10-28
chore: update @mswjs/interceptors to 0.17.6

kdav108:
Update dependency package `mswjs/interceptors` from 0.17.5 to 0.17.6 (which fixed a security vulnerability in one of its dependencies)

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
   "sideEffects": false,
   "dependencies": {
     "@mswjs/cookies": "^0.2.2",
-    "@mswjs/interceptors": "^0.17.5",
+    "@mswjs/interceptors": "^0.17.6",
     "@open-draft/until": "^1.0.3",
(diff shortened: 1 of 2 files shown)


kettanaito:
Hey, @kdav108. Thanks for opening this. I believe this update is not necessary: the dependency range for `@mswjs/interceptors` is set to pass through any patch version updates. We rely on `^0.17.5` which will install `0.17.6` whenever you install the library. 

You can try it out yourself by running `npm install msw` once again and checking the version of fourth-party dependency installed:

```
npm ls @mswjs/interceptors
// 0.17.6
```

Let me know if I'm missing something.

kdav108:
Ahh silly me, I must have done something wrong before, just checked again and the patch version `0.17.6` is being downloaded. Thanks @kettanaito, I'll close this PR.

PR #1477, opened 2022-11-25
fix(setupWorker): resolve the TS4094 error

### LINKED ISSUE #1454: Documentation example doesn't work with TypeScript
### Prerequisites

### Environment check

### Browsers

_No response_

### Reproduction repository

https://stackblitz.com/edit/node-ifch93?file=index.ts

### Reproduction steps

Use documentation example with typescript https://mswjs.io/docs/getting-started/integrate/browser#configure-worker

```
// src/mocks/browser.js
import { handlers } from './handlers'

// This configures a Service Worker with the given request handlers.
export const worker = setupWorker(...handlers)
```

### Current behavior

Typescript throws multiple 
`TS4094: Property 'SOMETHING' of exported class expression may not be private or protected.`

### Expected behavior

I expect typescript to be fine with that

gduliscouet-ubitransport:
Type the return of `setupWorker` with the interface instead of the class.

I also don't export the class `SetupWorkerApi` anymore because it was not used in other files in the project, and not exposd in the `index.ts` (only the interface is exposed)

I tested in local that it fixes https://github.com/mswjs/msw/issues/1454

### DIFF
--- a/src/setupWorker/setupWorker.ts
+++ b/src/setupWorker/setupWorker.ts
@@ -19,6 +19,7 @@ import { createFallbackStop } from './stop/createFallbackStop'
 import { devUtils } from '../utils/internal/devUtils'
 import { SetupApi } from '../SetupApi'
 import { mergeRight } from '../utils/internal/mergeRight'
+import { SetupWorkerApi as SetupWorker } from './glossary'
 
 interface Listener {
   target: EventTarget
@@ -26,7 +27,10 @@ interface Listener {
   callback: EventListener
 }
 
-export class SetupWorkerApi extends SetupApi<WorkerLifecycleEventsMap> {
+class SetupWorkerApi
+  extends SetupApi<WorkerLifecycleEventsMap>
+  implements SetupWorker
+{
   private context: SetupWorkerInternalContext
   private startHandler: StartHandler = null as any
   private stopHandler: StopHandler = null as any
@@ -221,8 +225,6 @@ export class SetupWorkerApi extends SetupApi<WorkerLifecycleEventsMap> {
  * @param {RequestHandler[]} handlers List of request handlers.
  * @see {@link https://mswjs.io/docs/api/setup-worker `setupWorker`}
  */
-export function setupWorker(
-  ...handlers: Array<RequestHandler>
-): SetupWorkerApi {
+export function setupWorker(...handlers: Array<RequestHandler>): SetupWorker {
   return new SetupWorkerApi(...handlers)
 }


filipw01 (dismissed):

src/setupWorker/setupWorker.ts
```diff
@@ -19,14 +19,15 @@ import { createFallbackStop } from './stop/createFallbackStop'
 import { devUtils } from '../utils/internal/devUtils'
 import { SetupApi } from '../SetupApi'
 import { mergeRight } from '../utils/internal/mergeRight'
+import { SetupWorkerApi as ISetupWorkerApi } from './glossary'
```
kettanaito: I think we need to export the `SetupWorkerApi` interface from `index.ts` just as we do `SetupServerApi` from `node/index.ts`. The fact that it wasn't exported before is a mistake.

src/setupWorker/setupWorker.ts
```diff
@@ -19,14 +19,15 @@ import { createFallbackStop } from './stop/createFallbackStop'
...
 
 interface Listener {
   target: EventTarget
   eventType: string
   callback: EventListener
 }
 
-export class SetupWorkerApi extends SetupApi<WorkerLifecycleEventsMap> {
+class SetupWorkerApi extends SetupApi<WorkerLifecycleEventsMap> {
```
kettanaito: Let's also force this class to implement the `SetupWorker` interface. Without this enforcement, the only area to catch the incongruence is in the `setupWorker` return type, which is rather late. 

```ts
import { SetupWorker } from './glossary'

export class SetupWorkerApi extends SetupApi implements SetupWorker {
  ...
}

export function setupWorker(): SetupWorker { ... }
```

- Rename `SetupWorkerApi` in `glossary` to `SetupWorker`
  - Do the same for `SetupServerApi` 
- Export both `SetupWorker` interface and `SetupWorkerApi` class
- Set `SetupWorker` as the return type for `setupWorker()` to fix the type issue.

Would this approach work? 🤔

src/setupWorker/setupWorker.ts
gduliscouet-ubitransport: In fact it was already the interface that was exported, and not the class !

image

src/setupWorker/setupWorker.ts
gduliscouet-ubitransport: I agree with the implements part, I hesitated to do that.

But as I said above the `SetupWorkerApi` interface is already exposed, so we should not change its name to avoid a breaking change.

The class was only exposed through `setupWorker` return type. And it cannot be used due to those TS declaration file limitations.

Tell me what you think and I'll update the PR

kettanaito: Makes sense. Let's split this change in two then:

1. Import `SetupWorkerApi` type, alias it to `SetupWorker`, adjust the return type of `setupWorker` function. Release as a fix.
2. Open a follow-up that exports the `SetupWorkerApi` (class) and renames the `SetupWorkerApi` (interface) in `glossary.ts` to `SetupWorker`. Do the same for `SetupServer*`. Release a breaking change. 

What do you think?

gduliscouet-ubitransport: Seems good. I had to check, but in fact exporting the class instead of the interface removes the TS error (I was expecting to still have it)

gduliscouet-ubitransport: I updated this one, I try to open the second one on top of this one

kettanaito (approved):
Great work! Thank you for the discussion and the fix for this issue, @gduliscouet-ubitransport 👏 Welcome to contributors.

kettanaito:
## Released: v0.49.1 🎉

This has been released in v0.49.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v0.49.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/0.49.1)

---

PR #1481, opened 2022-11-28
fix(HttpResponse): accept interfaces as json response-data type

christoph-fricke:
This PR introduces a fix for an issue discussed with the upcoming standard-api. The relevant discussion can be found here: https://github.com/mswjs/msw/discussions/1464#discussioncomment-4242424
TLDR: Objects that are typed with an interface cannot be assigned to `HttpResponse.json`.

This PR changes the generic of `HttpResponse.json` to extend `DefaultBodyType` instead of the similar yet slightly different union. I think it makes more sense (and fixes the issue), because the response generic of resolver functions and generic of `StrictResponse` also extend `DefaultBodyType`.

### DIFF
--- a/src/HttpResponse.ts
+++ b/src/HttpResponse.ts
@@ -51,14 +51,10 @@ export const HttpResponse = {
    * HttpResponse.json({ firstName: 'John' })
    * HttpResponse.json({ error: 'Not Authorized' }, { status: 401 })
    */
-  json<
-    BodyType extends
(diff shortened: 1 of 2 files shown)


kettanaito (approved):
Simply great! Thanks, @christoph-fricke 👏 

If you spot any other type issues with the `next` release, please let me know/open a pull request. Let's build the future of API mocking together.

Issue #1388, opened 2022-09-04
Support Node.js v18


ericmasiello:
### Prerequisites

- [X] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [X] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [X] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [X] I'm using Node.js version 14 or higher

### Node.js version

v18.8.0

### Reproduction repository

https://github.com/ericmasiello/msw-node-18

### Reproduction steps

1. Verify you are using node 18.8.0
2. `npm ci`
3. `npm run test`

Note that the tests _pass_ but they actually should not. I'm using a `toMatchInlineSnapshot()`. That data snapshotted is the live data coming from real Pokemon API. what *should* be snapshotted is the mocked response I created in `foo.test.ts`

### Current behavior

The tests pass because the msw is not intercepting the response. Therefore, the `toMatchInlineSnapshot` matches live data from the API.

### Expected behavior

The tests should fail because the data should match the mocked response in `foo.test.ts`.

milesrichardson:
Hey! I actually have a PR for this. Well, I have one simple PR, and then a series of more robust PRs. Let me summarize...

### Background

- Polyfilled `fetch` continues to be intercepted in Node 18 (mostly, I think)
- Requests to IPv6 addresses will cause errors, regardless of `fetch` implementation. This might happen in tests that use `localhost` on an IPv6 system, because in Node 18, the [resolution order has changed to prefer the order of OS resolution](https://github.com/nodejs/node/issues/40537#issuecomment-948047193), instead of the previous behavior of preferring IPv4. You can [revert to the previous behavior with the optional Node flag `node --dns-result-order=ipv4first`](https://nodejs.org/api/cli.html#--dns-result-orderorder). This exposes the actual bug in msw (and some of its dependencies), which is incorrect URL serialization of IPv6 hosts (should be surrounded by brackets but is not, e.g. `http://::1:80` should be `http://[::1]:80`).
- "Node-native fetch" (aka [Undici](https://undici.nodejs.org/#/?id=undicisetglobaldispatcherdispatcher)), which is [enabled by default since Node 18](https://nodejs.org/de/blog/announcements/v18-release-announce/#new-globally-available-browser-compatible-apis), is not intercepted
[...]

Issue #1297, opened 2022-06-16
Make worker autoupdate an explicit choice


kettanaito:
### Scope

Improves an existing behavior

### Compatibility

- [X] This is a breaking change

## Current behavior

This package has a `postinstall` script that auto-copies the worker script to the consumer's specified directory. Executing arbitrary scripts in NPM packages is not a good idea. 

## Expected behavior

The consumer should instead include the auto-update functionality in their `package.json` as an explicit command they wish to have any time they reinstall dependencies:

```json
{
  "scripts": {
    "postinstall": "npx msw update-worker"
  }
}
```

Possible CLI command names:

- update-worker
- autoupdate

kettanaito:
Well, `npx msw init` already copies the worker. I don't even think this change is needed, it can be achieved as-is right now:

```json
{
  "scripts": {
    "msw:update": "npx msw init ./public"
  }
}
```

IanVS:
Is the worker auto-update intentional?  If I have msw pinned to a specific version, why does this post-install step update the worker to the latest version, instead of the version I pinned it to?

neurosnap:
With the recent security vulnerabilities leveraging `postinstall` to inject malicious code into app environments, this seems like an important change to make.

- https://www.wiz.io/blog/shai-hulud-npm-supply-chain-attack
- https://www.nodejs-security.com/blog/npm-ignore-scripts-best-practices-as-security-mitigation-for-malicious-packages

karlhorky:
This also affects all shadcn/ui commands, when pnpm `strict-dep-builds: true` configured (signals made by pnpm team that this will become default in future):

```bash
$ echo -e "# Fail on pnpm ignored build scripts\n# - https://github.com/pnpm/pnpm/pull/9071\nstrict-dep-builds=true" >> ~/.npmrc

$ cat ~/.npmrc
# Fail on pnpm ignored build scripts
# - https://github.com/pnpm/pnpm/pull/9071
strict-dep-builds=true

$ pnpm dlx shadcn@latest --help
...
 ERR_PNPM_IGNORED_BUILDS  Ignored build scripts: msw

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
```

### Workaround

Adding [`--allow-build=msw`](https://pnpm.io/cli/dlx#--allow-build) to the `pnpm dlx` command works around the error

```bash
$ pnpm --allow-build=msw dlx shadcn@latest --help     
Usage: shadcn [options] [command]

[...]

PR #1485, opened 2022-11-30
feat: warn on unsupported versions of Node.js

mahish:
msw does not support Node.js v18. Silently. Took me some time to realised that. Lets help to save time to others. There should be some warning until https://github.com/mswjs/msw/issues/1388 is done. I believe this is it. Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "msw": "cli/index.js"
   },
   "engines": {
-    "node": ">=14 <17"
+    "node": ">=14"
   },
   "scripts": {
     "start": "tsup --watch",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "msw": "cli/index.js"
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=14 <17"
   },
   "scripts": {
     "start": "tsup --watch",
--- a/src/node/SetupServerApi.ts
+++ b/src/node/SetupServerApi.ts
@@ -49,6 +49,14 @@ export class SetupServerApi extends SetupApi<ServerLifecycleEventsMap> {
     })
     this.resolvedOptions = {} as RequiredDeep<SharedOptions>
 
+    const nodeMajorVersion = parseFloat(process.version.replace('v', ''))
+    if (nodeMajorVersion > 16) {
+      devUtils.warn(
+        'Detected the version of Node.js (%s) that is not officially supported. Some features may not work as expected.',
+        process.version,
+      )
+    }
+
     this.init()
   }
 


kettanaito:
Hey, @mahish. Thanks for looking into this. 

I agree that there should be a warning but `engines` violation throws an error and halts dependency installation. More to that, the way `engines` mismatches are handled depends on the package manager (e.g. Yarn will indeed print a warning). That's why we haven't specified the upper range of the `node` engine. 

On the related note, while we don't have official Node 18 support that doesn't mean you can't use MSW with that version of Node. It all comes down to _what_ you're using. For example, we don't support global `fetch` that comes in Node 18 but the library should still run without issues if you're using `http`-based request clients. I don't recall any breaking changes in those from Node's side. 

I'm tentative to set the upper `node` range because it will break builds for people who already use MSW in Node 18 without issues. The worst part is that there's no op-out from this, as they would have to disable entire `engines` check just to install one dependency. I'd argue that's a significantly worse experience than trying the library on a newer version of Node, seeing that it fails, and finding an open issue about that version of Node. 

> What we can do, however, is to detect the Node version ourselves and print a warning to the process. If I recall correctly, Node version is set on `process` global, so that shouldn't be difficult to do. 

What do you think about this? What was the use case that helped you realize Node 18 is not supported?

mahish:
Hey @kettanaito , thanks for the detailed response. Yep, I agree. And you are right https://nodejs.org/api/process.html#processversion does the job. I think it would really helpful. Specially if the support is not coming anytime soon ( I dont know its ETA). Since Im switching between different projects quite often, using different Node versions I can imagine I might get stuck again if I forgot to switch the version.
 
My use case was that I realised server requests are not even detected but browser requests are intercepted without any problem. I was runing v18 for quite some time already. So I was pretty confused and didnt think about the Node.js version  at first. Since I usually mock only server data in my project, I needed to confirm this with ad hoc browser request which finally helped me to understand the problem.

kettanaito:
I don't generally give any ETA on my open source because I do it in my free time and encourage every contributor to do the same. But I do think we will get Node 18 support sometime next year. 

I think the strategy we outlined is viable:

1. In our node code detect the current process's Node version.
2. If it's higher than 17, print a single warning that things may work unexpectedly (we can work on phrasing together).

mahish:
Sure.

This is probably very easy and minor thing for you bbut let me know if I can help you with it (for both 1. and 2.) Ill think about the phrasing.

mahish:
Took some inspiration from the existing console messages.

```js
`Intercepting server-side requests might not work properly using the current Node.js version: ${process.env.version}. Downgrading Node.js version to 16 might help if you encounter any problem. Read more https://github.com/mswjs/msw/issues/1388`
```

mahish:
Would be better to be more specific when it "does not work properly" but Im not fully aware of technical details.

kettanaito:
We can consider a warning message like this:

```
[MSW] Detected unsupported Node.js version (18.8.0). Some features may not work reliably. Read more in https://github.com/mswjs/msw/issues/1388.
```

I think it'd be a good tone to give people the ability to opt-out from these warnings. I hate to design a whole logic around this, so maybe just respect a magical env variable?

```js
if (!process.env.MSW_SUPPRESS_UNSUPPORTED_NODE_VERSION) {
  devUtils.warn('...')
}
```

So if the library works fine for those versions, people could get rid of the warning:
```json
{
  "test": "MSW_SUPPRESS_UNSUPPORTED_NODE_VERSION=1 jest"
}
```

What do you think?

mahish:
@kettanaito Sorry for delay, I had health issues. 

I would prefer to be little bit more specific in the message if possible to give a clue to developer what might not be reliable. I thought we know it is related only to server-side fetch method. If we mention this fact, or simpoly be more explicit about it, developers mocking only in browser would not need to care about it at all.

Are you suggesting to not check the Node version but use only the env variable?

Regarding the env variable to disable the warning. To be honest, I dont see much of advantage in it. I thought the warning would be printed on initial load/build only. Actually, I would find it safer if it is printed for everyone until the proper support exists

kettanaito:
Hey, @mahish. I'm sorry to hear about your health. Please get better and be safe ❤️ 

> Are you suggesting to not check the Node version but use only the env variable?

No, I meant to use an env variable as the means to opt-out from the warning. But what you're suggesting makes much more sense. Let's ensure we print it only _once_ on runtime. 

It'd be great to print this on `postinstall` but executing scripts on `postinstall` is unsafe and some package managers already block that by default. I think if we put it on the root scope of the package it will run exactly once. It may still be a nuisance to some people (trust me, it will be) but a single warning per the entire test run shouldn't be enough reason to provide an opt-out flag on our side, I hope.

kettanaito:
Unfortunately, I cannot find a way to print this warning only _once_. Even if I add it to the root entry, each test imports it on its own, resulting in multiple warnings. It's basically up to the test runner to control how things are imported/compiled, so MSW has no freedom here. 

I'm beginning to think that the DX regression of this feature may not outweigh the benefit it provides. Need to look into other possible ways to ensure the warning is printed only once. 

I wouldn't go for the `postinstall` script because I'm planning on moving away from it entirely (#1297).

kettanaito:
I don't see how we can implement this warning not to be annoying and appearing on each `msw` import. Alas, there doesn't seem to be a way. 

The devs should rely on the warnings/errors from their package manager whenever the "engines.node" contract is breached, and for the newer versions it's a trial-and-error approach.

PR #1487, opened 2022-12-02
docs(migrating): fix description for "ctx.errors" migration

christoph-fricke:
While giving the migration guide another read, I noticed that the `ctx.errors` description is the same as `ctx.data` above.

This PR replaces data with errors in the description, so it makes sense for the section.


kettanaito (approved):
A good catch! Thanks once again, @christoph-fricke!

PR #1530, opened 2023-02-03
fix: add "package.json" subpath export

iceu-bb:
Zrzut ekranu 2023-02-3 o 16 46 23

Add a fix for the warning that occurred when starting React Native metro server.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,7 @@
       "require": "./lib/node/index.js",
       "default": "./lib/node/index.mjs"
     },
-    "./package.json": {
-      "default": "./package.json"
-    }
+    "./package.json": "./package.json"
   },
   "bin": {
     "msw": "cli/index.js"
--- a/package.json
+++ b/package.json
@@ -16,6 +16,9 @@
       "types": "./lib/node/index.d.ts",
       "require": "./lib/node/index.js",
       "default": "./lib/node/index.mjs"
+    },
+    "./package.json": {
+      "default": "./package.json"
     }
   },
   "bin": {


package.json
```diff
@@ -16,6 +16,9 @@
       "types": "./lib/node/index.d.ts",
       "require": "./lib/node/index.js",
       "default": "./lib/node/index.mjs"
+    },
+    "./package.json": {
```
kettanaito: What's the difference between this and 

```json
{
  "./package.json": "./package.json"
}
```

I've seen [other libraries](https://github.com/ReactiveX/rxjs/blob/630d2b009b5ae4e8f2a62d9740738c1ec317c2d5/package.json#L67) using a direct path, so I assume that's equivalent to "default". Do you happen to know if there's a difference?

kettanaito (dismissed):
Hey, @iceu-bb! Thanks for looking into this. The fix looks great, I had but one question regarding the usage of "default" and whether we could drop it and still have the same behavior.

package.json
mattcosta7: @kettanaito yep, when just a string, it's default.  When an object is passed it's evaluated top to bottom, returning the first match!

kettanaito: @mattcosta7, your wizardry saves the realm once more! 🙇 Thanks, Matt.

kettanaito:
Welcome to contributors, @iceu-bb! 🎉

kettanaito:
## Released: v1.0.1 🎉

This has been released in v1.0.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v1.0.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/1.0.1)

---

PR #1532, opened 2023-02-05
test(playwright configuration): use fullyParallel flag to isolate tests

shwarcu:
This flag makes playwright run each single test in a separate context in parallel, providing easier isolation, teardown and better performance when running in parallel in many workers.

Reference https://playwright.dev/docs/test-parallel

### DIFF
--- a/test/browser/playwright.config.ts
+++ b/test/browser/playwright.config.ts
@@ -21,6 +21,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 2 : undefined,
   forbidOnly: !!process.env.CI,
   reporter: 'list',
+  fullyParallel: true,
 }
 
 export default config
--- a/test/browser/graphql-api/link.test.ts
+++ b/test/browser/graphql-api/link.test.ts
@@ -10,11 +10,11 @@ const server = new HttpServer((app) => {
   })
 })
 
-test.beforeAll(async () => {
+test.beforeEach(async () => {
   await server.listen()
 })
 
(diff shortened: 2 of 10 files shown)


kettanaito (approved):
This looks great! @shwarcu, is there any chance needed to the Playwright config? The `fullyParallel` flag sounds like something we should set there but it's not in the diff (double-checking).

test/browser/playwright.config.ts
```diff
@@ -21,6 +21,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 2 : undefined,
   forbidOnly: !!process.env.CI,
   reporter: 'list',
+  fullyParallel: true,
```
shwarcu, line 24: @kettanaito  it's here :D

shwarcu: Btw bare in mind that changing this way of running tests may unveil some setup-teardown issues that weren't caught before, but in long run I think it's crucial to have tests independent and it's better to find out about such problems/blockers sooner

kettanaito: I absolutely agree. Looks like a great addition to me.

kettanaito (approved):
🚀

kettanaito:
## Released: v1.1.0 🎉

This has been released in v1.1.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v1.1.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/1.1.0)

---

PR #1534, opened 2023-02-06
Experiments around webpack http server - use esbuild loader in webpack devserver

shwarcu:


kettanaito:
Thanks for addressing this, @shwarcu! I've already merged this approach in the playwright feature branch, I'm going to close this one.

PR #1436, opened 2022-10-17
feat!: adopt the global Fetch API

kettanaito:
> This is a breaking change.

- **Closes #1404**
- Closes #991, fixes #947 
- Closes #1422 
- Closes #1332 
- Fixes #1327
- Fixes #1336
- Potentially solves #1158
- Fixes #413
- Closes #1318 since it's no longer relevant (no `MockedRequest`)
- Fixes #1380 
- Fixes #1365
- Distributes the library as CJS and ESM for both browser and Node.js. Fixes #1384, fixes #1427
- Fixes #1442
- Closes #1388
- Closes #1207

## Migration guide

- [**MIGRATION GUIDE**](https://github.com/mswjs/msw/blob/feat/standard-api/MIGRATING.md)
- [Share your feedback](https://github.com/mswjs/msw/discussions/1464)

## Roadmap

- [x] https://github.com/mswjs/interceptors/pull/292
  - [x] https://github.com/remix-run/web-std-io/pull/21
  - [x] https://github.com/mswjs/interceptors/pull/329
  - [x] Bump `@mswjs/interceptors`
- [x] Fix typings tests (our tests pass but the compilation from `msw` fails)
- [x] Support type-safe request bodies (mainly JSON)
- [x] Support type-safe response bodies
- [x] Support `ReadableStream` as the mocked response body (internally: read/send request/response bodies as streams)
- [x] Fix `test/rest-api/cookies-inheritance.node.test.ts` ambiguous response body type
- [x] Restructure root-level exports
- [x] Publish Interceptors as CJS/ESM and update it here.
- [x] Fix `request.formData()` test (see https://github.com/nodejs/undici/issues/1979). Just updating to the latest Node 18 will fix this.
- [x] FormData: the node test hangs for some reason on `request.clone()`:
```
pnpm test:node rest-api/request/body/body-form-data.node.test.ts --detectOpenHandles

Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  MESSAGEPORT

      154 |     return this.predicate(
      155 |       request,
    > 156 |       await this.parse(request.clone(), resolutionContext),
          |                                ^
      157 |       resolutionContext,
      158 |     )
      159 |   }

      at RestHandler.test (../../src/handlers/RequestHandler.ts:156:32)
      at ../../src/utils/getResponse.ts:34:20
          at Array.map (<anonymous>)
      at filterAsync (../../src/utils/getResponse.ts:21:44)
[...]

Issue #156, opened 2020-05-13
Support mocking WebSocket APIs


Doesntmeananything:
Is it possible to use msw to mock server-sent events (SSE) and WebSocket (WS) connections?

My use case is to have a somewhat full client-side mocking story (including REST, SSE, and WS functionality), and since msw is such a joy to use when mocking out REST APIs, I was wondering if it makes sense to use it to mock more specialised server interactions.

Have you thought about this? I admit that I haven't looked much into whether it's possible just to use custom request handlers to add this functionality, emulating SSE and WS behaviour in some way. I wanted to know if you already had something in mind regarding this question. Thanks!

kettanaito:
Hey, @Doesntmeananything, thanks for bringing this topic up. I'd love to bring SSE and WS support to the users of MSW. I admit I haven't researched the topic yet, but would use this thread for this.

Technically, it comes down to the ability of Service Worker to intercept those requests/events. If the spec supports it, there shouldn't be much changes needed on the MSW side.

**Here's some useful resources:**

- https://github.com/w3c/ServiceWorker/issues/885, `EventSource` interception in Service Worker.
- https://github.com/w3c/ServiceWorker/issues/947, a discussion whether Service Worker should access WebSocket events.

Could you please try to set up a proof of concept, if those events can be intercepted in the worker's `fetch` event?

You're right about the custom request handler, we can use it to log all intercepted requests:

```js
setupWorker(
  {
    predicate(req) {
      console.log(req)
      // log captured requests, but bypass them
      return false
    },
    resolver: () => null
  }
)
```

If we confirm it working, I'd be glad to discuss the API for a dedicated request handler for WebSocket/SSE. I can read on their specification meanwhile.

Doesntmeananything:
Sounds like a plan! At a cursory glance, it does indeed seem quite doable. Let me PoC this, and I'll get back to you with my results as soon as I can.

Doesntmeananything:
Hi, @kettanaito! I've set up a (very quick and dirty) repository to test these interactions over at https://github.com/Doesntmeananything/msw-sse-ws.

My initial findings are the following:

- I am able to intercept SSE connections, however this occurs only when a stream ends. For example, if a stream consists of 3 messages, logging happens only after all 3 messages have been received, including the final `end` event.
- I am not able to intercept WS connections. I tried to make sure that the WS connection is established from the client only after the mock service worker had been initialised, but it didn't seem to help. I want to investigate this further.

[...]

PR #1551, opened 2023-02-21
fix(worker): remove bypassing of server-sent events

piotr-cz:
Server-sent Events are now working fine and there is no need to bypass them.

Fixes workaround from https://github.com/mswjs/msw/issues/156#issuecomment-1189807742

Implementation examples
- Migration guide > [Support ReadableStream mocked responses](https://github.com/mswjs/msw/blob/feat/standard-api/MIGRATING.md#support-readablestream-mocked-responses)
- [test/rest-api/response/body/body-readable-stream.mocks.ts](https://github.com/mswjs/msw/pull/1436/files#diff-cc6c833284918f1602b3c897fb29381fd3803581ae18da15e795e8612360bef9)

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -88,11 +88,6 @@ self.addEventListener('fetch', function (event) {
   const { request } = event
   const accept = request.headers.get('accept') || ''
 
-  // Bypass server-sent events.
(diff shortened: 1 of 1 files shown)


kettanaito (approved):
Awesome! Thanks, @piotr-cz.

PR #1584, opened 2023-04-04
feat: split the library into "core", "browser" and "node"

kettanaito:
- Splits the library modules into `core`, `browser`, `node`, and `native`.
- Adjusts TSUP config to replace `~/core` import path with the relative import relatively to the _emitted_ files.
- Appends `.mjs` or `.js` to the resolved `~/core` imports based on whether the output file is ESM or not. 
- Patches the emitted `.d.ts` files to resolve `~/core` imports in them relatively to the emitted files. 
- Forces browser module resolution for `*.mocks.ts` test files. 

A really cool side-effect of this is that MSW can now be used in both the browser and Node using the same root-level `msw` import!

```js
// some.node.file.js
import { setupServer } from 'msw'
```

## Roadmap

- [x] Fix failing unit tests
- [x] Fix failing Node tests
- [x] Fix failing browser tests
- [x] Move worker-specific `core/utils` to `src/browser/...` so they don't end up in the Node bundle accidentally
- [x] Add automation to validate this setup keeps working


kettanaito:
Some tests are failing locally, I don't expect these changes to be ready just yet.

config/scripts/patch-ts.js
```diff
@@ -0,0 +1,29 @@
+const fs = require('fs')
+const path = require('path')
+const { replaceCoreImports } = require('../replaceCoreImports')
+
+async function patchTypeDefs() {
```
kettanaito, line 5: Type defs emitted by TS still have the `~/core` imports in them. Nothing resolves them to actual `paths` defined in tsconfig. Well, someone must.

src/browser/index.ts
```diff
@@ -1,7 +1,7 @@
-// Include the root-level common exports so that the browser bundle
-// can export everything from a single module ("msw" import root).
-export * from '..'
+// Re-export the code MSW API (e.g. "rest", "RequestHandler") because
+// they are available from the "msw" (browser) root-level export.
+export * from '../core'
```
Andarist: shouldnt this be something like?
```suggestion
export * from '~/core'
```

tsup.config.ts
```diff
@@ -1,59 +1,80 @@
...
-  name: 'node',
-  platform: 'node',
-  entry: ['./src/index.ts', './src/node/index.ts'],
-  inject: ['./config/polyfills-node.ts'],
+const coreConfig: Options = {
+  name: 'core',
+  platform: 'neutral',
+  entry: glob.sync('./src/core/**/*.ts', {
+    ignore: '**/*.test.ts',
+  }),
```
Andarist, line 25: clever

src/browser/index.ts
kettanaito: I thought so initially but for some strange reason, TS replaces the `~/core` import to an absolute path on the disk in the emitted files. I don't know why that happens but keeping a relative import here is fine since the depth of source and lib directories is the same for this module.

tsup.config.ts
kettanaito: I was surprised to learn esbuild doesn't support negative patterns and doesn't have an `ignore` option.

kettanaito:
The implementation itself is ready but I'm reluctant to merge it yet because of the degradation of the developer experience this change implies. 

Shortly, since `msw` root-level import will now load different runtime and type exports, the consumer needs to explicitly mark their TS modules that they are meant for the browser. This way, TS will correctly resolve `msw` exports when you're writing code meant to execute in the browser:

```ts
// src/pages/index.tsx
```

Without this explicit resolution, TS has no clue what is your module meant for, and so it assumes the default `main` export of the package (or the corresponding `exports['.']`, which is `node`. This breaks the type checking without any effect on the runtime. 

The way people can fix that is by creating a nested `tsconfig.json` and teaching TS to treat certain subset of modules as browser modules. 

> This is also not applicable for older versions of TS (< 4.8 I believe), where `customConditions` is not a thing. On those versions, the consumer would have to create a custom `msw` path and resolve it manually to `node_modules/msw/lib/browser`, which is brittle because it relies on implementation details of the library). 

## Alternative 

Alternatively, the library can expose its browser exports under `msw/browser` so no tooling would confuse which code/types to load. The downside of this is having to update all the imports.

kettanaito:
I think it may be a good idea to consider a uniform API for both browser and server so the usage surface is identical.

```js
// In both browser and Node.
import { setup } from 'msw'

setup()
```

If the public API between environments is the same, then it wouldn't matter what module resolution is being used (type-wise at least).

Issue #1404, opened 2022-09-12
Adopt Fetch API primitives


kentcdodds:
### Scope

Improves an existing behavior

### Compatibility

- [X] This is a breaking change

### Feature description

I *think* this could be done in a way that is non-breaking. But I think it'd be great to be able to write my handlers similar to how I write my Remix loaders/actions. They receive an object with `{ request, params, context }` and return a `Response`. Could even use the Remix ponyfill for Request/Response: https://github.com/remix-run/web-std-io

kettanaito:
Thanks for writing this down, @kentcdodds. I've been thinking about using native Fetch's `Request` and `Response` for some time now. I will provide this thread with more thoughts/concerns around this change in the future.

kettanaito:
First of all, I think this is a good change. Embracing the platform is something MSW has been doing almost since the beginning, both when using the Service Worker API as well as operating with Fetch `Response` instances in the browser. I think we should also embrace these APIs for public usage (currently, used only internally). Overall, this is the direction that our public API will take sooner or later.

In a very drafty fashion, this is the public API I imagine:

```js
// handlers.js
import { rest } from 'msw'

export const handlers = [
  rest.get('/user', () => {
    const user = JSON.stringify({ name: 'John' })
    return new Response(user)
  })
]
```

Now, let’s talk about some concerns with such API change.

## Insufficient data
Both `Request` and `Response` instances are insufficient to describe real-world requests and responses. Some of the examples of the missing data are:

- Request path parameters (`/user/:id`);
- Request cookies (represented as a plain string in the `Cookie` header);
- Response timings (cannot control response delay);

The fetch API also lacks certain convenience features such as responding with a JSON body quickly. I think it’s too verbose to be consumed directly. 

## Environment support
MSW still supports versions of Node.js that have no native `Request` or `Response` classes. I believe those are added in Node.js 18 with the addition of the native `fetch` but I may be mistaken. This means that our users wouldn’t be able to construct `new Response()` in handlers because that class does not exist in Node.js.

Even outside of support, Node.js handles requests and responses differently based on the request issuing actor:

- `http.Client`: request (`http.OutgoingMessage`), response (`http.IncomingMessage`);
- `XMLHttpRequest`: abstract internal representations. However, it’s fair to say XHR support in Node.js is purely JSDOM-oriented and we may look into whether it’s possible to drop it and use `http` interception only (`XHR` is polyfilled by `http` in JSDOM).

[...]

PR #1585, opened 2023-04-04
feat: support batch graphql request handling

### LINKED ISSUE #510: Support batched GraphQL queries

johnboulder:
Closes #510 

This might be more of a bandaid than a final solution, but I figure something is better than nothing. For the team I'm working with, fixing this is kindof critical to the work we're doing (which I hope just clarifies why I'm approaching it this way). It seemed like the best way to fix the problem without making a larger set of changes. 
I've really only been looking at this for the past few days so I'm just doubting whether this contribution is thorough enough. Everything I've tested with this works fine though, so it feels like a nice bandaid. 

Approach:
The core of the change is in `handleRequest()`. What I've done is made it so it just checks if the incoming request is a batched graphql request. If it is, we iterate over the body creating individual requests and calling `getResponse()` with them. Once that finishes executing, the code just wraps that up in an array, serializes it, and returns it as the body of the response. I get the impression that this might be circumventing the logic that comes after calling `handleRequest()`, but I struggled to find documentation around that in particular, and it seems like it might be hard to treat a batched request exactly the same as a normal graphql request. 

I feel like this just a nice, temporary solution for the time being while folks decide on something more thorough.


johnboulder:
@kettanaito any thoughts on this?

Also, sorry to tag you directly, I'm just not sure who else to maybe tag on this lol. There's nobody else that's even close to as active as you are, so it's not clear who is actually active in contributing...

dani-a-dolan:
@kettanaito

kettanaito:
Hi, @johnboulder. First things first, thanks for proposing and implementing this! 

What concerns getting this merged and released, the library is on hold for new features such as this one because I don't have the capacity to migrate those features to v2.0 that's in its final stages of development at the moment. 

This means that I will get to this pull request and we will, very likely, merge it once v2.0 is out (i.e. #1404). Expect some time this summer if everything goes smoothly.

johnboulder:
Hey! Good to hear from you! That's awesome news! My team is currently using a forked release that includes my changes, and we'd love to get off it as soon as possible. So I'll be happy to make any updates to this (if they're needed) as soon as v2.0 is released. 

If v2.0 comes out and I don't follow up with resolving any merge conflicts on this feel free to tag me. I may just not have noticed the release.

kettanaito:
## Released: v2.1.3 🎉

---

PR #1603, opened 2023-04-29
fix: use ESM-friendly mirrors of "statuses" and "js-levenshtein" 

thepassle:
Tests pass:

image

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,5 +1,10 @@
 declare const SERVICE_WORKER_CHECKSUM: string
 
 declare module '@bundled-es-modules/cookie' {
-  export * from 'cookie'
+  export * as default from 'cookie'
+}
+
+declare module '@bundled-es-modules/statuses' {
+  const source_default: any
+  export { source_default as default }
 }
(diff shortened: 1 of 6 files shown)


global.d.ts
```diff
@@ -1,5 +1,10 @@
 declare const SERVICE_WORKER_CHECKSUM: string
 
 declare module '@bundled-es-modules/cookie' {
-  export * from 'cookie'
+  export * as default from 'cookie'
+}
+
+declare module '@bundled-es-modules/statuses' {
+  const source_default: any
```
kettanaito, line 8: Maybe let's annotate this export? Just to be safe.

kettanaito (approved):
These look fantastic. Thank you so much for all the help with ESM, @thepassle. 

I've left a comment about annotating the default export of statuses. That should be possible, I think.

global.d.ts
kettanaito: I will add these type definitions, we're already depending on `@types/statuses`.

kettanaito:
I've pushed to the remote and since it pointed to the same branch (just on fork), GitHub decided to merge this pull request. All's good though, the changes are great.

PR #1606, opened 2023-05-03
fix: stream response chunks as they come in Node.js

kettanaito:
- Implemented in https://github.com/mswjs/interceptors/pull/373

This needs the Interceptors dependency bump to have the newly added tests pass.


test/node/rest-api/response/body-stream.node.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    })
+
+    response.once('end', () => {
+      const textChunks = chunks.map((chunk) => chunk.buffer.toString('utf8'))
+      expect(textChunks).toEqual(['first', 'second', 'third'])
+
+      // Ensure that the chunks were sent over time,
+      // respecting the delay set in the mocked stream.
+      const chunkTimings = chunks.map((chunk) => chunk.timestamp)
+      expect(chunkTimings[1] - chunkTimings[0]).toBeGreaterThanOrEqual(500)
```
kettanaito: This will fail at the moment because the Interceptors used to consume the entire response stream before setting the `this.res = response` and emitting the `response` event on the `ClientRequest` instance. 

That's no longer the case with the fix.

PR #1608, opened 2023-05-05
fix: prevent "abort" listener memory leak in RequestHandler

### LINKED ISSUE #1207: Return instructions from the resolver
### Scope

Improves an existing behavior

### Compatibility

### Feature description

Instead of a response resolver returning a `MockedResponse` object, I propose it returns an abstract instruction object. Such an instruction object tells the library what to do next with the intercepted request.

- Originates from #1204 

## Instruction types

- Fallthrough. Ignore this handler, skip to the next matching handler, if any.
- Passthrough (see #1204). Perform the intercepted request as-is.
- Use mocked response (currently `res()`).
- Return a network error (currently `res.networkError()`).
- Any arbitrary custom instruction in the future to be handled by `handleRequest`.

## Motivation

There are multiple sources that may affect the resolution of an intercepted request. You may compose a mocked response via `res()`, return a network error, or instruct the library to explicitly perform this request as-is. Pushing all of those different intentions into a single `MockedResponse` object is not maintainable. 

Even the behaviors like `once` do not belong on the mocked response object. Instead, things like `once` must be _modifiers_ to instructions, not even instructions themselves. This will allow us to return any instructions with a modifier, such as "return a network error once" (see #413).

kettanaito:
- Closes #1207 
- Prevents the "abort" request listener leak in the `RequestHandler` class by cloning the incoming request _once_, and then doing subsequent clones from that main clone.

> Context: whenever a `Request` is cloned, it gets a new `abort` listener to propagate the original request abort event to the cloned instance. Since we were processing _the same_ request through multiple handlers, cloning it _multiple times_, every `request.clone()` invocation appended the `abort` listener, resulting in a memory leak. 

- Refactors the request handler lookup algorithm (finally). The main changes are:
  - Returning `undefined` from a response resolver _no longer prints a warning_. The devs must read the docks and understand that if they return nothing, nothing will happen.

### DIFF
--- a/src/browser/setupWorker/start/createFallbackRequestListener.ts
+++ b/src/browser/setupWorker/start/createFallbackRequestListener.ts
@@ -26,7 +26,7 @@ export function createFallbackRequestListener(
       options,
       context.emitter,
       {
-        onMockedResponse(_, { handler, request, parsedRequest }) {
+        onMockedResponse(_, { handler, parsedRequest }) {
           if (!options.quiet) {
             context.emitter.once('response:mocked', (response) => {
               handler.log(request, response, parsedRequest)
--- a/src/browser/setupWorker/start/createFallbackRequestListener.ts
+++ b/src/browser/setupWorker/start/createFallbackRequestListener.ts
@@ -19,6 +19,8 @@ export function createFallbackRequestListener(
   })
 
   interceptor.on('request', async (request, requestId) => {
+    const requestCloneForLogs = request.clone()
+
     const response = await handleRequest(
       request,
       requestId,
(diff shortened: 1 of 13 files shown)


kettanaito:
## Error [TypeError]: Cannot read properties of undefined (reading 'clone')

Something goes off with `serializeRequest` and `getPublicUrlFromRequest`: they receive `undefined` instead of a request instance. I suspect that some request clones don't go as planned in the browser only.

kettanaito:
✅ Had to adjust the `onMockedResponse()` callback signature because it no longer returns a `request` in the argument object
✅ Had to clone the request before the processing so the `handler.log()` function would be able to read the request's body
✅ Had to adjust some browser tests because they were still expecting the warning when a resolver returns undefined

PR #1617, opened 2023-05-12
feat: support cross-process interception via `setupRemoteServer`

kettanaito:
> This is an experimental feature. It's unlikely to ship before 2.0. 

## Intention

Introduce an API that allows one process to modify the traffic of another process. The most apparent application for this is testing server-side behaviors of a JavaScript application:

```js
// app.js
forwardNetworkToRemote()

export const loader = async () => {
  const res = await fetch('https://example.com/resource')
}
```

```js
// app.test.js
it('fetches the user server-side', async () => {
  let a = listenToRemoteNetwork(targetProcess)
  modifyNetwork(a)
  // ...render
  // ...assert
})
```

> This is an example API. For the exact proposed API, keep reading.

This API is designed exclusively for use cases when the request-issuing process and the request-resolving process (i.e. where you run MSW) are two different processes. 

> [!IMPORTANT]
> While the feature itself is unopinionated in how it's used, in practice it implies that **you either have a fixed list of handlers OR you provide test-app isolation by spawning an app instance per _test case_**. That is so handler overrides in one test don't affect the remote network for other tests. The app becomes a shared state here with no means to distinguish who is the consumer (test -> client -> server chain loses any identifiers).

## Proposed API

With consideration to the existing MSW user experience, I suggest we add a `setupRemoteServer()` API that implements the `SetupApi` interface and has a similar API to `setupServer`. The main user-facing distinction here is that `setupRemoteServer` is affecting a _remote_ process, as indicated by the name. 

```js
import { http } from 'msw'
import { setupRemoteServer } from 'msw/node'

const remote = setupRemoteServer(...initialHandlers)

// Notice: async!
beforeAll(async () => await remote.listen())
afterEach(() => remote.resetHandlers())
afterAll(async () => await remote.close())
```

The `.listen()` and `.close()` methods of the remote server _become async_ since they now establish and terminate an internal server instance respectively. 

> Similar to the `setupServer` integration, it would be recommended to call `setupRemoteServer` _once_ as a part of your global testing setup. Closing the WebSocket server after each test suite will have performance implications since each next test suite would wait while `remote.listen()` spawns that server again. 

You can then operate with the `remote` server as you would with a regular `setupServer`, keeping in mind that it doesn't affect the current process (your test) but instead, any remote process that runs `setupServer` (your app).

```js
it('handles user errors', () => {
  // Appending and removing request handlers is sync
  // because they are stored in the current (test) process.
  remote.use(
    http.get('/user', () => {
      return new Response(null, { status: 500 })
    })
  )

  // ...interact and assert your app.
})
```

By fully extending the `SetupApi`, the `setupRemoteServer` API provides the user with full network-managing capabilities. This includes defining initial and runtime request handlers, as well as observing the outgoing traffic of a remote process using the Life-cycle API (`remote.events.on(event, listener)`). I think this is a nice familiarity that also provides the user with more power when it comes to controlling the network. 

## Implementation

I've considered multiple ways of implementing this feature. Listing them below.

### (Chosen) WebSocket server

The `setupRemoteServer` API can establish an internal WebSocket server that can route the outgoing traffic from any server-side MSW instance anywhere and deliver it to the remote server to potentially resolve. 

Technically, the WebSocket server acts as a resolution point (i.e. your handlers) while the remote MSW process acts as a request supplier (similar to how the Service Worker acts in the browser). 

Very roughly, this implies that the regular `setupServer` instances now have a fixed request handler that tries to check if any outgoing request is potentially handled by an existing remote WebSocket server:

```js
// setupServer.js
await handleRequest(
  request,
  requestId,
  [
    // A very basic idea on how a "remote" request handler works.
    http.all('*', async ({ request }) => {
      wsServer.emit('request', serializeRequest(request))
      await wsServer.on('response', (serializedResponse) => {
        return deserializeResponse(serializedResponse)
      })
    }),
    ...this.currentHandlers,
  ]
)
```

> Unlike request handler (i.e. function) serialization, it is perfectly fine to serialize `Request` and `Response` instances and transfer them over any message channel, like a WebSocket transport. 

If no WebSocket server was found or establishing a connection with it fails within a sensible timeout period (~500ms), the `setupServer` instance of the app continues to operate as normal. 

> Alternatively, we can skip the WebSocket server lookup altogether and make it opt-in via some `remote: true` option on the app's side. 

### IPC

The test process and the app process can utilize IPC (interprocess communication) to implement a messaging protocol. Using that protocol, the app can signal back any outgoing requests and the test can try resolving them against the request handlers you defined immediately in the test. 

This approach is similar to the WebSocket approach above with the exception that it relies on IPC instead of a standalone running server. With that, it also gains its biggest disadvantage: _the app process must be a child process of the test process_. This is not easy to guarantee. Depending on the framework's internal implementation, the user may not achieve this parent/child relationship, and the IPC implementation will not work. 

Given such a demanding requirement, I've decided not to use this implementation.

## Limitations

- `useRemoteServer()` affects the network resolution for the entire app. This means that you cannot have multiple tests that override request handlers for the same app at the same time. I think this is more than reasonable since you know you're running 1 app instance that can only behave in a single way at a single point in time. Still, I expect users to be confused when they parallelize their E2E tests and suddenly see some network behaviors leaking across the test cases. 

## Concerns

- Can we rely on a fixed local port to always be available?
- Is it safe to introduce a WebSocket server that will be, effectively, routing HTTP messages over the local network (during tests only)? 
  - Yes. If someone can intercept that WebSocket communication, they are already in your machine and can do things far worse than that. 
- Is it clear that `setupRemoteServer` only affects the _server-side_ network behavior of any running application process with the server-side MSW integration? To affect the client-side network behavior from a test you have to 1) have `setupWorker` integration in the app; 2) set a global `window.worker` instance; 3) use `window.worker.use()` to add runtime request handlers. This stays as it is right now, no changes here. 

---

- [Remix example repo](https://github.com/mswjs/lab-remote-handlers)

The API is TBD and is subjected to change. 

## Roadmap

- [x] Ensure the sync server connection is awaited before the first request handler runs.
- [x] Introduce serialize/deserialize utilities for requests and responses (used both in the worker and in the WS sync layer now).
- [x] Fix listeners' memory leaks on hot updates (clean up listeners).
- [x] Make the WS events map type-safe
- [x] Rely on the internal request header when bypassing Socket IO connection requests in the `rest.all()` handler. 
- [x] Handle socket timeout and errors when awaiting for the `response` in `setupServer`.
  - CPRI doesn't use websockets anymore.
- [x] Support `ReadableStream` from the remote request handler (may consider transferring `ReadableStream` over the WS messages instead of `ArrayBuffer`, if that's allowed). 
  - This may not be needed, in the end, but if we can pull off `ReadableStream` transfer over WebSockets that would be great. 
- [x] Support all Life-cycle events.
- [x] Support setting a custom WebSocket server port number through environment variables.
- [x] Make the `remotePort` and `port` an implementation detail of `setupRemoteServer` and `setupServer({ remote: true })`. The developer mustn't care about those.
- [x] Do not spread the list of user-defined request handlers to prepend the fixed remote server handler (spreading of large lists may have performance implications). 
  - Not an issue until proven otherwise; have no wish to optimize prematurely. 
- [x] Solve the test/app catch 22 by attaching a self-replicating one-time handlers only for the first-time requests (those fetched when the testing framework pings your app). 
- [x] Fix: failing `use()` test (may have something to do with the handlers management refactoring as a part of the `server.boundary()`).
- [x] Support differentiating between requests done in different Playwright workers (see [this](https://github.com/mswjs/msw/pull/1617#issuecomment-2332499183)).
- [x] **Add more tests, specifically for different response body types.** Also life-cycle events tests for `setupServer` (that it doesn't emit them for internal requests).
- [x] Consider adding `setupWorker` support (see https://github.com/mswjs/msw/pull/1617#issuecomment-2356776359).
  - Out of scope. This is likely too much for the initial scope. May consider iterating on it to bring remote interception to the browser in the future. 
- [x] Consider dropping `socket.io` in favor of `ws` if we don't need anything socketio-specific
- [x] Silent WebSocket connection errors if the remote mode is enabled but the WebSocket server is not running/fails to connect. Print a warning and continue in regular mode. 
- [x] Don't send life-cycle events for the WebSocket connection HTTP request to the WebSocket server (`forwardLifeCycleEvents()`).
- [x] Support `ListenOptions` on `setupRemoteServer()`, like `onUnhandledRequest`.
- [x] Fix the `spyOnLifeCycleEvents` test utility importing `vitest` when it can also be run in the browser. 
- [ ] Fix forwarded LCE order sensitivity https://github.com/mswjs/msw/pull/1617#issuecomment-2580999914
- [ ] Protect the internal HTTP connection form malicious actors (https://github.com/mswjs/msw/pull/1617#pullrequestreview-2937795837).
- [ ] Implement the `RemoteInterceptor`. Then, use it in MSW via `InterceptorNetworkSource`.
- [ ] ⚠️  **Clean up the code**. `/new` has to be merged with `/core`.

## Blockers 

- [x] https://github.com/socketio/socket.io/issues/4621 (type definitions for `socket.io-parser` are broken for the CJS build).
- [x] Modern Remix requires sourcemaps. I added those for Interceptors but adding them to MSW breaks due to tsup: https://github.com/egoist/tsup/issues/1073
  - Fixed in https://github.com/mswjs/msw/pull/1958
- [x] Support different tests provisioning overrides for the same running app (https://github.com/mswjs/msw/pull/1617#issuecomment-2332499183). 
- [x] Implement life cycle events forwarding to remote.


src/node/setupRemoteServer.ts
```diff
@@ -0,0 +1,212 @@
...
+${`${pragma} ${header}`}
+  Declaration: ${callFrame}
+`)
+    })
+  }
+
+  public async close(): Promise<void> {
+    const { syncServer } = globalThis
+
+    invariant(
```
kettanaito: Should use `devUtils` here to maintain a consistent library error format.

src/node/setupRemoteServer.ts
```diff
@@ -0,0 +1,212 @@
...
+        )
+      })
+
+      /**
+       * @todo Have the socket signal back whichever response
+       * was used for whichever request. Include request ID
+       * and somehow let this API know whether the response was
+       * the mocked one or note.
+       */
+      // socket.on('response', (serializedResponse) => {
```
kettanaito: The client socket must send some read-only events indicating that the response has occurred.

src/node/setupRemoteServer.ts
```diff
@@ -0,0 +1,212 @@
...
+  SerializedRequest,
+  SerializedResponse,
+  deserializeRequest,
+  deserializeResponse,
+  serializeRequest,
+  serializeResponse,
+} from '~/core/utils/request/serializeUtils'
+import { LifeCycleEventEmitter } from '~/core/sharedOptions'
+
+export const SYNC_SERVER_URL = new URL('http://localhost:50222')
```
kettanaito: I wonder if we can rely on a fixed port to always be available. I'm tempted to say yes but we should probably handle the scenario when this part is busy.

kettanaito:
## ESM in DeferredPromise

Looks like the package doesn't ship as ESM, breaking MSW's ESM build, respectively. Need to add an ESM support there.

kristojorg:
Interesting approach with runtimes and interceptors. I am worried that maintaining so many runtimes is already complex for users to understand, and will only grow more complex as frameworks develop. In order to understand how to mock a request in Next.js or Remix, a user has to understand where the request is being issued (client or server), what process that code is running in (is it a separate node process, is it connected to my test runner somehow?), and what runtime it is using (browser, Nodejs or "edge" like Next.js middleware). I wonder if a unified mock server could simplify the mental model for writing tests. Something like the http-middleware from v1. 

Drawbacks:
- A mock server can only intercept requests to one domain.
- The mock server would still need to be started by the test runner for test-time mocking.
- Maintenance can be challenging but could be eased with a fetch-native API and MSW integration.

An alternative would be to try and hide these runtime and communication details from users more, possibly with adapters for each framework. 

Eager to dive into the code and contribute. Great work as always!

nickserv:
I believe Artem's intention is for handlers to be interchangeable between client and server environments. You should be able to assign handlers to a constant and call it with both `setupRemoteServer` and `setupServer`, though I think it would be more convenient for `setupRemoteServer` to do this automatically.

kettanaito:
> I am worried that maintaining so many runtimes is already complex for users to understand, and will only grow more complex as frameworks develop.

To be clear, this feature introduces no additional runtimes. The inter-process interception is still meant for Node.js but you can, technically, combine tests in Node and runtime in something like React Native and that should still work. 

You only need MSW in your development runtime. And that is still JavaScript, which is either the browser or Node, so in this regard nothing changes, even if you're developing edge functions or Cloudflare Workers, etc. _You are still developing/testing them in Node.js_. 

> In order to understand how to mock a request in Next.js or Remix, a user has to understand where the request is being issued (client or server), what process that code is running in (is it a separate node process, is it connected to my test runner somehow?), and what runtime it is using (browser, Nodejs or "edge" like Next.js middleware).

The user doesn't have to know any of that. Request handlers remain environment-agnostic, the only thing that changes is this:

- If I want to affect a network of _another process_, I reach out for this new special `setupRemoteServer()` function _in my tests_. 

This feature doesn't discard previously existing practices, it simply expands the capabilities of your tests for better server-side network control.

> I wonder if a unified mock server could simplify the mental model for writing tests. Something like the http-middleware from v1.

Standalone HTTP server for mocking has a number of disadvantages and I would only recommend it in particular scenarios. That's why you still have `@mswjs/http-middleware` and V1/V2 combination if you need that setup. It should not be the default. 

This feature unifies the mocking experience even more, because if you take a look at its implementation, you can notice how very little changes here. MSW is a very simple tool. It needs a source of network to get requests, and it resolves those requests against the list of request handlers. It's been like that 5 years ago, and it's still like that today. What changes is that the new `setupRemoteServer` API allows an independent process to become the source of the network, similar to how the Service Worker API allows the browser to become the source of the network. I find it rather beautiful. 

> An alternative would be to try and hide these runtime and communication details from users more, possibly with adapters for each framework.

I appreciate your advice but time has taught me that this is a slippery slope. What may seem like a better direction often leads to indirect complexity leaking to the end user. What I tend to do instead, is base my decisions on the established principles of MSW. Those served me well, they still do today and will continue so after this feature is merged. I don't want folks to configure adapters or runtimes. That's completely unnecessary and MSW has proven it so. 

But that's the beauty in its API design: _you still can have all that if you want_. You can abstract to whichever degree you feel comfortable, which includes designing your configurations, publishing your setup to the internal package registry for other teams, making it HTTP server-first, etc. Support strong defaults, allow opinionated customization. That's the key. 

> Eager to dive into the code and contribute. Great work as always!

Thanks! I will certainly need some help with testing this once the implementation is ready. Hope to give it more attention later this/next month.

kettanaito:
@nickmccurdy is completely correct. All `setupRemoteServer` does is introduce an additional source for the network requests. Here are some possible usage combinations for this setup that can help you understand the intention behind this feature better:

## Setup 1: Test-driven network

- App: `setupServer()` + no request handlers.
- Test: `setupRemoteServer()` + request handlers.

This way the test controls the server-side traffic of the tested application entirely. 

## Setup 2: Mixed network

- App: `setupServer()` + happy-path handlers.
- Test: `setupRemoteServer()` + per-test overrides with `.use()` to test deviating behaviors like network errors or error responses.

This is what you'd call a traditional testing setup for Node.js right now. It's still applicable to this inter-process testing story in the same shape. 

## Setup 3: App-driven network

- App: `setupServer()` + request handlers.
- Test: `setupRemoteServer()` + no handlers at all.

This allows to strategically turn on/off the network in your app for whichever requests you wish, while the tests (or any other process, like a script) can _monitor_ the outgoing traffic.

kettanaito:
Also, a thought:

```diff
-setupRemoteServer()
+setupServer({ remote: true })
+setupServer({ remote: options })
```

kettanaito:
I think the remote style of connection must be an explicit choice by providing the `remote: true` option on the `server.listen()` call on the _application's_ side:

```js
// app/entry.server.js
server.listen({
  // By doing this, I tell MSW that I expect this application's network
  // to be controlled by the remote instance of MSW (e.g. in my tests). 
  remote: true
})
```

This way we'd save a bit on roundtrips of having to wait for the remote request handler if the `remote` is set to `false`, and have that entire fixed handler skipped altogether. 

This is also where we may introduce some identity checks. It'd be a good idea to allow the developer to specify what exact process to connect to. Granted, that should be controlled on the `setupRemoteServer()`'s side (e.g. in the tests), and it needs some reliable way to establish said identification since things like `process.pid` are non-reproducible and random by design, fitting poorly for this purpose.

kettanaito:
## `ReadableStream` support in `WebSocket`

### `WebSocketStream` API in Chrome

- [`WebSocketStream` from Google labs](https://developer.chrome.com/docs/capabilities/web-apis/websocketstream)

Chrome is working on the spec for a new `WebSocketStream` API that would be nice to use here but it's not yet even experimental (simply doesn't exist in the stable Chrome). 

Looks like for now, we'd have to map stream pushes to the WebSocket events manually. 

### `websocket-stream` package

The [`websocket-stream`](https://github.com/max-mapper/websocket-stream) package seems to implement what we need. It promises both Node.js and browser support, which is nice. My only concern is that I don't see it directly mentioning you can use `ReadableStream` with it. I hope you can.

kettanaito:
## Update the `.use()` API is failing

Looks like the `setupServer()` is having trouble connecting to the remote WebSocket server for some reason. Needs investigation. 

### Reproduction steps

```
pnpm build
pnpm test:node pnpm test:node test/node/msw-api/setup-remote-server/use.node.test.ts
```

Expected both tests to pass. The second one, the one using the override, is now failing.

chrisb2244:
I think one of the tests fails because the test wasn't updated to pass the port value to the `listen` call.

Separately, the very long repeated logs can be solved (iirc) by testing for socket requests and passing the results through (#2041 has changes to `SetupServerCommonApi.ts` from around line 73 in 3ea224f which I believe was to prevent this recursive problem).

Phoenixmatrix:
Just saw this PR. Since we're doing playwright integration testing in Remix and want to mock our actual backend (not the Remix backend for frontend), we're kind of reimplementing a poor man's version. Looking forward to when its picked back up!

chrisb2244:
@Phoenixmatrix I'm unsure if you're willing to try unpublished versions, but if you wanted to try out my PR at https://github.com/mswjs/msw/pull/2041 I'd be grateful for any advance feedback (@kettanaito is I think busy with various other PRs and activities currently, but if you can point out any problems or confirm that it works as you'd expect, I can make changes if needed to try and smooth the merge).

ar4hc:
FYI:
There is a possible workaround for this i found during my search for this problem:
[Test your SvelteKit app end-to-end with Playwright and MSW](https://spuxx.dev/blog/2024/sveltekit-playwright-msw/)

spuxx-dev:
> FYI: There is a possible workaround for this i found during my search for this problem: [Test your SvelteKit app end-to-end with Playwright and MSW](https://spuxx.dev/blog/2024/sveltekit-playwright-msw/)

Glad that my post was of some help to you! :smile:

kettanaito:
## Updates

I've had some time to work on this. Sharing the updates below. 

- Simplified and fixed the client sync connection in `setupServer`. Introduced the `beforeRequest` async hook that allows `setupServer` to fire a side effect (sync server connection) before the first intercepted request is being handled.
- `RemoteRequestHandler` now correctly support unhandled requests. This is done by moving the response await logic to the `parse` phase, and then deciding whether the remote handler is matching in `predicate` based on the received mocked response from the remote. This way, the remote request handler is no longer all-matching, but instead the remote process decides whether the remote handler will match at all. 
- Unified how `x-msw-intention` header works. Made the header name a constant, made its possible values an enum. Updates the usage code.
- Polished the tests.

## On test/app relationship conundrum

I believe I found a way to fix the test/app catch 22 problem (app depending on test to handle initial `GET /`, test won't run until the app responds to `GET /` during the testing framework pinging the app to know if it's running). 

We can circumvent this problem by attaching a permissive request handler that will resolve _all requests as `200 OK`_. This is relevant only for the ping request made by the testing framework. The framework doesn't care about the validity of your homepage, only that the app responded (i.e. is running). 

We can try utilizing one-time request handlers for this. Basically, the first time any resource is loaded, it gets `200 OK` as a mocked response. The subsequent loads (those happening then you visit your app in tests) are not affected. 

We don't have the means to create such handlers as of now, since a one-time handler will self-terminate on the first matching resource. We'd probably need to create something like

```js
const handlers = [
  http.all('*', ({ request }) => {
    handlers.unshift(http[request.method](request.url, () => new Response(), { once: true }))
  }
]
```

> So a permissive handler that will prepend one-time handlers specifically for intercepted requests. Then, this permissive handler would have to self-terminate as well somehow. This will be done in the `SetupServer` context, so we can use `this.handlerController.prepend()` to prepend such handlers.

kettanaito:
Before I am to ship a questionable workaround to that catch 22 problem, I will try to solve it on the ecosystem level first. I've already raised this question in the Playwright repo (https://github.com/microsoft/playwright/issues/28893), that making a HEAD/GET request to ping the app server is undesirable. 

I've also opened https://github.com/jeffbski/wait-on/issues/163 to `wait-on`, which is a dependency in `start-server-and-test` widely used in Cypress E2E testing, to consider the same thing (switching from a HEAD/GET request to `net.connect()` ping). Let's see where this leads us.

Phoenixmatrix:
I discussed it in the other PR, but do you have thoughts on how this will handle multiple Playwright workers? There's going to be multiple socket servers (running within the test workers), but only one "app server", where msw/node is running, which then needs to know which worker is making the request to handle to connect to the appropriate web socket server.

kettanaito:
@Phoenixmatrix, I must've missed your comment. That's a great concern! 

At the present state, there's no support for that. We need to design one. 

Initial guess: I think it can be solved by using `process.pid` as an identifier for individual workers that Playwright spawns. The workers won't be making any requests as that would be your app, but instead there has to be some mapping between these:

```js
// my.test.ts

test('...', async ({ page }) => {
  // 2. ...should be handled by these handlers.
  remote.use(http.get('/resource', resolver))

  // 1. Make MSW understand that any server-side requests
  // issued as a part of this below...
  await page.goto('/')
})
```

This will be tricky because the app is just one, while there may be multiple tests running in parallel in different workers. 

Basically, we need some sort of ID to be the same between the test (the worker) and the app's runtime (a browser tab, ideally). The biggest challenge here is to achieve this while not shipping any Playwright-specific logic! 

Need to think and explore this in more detail before arriving at any conclusion. If you have any thoughts, please share.

Phoenixmatrix:
Thanks! What we did here is having a userland implementation (since I didn't want to fork MSW), and I added ecosystem specific adapters. One to decorate the Remix context (we use Remix), one as a Wretch plugin (Wretch being the http client we use), and a Playwright helper in charge of setting a header. Then the header has the port of the socket server, playwright ensures all requests have it, Remix forwards it via context, and the http client built from context adds the header. Then my socket client looks for the header to connect to the right server.

What I was thinking was to leave those implementations details to framework authors or as an implementation detail/recipe. Then MSW only needs to do 2 things:

A) expose the createServer function (already done)
B) provide users of MSW with a predicate in the form `(r: Request) => ClientConfig`, where ClientConfig would be, in its simplest form, an object of the form `{ port: number}`. Then people can create a server wherever they want, get its port, forward it along their environment following a community provided recipe, and then do something like `(r) => { port: Number(r.headers["my-cool-header-with-port")) }`

Then community packages could be provided to streamline this even more (I made a "msw-remix-remote-server" in our internal monorepo with all the pieces).

Is it elegant? No, not one bit.
But it works for every use case and environment I can think of, without polluting MSW with implementation specific stuff (it also solves your catch 22 with Playwright above, since you simply don't handle a request if the predicate returns a falsy value or an empty config object)

If you can figure out something better, then definitely go a different route, but the above work well in a medium sized production app right now, so I guess its the "worse case scenario if we can't figure out any better". The pid strategy does look interesting.

kettanaito:
What is interesting about this issue is that it's not Playwright-specific. It's a fundamental functionality we should provide as a part of the feature itself. And yeah, it has to be framework-agnostic. Your tests are likely running against the same instance of the app so, naturally, you want some sort of isolation on a per-test basis. There are a few things we can use as an ID of the page, but the problem is that there's no easy way to associate that particular page with a test that opens it. 

Even a single test suite can have different behaviors for the same request:

```js
test('one', async () => {
  await page.goto('/')
})

test('two', async () => {
  remote.use(http.get('/resource', resolver))
  await page.goto('/')
})

test('three', async () => {
  remote.use(http.get('/resource', differentResolver))
  await page.goto('/')
})
```

And since the app doesn't share anything with the test's scope, we cannot use things like `server.boundary()` here to provide the said isolation. Tricky, tricky.

Phoenixmatrix:
Yup! Not only is it not Playwright specific, but even in Playwright, there might be other things (in our case, Remix and our http client) that happen in between and that may not be able to be monkeypatched. That's why I think it has to be up to the environment to forward along the required information with the request, and let the user of MSW teach it how to fetch that information back. 

So for Playwright, we used:

```
// in beforeEach
context.setExtraHTTPHeaders({
  "x-msw-server-port": String(address.port) //address.port comes from the remoteServer instance
});
```

And then tweak the environment to forward that header along across all the hops (on our case the above takes care of the Remix useFetcher, but we had to add code for our http client running on the server). If its a microservice app, there could be multiple hops in multiple programming languages before hitting the Node server we're trying to mock!

What IS cool, is that the handler itself runs in the Playwright worker, so the handlers are automatically isolated from each other. So if I have state used across handlers in a single spec files, and want to share them across these, but not across other handlers in another file, it is now trivial. My handlers are isolated in their processes thanks to the remote server architecture.

It dramatically simplifies our tests.

kettanaito:
Your suggestion is interesting, it just flips the request flow upside-down. You cannot access a request reference in `server.listen()`. We can do whatever we want in the handlers though, particularly in the internal `RemoteRequestHandler`. I think I will take the multiple Playwright tests above as the base scenario. If we can solve that, we can likely solve any concurrency issues with this feature.

Phoenixmatrix:
>  You cannot access a request reference in server.listen()

You cannot, that's why I was suggesting a callback. Pass a callback in listen() as an option, and send THAT down to the internals to be used by the handler, to be evaluated on each request.

Alternatively, the way I handled it in our code base, is I just created a special "remoteRequestHandler" factory similar to the one in this PR and takes the callback as argument, but its up to the user of MSW to add it to their handlers array in the Node server. Then it can be configured however they want, and added wherever they want.

For us, we needed it a the beginning of our handlers because we want it to override our old static handlers when a handler is defined in Playwright. Some other people might want the opposite.

Less magical, but powerful.

kettanaito:
I meant conceptually, not just implementation-wise. MSW follows a list of strict rules of what can be done where, and I'd like to keep following those rules. Sure, they make my life way harder. But as a result, everyone gets a consistent usage experience. One of such rules is that the only way to affect the network is through a request handler. I'm not in favor of userland packages violating those rules either, as I wouldn't want your MSW experience to suddenly become different if you are using a userland package. 

There are fundamentally two issues here:

1. Prevent individual tests from using the `remote` instance as a shared state. _This_ can be solved with `server.boundary()` since Playwright tests are running in Node.js. Won't work for Cypress tests that run in the browser (no `AsyncLocalStorage` support there just yet). 
2. Associate certain page runtimes with certain network behaviors. 

_Edit:_ Yeah, my head spins from all of this. You do `page.goto()` but that hits the _server_ first before any page is created.

kettanaito:
Here's one idea:

```ts
// my.test.ts
test('...', async ({ page }) => {
  // `server.use()` generates a unique ID and wraps all its runtime handlers with it.
  // You get that ID back.
  const contextId = remote.use(override)

  // Next, you can decorate any request with that ID.
  // The remaining part is that the `setupServer()` counterpart would read that ID
  // and forward it back to the `remote` to use in predicate.
  await page.goto('/', { referer: contextId })
})

// app.js
// The problem with the app, is that "GET /" that triggers the server
// won't share any of its data, like headers, with any resources the server
// has to fetch to handle the "GET /" request.
```

Phoenixmatrix:
I like that api.  Though how does the counterpart knows the port of the socket servers that are available to connect to? With just the contextIds they still wouldn't know which socket servers are available, right?

Unless you went all in and have a socket -server- in the app on a known port and the remotes running in playwright would use that when you call remote.use to announce themselves (I also considered doing this via local files or named pipes in environments that support them)

PS: the hardest thing when discussing this is the terminology, haha. In my code, having a socket client running in a server handler and a socket server running in a client test makes my head hurts, and code reviewers cry.

kettanaito:
> Though how does the counterpart knows the port of the socket servers that are available to connect to?

The socket server is the same, there's only one being spawned. The request differentiation based on the context ID is done in the response resolvers. If a resolver receives a request decorated with a context ID other than the one associated with this resolver, it ignores it. All handlers still run for all requests, but only the same-context handlers take effect, providing that context collocation.

kettanaito:
## Support in `setupWorker`

One of the community members have raised a great point: it would be nice to be able to apply handlers to a _browser_ process as well. There is no way to pass runtime handlers from a Playwright test to the browser unless you expose both `worker` and `http`/`graphql` to create the handlers. You cannot serialize functions in `page.evaluate()`. 

I should consider extending `setupWorker` to also support the `remote: true` option. If provided, it will behave similarly to `setupServer`'s handling of that option. Any intercepted client-side requests (after they arrive from the worker) will be sent over WebSocket to the remote sync server to handle first. 

```ts
// your/app.ts
const worker = setupWorker({ remote: true })
```
```ts
// playwright.test.ts
import { http } from 'msw'

const remote = setupRemoteServer()
// ...start and stop it.

test('', async () => {
  remote.use(
    http.get('/resource', override)
  )
})
```

This will finally solve the DX gap in modifying client-side network behavior in E2E tests!

niccholaspage:
I’ve went ahead and made a sample repo with Remix, Playwright, and MSW that does both server and client side data fetching from the same URL, which can be used as a test bed for this functionality: https://github.com/niccholaspage/remix-msw-playwright-sandbox

I will try to use this PR over the next couple days with it and see how it works out, and try to think through how we could achieve isolation of handler overrides per test like we’ve been chatting about.

SebastianSedzik:
@kettanaito When integrating `setupRemoteServer` into a simple repository (Express + MSW + Playwright) ([repo](https://github.com/SebastianSedzik/msw-remote-server-playwright-example), [PR](https://github.com/SebastianSedzik/msw-remote-server-playwright-example/pull/2)), I noticed that the current implementation requires `setupRemoteServer` to be started before `setupServer`. Is this the intended behaviour?

It might be beneficial to allow `setupServer.listen({ remote: true })` to run without errors when `setupRemoteServer` is not running yet, and allow attaching `setupRemoteServer` later. This could simplify integration with Playwright's [webServer](https://playwright.dev/docs/test-webserver) option, (following a flow like: start the mocked server app (`setupServer.listen({ remote: true })`) -> (before all tests) call `setupRemoteServer` -> run the test suite). WDYT?

kettanaito:
Hi, @SebastianSedzik. Thanks for sharing your feedback!

Right now, we have the design that you need to start the remote part _before_ you start the regular `setupServer`. That is done so the network behavior is deterministic, and the remote counterpart can control it from the start, as opposed to kicking in somewhere in the middle of the app's/test's runtime. 

`server.listen()` is _synchronous_, so there is no way for you to await it having the remote counterpart. Leaving your test running while the internal WebSocket connection is being established means flaky tests. 

There is an issue with the E2E integration as those testing frameworks would often do a `GET` request to ping your app (where they should've just established a socket connection, but well), which creates a catch 22 between the app runtime and the test's runtime. I'm thinking of solving it by ignoring the initial server-side requests with `setupServer`, if that proves possible.

SebastianSedzik:
@kettanaito Thank a lot for detailed clarification!
This design does appear more deterministic, but it limits the ability to connect multiple remote parts within the same application/process (`setupServer`). This could be useful when running Playwright tests using multiple workers (as they cannot share a class instance like `remoteServer` between them, and needs to create different instance per worker).
Consequently, users will either be restricted to running tests with a single worker or will need to spawn a separate application for each worker, which may lead to higher resource consumption depending on the application.

arjenbloemsma:
@kettanaito Thanks for all you hard work. Much appreciated.

Will this feature also allow to mock calls made by Next.js api routes?

So imagine I have a client component that calls a nextjs api route on the server, which in turn will make a call to a third party app. I would like to intercept the call to the 3rd party app with MSW and mock it.

Or is there already another solution for that which I've overlooked?

kettanaito:
@SebastianSedzik, yes, the intention is to iterate on the current design to allow multiple remote connections to the same `setupServer` instance. There are still problems with differentiating between those connections as there is no way to know what test has triggered a request to respond to it appropriately. 

@arjenbloemsma, 

> Will this feature also allow to mock calls made by Next.js api routes?

This feature will allow you to control the network in one Node.js (e.g. the server-side of your Next.js) from another (e.g. your test). 

> So imagine I have a client component that calls a nextjs api route on the server, which in turn will make a call to a third party app. I would like to intercept the call to the 3rd party app with MSW and mock it.

For in-application usage, this already works. Check out this [Next.js + MSW example](https://github.com/mswjs/examples/pull/101) for the full setup to enable this.

kettanaito:
I have some thoughts regarding the application spawn order and identifying individual tests/workers. Will share once I have a prototype confirming my ideas. Looks promising.

spuxx-dev:
This is really great. Thank you for your hard work! 🎉

kettanaito:
The tests are failing because I'm on pnpm 9 and #2211 isn't merged yet.

kettanaito:
## Blockers

- [x] https://github.com/mswjs/msw/pull/2211
- [x] Update Vitest to v2 to support disposable objects.

kettanaito:
## Update

Alright, there's a lot to unpack. In short, I've got the MVP of CPRI completely working. This includes the test/app paradox solved, the association between the app's runtime and a particular test/worker closure, and the ability to control the server-side network from the test.

Documenting those decisions for posterity.

### Test/app paradox

Solved by two things:

1. You have to build your app _before_ the entire test run. Use the global setup hooks to achieve that.
2. You have to start individual instances of your built app in each individual test case. 

Not only does this solve the test/app paradox because it guarantees the fixed order of things (first test, then `remote`, then your app spawns), it allows the app to be spawned with a custom `contextId`, which is important for the next thing.

### Test/app binding

Even when spawned within individual test cases, your app still communicates with a _single_ instance of `remote` (to save bandwidth on spawning a ton of WebSocket servers, those are redundant). This makes `remote` a _shared state_ across different test scenarios, which isn't nice (but it is nice to have it as a shared fallback!).

This is solved by using the `remote.boundary()` API. Similar to the existing `server.boundary()`, the remote boundary utilizes `AsyncLocalStorage` in Node.js to bind request handler overrides to the test's closure, preventing the shared state issue and allowing for concurrent test runs. In the remote context, the boundary has additional behavior because _request resolution doesn't happen within the test's closure_ like it does for regular Node.js tests. Instead, it happens in the `request` listener of the WebSocket server, which is "visually" in `await remote.listen()`.

Test/app binding occurs due to `remote.contextId` grabbing the unique ID of the particular boundary and storing it in a `Map<ContextId, () => Context>` internal map of the `remote`. During the request resolution (i.e. the `request` event handling), the API can lookup any async context by its ID and get the list of relevant handlers. Only those handlers will be used (the happy-path handlers from `setupRemoteServer` are still applied as with `server.boundary()`). 

## What's next?

A ton of todo's to cover! The public API isn't the most ergonomic thing, and I'm considering simplifying it quite a bit while exposing the customization options to the end developer. There are also some security considerations I want to address, mostly concerning safe defaults of the WebSocket server CORS. 

This is also a good time to [sponsor](https://github.com/sponsors/mswjs) this effort. Thank you.

kettanaito:
## Bug: Empty buffer sent to WebSocket connection

```
[setupRemoteServer] ws client ERROR! Error: got binary data when not reconstructing a packet
    at Decoder.add (/msw/node_modules/.pnpm/socket.io-parser@4.2.4/node_modules/socket.io-parser/build/cjs/index.js:161:23)
    at Client.ondata (/msw/node_modules/.pnpm/socket.io@4.7.5/node_modules/socket.io/dist/client.js:182:26)
    at Socket.emit (node:events:518:28)
    at Socket.onPacket (/msw/node_modules/.pnpm/engine.io@6.5.4/node_modules/engine.io/build/socket.js:120:22)
    at WebSocket.emit (node:events:518:28)
    at WebSocket.onPacket (/msw/node_modules/.pnpm/engine.io@6.5.4/node_modules/engine.io/build/transport.js:94:14)
    at WebSocket.onData (/msw/node_modules/.pnpm/engine.io@6.5.4/node_modules/engine.io/build/transport.js:103:14)
    at WebSocket.<anonymous> (/msw/node_modules/.pnpm/engine.io@6.5.4/node_modules/engine.io/build/transports/websocket.js:20:19)
    at WebSocket.emit (node:events:518:28)
    at Receiver.emit (node:events:518:28)
    at Receiver.dataMessage (/msw/node_modules/.pnpm/ws@8.11.0/node_modules/ws/lib/receiver.js:514:14)
    at Receiver.getData (/msw/node_modules/.pnpm/ws@8.11.0/node_modules/ws/lib/receiver.js:446:17)
    at Receiver.startLoop (/msw/node_modules/.pnpm/ws@8.11.0/node_modules/ws/lib/receiver.js:148:22)
    at Receiver._write (/msw/node_modules/.pnpm/ws@8.11.0/node_modules/ws/lib/receiver.js:83:10)
    at writeOrBuffer (node:internal/streams/writable:564:12)
    at _write (node:internal/streams/writable:493:10)
    at Receiver.Writable.write (node:internal/streams/writable:502:10)
    at Socket.socketOnData (/msw/node_modules/.pnpm/ws@8.11.0/node_modules/ws/lib/websocket.js:1272:35)
    at Socket.emit (node:events:518:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Socket.Readable.push (node:internal/streams/readable:390:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
```

No idea why this happens. Something down the line sends an empty buffer to the server, SocketIO tries to emit the `message` with that buffer, and fails on the top-most frame. 


### Root cause

https://github.com/mswjs/msw/blob/fa8a197434dcc446ebffbd1795356afa31afcf62/src/node/SetupServerApi.ts#L191

Forwarding the `response:bypass` event for the WebSocket handshake to the remote causes this error:

```
{
  type: 'response:bypass',
  payload: {
    response: {
      __serializedType: 'response',
      status: 101,
      statusText: 'Switching Protocols',
      headers: [Array],
      body: [ArrayBuffer]
    },
    request: {
      __serializedType: 'request',
      method: 'GET',
      url: 'http://localhost:56957/socket.io/?EIO=4&transport=websocket',
      headers: [Array],
      body: undefined
    },
    requestId: '16ddfd8d917a6'
  }
}
```

Note that `body` of the response if an empty `ArrayBuffer`. Sending an empty buffer errors the SocketIO parsing phase.

kettanaito:
## Bug: Deep objects cannot be emitted with SocketIO

Not sure if it's MSW interfering in some way, but emitting the `request` event with the deep object as an argument never arrives to the server. Shallow objects work fine. 

Would be nice to add an integration test to Interceptors to confirm or debunk this. 

## ✅ Solved

Moved away from WebSockets to HTTP for the internal server.

kettanaito:
## Test failures

```
 FAIL  test/node/msw-api/setup-remote-server/response.body.test.ts [ test/node/msw-api/setup-remote-server/response.body.test.ts ]
 Test Files  2 failed | 84 passed (86)
Error: listen EADDRINUSE: address already in use ::1:56957
```

This is caused by migrating to HTTP for the internal server. Since Vitest likely runs some of these tests in parallel, it attempts to call `remote.listen()`, which attempts to create multiple internal HTTP servers on the same port, which, naturally, fails. 

This is a good precursor into figuring out whether the internal server port is fixed or random. I likely lean toward making it random and passing it along with `remoteContext` through `remote.boundary()` (but that will make the boundary required; or we can use a fallback port). 

### Solution

Solved by (1) spawning the internal server at a random port; (2) propagating the entire server URL through _environment_ variable (`getRemoteContext()` + per-test app spawning). Boundary is not required.

kettanaito:
## Task: Ignore internal requests in life-cycle reporting

Internal MSW requests is a new thing with this pull request, such as a request to the remote HTTP server, or internal requests that forward life-cycle events. Those internal requests bubble to the life-cycle event emitting, and result in entries there. 

Need to add some flag to ignore certain requests from the life-cycle API altogether. It may be an `accept` header once again. 

### ✅ Solved by introducing a check for the `msw/internal` accept header value.

kettanaito:
## 🐞 ECONNREFUSED on life-cycle event forwarding

Sometimes the following error is thrown while using the new API:

```
Error: connect ECONNREFUSED ::1:50004
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1555:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '::1',
  port: 50004
}


stderr | Socket.<anonymous> (msw/test/node/msw-api/setup-remote-server/utils.ts:25:43)
msw/lib/node/index.js:556
        new Error(
        ^

Error: Failed to forward life-cycle event "request:match" for request "GET https://example.com/resource": unexpected error. There's likely additional information above.
    at ClientRequest.<anonymous> (msw/lib/node/index.js:556:9)
```

kettanaito:
## 🐞 Life-cycle event forwarding order sensitivity

LCE are sensitive to order. But when forwarded, some events may arrive faster over HTTP, breaking that order:

```diff
 FAIL  test/node/msw-api/setup-remote-server/life-cycle-event-forwarding.node.test.ts > emits correct events for the request handled in the test process
AssertionError: expected [ …(4) ] to deeply equal [ …(4) ]

- Expected
+ Received

  Array [
    Array [
      "[request:start] GET https://example.com/resource ade0140740d15",
    ],
    Array [
-     "[request:match] GET https://example.com/resource ade0140740d15",
+     "[request:end] GET https://example.com/resource ade0140740d15",
    ],
    Array [
-     "[request:end] GET https://example.com/resource ade0140740d15",
+     "[request:match] GET https://example.com/resource ade0140740d15",
    ],
    Array [
      "[response:mocked] GET https://example.com/resource ade0140740d15 200 {\"mocked\":true}",
    ],
  ]
```

Need to solve this.

src/node/setupRemoteServer.ts
```diff
@@ -0,0 +1,402 @@
...
+        return
+      }
+
+      const request = new Request(requestUrl, {
+        method: incoming.method,
+        body:
+          incoming.method !== 'HEAD' && incoming.method !== 'GET'
+            ? (Readable.toWeb(incoming) as ReadableStream<unknown>)
+            : null,
+      })
```
alitas:
```suggestion
        duplex: incoming.method !== 'HEAD' && incoming.method !== 'GET'
            ? 'half'
            : undefined,
      })
```
We can't send a stream in the body without explicitly defining the duplex, otherwise you get `Uncaught TypeError: RequestInit: duplex option is required when sending a body.`. See https://github.com/nodejs/node/issues/46221

kettanaito: Oh, it's this issue again! Thanks for pointing it out.
We can also set `duplex` all the time, as it doesn't seem to error in case of requests without any body.

kettanaito: Pushed the fix at 3c76156efe769f6d9406314909e82070f90d8649

alitas: I'm also getting the following error, followed by a server crash, possibly because the request body is being used elsewhere:

```
TypeError: unusable
    at Request.clone
    at serializeFetchRequest
    at handleLifeCycleEvent
```

alitas: Looks like the ReadableStream is closed: `ReadableStream { locked: false, state: 'closed', supportsBYOB: false }`

kettanaito: Can you tell me more about the use case? I'm not hitting this one.

AriPerkkio (commented):
> ## Concerns
>
> * Is it safe to introduce a WebSocket server that will be, effectively, routing HTTP messages over the local network (during tests only)?
>   
>   * Yes. If someone can intercept that WebSocket communication, they are already in your machine and can do things far worse than that.

When adding new WebSocket server, please double-check that it's protected by Cross-site WebSocket hijacking too. I'm mentioning this explicitly as it's been common flaw in building and testing tools.

- https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6
- https://github.com/vitest-dev/vitest/security/advisories/GHSA-9crc-q9x8-hgqq

tldr; is that any website was able to connect to Vite and Vitest websocket server by opening connection to `localhost`, e.g. `new WebSocket('ws://localhost:51204/__vitest_api__')`. The HTTP APIs would reject such connections automatically, but WS didn't. Imagine someone hosting `cool-msw-tips.com` that checks if MSW related WS servers are open.

MilosRasic:
Great stuff. I really hope this makes it into a release.

A question, though. Wouldn't it be easier to expose a simple HTTP API to control mocking remotely? With websockets, a service that doesn't expose a websocket API suddenly has to worry about having the infra to support websockets only because it uses them for testing.

I'm primarily looking at this from the angle of testing your service when it's deployed to a testing environment, not only running the service and the tests on your machine.

kettanaito:
@MilosRasic, I've started this feature over WebSockets but has long moved to using HTTP as the protocol for remote interception. I plan on releasing it that way, too. We will likely migrate to WebSockets anyway to support intercepting WebSockets as event-driven protocol will capture that use case more elegantly while still allowing for HTTP interception. 

These protocols are meant as purely internal though. Regardless of what MSW choses to use under the hood, it will not something you will be interacting with in any way. 

**Your request handlers remain the only means to control the network**. These HTTP or WebSocket servers are internal and serve as the _source_ of the network, not the means to control it.

MilosRasic:
@kettanaito I think I understand, but now I'm not sure any more :slightly_smiling_face: 

If my service is running on Machine A and my tests on Machine T, the two would communicate using the protocol you have chosen for implementation, correct?

If that happens to be WebSockets, I could encounter issues if my infrastructure wasn't set up to support them, since my service on Machine A doesn't expose a WebSockets API as part of its own API.

That's the only concern I have.

kettanaito:
While that's technically possible, both HTTP and WebSocket are globally available protocols in Node.js. MSW doesn't support custom environments. If you end up running one, it becomes your responsibility to procure all the necessary things for MSW to work (or just skip the features that don't). Does this make sense?

MilosRasic:
Thanks @kettanaito 

Upon further consideration, I don't think either HTTP or WebSocket control would be a match for testing in a real production-like environment. If MSW's internal HTTP request or WebSocket connection is handled by one instance of the service under test in my testing environment, only that instance will be configured to use the MSW mock. How can I guarantee that the test request will be handled by the same instance?

I don't think it can work, but this will still be handy for those who wish to test only locally or in CI.

kettanaito:
@MilosRasic, not sure which use case you have in mind, but cross-process request interception is agnostic of how your system/test is structured. Neither is it limited to tests. You use it to intercept and control network of one process from another. 

Speaking of testing in particular, CPRI does require you to provision test case - app instance isolation to prevent your often _single_ running app instance from acting as a shared state (imagine two test cases with CPRI listing different network behaviors. Ouch!). That's precisely why test cases in this pull request rely on `server.boundary()` as one part of provisioning that isolation. 

I intend CPRI primarily for testing full-stack behaviors, like React Server Components. In that context, you'd want to do two things to ensure reliable tests:

1. Spawn a new instance of your app _per test case_.
2. Rely on `server.boundary` + `remoteContext` from MSW to bind test case to a particular app runtime.

You can see both of these in action in [`next-rsc-testing`](https://github.com/kettanaito/nextjs-rsc-testing) example repo. Hope this helps.

lewnelson:
Hey @kettanaito I was just reading through the history of this pr as I've just published a somewhat similar solution that leverages MSW called [Mocky Balboa](https://docs.mockybalboa.com/). The core use case is to mock outbound http requests on your SSR frameworks server runtime from your test suite.

I'm not sure if you're still continuing on the path of using WebSockets for the `client -> server` communication, but one thing I noticed is you've chosen to use [ws](https://npmjs.com/package/ws). I initially went with `ws` as well, given it's widespread use for Node.js applications of WebSockets. However you might want to take a look at [isomorphic-ws](https://www.npmjs.com/package/isomorphic-ws) for the client implementation as it can run in Node.js as well as in browsers.

[Playwright](https://playwright.dev/) uses a Node.js runtime for their test workers, but [Cypress](https://www.cypress.io/) executes test code in the browser. I'd imagine a primary use case for this feature is going to be for UI testing applications running on SSR frameworks.

dziamid:
On our project we have invested a lot of effort of using MSW in Storybook and it's such a good DX. However, with Cypress it's becoming a nightmare to configure handlers on a remove process w/o this API. 

I'm willing to fork and help fix this PR. Before I start, does anyone have a working fork of this branch? Last commit seems to be totally broken. 

Running `pnpm build`:

```bash
src/node/setupRemoteServer.ts(278,11): error TS2552: Cannot find name 'closeSyncServer'. Did you mean 'syncServer'?

```

kettanaito:
@lewnelson, hi! Thanks for the heads up. I think I will keep using WebSockets for the inter-process communication because it makes it more fluent to support sending over WebSocket mock instructions. As for the libraries, I will either use the global `WebSocket`, which is available both in the browser or in Node.js, or a third-party on top of the global `WebSocket` for convenience alone.

@dziamid, hi 👋 Appreciate the initiative to help. Would you be open to [joining my Discord](http://kettanaito.com/discord) and discussing the direction of this API together? This is a big feature in MSW and we should align the effort so it doesn't go wasted. Meanwhile, I will update this branch as soon as I have a moment. Looking forward to discussing this with you!

kettanaito:
## Exploration: Using `capnweb` 

`capnweb` is a promising RPC package that does allow one process to tap into the code of another. Sadly, it's insufficient as of now to power `setupRemoteServer` for the following reasons:

- It doesn't serialize complex objects, like `Request` or `Response`. The support for custom serialization is also missing (see https://github.com/cloudflare/capnweb/issues/32#issuecomment-3369234566). 
- It doesn't transfer complex objects _at all_. You need to wrap all of them in `RpcTarget` and describe what they do. That's cumbersome. 
- It doesn't seem to support streaming data since `RpcSessionImpl.send()` produces a string batch to send over the transport. That's a huge bummer since `.postMessage()` _does_ support streams as transferables! 
- It might be an overkill as I feel that Capn Web is meant for something else that we're trying to achieve here. 

My initial idea was to expose an interface so that `setupServer()` (sender) could tap into a remote process and grab the handlers provided to `setupRemoteServer()` (receiver) via Capn Web. I was not successful in that prototype and now I know at least a few reasons why. I suspect that nothing was transferred as the library cannot transfer class instances, in general.

Feels like we'd be far better off with a simple IPC implementation. I just hoped we could ditch request/response serialization and instead make the logic fancier. Perhaps, that was a wrong thing to hope for. 

On the bright side, req/res serialization is something we could potentially reuse for the main-worker threads communication for `setupWorker()`. Would require a build step and is definitely outside of this PR's scope but something to consider.

jonlambert:
Hey @kettanaito! Sorry to ask the age-old annoying question...

This would be suuuuch a game changer for our team. Are there any plans to merge this functionality? I can see it was added to the 3.0 milestone but can't see it in the list!

kettanaito:
Hi @jonlambert. No worries. This is going to be a gamechanger for all teams, I can imagine!

This pull request as-is is likely dead weight. It was foundational in figuring out the `defineNetwork` architecture that I have since surgically extracted and merged into `main` (see #2650). 

Instead of continuing with this branch, I've opened another feature branch that rebuilds this on top of the new architecture. There are a couple of decisions that were not addressed with this feature that I'm still working on:

- Who should decide the remote `port` between the two parties?
- Is there a room for the `port` to be implicitly set? How?
- Still (laughably) unsure which side should establish the connection and which one should connect. Ended up on the _app_ being responsible for the connection (app is always Node.js) and anything else is the client (which potentially opens `setupWorker()` to be compatible with this API as well).

And a few dozens smaller decisions sprinkled all around. 

## Availability

The biggest blocker right now is my availability. I have a ton of work at Epic Web, I have a few pull request I _really_ need to finish. Realistically, this is coming in 3.0, which might release sometime later this year if all goes smoothly. 

## What you can do to help

If this feature would be beneficial to your team, consider sponsoring my work on it! That would allow me to dedicate proper time to the cross-process interception and finish it for everybody, with your company as a proud patron (I will make sure everyone knows who made this possible). I'd then release this under an experimental version so you can start using it without having to wait for 3.0. Reach out to me on Discord or Twitter if this is something you would be interested in. Thank you!

PR #1622, opened 2023-05-16
fix: log `ReadableStream` responses as text/json

### LINKED ISSUE #1620: Log response body in plain text/json instead of a ReadableStream
### Scope

Improves an existing behavior

### Compatibility

### Feature description

The MSW log when a query is handled used to let me see the body that was returned in plain JSON.

Now I get a `body: ReadableStream { locked: false }` in the console that is of little value to me when developing.

image

I think it'd would make this log more useful if it was human readable.

esquevin:
This logs a promise giving you access to the content of the body in the logs

image

fix #1620


src/utils/logging/prepareResponse.ts
```diff
@@ -8,9 +8,24 @@ import { parseBody } from '../request/parseBody'
...
   const responseHeaders = objectToHeaders(res.headers)
 
+  // Parse a response JSON body for preview in the logs
+  let parsedBody: ReturnType<typeof parseBody> | Promise<any> = parseBody(
+    res.body,
+    responseHeaders,
+  )
+
+  if (parsedBody instanceof ReadableStream) {
+    const response = new Response(parsedBody, { headers: responseHeaders })
```
kettanaito: Alas, this won't work since `prepareResponse` is called in _any_ environment, including Node. The current version of MSW still promises support for Node 16, which doesn't have the Fetch API available globally. This will break for those consumers. 

We have to depend on something different, it seems.

esquevin: Yeah :/ that's sad

kettanaito: Actually, I just checked and this `prepareResponse` pipeline is only run in the browser. We should be safe relying on the `Response` constructor here, given we don't get any type errors in general. 

Testing this locally now...

kettanaito:
I'm proposing to address this on a higher level, at the `serializeResponse()` function. That function _always_ receives a response representation (not an actual Fetch API Response, that caught me) whose `body` is always a `ReadableStream`. So, let's always read it as text and then let the `prepareResponse()` use `parseBody()` that will, eventually, parse JSON response bodies into objects. 

This way `prepareResponse()` is unchanged, no mutable state there either. `prepareBody` is unchanged too. The only tradeoff of this proposal is that `serializeResponse` becomes an async function, which is fine if we look at its consumption: only in the `response:mocked` life-cycle event listeners. 

Oh, another upside for this is that the logged response now is plain text/Object, no Promises. They are tackled on the `serializeResponse()` level (thus the async/await)!

src/utils/logging/serializeResponse.ts
```diff
@@ -1,11 +1,16 @@
...
-    statusText: source.statusText,
-    headers: flattenHeadersObject(headersToObject(source.headers)),
-    body: source.body,
+    status: response.status,
+    statusText: response.statusText,
+    headers: flattenHeadersObject(headersToObject(response.headers)),
+    // Serialize the response body to a string
+    // so it's easier to process further down the chain in "prepareResponse" (browser-only)
+    // and "parseBody" (ambiguous).
+    body: await response.text(),
```
kettanaito, line 14: ## Beware: Not an actual `Response`!

Using a normal `Response` here, we'd have to clone it first so every consumer down the chain would still be able to read it. But this is _not_ a Fetch API `Response` instance here. It's our internal, custom thing since as per 1.2 we are still running in ambiguous, old environments without the global Fetch API. 

Won't be a thing since 2.0.

kettanaito: Consequentially, this doesn't block the consumers from reading the response body again.

esquevin:
> I'm proposing to address this on a higher level, at the `serializeResponse()` function. That function _always_ receives a response representation (not an actual Fetch API Response, that caught me) whose `body` is always a `ReadableStream`. So, let's always read it as text and then let the `prepareResponse()` use `parseBody()` that will, eventually, parse JSON response bodies into objects.
> 
> This way `prepareResponse()` is unchanged, no mutable state there either. `prepareBody` is unchanged too. The only tradeoff of this proposal is that `serializeResponse` becomes an async function, which is fine if we look at its consumption: only in the `response:mocked` life-cycle event listeners.
> 
> Oh, another upside for this is that the logged response now is plain text/Object, no Promises. They are tackled on the `serializeResponse()` level (thus the async/await)!

Yeah I tried something with that PR but I don't know / understand MSW codebase well enough to do something appropriate. This PR should probably be closed in favor of something else.

kettanaito:
No need to close this, I pushed the proposed changes already. They still address the issue so this pull request feels like an appropriate place. Waiting for the CI to pass.

kettanaito:
Let's go! 🚀 Thank you for helping me fix this, @esquevin! Welcome to contributors.

esquevin:
Thx to you ! MSW is truly awesome to use

src/utils/logging/serializeResponse.ts
mattcosta7: @kettanaito after updating to 1.2.2 it looks like i'm getting errors here 

```
Uncaught (in promise) TypeError: Failed to execute 'text' on 'Response': body stream already read
    at serializeResponse (index.js:1616:1)
    at Emitter.<anonymous> (index.js:1660:1)
    at Emitter.onceListener (Emitter.js:40:1)
    at Emitter.js:75:1
    at Array.forEach (<anonymous>)
    at Emitter.emit (Emitter.js:74:1)
    at source.emit (index.js:2036:1)
    at index.js:1756:1
    at ServiceWorkerContainer.<anonymous> (index.js:2140:1)
serializeResponse @ index.js:1616
(anonymous) @ index.js:1660
onceListener @ Emitter.js:40
(anonymous) @ Emitter.js:75
emit @ Emitter.js:74
source.emit @ index.js:2036
(anonymous) @ index.js:1756
(anonymous) @ index.js:2140
await in (anonymous) (async)
onceListener @ Emitter.js:40
(anonymous) @ Emitter.js:75
emit @ Emitter.js:74
source.emit @ index.js:2036
(anonymous) @ index.js:1756
(anonymous) @ index.js:2140

```

mattcosta7: It seems that `quiet: true` gets rid of those issues though, since it no longer tries to read that response twice?

kettanaito: @mattcosta7, do you have a reproduction case for this? 

I was concerned about this too but the `response` we use as of 1.x is _not_ a Fetch AIP Response. It doesn't even have the `.clone()` method and is meant purely for the internal representation of responses. Thus, there's no issue of reading the body multiple times. 

I'd like to look closer at this.

mattcosta7: I'll try to get something together when I can.  it seems to happen with a pretty minimal case (in the work app i'm looking at now)

```
rest.post('<url>', async (req, res, ctx) => {
   return res(ctx.status(200))
})
```

that alone caused this in the work project I'm looking at now

mattcosta7: Interesting - haven't managed a repro on this, despite the 'handler' being super thin.   feel free to disregard and i'll keep trying to understand how/where it's failing.

in the meantime, `quiet` avoids the issue for us and keeps the console a bit clearer so that's ok

strativd: Hey guys! Hopping in the chat because I'm having the same issue with v1.2.2:


<details>
<summary><b>👀 Console error</b></summary>

---------------------------------

```
index.js:1402 [MSW] 07:22:18 query AuthUser (200 OK)
index.js:1616 Uncaught (in promise) TypeError: Failed to execute 'text' on 'Response': body stream already read
    at serializeResponse (webpack-internal:///(app-client)/../../node_modules/.pnpm/msw@1.2.2_typescript@4.8.4/node_modules/msw/lib/index.js:1616:27)
    at Emitter.eval (webpack-internal:///(app-client)/../../node_modules/.pnpm/msw@1.2.2_typescript@4.8.4/node_modules/msw/lib/index.js:1660:25)
    at Emitter.onceListener (webpack-internal:///(app-client)/../../node_modules/.pnpm/strict-event-emitter@0.4.6/node_modules/strict-event-emitter/lib/Emitter.js:40:22)
    at eval (webpack-internal:///(app-client)/../../node_modules/.pnpm/strict-event-emitter@0.4.6/node_modules/strict-event-emitter/lib/Emitter.js:75:22)
    at Array.forEach (<anonymous>)
    at Emitter.emit (webpack-internal:///(app-client)/../../node_modules/.pnpm/strict-event-emitter@0.4.6/node_modules/strict-event-emitter/lib/Emitter.js:74:19)
    at source.emit (webpack-internal:///(app-client)/../../node_modules/.pnpm/msw@1.2.2_typescript@4.8.4/node_modules/msw/lib/index.js:2036:20)
    at eval (webpack-internal:///(app-client)/../../node_modules/.pnpm/msw@1.2.2_typescript@4.8.4/node_modules/msw/lib/index.js:1756:23)
    at ServiceWorkerContainer.eval (webpack-internal:///(app-client)/../../node_modules/.pnpm/msw@1.2.2_typescript@4.8.4/node_modules/msw/lib/index.js:2140:17)
serializeResponse @ index.js:1616
eval @ index.js:1660
onceListener @ Emitter.js:40
eval @ Emitter.js:75
emit @ Emitter.js:74
source.emit @ index.js:2036
eval @ index.js:1756
eval @ index.js:2140
await in eval (async)
onceListener @ Emitter.js:40
eval @ Emitter.js:75
emit @ Emitter.js:74
source.emit @ index.js:2036
eval @ index.js:1756
eval @ index.js:2140

```

---------------------------------

</details>

The following solutions worked for me:

1. use option `quiet: true`
2. downgrade to v1.2.1

With v1.2.2 and quiet set to false, the console error occurs on the second request to the same endpoint. In our case we are making two requests to an `AuthUser` query in GraphQL. The first request is logged to the console and the second request-log is followed by the error. That might be why the error says "body stream already read" — I hope that's helpful!

I'm on a work repo so I can't provide a reproduction env easily right now. That said, our implementation isn't special. I am seeing the console error while mocking responses (in local development) to our GraphQL API within a Next.js app.

PR #1676, opened 2023-07-30
chore: specify pnpm version in package json

bartektelec:
specifies the pnpm version to use when contributing
enables contributors to easily switch to the correct version using `corepack`

update:
looks like the codesandbox demo runs of `pnpm@7.26.3`, do we want to lock the pnpm version to the major? `pnpm@^7.0.0`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,8 @@ $ pnpm install
 $ pnpm start
 ```
 
-> Please use [PNPM][pnpm-url] version specified in the `package.json` file while working on this project. You can use [corepack][corepack-url] to manage multiple pnpm versions.
+> Please use [PNPM][pnpm-url] version 7.12 while working on this project.
+> Guide on how to install a specific PNPM version can be [found here][pnpm-install-guide-url].
 
 ## Git workflow
 
@@ -272,4 +273,4 @@ $ pnpm build
 [pnpm-url]: https://pnpm.io/
 [jest-url]: https://jestjs.io
 [page-with-url]: https://github.com/kettanaito/page-with
-[corepack-url]: https://github.com/nodejs/corepack
+[pnpm-install-guide-url]: https://pnpm.io/7.x/installation#installing-a-specific-version
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ $ pnpm install
 $ pnpm start
 ```
 
-> Please use [PNPM][pnpm-url] while working on this project.
+> Please use [PNPM][pnpm-url] version specified in the `package.json` file while working on this project. You can use [corepack][corepack-url] to manage multiple pnpm versions.
 
 ## Git workflow
 
@@ -272,3 +272,4 @@ $ pnpm build
 [pnpm-url]: https://pnpm.io/
 [jest-url]: https://jestjs.io
 [page-with-url]: https://github.com/kettanaito/page-with
+[corepack-url]: https://github.com/nodejs/corepack
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Seamless REST/GraphQL API mocking library for browser and Node.js.",
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
+  "packageManager": "pnpm@7.12.2",
   "exports": {
     ".": {
       "default": "./lib/index.js"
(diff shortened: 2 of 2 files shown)


kettanaito:
@bartektelec, we need to lock PNPM version to the one used on the CI to make sure builds are passing (lock mismatches due to PNPM version mismatches throw):

https://github.com/mswjs/msw/blob/e87e00b5de5b9145d6010793d2049b242f999c16/.github/workflows/ci.yml#L24-L26

If the sandbox is using an older/newer version, it likely has to be pinned to but we don't have to solve it. The need to have a stable version is to:

- Make sure it's signaled for new contributors cloning the project locally;
- Make sure new contributors don't commit the lockfile generated on a different PNPM version, breaking the build.

CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ $ pnpm install
 $ pnpm start
 ```
 
-> Please use [PNPM][pnpm-url] while working on this project.
+> Please use [PNPM][pnpm-url] version specified in the `package.json` file while working on this project. You can use [corepack][corepack-url] to manage multiple pnpm versions.
```
kettanaito: We can drop the version mention from here. When a new contributor runs `npm install`, it will check the `engines` automatically, throwing if `pnpm` is not installed or the non-matching version is installed. I do hope that's the behavior of `engines`. 

Also, I'd drop any recommendations of third-party tooling to help manage package managers. Those are implicit knowledge and are not specific to MSW. 

> TIL `corepack` is a new Node.js API but it's still experimental so I wouldn't mention it. Even if it wasn't, I think it's still implicit knowledge we don't have to propagate as a part of our docs.

kettanaito: Instead, to ease the contributing experience, we can:

- Clearly state that "Please use PNPM version 7.12".
- Include a curl URL that installs the required version directly. Following the installation from PNPM's docs will always lead to the latest version installed, which will get people confused.

kettanaito (changes requested):
Thanks for opening this! I left a few suggestions on how to make a new contributing experience a bit more streamlined in regards to the correct PNPM version. Let me know what you think.

CONTRIBUTING.md
bartektelec: I have now updated that file to point to the official pnpm docs for a guide on how to install a specific version so the contributers can choose themselves which way of installing they prefer. 
The `packageManager` field in `package.json` is the one used by corepack in case anyone want to use it. 
The field requires to set the exact version (including the patch number) of package manager to be used when corepack is enabled. `7.12.2` was the last from `7.12.x`.

Also added that `7.12` is the one to use, removed the `corepack` mention

bartektelec: I had a brief look at the codesandbox CI docs but didn't find anything of how we could pin the `pnpm` version there.

Some issues on their repo mention the same problem and a possible workaround is to install a specific version of `pnpm` through a custom `installCommand` script, what do you think about that?

kettanaito (approved):
This looks fantastic! Thank you for improving the experience, @bartektelec! 👏

kettanaito:
Welcome to contributors! 🎉

kettanaito:
## Released: v1.2.4 🎉

This has been released in v1.2.4!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v1.2.4)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/1.2.4)

---

auskast:
Specifying the `pnpm` version in the `engines` field causes consumers of the package not running the specified `pnpm` version to not be able to install the library. The `packageManager` field should be sufficient for development purposes.

Error message:
```
 ERR_PNPM_UNSUPPORTED_ENGINE  Unsupported environment (bad pnpm and/or Node.js version)

This error happened while installing a direct dependency of /tmp/renovate/repos/<redacted>

Your pnpm version is incompatible with "registry.npmjs.org/msw/1.2.4".

Expected version: ~7.12
Got: 8.6.12

This is happening because the package's manifest has an engines.pnpm field specified.
To fix this issue, install the required pnpm version globally.

To install the latest version of pnpm, run "pnpm i -g pnpm".
To check your pnpm version, run "pnpm -v".
```

kettanaito:
Thanks for reporting this, @auskast. I've addressed this issue in #1713. The fix will be automatically released in the upcoming days.

PR #1711, opened 2023-08-25
feat: support TypeScript 5.2

chentsulin:
Loosen peer dependency constraint to unblock projects that use msw from upgrading to TypeScript 5.2.x.

### DIFF
--- a/test/typings/tsconfig.5.2.json
+++ b/test/typings/tsconfig.5.2.json
@@ -1,0 +1,3 @@
+{
+  "extends": "./tsconfig.4.7.json"
+}
--- a/test/typings/tsconfig.5.2.json
+++ b/test/typings/tsconfig.5.2.json
@@ -1,3 +1,7 @@
 {
-  "extends": "./tsconfig.4.7.json"
+  "extends": "./tsconfig.4.7.json",
+  "compilerOptions": {
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext"
+  }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ts: ['4.4', '4.5', '4.6', '4.7', '4.8', '4.9', '5.0', '5.1']
+        ts: ['4.4', '4.5', '4.6', '4.7', '4.8', '4.9', '5.0', '5.1', '5.2']
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
     "webpack-http-server": "^0.5.0"
   },
   "peerDependencies": {
-    "typescript": ">= 4.4.x <= 5.1.x"
+    "typescript": ">= 4.4.x <= 5.2.x"
   },
   "peerDependenciesMeta": {
     "typescript": {
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "chokidar": "^3.4.2",
     "cookie": "^0.4.2",
     "graphql": "^15.0.0 || ^16.0.0",
-    "headers-polyfill": "^3.1.2",
+    "headers-polyfill": "^3.2.0",
     "inquirer": "^8.2.0",
     "is-node-process": "^1.2.0",
     "js-levenshtein": "^1.1.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -44,7 +44,7 @@ specifiers:
   fs-extra: ^10.0.0
   fs-teardown: ^0.3.0
   graphql: ^15.0.0 || ^16.0.0
-  headers-polyfill: ^3.1.2
+  headers-polyfill: ^3.2.0
   inquirer: ^8.2.0
   is-node-process: ^1.2.0
   jest: ^29.4.3
@@ -82,7 +82,7 @@ dependencies:
   chokidar: 3.4.1
   cookie: 0.4.2
   graphql: 16.6.0
-  headers-polyfill: 3.1.2
+  headers-polyfill: 3.2.0
   inquirer: 8.2.5
   is-node-process: 1.2.0
   js-levenshtein: 1.1.6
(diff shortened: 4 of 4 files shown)


test/typings/tsconfig.5.2.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "./tsconfig.4.7.json"
```
kettanaito, line 2: Does it make sense to extend `tsconfig.5.0.json`, if we have one? I feel 5.2 and 4.7 are rather far apart.

kettanaito:
I've updated the branch against the latest main.

kettanaito:
Looks like we are getting a consistent TS compilation error:

```sh
Error: src/utils/request/MockedRequest.ts(119,34): error TS2345: Argument of type 'HeadersPolyfill' is not assignable to parameter of type 'Headers'.
  Property 'getSetCookie' is missing in type 'HeadersPolyfill' but required in type 'Headers'.
```

I suspect that something might have changed regarding the validation in this regard, or the `Headers` type definitions have changed (do those ship alongside TS versions? It should come from the `dom` library).

mattcosta7:
> (do those ship alongside TS versions? It should come from the `dom` library).

`lib.d.ts` changes do ship with TS versions, usually noted in the breaking changes section on the release announcement - https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#breaking-changes-and-correctness-fixes

DOM type updates from 5.2 are here
- https://github.com/microsoft/TypeScript/pull/54725

looks like https://github.com/microsoft/TypeScript/pull/54725/files#diff-dc0eab937d15e62545da3ed7b4f40ad6b24f15dd88fbc6ceda2bfb4ed8356eb0R13429 is the line that added the method in question

related mdn docs - https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie

mattcosta7:
@kettanaito  - https://github.com/mswjs/headers-polyfill/pull/57 should implement the method this PR fails type checks from, on the header-polyfill package

kettanaito:
Thank you for opening a pull request in such a timely manner, @mattcosta7 🙌 

My main concern is what would happen with the older versions of TypeScript where `getSetCookie` is not present on `Headers` in the `lib.dom`? I do expect its presence to cause an error. Since we are maintaining a rather wide range of TypeScript versions, we have to take that into account as we don't want to break the types compilation for our users on older TypeScript versions.

Could someone please give it a try on TS < 5.2? I expect it either to break or TS to be smart about it and treat a class with extra methods as satisfying the class with narrower expected methods (I hope very much that's the case).

kettanaito:
A new version of `headers-polyfill` has been published so this pull request can be updated with it and have the typings tests pass.

chentsulin:
```
Run pnpm tsc --version
Version 5.2.2

> msw@1.2.5 test:ts /Users/runner/work/msw/msw
> ts-node test/typings/run.ts

Using tsconfig at "/Users/runner/work/msw/msw/test/typings/tsconfig.5.2.json"
error TS51[10](https://github.com/mswjs/msw/actions/runs/5995405284/job/16258365832?pr=1711#step:8:11): Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.
 ELIFECYCLE  Command failed with exit code 2.
Error: Process completed with exit code 1.
```

mattcosta7:
> ```
> Run pnpm tsc --version
> Version 5.2.2
> 
> > msw@1.2.5 test:ts /Users/runner/work/msw/msw
> > ts-node test/typings/run.ts
> 
> Using tsconfig at "/Users/runner/work/msw/msw/test/typings/tsconfig.5.2.json"
> error TS51[10](https://github.com/mswjs/msw/actions/runs/5995405284/job/16258365832?pr=1711#step:8:11): Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.
>  ELIFECYCLE  Command failed with exit code 2.
> Error: Process completed with exit code 1.
> ```

may need to change the tsconfig.5.2.2 file to use a different `module` - by default I think it uses commonjs which isn't compat in 5.2.2 with moduleResolution node 16 I guess

kettanaito:
@mattcosta7, that's odd. I thought that module resolution has changed as far back as 4.9/5.0. But agree on the strategy, we need to introduce a new base config for this version of TS with the right module resolution set.

mattcosta7:
> @mattcosta7, that's odd. I thought that module resolution has changed as far back as 4.9/5.0. But agree on the strategy, we need to introduce a new base config for this version of TS with the right module resolution set.

typescript added this error in 5.2 - https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#module-and-moduleresolution-must-match-under-recent-node-js-settings

so it might not have correct before, but only causes issues in 5.2+

See the PR that added this here: https://github.com/microsoft/TypeScript/pull/54567

kettanaito:
Thanks for sharing, @mattcosta7. 

I've noticed 5.2 extended 4.7 config that had `moduleResolution: Node16` set already. It appears that because of that `moduleResolution` was supposed to be set to `Node16` as well. 

I set explicit module and moduleResolution values to `NodeNext` and pushed the changes. Let me know if that's alright or we should keep `Node16` for 5.2 as well. I understand it'd he nice to type test MSW with various module/moduleResolution combinations but we don't have the infrastructure for that yet. I'd recommend we go with the most common settings people will have.

kettanaito:
I've also noticed that `tsup` is rather old and requires old TS versions but upgrading it caused test failures in response patching with GraphQL (Jest's error about ESM support). Looks like a bigger endeavor, we don't have to address it now.

kettanaito:
The tests are passing. Let me know if any more changes are intended here. Otherwise, we can merge and release.

kettanaito (approved):

chentsulin:
I think we're ready to merge and release this.

kettanaito:
Thanks for seeing this improvement through, @chentsulin, @mattcosta7!

azangru:
Could you please release this change to npm?

thomasburguiere:
Hello, thanks a lot for the changes, any ETA on the npm release ? 🙂

kettanaito:
## Released: v1.3.0 🎉

This has been released in v1.3.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v1.3.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/1.3.0)

---

PR #1737, opened 2023-09-14
chore: update MIGRATING.md

Kosai106:
Cleanup the migration guide

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -99,8 +99,6 @@ A response resolver now exposes a single object argument instead of `(req, res, 
 To mock responses, you should now return a Fetch API `Response` instance from the response resolver. You no longer need to compose a response via `res()`, and all the context utilities have also [been removed](#context-utilities).
 
 ```js
-import { http } from 'msw'
-
 http.get('/greet/:name', ({ request, params }) => {
   console.log('Captured %s %s', request.method, request.url)
   return new Response(`hello, ${params.name}!`)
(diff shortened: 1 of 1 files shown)


MIGRATING.md
```diff
@@ -99,6 +99,8 @@ A response resolver now exposes a single object argument instead of `(req, res,
 To mock responses, you should now return a Fetch API `Response` instance from the response resolver. You no longer need to compose a response via `res()`, and all the context utilities have also [been removed](#context-utilities).
 
 ```js
+import { http, Response } from 'msw'
```
kettanaito: `Response` is a global class, we don't need to import it. Besides, `msw` doesn't export it, to begin with :D You can drop it.

kettanaito (changes requested):
Hey, @Kosai106 👋 Thanks for making these guidelines a bit more clear. I've got only one remark regarding the `Response` global class. Otherwise looks great.

MIGRATING.md
Kosai106: Ah, fair! In that case there are some instances from before my PR which I'll drop too. :smile:

MIGRATING.md
```diff
@@ -353,7 +361,7 @@ export const handlers = [
 When you provide an object as the `ResponseInit.headers` value, you cannot specify multiple response cookies with the same name. Instead, to support multiple response cookies, provide a `Headers` instance:
 
 ```js
-import { Headers, HttpResponse, rest } from 'msw'
+import { Headers, http, HttpResponse } from 'msw'
```
kettanaito: While we're at it, maybe let's clean up other global classes like `Headers` too? No need to import them from MSW anymore, neither does it export them with 2.0.

kettanaito (approved):

kettanaito:
Thanks, @Kosai106 👏 Welcome to contributors.

PR #1741, opened 2023-09-16
chore: upgrade to "headers-polyfill" v4

kettanaito:


src/core/bypass.ts
```diff
@@ -1,5 +1,4 @@
 import { invariant } from 'outvariant'
-import { Headers } from 'headers-polyfill'
```
kettanaito, line 2: Not using `Headers` polyfill in tests anymore—relying on the global API from Jest. In fact, MSW doesn't use the polyfill itself at all.

tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "declaration": true,
     "declarationDir": "lib/types",
     "noEmit": true,
-    "lib": ["es2017", "ESNext.AsyncIterable", "dom", "webworker"],
+    "lib": ["DOM", "DOM.Iterable", "ESNext.AsyncIterable"],
```
kettanaito, line 13: TIL, use `dom.iterable` to get iterable methods like `keys()`, `values()`, and `entries()` on the `Headers` class.

PR #1755, opened 2023-09-29
Ignore type check in generated file

kmkzt:
A type check error occurred in the file generated when running msw init, so it was corrected to be ignored.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -1,5 +1,6 @@
 /* eslint-disable */
 /* tslint:disable */
+// @ts-nocheck
 
 /**
  * Mock Service Worker (<PACKAGE_VERSION>).


kettanaito:
Hi, @kmkzt. Can you tell me more why the `.js` file in a public directory was type-checked, to begin with? You should configure TypeScript to exclude `*.js` files (very likely) and certainly ignore the `public `directory of your website. This looks like an error on your end but I'm eager to learn more.

kettanaito:
I will close this one. 

## Solution

**Configure TypeScript/Prettier/ESLint/other tooling to ignore static assets in your `/public` directory**. This is the right project setup and if the worker script from MSW causes troubles, make sure that your tooling doesn't attempt to check assets (it really shouldn't).

kmkzt:
@kettanaito 
I apologize for the delayed response.

In my opinion, it would be convenient to ensure that the type check passes, irrespective of the build settings, considering it's a generated file. Therefore, I would like to propose this idea.

Additionally, thanks you sharing the solution, and I believe it's the correct approach.

PR #1762, opened 2023-10-06
fix(worker): replace `Math.random()` with `crypto.randomUUID()`

laryro:
Replace  `Math.random()` with `with crypto.randomUUID()` in `mockServiceWorker.js` file.

It's known that Math.Random isn't secure (see https://v8.dev/blog/math-random). Even if this file isn't supposed to be deployed in production, some static code analysis tools may flag this as a critical issue, breaking CI pipelines.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -107,7 +107,7 @@ self.addEventListener('fetch', function (event) {
   }
 
   // Generate unique request ID.
-  const requestId = Math.random().toString(16).slice(2)
+  const requestId = crypto.randomUUID()
   event.respondWith(handleRequest(event, requestId))
 })
 


src/mockServiceWorker.js
```diff
@@ -112,7 +112,7 @@ self.addEventListener('fetch', function (event) {
   }
 
   // Generate unique request ID.
-  const requestId = Math.random().toString(16).slice(2)
+  const requestId = crypto.randomUUID().slice(0, 8)
```
kettanaito: Can you please tell me more about why we need to `.slice()` the generated UUID? 🤔

kettanaito:
Hi, @laryro! Thanks for opening this, I think it's a great improvement. One thing I may add, MSW treats any changes to the worker file is breaking changes. Would you mind rebasing this pull request against the `feat/standard-api` branch? We would get the change released in the next major version then. Thanks!

src/mockServiceWorker.js
laryro: Ah, good catch, for some weird reason I thought the previous command outputted 8 chars and I tried to replicate the same behaviour - which is actually incorrect as `Math.random().toString(16).slice(2)` results in 13 characters 😅 

Anyways, I'll remove the `slice` and leave the original output (which is 36 characters long) 😬

laryro:
> Hi, @laryro! Thanks for opening this, I think it's a great improvement. One thing I may add, MSW treats any changes to the worker file is breaking changes. Would you mind rebasing this pull request against the `feat/standard-api` branch? We would get the change released in the next major version then. Thanks!

@kettanaito I rebased my branch and pointed it to `feat/standard-api`, as well as removed the `slice` method per our convo. 
Lemme know if you need anything else 😄

kettanaito (approved):
This looks fantastic! Thank you for working on this, @laryro, and welcome to contributors! 👏

PR #1765, opened 2023-10-10
fix(setupServer): set max listeners on the "request.signal"

### LINKED ISSUE #1760: v2 - Max listeners warning in Node.js (likely the cloning issue)
### Prerequisites

### Environment check

### Node.js version

v18.14.0

### Reproduction repository

—

### Reproduction steps

https://github.com/mswjs/msw/pull/1436#issuecomment-1746726819

### Current behavior

Warns.

### Expected behavior

Doesn't warn.

kettanaito:
- Fixes #1760

Fixes the issue when having a large number of request handlers would result in the following error from Node when processing an intercepting request:

```
(node:3158) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal].
```

## Why was this error printed?

When a large number of event listeners is added to any target, Node preemptively prints an error letting you know that something may be off (a potential memory leak). However, in the case of the request handlers, we are expecting at least a N `abort` event listener to be added to the original `request.signal` because every request handler clones the intercepted request at least once (so it could read its body when logging it).

## How is this fixed?

By using `require('events').setMaxListeners()` on the `request.signal`.

### DIFF
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -156,7 +156,7 @@ export abstract class RequestHandler<
     resolutionContext?: ResponseResolutionContext
   }): Promise<boolean> {
     const parsedResult = await this.parse({
-      request: args.request.clone(),
+      request: args.request,
       resolutionContext: args.resolutionContext,
     })
 
@@ -186,6 +186,8 @@ export abstract class RequestHandler<
       return null
     }
 
+    // Clone the request instance before it's passed to the handler phases
+    // and the response resolver so we can always read it for logging.
     const mainRequestRef = args.request.clone()
 
     // Immediately mark the handler as used.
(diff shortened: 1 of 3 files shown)


src/node/SetupServerApi.ts
```diff
@@ -50,6 +51,18 @@ export class SetupServerApi
...
     this.interceptor.on('request', async ({ request, requestId }) => {
+      // Bump the maximum number of event listeners on the
+      // request's "AbortSignal". This prepares the request
+      // for each request handler cloning it at least once.
+      // Note that cloning a request automatically appends a
+      // new "abort" event listener to the parent request's
+      // "AbortController" so if the parent aborts, all the
+      // clones are automatically aborted.
+      setMaxListeners(
+        Math.max(defaultMaxListeners, this.currentHandlers.length),
```
kettanaito, line 62: We must never set fewer max listeners than `defaultMaxListeners`, even if there are fewer request handlers, because that may produce false positive "possible memory leak" errors from Node. 

Node sets the max listeners to 10 by default, if I'm not mistaken. If there are fewer request handlers, keep the default max listeners count.

src/core/handlers/RequestHandler.ts
```diff
@@ -194,11 +196,11 @@ export abstract class RequestHandler<
     this.isUsed = true
 
     const parsedResult = await this.parse({
-      request: mainRequestRef.clone(),
+      request: args.request,
```
kettanaito, line 199: No more cloning of the intercepted request for the internal phases. Each request handler class must clone the given request by itself if it intends to read its body.

PR #1766, opened 2023-10-11
chore: re-add inquirer

mattcosta7:
Re-adds `inquirer` to package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "formdata-node": "4.4.1",
     "graphql": "^16.8.1",
     "headers-polyfill": "^4.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -124,6 +124,7 @@
     "formdata-node": "4.4.1",
     "graphql": "^16.8.1",
     "headers-polyfill": "^4.0.1",
+    "inquirer": "^8.2.0",
```
mattcosta7, line 127: 9+ is latest, but re-introducing the version that was accidentally removed

kettanaito (approved):
Thanks, @mattcosta7 👏

PR #1770, opened 2023-10-15
test(browser): add ReadableStream response body test

kettanaito:


kettanaito:
Hey, @thepassle. I've added this browser integration test for a readable stream mock response body and it doesn't require setting the `duplex` option either. Could you give this a quick look if everything's alright? Thanks.

thepassle:
Ah, the duplex is required in `Request`, not `Response`:

```js
a = new Request('/api/foo', {
  method: 'POST',
  body: new ReadableStream()
});
```

results in:
Uncaught TypeError: Failed to construct 'Request': The `duplex` member must be specified for a request with a streaming body

See also: https://developer.chrome.com/articles/fetch-streaming-requests/#half-duplex

kettanaito:
@thepassle, yeah, that's right. You need that option to _send_ a readable stream. That needs no testing on MSW's side because the way you construct the request is up to you and it has to be correct. Thanks for verifying that 👍

thepassle:
Yeah, looks like I got things mixed up :) my bad!

PR #1771, opened 2023-10-15
fix: transfer response buffer on safari

### LINKED ISSUE #1757: DataCloneError: The object cannot be cloned in Safari v17 (using fetch-based MSW)
### Prerequisites

### Environment check

### Browsers

Safari

### Reproduction repository

https://github.com/thepassle/msw-safari-repro

### Reproduction steps

- `npm i`
- `npm start`
- open browser in safari 17

### Current behavior

repro

### Expected behavior

No error and the object is cloned successfully

kettanaito:
- Fixes #1757
- Related to #1436 
- Originally reported [here](https://github.com/mswjs/msw/pull/1436#issuecomment-1762701825)

## Motivation

Fixes an issue where sending a `ReadableStream` via `messageChannel.port1.postMessage()` throws the following error on Safari:

```
DataCloneError: The object can not be cloned.
```

## Solution

MSW will run a quick supports check for `ReadableStream` as transferable. If it fails, MSW will read the mocked response's `ArrayBuffer` body and send it _at once_ to the worker. This still means that mocking streams won't work in Safari as reading the response body will exhaust its stream so it loses the chunks/timings information when the client will receive it. 

That's acceptable since it's an issue in Safari and must be addressed there.

### DIFF
--- a/src/browser/setupWorker/setupWorker.ts
+++ b/src/browser/setupWorker/setupWorker.ts
@@ -20,6 +20,7 @@ import { SetupApi } from '~/core/SetupApi'
 import { mergeRight } from '~/core/utils/internal/mergeRight'
 import { LifeCycleEventsMap } from '~/core/sharedOptions'
 import { SetupWorker } from './glossary'
+import { supportsReadableStreamTransfer } from '../utils/supportsReadableStreamTransfer'
 
 interface Listener {
   target: EventTarget
@@ -144,6 +145,9 @@ export class SetupWorkerApi
       },
       useFallbackMode:
         !('serviceWorker' in navigator) || location.protocol === 'file:',
+      supports: {
+        readableStreamTransfer: supportsReadableStreamTransfer(),
+      },
     }
 
     /**
--- a/src/browser/setupWorker/setupWorker.ts
+++ b/src/browser/setupWorker/setupWorker.ts
@@ -143,9 +143,9 @@ export class SetupWorkerApi
           })
         },
       },
-      useFallbackMode:
-        !('serviceWorker' in navigator) || location.protocol === 'file:',
       supports: {
+        serviceWorkerApi:
+          !('serviceWorker' in navigator) || location.protocol === 'file:',
         readableStreamTransfer: supportsReadableStreamTransfer(),
       },
     }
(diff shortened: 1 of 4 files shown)


src/browser/setupWorker/start/createRequestListener.ts
```diff
@@ -47,18 +48,29 @@ export const createRequestListener = (
...
-            const responseStream = responseClone.body
 
-            messageChannel.postMessage(
-              'MOCK_RESPONSE',
-              {
+            /**
+             * @note Safari doesn't support transferring a "ReadableStream".
+             * Check that the browser supports that before sending it to the worker.
+             */
+            if (supportsReadableStreamTransfer()) {
```
kettanaito: I dislike checking this on _every_ response listener. 

This should be determined _once_ because its value is bound to the current browser and cannot change on runtime. 

We should move this somewhere to the `setupWorker()` call, I think. It's computed once, even before handling the first response, and stored in the worker context. Something like `worker.context.supportsReadableStreamTransfer` can work.

src/browser/setupWorker/start/createRequestListener.ts
```diff
@@ -47,18 +48,29 @@ export const createRequestListener = (
...
+                'MOCK_RESPONSE',
+                {
+                  ...responseInit,
+                  body: responseStream,
+                },
+                responseStream ? [responseStream] : undefined,
+              )
+            } else {
+              // As a fallback, send the response body buffer to the worker.
+              const responseBuffer = await responseClone.arrayBuffer()
```
kettanaito, line 67: @thepassle, would this work for your use case if we read the mocked response body as `ArrayBuffer` instead of `Blob`? I find it to be more correct.

src/browser/setupWorker/setupWorker.ts
```diff
@@ -142,8 +143,11 @@ export class SetupWorkerApi
           })
         },
       },
-      useFallbackMode:
-        !('serviceWorker' in navigator) || location.protocol === 'file:',
+      supports: {
+        serviceWorkerApi:
```
kettanaito, line 147: This is a good chance to move `useFallbackMode` to `supports.serviceWorkerApi` since now we check for two things on the side of browser's support.

src/browser/setupWorker/start/createRequestListener.ts
thepassle: looks like it works: 
image

thepassle:
tried this out locally and it seems to work, lgtm

src/browser/setupWorker/start/createRequestListener.ts
kettanaito: Awesome! 

In the future, we should print a warning if the developer is mocking a `ReadableStream` body while transferring that stream isn't supported by their browser. We will still use the fallback to `ArrayBuffer` but the developer may be confused lest we show a warning of what's happening.

PR #1776, opened 2023-10-18
chore: add more browsers to playwright config

thepassle:


thepassle:
Welp, looks like a bunch of failing tests 😅 That's a lot more than I was expecting though, I was expecting maybe a couple. Im not sure if I'll have time to address all of these today, but ill try to dive in sometime this week

kettanaito:
@thepassle, we can address those together in our spare time. MSW as it is doesn't promise nor offer any particular browser support. I always saw MSW as a devtool, and if it doesn't work in one browser simply switch to another and work with it there (similar to how every browser has different devtools available). 

That being said, I do find great value in cross-browser testing and we should pursue this endeavor. I think we can improve how browser tests run in CI so it's more apparent what browser corresponds to what failed tests, as well as parallelize cross-browser runs since GitHub Actions has 1 thread and will run all tests for all browsers _sequentially_. Browser tests already take the most time on CI, and if we add more browser targets, we have to likely introduce separate GitHub workflows for those so they can run in parallel and save us some time.

thepassle:
Yeah there's no rush to getting this in, but it would definitely be nice in general to improve the browser support for the library.

> MSW as it is doesn't promise nor offer any particular browser support. I always saw MSW as a devtool, and if it doesn't work in one browser simply switch to another and work with it there (similar to how every browser has different devtools available).

At ING we encourage our developers to test their code on multiple browsers, developers can do this manually during local dev, but also often teams use automated testing via `@web/test-runner` to run their tests on multiple browsers. While a lot of JS syntax-compatibility related things can be transpiled away at build time, not everything is statically analyzable, or detectable, or polyfillable. As a bank, browser compat is very important to us; also for local development.

While I agree that MSW shouldn't have to bend over backwards to support IE11, but I don't think its unreasonable to at least make sure the library works on the latest version of the major browsers (chromium/firefox/safari).

> That being said, I do find great value in cross-browser testing and we should pursue this endeavor.

Sounds good! Like I mentioned, there's no rush to get this in. Ill try to find some time this week to start chipping away at this :)

kettanaito:
> While I agree that MSW shouldn't have to bend over backwards to support IE11, but I don't think its unreasonable to at least make sure the library works on the latest version of the major browsers (chromium/firefox/safari).

Absolutely agree.

kettanaito:
Circling back to this one. Given we don't ship any API specific to certain browser vendors, I'm not sure what running against different browsers would accomplish other than implicitly test how those browsers implement the Service Worker API. My concern is that if we catch an issue, say, in Safari, and it has nothing to do with the library but how Safari adheres to the spec, what is the action here? Should that be considered a failing test on our side?

I'd say no. 

For example, there's already a well-known limitation in Firefox that it doesn't trigger the `fetch` event in the worker on XMLHttpRequest on the page. As a result, we might have tests failing in Firefox but it yields no actionable input on our side. We shouldn't tailor to browser bugs as we shouldn't tailor to request client bugs. 

I will close this for now but always happy to discuss.

PR #1791, opened 2023-10-23
feat: add dangerouslyRunInProduction flag to worker and server options

### LINKED ISSUE #1703: Throw when run in production
### Scope

Adds a new behavior

### Compatibility

- [X] This is a breaking change

### Feature description

I propose MSW, both `setupWorker` and `setupServer` throw an exception when detected being run in `NODE_ENV === 'production'`. 

## Motivation

When MSW gets to a production environment that's a mistake on the developer's end 9 times out of 10. I believe this would be a good default behavior to have to prevent this kind of mistake. 

## Implementation

Apart from raising an exception, this feature should also introduce an opt-out mechanism for those who explicitly wish to run MSW in production. We must still permit to run MSW anywhere the end user want but they need to explicitly state their intention. I know that a lot of teachers rely on MSW in production to teach their workshops. I want them to keep doing that without hindrance.

```js
worker.start({ dangerouslyRunInProduction: true })
server.listen({ dangerouslyRunInProduction: true })
```

When the `dangerouslyRunInProduction` is provided and equals `true`, suppress the said exception.

scr2em:
- Closes #1703


scr2em:
@kettanaito this actually breaks the current tests, do we need to enable dangerouslyRunInProduction in the package tests ?

src/browser/setupWorker/glossary.ts
```diff
@@ -195,6 +195,11 @@ export interface StartOptions extends SharedOptions {
    * of all registered Service Workers on the page.
    */
   findWorker?: FindWorker
+
+  /**
+   * A flag to enable MSW in production (NODE_ENV === 'production'), otherwise it will throw an error
+   */
+  dangerouslyRunInProduction?: boolean
```
kettanaito: `StartOptions` here extends `SharedOptions` so there's no need to add this new property here as well.

src/browser/setupWorker/setupWorker.ts
```diff
@@ -177,6 +178,14 @@ export class SetupWorkerApi
       options,
     ) as SetupWorkerInternalContext['startOptions']
 
+    invariant(
+      !this.context.startOptions.dangerouslyRunInProduction && isProduction(),
+      devUtils.formatMessage(
+        'The flag dangerouslyRunInProduction is false but you are in a production environment',
+      ),
+      'https://github.com/mswjs/msw/issues/1703',
```
kettanaito, line 186: I think we should rather feature a link to the docs explaining this flag and what it does. I will document it.

src/browser/setupWorker/setupWorker.ts
```diff
@@ -177,6 +178,14 @@ export class SetupWorkerApi
       options,
     ) as SetupWorkerInternalContext['startOptions']
 
+    invariant(
+      !this.context.startOptions.dangerouslyRunInProduction && isProduction(),
+      devUtils.formatMessage(
+        'The flag dangerouslyRunInProduction is false but you are in a production environment',
```
kettanaito: What do you think if we phrase this error a bit better? Perhaps something like this?

- Failed to call "setupWorker" in a production environment. Please make sure you enable API mocking conditionally so it doesn't leak to production. 

Don't even mention the flag that controls it to emphasize the importance of this behavior. Still include a link to the docs about the flag for those who know what they are doing and want to suppress this behavior.

src/browser/setupWorker/setupWorker.ts
scr2em: @kettanaito Awesome, that would be more appropriate.

src/browser/setupWorker/setupWorker.ts
scr2em: @kettanaito let me know the link when you finish it

kettanaito:
Updated the branch. Still need some time to review and document this.

kettanaito:
Hey, @scr2em 👋 Thanks for helping me with this. 

I will close this one for now. We've not had a single issue on my memory when someone has deployed MSW to prod by mistake. Sometimes, that's what people want, depending on what they are building. 

Let's leave it this pull request for reference.

PR #1824, opened 2023-11-01
fix: correctly check "undefined" as response body type

### LINKED ISSUE #1823: Typescript error: StrictResponse of generic is not assignable to AsyncResponseResolverReturnType
### Prerequisites

### Environment check

### Browsers

_No response_

### Reproduction repository

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgCRjMAlApgZ0gOxywBo4ARLAMwEMBXAGxgCEIATATwBV2wS4ALNGDgBfOJSgQQcAOQgcAdxkBuAFDrKtfAGMYwCPjgQA1gHEsMZNXyt6WKAB5O2PAaJwsADxhYbOOAByBgEM9NQARnYOFDQMzGxcPFgAfMkAFL6sAAoQwPgwAFxwODBQeQDmAJSIqnBwUBa0UIZpDa6EWEXOuARE1QC8yTV1dYLoAHTlFhk2OXkwpGkDQ6joLr1Y4wBWOAatPW5YlZW1onCqIuoA9FdwAKKe1OB2cLQ41FOq2m7wXk9gdgAshBtMYrDY7FA4P0jGYLODbPYHAhqGgyuFaD4ivhaCBwvYROkAERXP7PLCkmyQeZEk5fH4CayIqD+GEAbVOZIBWGBoIRkLSKLRwAxWLgAEYRCcALpAA

### Reproduction steps

I'm trying to create wrappers around the http.get method to be able to couple endpoints in our backend with their return types. Here is an example of one such wrapper:

```typescript
import { HttpResponse, DefaultBodyType, http } from 'msw';

function okGetHandler<TResponse extends NonNullable<DefaultBodyType>>(endPoint: string) {
  return (response: TResponse) => {
    http.get(endPoint, () => HttpResponse.json(response))
  } 
}
```

### Current behavior

This causes a Typescript error: 
Type 'StrictResponse<TResponse>' is not assignable to type 'AsyncResponseResolverReturnType<TResponse>'.
  Type 'StrictResponse<TResponse>' is not assignable to type 'TResponse extends undefined ? Response : StrictResponse<TResponse>'

### Expected behavior

I expect StrictResponse<TResponse> to be assignable to AsyncResponseResolverReturnType<TResponse>.

There is a easy workaround which is to simply use the native Response like so:

```typescript
export function okGetHandler<TResponse extends NonNullable<DefaultBodyType>>(endPoint: string) {
    return (response: TResponse) =>
        http.get(
            url + endPoint,
            () => new Response(JSON.stringify(response), { headers: { 'Content-Type': 'application/json' } }),
        )
}
```
 this is not as readable however.

Is there some extends clause that I'm missing on TResponse?

kettanaito:
- Fixes #1823

## Why still check if `[undefined`] and not `[never]`?

Because `never` is an explicit "this response must _not_ have any body". This is a narrower response body type and it must use the `StrictResponse<never>` internal type if it's provided. 

`undefined` is the _default_ (fallback) response body type. This means that by default, we don't know what type the response body will be, so `undefined` is equivalent to a plain `Response`.

### DIFF
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -30,7 +30,7 @@ export interface RequestHandlerInternalInfo {
 export type ResponseResolverReturnType<
   BodyType extends DefaultBodyType = undefined,
 > =
-  | (BodyType extends undefined ? Response : StrictResponse<BodyType>)
+  | ([BodyType] extends [undefined] ? Response : StrictResponse<BodyType>)
   | undefined
   | void
 
@@ -265,7 +265,7 @@ export abstract class RequestHandler<
 
           // Clone the previously stored response from the generator
           // so that it could be read again.
-          return this.resolverGeneratorResult.clone()
+          return this.resolverGeneratorResult.clone() as StrictResponse<any>
         }
 
         if (!this.resolverGenerator) {
--- a/test/typings/regressions/response-body-type.test-d.ts
+++ b/test/typings/regressions/response-body-type.test-d.ts
@@ -1,0 +1,10 @@
+/**
+ * @see https://github.com/mswjs/msw/issues/1823
+ */
+import { http, Path, HttpResponse, DefaultBodyType } from 'msw'
+
+function myHandler<CustomResponseBodyType extends DefaultBodyType>(path: Path) {
+  return (response: CustomResponseBodyType) => {
+    http.get(path, () => HttpResponse.json(response))
+  }
+}


src/core/handlers/RequestHandler.ts
```diff
@@ -30,7 +30,7 @@ export interface RequestHandlerInternalInfo {
 export type ResponseResolverReturnType<
   BodyType extends DefaultBodyType = undefined,
 > =
-  | (BodyType extends undefined ? Response : StrictResponse<BodyType>)
```
kettanaito, line 33: `T extends undefined` is not the correct way to check if a generic was provided in TypeScript. Instead, one must resort to a hack:

```ts
[T] extends [never] ? 'NOT PROVIDED' : 'PROVIDED'
```

The previous implementation was giving false positives, making the custom response body type and the body type in `ResponseResolverReturnType` incompatible.

test/typings/regressions/response-body-type.test-d.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @see https://github.com/mswjs/msw/issues/1823
+ */
+import { http, Path, HttpResponse, DefaultBodyType } from 'msw'
+
+function myHandler<CustomResponseBodyType extends DefaultBodyType>(path: Path) {
+  return (response: CustomResponseBodyType) => {
+    http.get(path, () => HttpResponse.json(response))
```
kettanaito, line 8: Also adding a typing regression test to make sure this doesn't happen again.

src/core/handlers/RequestHandler.ts
```diff
@@ -265,7 +265,7 @@ export abstract class RequestHandler<
 
           // Clone the previously stored response from the generator
           // so that it could be read again.
-          return this.resolverGeneratorResult.clone()
+          return this.resolverGeneratorResult.clone() as StrictResponse<any>
```
kettanaito, line 268: A bit hacky. Suggestions are welcome.

kettanaito:
## Released: v2.0.3 🎉

This has been released in v2.0.3!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.3)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.3)

---

PR #1826, opened 2023-11-02
fix: export missing "GraphQLResolverExtras" and "GraphQLResponseBody" types

jb-1980:
### Problem
The resolver function for GraphQL is typed as
```ts
type resolver = ResponseResolver<
    GraphQLResolverExtras<Variables>,
    null,
    GraphQLResponseBody<Query>
  >
```
I want to be able to have a wrapper that takes in a resolver function, but I can't properly type it without being able to import the `GraphQLResolverExtras` and `GraphQLResponseBody` types.

### Solution
Export the needed types.

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -38,6 +38,8 @@ export type {
   GraphQLVariables,
   GraphQLRequestBody,
   GraphQLJsonRequestBody,
+  GraphQLResolverExtras,
+  GraphQLResponseBody,
 } from './handlers/GraphQLHandler'
 
 export type { Path, PathParams, Match } from './utils/matching/matchRequestUrl'


kettanaito:
Thanks for addressing this, @jb-1980. 

That being said, I think this needs a proper discussion. I dislike the idea of exposing types like `GraphQLExtras` to the developer. They are way too internal to work with, and they expand the public API surface to implementation details, imo. 

Instead, it'd be nice to expose developer-facing types like `HttpResolver` and `GraphQLResolver` that would be designed to be worked with (extended/overridden/etc). 

This was also brought up in #1802 which makes my point for developer-facing types all the more relevant.

jb-1980:
Thanks for the quick response @kettanaito. I think a discussion is great.

If I understand you correctly, you want to do something like this:
```ts
// in some file deep in the package, like `handlers/GraphQLHandler`
export type GraphQLResolver<Variables, Query> = ResponseResolver<
  GraphQLResolverExtras<Variables>,
  null,
  GraphQLResponseBody<Query>
>

// then in core/index
export {GraphQLResolver} from "./handlers/GraphQLHandler"

// And package consumers can then import and use
import { type GraphqQLResolver, graphql } from "msw"

const MySpecialGraphql = graphql.link("http://to.my.gql.server")

const wrappedHandler = (otherArgs: any, resolver: GraphQLResolver) => {
  // do stuff with `otherArgs`
  MySpecialGraphql.query("MyQuery", resolver)
}
```

If that is the direction you are leaning, I think it is a great idea. Much better than exporting the low-level types for the reasons you mentioned.

What are some of your ideas on design that would allow a developer to work with the type? Is it more than allowing them to pass through the `Variables` and `Query` types? I don't fully grok all the lower-level workings of the package, so from my perspective the ability to pass those in is sufficient. I would be interested to know what your thoughts are.

Whatever direction you want to take, I would be happy to put in the work and make it happen.

Wroud:
I also have a case for this types:
Screenshot 2023-12-12 at 2 51 30 PM
then later I can use it like this:
Screenshot 2023-12-12 at 2 52 22 PM
it will be much better to have `GraphQLResolver<Variables, Query>` exported type

kettanaito:
Hi, folks 👋 

We've just released an improvement to annotating request handlers and response resolvers (#1961). Please read the new [Using with TypeScript](https://mswjs.io/docs/best-practices/typescript) page in our docs as well. I believe it solves these use cases without the need to export MSW internals like `GraphQLResolverExtras`. 

I will close this pull request in favor of that. Once again, thank you for raising this, @jb-1980! Please, if you have any other use cases in TypeScript that are difficult to achieve right now, let me know!

PR #1833, opened 2023-11-04
fix(Safari): send explicit null response bodies

### LINKED ISSUE #1827: Safari `readableStreamTransfer` workaround turns `null` body into empty `ArrayBuffer`, requests fail
### Prerequisites

### Environment check

### Browsers

Safari

### Reproduction repository

https://github.com/oxidecomputer/msw2-safari-bug/

### Reproduction steps

`npm install`, `npm run dev`, open in safari, see error on page and in dev console.

### Current behavior

In Safari only, handler responses with `null` body produce the following error:

```
FetchEvent.respondWith received an error: TypeError: Response cannot have a body with the given status.
```

The error is thrown here on the `new Response(...)`

https://github.com/mswjs/msw/blob/008a8c69610709308fee5226e24f627717882bcb/src/mockServiceWorker.js#L275-L292

because `response.body` is an `ArrayBuffer` even though it was `null` in the handler definition. When the status is `204` (or other null body status) but the body is something other than `null`, the `Response` constructor throws an error:

image

### Suspected cause

I believe this is caused by this workaround for the lack of `readableStreamTransfer` in Safari:

https://github.com/mswjs/msw/blob/008a8c69610709308fee5226e24f627717882bcb/src/browser/setupWorker/start/createRequestListener.ts#L52-L73

Specifically, the line 

```ts
const responseBuffer = await responseClone.arrayBuffer()
```

will give a non-null empty `ArrayBuffer` even when `responseClone.body` is `null`. (The screenshot is in Firefox, but it's the same in Chrome and Safari.)

image

I believe the fix (or _a_ fix) would be something like 

```ts
const responseBuffer = responseClone.body === null ? null : await responseClone.arrayBuffer()
```

I'm happy to make that PR, I just wanted to write this all down after a fun day of tracking it down with @zephraph.

### Expected behavior

`new Response(null, { status: 204 })` works in Safari like it does anywhere else.

kettanaito:
- Fixes #1827

### DIFF
--- a/src/browser/setupWorker/start/createRequestListener.ts
+++ b/src/browser/setupWorker/start/createRequestListener.ts
@@ -54,21 +54,31 @@ export const createRequestListener = (
              * Check that the browser supports that before sending it to the worker.
              */
             if (context.supports.readableStreamTransfer) {
-              const responseStream = response.body
+              const responseStreamOrNull = response.body
+
               messageChannel.postMessage(
                 'MOCK_RESPONSE',
                 {
                   ...responseInit,
-                  body: responseStream,
+                  body: responseStreamOrNull,
                 },
-                responseStream ? [responseStream] : undefined,
+                responseStreamOrNull ? [responseStreamOrNull] : undefined,
               )
             } else {
-              // As a fallback, send the response body buffer to the worker.
-              const responseBuffer = await responseClone.arrayBuffer()
+              /**
+               * @note If we are here, this means the current environment doesn't
+               * support "ReadableStream" as transferable. In that case,
+               * attempt to read the non-empty response body as ArrayBuffer, if it's not empty.
+               * @see https://github.com/mswjs/msw/issues/1827
+               */
+              const responseBufferOrNull =
+                response.body === null
+                  ? null
+                  : await responseClone.arrayBuffer()
+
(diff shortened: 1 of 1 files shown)


src/browser/setupWorker/start/createRequestListener.ts
```diff
@@ -54,21 +54,31 @@ export const createRequestListener = (
...
-              // As a fallback, send the response body buffer to the worker.
-              const responseBuffer = await responseClone.arrayBuffer()
+              /**
+               * @note If we are here, this means the current environment doesn't
+               * support "ReadableStream" as transferable. In that case,
+               * attempt to read the non-empty response body as ArrayBuffer, if it's not empty.
+               * @see https://github.com/mswjs/msw/issues/1827
+               */
+              const responseBufferOrNull =
+                response.body === null
```
kettanaito, line 75: Here's the actual fix.

kettanaito:
## Released: v2.0.3 🎉

This has been released in v2.0.3!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.3)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.3)

---

PR #1838, opened 2023-11-07
fix: narrow `body` type for `HttpResponse.json()` to be serializable json

kettanaito:
## Changes

- The `body` type of the `HttpResponse.json()` static method _must_ be a serializable JSON. Narrow the argument type to reflect that. 
- Exports the new `JsonBodyType` type from the core.

### DIFF
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -19,6 +19,14 @@ export type DefaultBodyType =
   | null
   | undefined
 
+export type JsonBodyType =
+  | Record<string, any>
+  | string
+  | number
+  | boolean
+  | null
+  | undefined
+
 export interface RequestHandlerDefaultInfo {
   header: string
 }
--- a/src/core/HttpResponse.test.ts
+++ b/src/core/HttpResponse.test.ts
@@ -10,6 +10,7 @@ it('creates a plain response', async () => {
   expect(response.statusText).toBe('Moved Permanently')
   expect(response.body).toBe(null)
   expect(await response.text()).toBe('')
+  expect(Object.fromEntries(response.headers.entries())).toEqual({})
 })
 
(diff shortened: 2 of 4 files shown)


src/core/handlers/RequestHandler.ts
```diff
@@ -19,6 +19,14 @@ export type DefaultBodyType =
   | null
   | undefined
 
+export type JsonBodyType =
+  | Record<string, any>
```
mattcosta7, line 23: nit: do we want to be slightly more prescriptive than `Record<string, any>` here, `Record<string, JsonBodyType>` might be the most correct option?

also I think we'd need to account for `Array<JsonBodyType>` too? (and potentially their readonly varieties?)

kettanaito: Surprisingly, TypeScript is permissive about this and arrays [extend](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwhApgBygeQEYBW60AzjmADYCeAFAG7IUjoBciASsXAE4AmAPCSjcYYAOYAaFGCoA+AJSIA3gF8AUGtQZs+IqXLUaq+ZrRZchYlDKVaAbQC6Jred1WbhuwHJGJOF6kvYFF0LycgA) `Record<string, any>` quite well. It's the `any` part that does the trick.

As to narrowing down the type from `any` to a recursive `JsonBodyType`, I'm not so sure about it. It sounds like the right thing to do but there's always corner cases people run into, and it's hard to predict what will happen if we make this body type _that_ strict.

kettanaito:
## Released: v2.0.4 🎉

This has been released in v2.0.4!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.4)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.4)

---

PR #1858, opened 2023-11-15
fix(react-native): import "node:events" lazily

### LINKED ISSUE #1847: Expo / React Native not having support for 'events'
### Prerequisites

### Environment check

### Node.js version

21.1.0

### Reproduction repository

https://github.com/Cocolis-1/blank-expo/tree/feat/msw

### Reproduction steps

Try to start the bundler using yarn start and start your dev client.

### Current behavior

You should see the following error:
```
iOS Bundling failed
The package at "node_modules/msw/lib/native/index.js" attempted to import the Node standard library module "events".
It failed because the native React runtime does not include the Node standard library.
Learn more: https://docs.expo.dev/workflow/using-libraries/#using-third-party-libraries
```

### Expected behavior

MSW should not depend on node:events (if possible) so that React Native users can use it.
I think the commit introducing node:events is [this one](https://github.com/mswjs/msw/commit/1033f651291f67a0ce509af965fdd0aa50fa7509).
Let me know if I can help in any way!

Thanks!

kettanaito:
- Fixes #1847 

## Changes

- Adds a `context` on the `SetupServerApi` class. 
- Adds `context.nodeEvents()` getter that lazily imports `node:events` and returns `undefined` if the import failed (e.g. in React Native). 
- Moves the `AbortSignal` max listeners handling from the request listener to a private method of the `SetupServerApi` class. Makes it fail-fast.

### DIFF
--- a/src/node/SetupServerApi.ts
+++ b/src/node/SetupServerApi.ts
@@ -1,4 +1,3 @@
-import { setMaxListeners, defaultMaxListeners } from 'node:events'
 import { invariant } from 'outvariant'
 import {
   BatchInterceptor,
@@ -13,7 +12,7 @@ import { RequiredDeep } from '~/core/typeUtils'
 import { mergeRight } from '~/core/utils/internal/mergeRight'
 import { handleRequest } from '~/core/utils/handleRequest'
 import { devUtils } from '~/core/utils/internal/devUtils'
-import { SetupServer } from './glossary'
+import { SetupServer, SetupServerInternalContext } from './glossary'
 import { isNodeExceptionLike } from './utils/isNodeExceptionLike'
 
 const DEFAULT_LISTEN_OPTIONS: RequiredDeep<SharedOptions> = {
@@ -24,6 +23,7 @@ export class SetupServerApi
   extends SetupApi<LifeCycleEventsMap>
   implements SetupServer
 {
+  private context: SetupServerInternalContext
   protected readonly interceptor: BatchInterceptor<
     Array<Interceptor<HttpRequestEventMap>>,
     HttpRequestEventMap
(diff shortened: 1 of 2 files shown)


src/node/SetupServerApi.ts
```diff
@@ -47,48 +48,22 @@ export class SetupServerApi
     this.init()
   }
 
+  private createContext(): SetupServerInternalContext {
+    return {
+      get nodeEvents() {
+        return import('node:events')
```
kettanaito, line 54: I _think_ this Promise should only ever fire once. I'm not sure if I'm structuring this correctly but the intention is to try to import this module _once_ so each outgoing request (request listener) doesn't pend on this import, increasing the response time).

mattcosta7 (approved):
I think this makes sense, hopefully that helps work around the issues with this api in react-native like environments. 

Re the import promise only being called once, I think it should resolve immediately since `import` calls are cached as far as I know, so that should be fine

DogukanTansuk:
@mattcosta7 Can we release a beta version with this change? I could try and give feedback.

kettanaito:
@DogukanTansuk, you can always install a package from a particular branch/commit. I believe all modern package managers support that (NPM, Yarn, PNPM). You can install it that way and let me know if this fixes the issue? Thanks!

kettanaito:
I'm promoting this to a release candidate because the change is backward-compatible and has no effect on the existing `msw/node` integrations. Folks would have to give this a try for `msw/native` and report back whether the issue is fixed. That's the optimal course of action until we have an official React Native example we can run in the existing smoke tests pipeline.

kettanaito:
## Released: v2.0.7 🎉

---

PR #1865, opened 2023-11-16
fix(deferNetworkUntil): restore module patches on registration rejections

### LINKED ISSUE #1573: MSW doesn't restore fetch is the worker is stopped before "deferNetworkRequestUntil" resolves
### Discussed in https://github.com/mswjs/msw/discussions/1570

<sup>Originally posted by **jkieboom** March 21, 2023</sup>
Asking as a discussion because I don't have a standalone repro case yet. I'm running into an issue where the `window.fetch` patching will run on top of an already patched `fetch` (from `msw` itself):

https://github.com/mswjs/msw/blob/115b435aef089a3c4fe9a618d5756c9cb944fe13/src/utils/deferNetworkRequestsUntil.ts#L23-L28

I'm setting up MSW in `beforeEach` (teardown in `afterEach`) and if a test does not make any network request that would wait for MSW to be setup, it is possible for the `fetch` patching to happen twice. It looks like `stop` does not "abort" the network request deferral.

I'm currently working around the issue by polling for "window.fetch" to be the native fetch again, after calling `stop`, but I feel this is a bug in `msw` fundamentally.</div>

kettanaito:
- Fixes #1573 

## Changes

- The deferred network feature now correctly restores module patches if the predicate promise rejects (e.g. worker registration failed). 
- The deferred network feature now correctly restores module patches if no requests were made using those modules (previously, the patches remained until the first request). 
- [ ] Abort the network deferral if `worker.stop()` has been called sooner before the predicate promise had a chance to resolve (the original reported issue).


src/browser/setupWorker/glossary.ts
```diff
@@ -98,7 +99,7 @@ export interface StrictEventListener<EventType extends Event> {
 }
 
 export interface SetupWorkerInternalContext {
-  isMockingEnabled: boolean
+  state: 'idle' | 'activating' | 'activated' | 'cancelled'
```
kettanaito, line 102: There's more to the worker state than `isMockingEnabled: boolean` now.

src/browser/setupWorker/glossary.ts
```diff
@@ -197,11 +198,13 @@ export interface StartOptions extends SharedOptions {
   findWorker?: FindWorker
 }
 
-export type StartReturnType = Promise<ServiceWorkerRegistration | undefined>
+export type StartReturnType = Promise<ServiceWorkerRegistration>
```
kettanaito, line 201: The public return type is still a plain Promise.

src/browser/setupWorker/glossary.ts
```diff
@@ -197,11 +198,13 @@ export interface StartOptions extends SharedOptions {
...
+
 export type StartHandler = (
   options: RequiredDeep<StartOptions>,
   initialOptions: StartOptions,
-) => StartReturnType
+) => DeferredPromise<ServiceWorkerRegistration>
```
kettanaito, line 206: The internal return type is a deferred promise so we can reject the pending activation promise at any time (e.g. when calling `stop()`).

src/browser/setupWorker/setupWorker.ts
```diff
@@ -36,6 +37,8 @@ export class SetupWorkerApi
   private startHandler: StartHandler = null as any
   private stopHandler: StopHandler = null as any
   private listeners: Array<Listener>
+  private workerActivationPromise: DeferredPromise<ServiceWorkerRegistration> | null =
```
kettanaito, line 40: Since the activation promise can be rejected now, it fulfills to either the registration reference or `null`.

src/browser/setupWorker/setupWorker.ts
```diff
@@ -177,11 +180,32 @@ export class SetupWorkerApi
...
     ) as SetupWorkerInternalContext['startOptions']
 
-    return await this.startHandler(this.context.startOptions, options)
+    this.workerActivationPromise = this.startHandler(
+      this.context.startOptions,
+      options,
+    )
+
+    this.workerActivationPromise.catch(() => {
+      this.context.state = 'cancelled'
```
kettanaito, line 189: No matter who rejects the activation promise, always transition to the `cancelled` state.

src/browser/setupWorker/start/createStartHandler.ts
```diff
@@ -13,131 +14,158 @@ export const createStartHandler = (
...
+    workerActivationPromise.then(async () => {
+      // Print the activation message once the worker has been activated.
+      await enableMocking(context, options)
+        .then(() => {
+          context.state = 'activated'
+        })
+        .catch((error) => {
           throw new Error(`Failed to enable mocking: ${error?.message}`)
         })
+    })
```
mattcosta7, line 160: I'm new to this registration process

But if there isn't a pending registration, do we never `enableMocking` since it appears the only call to resolve `workerActivationPromise` occurs in the state change listener now, hwere we used to `enableMocking` even when that hadn't resolved?

src/browser/setupWorker/start/createStartHandler.ts
```diff
@@ -13,131 +14,158 @@ export const createStartHandler = (
...
+              'statechange',
+              stateChangeListener,
+            )
+          }
         }
 
-        // Print the activation message only after the worker has been activated.
-        await enableMocking(context, options).catch((error) => {
+        pendingWorker.addEventListener('statechange', stateChangeListener)
+      }
```
mattcosta7, line 148:
```suggestion
      }
      workerActivationPromise.resolve(pendingWorkerRegistration)
```

I think we need this line here to catch this?

mattcosta7: if not, we may need some other way to handle this because fresh page loads aren't ever falling into this state change handler, which means that the `enableMocking` calls are never landing.

src/browser/setupWorker/start/createStartHandler.ts
```diff
@@ -13,131 +14,158 @@ export const createStartHandler = (
...
+        pendingWorkerRegistration.installing ||
+        pendingWorkerRegistration.waiting
+
+      // Wait until the worker is activated.
+      // Assume the worker is already activated if there's no pending registration
+      // (i.e. when reloading the page after a successful activation).
+      if (pendingWorker) {
+        const stateChangeListener = () => {
+          if (pendingWorker.state === 'activated') {
+            workerActivationPromise.resolve(pendingWorkerRegistration)
```
mattcosta7, line 138: we never resolve this when the worker wasn't `pending` which meant we never `enabledMocking`

kettanaito:
## Released: v2.1.0 🎉

---

Issue #1804, opened 2023-10-27
Custom request predicate function


kettanaito:
### Scope

Adds a new behavior

### Compatibility

- [ ] This is a breaking change

### Feature description

Right now, the `predicate` part of the request handler is usually a string or a RegExp (in some cases, a more complex data structure, like a `TypedDocumentNode` for GraphQL operations). That's handy but sometimes developers want to have completely custom predicates. For example:

- Only intercept requests with a particular search parameter;
- Only intercept requests with a certain header/body.

We encourage implementing [Custom request predicate](https://mswjs.io/docs/best-practices/custom-request-predicate) via the higher-order resolver function. While that's flexible, it's a bit upside-down from the request handler's perspective since the resolver is run _after_ the predicate phase. 

## Proposal

Expose a `({ request: Request }) => boolean` function as the accepted value for the `predicate` argument of any request handler.

```js
http.get(({ request }) => request.url.includes('foo'), resolver)
```

The previously accepted predicate values remain, this is an additive feature. 

This will allow for more streamlined request predicates, even support reusing them separately from the response resolver logic. 

```js
http.post(async ({ request, ...resolverArgs }) => {
  // Must clone the request, lest its body stream will be
  // locked in the response resolver.
  const body = await request.clone().json()
  return body.property === 'value'
}, resolver)
```

## Internal changes

In order to expose as much request info as possible, we'd have to move the `extendResolverArgs()` phase _before_ the call to the `predicate()` method of the `RequestHandler`. That shouldn't be a big deal as both existing handlers extend the resolver arguments with the data derived from the `parsedResult`.

kettanaito:
## Released: v2.11.0 🎉

This has been released in v2.11.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.11.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.11.0)

Make sure to always update to the latest version (`npm i msw@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

PR #1871, opened 2023-11-17
fix: skip parsing GraphQL requests if the request url doesn't match

mattcosta7:
When we have a specified graphql endpoint, and that endpoint doesn't match we don't need to parse the request at all.  This avoids weird console.errors that may apply when the user mixes rest and d graphql handlers. This fix cannot be easily applied to the base `graphql` handler since it doesn't have a dedicated endpoint, but that's a decent tradeoff for now, as it provides a way to better scope those errors and avoid user confusion in the short term.

Maybe we should consider forcing graphql calls to `link` a specific endpoint prior to handling which would avoid this issue

@kettanaito curious for thoughts on this type of thing.  

It would help me cleanup some code in a work related project, if we didn't have to add new overrides for the graphql cases, and could just avoid matching on the endpoint if graphql.link was given


mattcosta7:
Will come back to this in a little bit to cleanup the tests - want to add a bit more validation to them that both the graphql and http handlers work well, but my laptop died before I could push and I don't have a power cable with me


My use case I'm optimizing for is when using graphl.link and mixing http request with body parameter 'query' with graphql queries, in a test suite that will fail on calls to console in tests (except where explicitly expected)

src/core/handlers/GraphQLHandler.ts
```diff
@@ -156,10 +161,8 @@ Consider naming this operation or using "graphql.operation()" request handler to
       return false
     }
 
-    const hasMatchingUrl = matchRequestUrl(
-      new URL(args.request.url),
-      this.endpoint,
-    )
+    const hasMatchingUrl = this.hasMatches(args)
```
mattcosta7: The diff makes this look like it's called in the same function body as the call above, but it is not

kettanaito:
Hi, @mattcosta7. Thanks for opening this. 

As an optimization route, I think this has a place to be. From the `RequestHandler` perspective, the parsing phase always happens before the predicate phase since it provides the predicate with additional information (such as HTTP request `params` and `cookies` and GraphQL request `query` and `variables`). I understand that in cases when we know the exact endpoint, we may want to short-circuit the parsing phase if we know the endpoint doesn't match. But we have to be careful about decisions like that so the promise we establish in `RequestHandler` remains true. 

Overall, I wrote about the [Request handler phases](https://mswjs.io/docs/api/bypass) and I think it's a good read in the context of this change. 

So, with the proposed change, we will kind of do the predicate _before_ parsing to opt-out of parsing if the actual request URL doesn't match the GraphQL handler's endpoint. As things are right now, this does make sense but let's also consider #1804, which I'd like to ship rather soon to allow people to have custom request predicate functions. It looks like in that context, opting out from the parsing phase before the predicate phase won't be possible (as we don't know what the user-provided predicate may be for a GraphQL request handler). 

For example, consider a use case where I want to create a request handler for GraphQL requests that contain a `user` field in their query:

```js
// Note, this already features a yet non-existing 
// custom predicate API.
graphql.query(({ query }) => query.includes('user'), resolver)
```

This highlights the need for the parsing (`query`) to happen before the predicate (`query.includes('user')`). 

But I _think_ it shouldn't be a problem due to the nature of endpoint-based GraphQL interception:

- For handlers like `graphql.query` and `graphql.mutation` (that are endpoint-less), the preemptive URL matching will _always pass_ (since their URL predicate is literally `/.+/`). 
- For endpoint-scoped handlers created via `graphql.link(endpoint)`, the user already has the expectation of: 1) the endpoint must match; 2) _then_, my custom predicate must match. So it looks safe to ignore the custom predicate (i.e. parsing) if the endpoint doesn't match. 

Do you have any thoughts about this given this additional info?

src/core/handlers/GraphQLHandler.ts
kettanaito: Yeah, I noticed that but it's okay 👍

mattcosta7:
Thanks for the discussion here!  Traveling for the rest of the day and much of the next week but I'll get more thoughts here soon

mattcosta7:
> Hi, @mattcosta7. Thanks for opening this.
> 
> As an optimization route, I think this has a place to be. From the `RequestHandler` perspective, the parsing phase always happens before the predicate phase since it provides the predicate with additional information (such as HTTP request `params` and `cookies` and GraphQL request `query` and `variables`). I understand that in cases when we know the exact endpoint, we may want to short-circuit the parsing phase if we know the endpoint doesn't match. But we have to be careful about decisions like that so the promise we establish in `RequestHandler` remains true.

Yep, this makes sense!  I've slightly modified this implementation to handle the `match` on the url as part of the `parse` method, and return it from there, a bit more similar to how the HttpHandler does this.   In the event that there's an endpoint provided via `graphql.link` we'll avoid request parsing, because we've determined that the handler doesn't match already, so there's no need to continue with the graphql parsing attempt, since. we can't match.  As you mentioned later, this has no impact on endpoint-less graphql requests (I'm curious a bit as to why we'd need endpoint-less requests, given that every graphql server must live somewhere on a network - but that's not something we need to deal with at this point)

> Overall, I wrote about the [Request handler phases](https://mswjs.io/docs/api/bypass) and I think it's a good read in the context of this change.
> 
> So, with the proposed change, we will kind of do the predicate _before_ parsing to opt-out of parsing if the actual request URL doesn't match the GraphQL handler's endpoint. As things are right now, this does make sense but let's also consider #1804, which I'd like to ship rather soon to allow people to have custom request predicate functions. It looks like in that context, opting out from the parsing phase before the predicate phase won't be possible (as we don't know what the user-provided predicate may be for a GraphQL request handler).
> 
> For example, consider a use case where I want to create a request handler for GraphQL requests that contain a `user` field in their query:
> 
> ```js
> // Note, this already features a yet non-existing 
> // custom predicate API.
> graphql.query(({ query }) => query.includes('user'), resolver)
> ```
> 
> This highlights the need for the parsing (`query`) to happen before the predicate (`query.includes('user')`).
> 
> But I _think_ it shouldn't be a problem due to the nature of endpoint-based GraphQL interception:
> 
> * For handlers like `graphql.query` and `graphql.mutation` (that are endpoint-less), the preemptive URL matching will _always pass_ (since their URL predicate is literally `/.+/`).
> * For endpoint-scoped handlers created via `graphql.link(endpoint)`, the user already has the expectation of: 1) the endpoint must match; 2) _then_, my custom predicate must match. So it looks safe to ignore the custom predicate (i.e. parsing) if the endpoint doesn't match.
> 
> Do you have any thoughts about this given this additional info?


Yep!  I think this is correct. 

For 'endpoint-less' queries, the custom predicate approach should still be valid, and not impact much else, since we'll only skip parsing the request body once the request _cannot match_ - and with those, they will always match. 

I don't think it would make sense that a handler defining an endpoint (via link in this case) would then want a custom predicate that totally ignored it, but instead only be called for requests to that endpoint

src/core/handlers/GraphQLHandler.ts
```diff
@@ -31,6 +31,23 @@ export interface GraphQLHandlerInfo extends RequestHandlerDefaultInfo {
...
+export type GraphQLRequestParsedResult = {
+  match: Match
+} & (
+  | ParsedGraphQLRequest<GraphQLVariables>
+  /**
+   * An empty version of the ParsedGraphQLRequest
+   * which simplifies the return type of the resolver
+   * when the request is to a non-matching endpoint
+   */
+  | {
```
mattcosta7, line 43: we could also nest instead of merging match and these properties. I don't have a strong preference, besides not wanting to name the object containing these here.

it could be `graphqlAttributes` or `parsedGraphqlRequest`  but this is already a parsed result, so it feels a bit weird

src/core/handlers/GraphQLHandler.ts
mattcosta7: modified this now, so `match` gets returned from parse, and passed to predicate.  We do this in the http handler as well, so it adds a bit of uniformity between those

kettanaito:
> I don't think it would make sense that a handler defining an endpoint (via link in this case) would then want a custom predicate that totally ignored it, but instead only be called for requests to that endpoint

Agree. That'd be a no-op and intention conflict.

src/core/handlers/GraphQLHandler.test.ts
```diff
@@ -160,6 +160,12 @@ describe('parse', () => {
       })
 
       expect(await handler.parse({ request })).toEqual({
```
kettanaito, line 162: We also need to add the URL matching unit tests to the GraphQL handler:

- parses a graphql request if the given endpoint matches (for graphql.link)
- skips parsing if the given endpoint does not match (for graphql.link)

I assume the endpoint-less requests should be covered by the existing tests but an extra look at those won't hurt.

mattcosta7: added cases for this here: https://github.com/mswjs/msw/pull/1871/files#diff-81d8e35b978b65087de9c9c140e9cc932591997c3d5a464106b07d6fb0094fe5R374-R527

we could add more, but it seems like they'd end up being relatively duplicative of other cases already covered. this may already bit a bit more than necessary

src/core/handlers/GraphQLHandler.test.ts
```diff
@@ -26,19 +26,19 @@ const resolver: ResponseResolver<GraphQLResolverExtras<{ userId: string }>> = ({
 
 function createGetGraphQLRequest(
   body: GraphQLRequestBody<any>,
-  hostname = 'https://example.com',
+  graphqlEndpoint = 'https://example.com',
```
mattcosta7, line 29: kind of nitty, but I change the argument name here

hostname implies `example.com` to me, and I wanted it to be clear this is actual an endpoint.  this could be `graphqlUrl` or `graphqleServerHref` or something, but I think this naming is probably clear enough?

kettanaito: Bad naming on my part in the past, great correction.

src/core/handlers/GraphQLHandler.test.ts
kettanaito: Thanks! My thinking as well, no need to duplicate what's already covered by existing unit tests.

src/core/handlers/GraphQLHandler.test.ts
```diff
@@ -322,6 +370,161 @@ describe('parse', () => {
...
+        operationType: 'query',
+        operationName: 'GetUser',
+        query: GET_USER,
+        variables: {
+          userId: 'abc-123',
+        },
+      })
+    })
+
+    test('parses a request when the graphql.link hostname does not match', async () => {
```
kettanaito: Shouldn't this say "does not part the request" since the parsing is effectively skipped (no `query`, `variables`, etc. produced)?

src/core/handlers/GraphQLHandler.ts
kettanaito: As long as the existing typings tests pass, I have no objection how we handle this.

src/core/handlers/GraphQLHandler.test.ts
mattcosta7: The handler.parse succeeds successfully, but we don't parse the request body out. 

I'll work on the language here to make that a bit more clear.

mattcosta7: https://github.com/mswjs/msw/pull/1871/commits/064cfe1ed02b96583526a5c60592a98ede0914d4

updated. maybe a bit verbose now, but hopefully expresses that well.  I wanted to avoid it reading like `handler.parse` failed or returned nothing, when it's mostly focused on not returning the gql stuff from the request

kettanaito: Thank you!

kettanaito (approved):
No objections from my side. Superb work 👏

kettanaito:
One thing, don't merge this yet. I'd like to pair it with #1865.

mattcosta7:
> One thing, don't merge this yet. I'd like to pair it with #1865. 

Sounds good!  Feel free to merge whenever it makes sense to!

kettanaito:
## Released: v2.0.9 🎉

---

PR #1879, opened 2023-11-22
fix: node:events segfault

### LINKED ISSUE #1868: segfault in jest with msw@2.0.7
### Prerequisites

### Environment check

### Node.js version

v18.16.1

### Reproduction repository

https://github.com/mswjs/examples/tree/main/examples/with-jest

### Reproduction steps

```
cd examples/with-jest

nvm use 18.16.1

npm i msw@2.0.7

npm run test
```

### Current behavior

The tests crash with a segfault.

with-jest % npm run test   

> test
> jest

 RUNS  ./example.test.ts
zsh: segmentation fault  npm run test

### Expected behavior

I expect the tests to pass without segfault.

```
nvm use 18.16.1

npm i msw@2.0.6

npm run test
```

- works fine, so it definitely has something to do with the last release of msw.

This testcase in test repo also somehow does get fixed by upping the version of node, but the one in my project (which is more complicated, so i don't really want to try and recreate it exactly) does not.

mattcosta7:
fixes https://github.com/mswjs/msw/issues/1868

In https://github.com/mswjs/msw/pull/1858 we tried to work around the fact that `node:events` is not defined in `react-native` by default, leading to that environment throwing on setup. 

The fix applied previously was to dynamically import and test for those properties before accessing.  This is entirely functional, and should work, but from what I can gather from discussion and separate research, some node/jest combinations (and I wasn't really sure where this comes into play directly) cause tests to segfault on dynamic imports of builtin modules.   

While this is not something wrong with `msw` I think if we can find an alternative, we should try that out.  We also don't have much of a test suite for the `native` export, and maybe need to figure out how to support that - not sure if vite has a good path setup for that yet, but I think we need something to do it. 

I've suggested an alternative strategy here. I don't love it, but It's a non-breaking change and should hopefully help in this case. 

Ideally, an additional constructor argument to `SetupServerApi` would probably be the way to go, but since we expose `SetupServerApi` that would be a breaking change.  I think this provides enough value to ship now, and we can make those changes at the next major, if we decide to do this now.  Maybe we could do this purely additively as well, and i'll open a [second PR](https://github.com/mswjs/msw/pull/1880) to this one that tries that approach also.

There may be other alternatives here to explore, however this seemed like one of the simpler options overall

I also added a simple test for `msw/native`'s call to setupServer, just to have one.  it's in the wrong place (in the `test/node` folder) and isn't a native-like environment, but this seemed like the most straightforward place for adding this simple test case, for now

### DIFF
--- a/src/node/SetupServerApi.ts
+++ b/src/node/SetupServerApi.ts
@@ -1,18 +1,18 @@
-import { invariant } from 'outvariant'
 import {
   BatchInterceptor,
   HttpRequestEventMap,
   Interceptor,
   InterceptorReadyState,
 } from '@mswjs/interceptors'
+import { invariant } from 'outvariant'
 import { SetupApi } from '~/core/SetupApi'
 import { RequestHandler } from '~/core/handlers/RequestHandler'
 import { LifeCycleEventsMap, SharedOptions } from '~/core/sharedOptions'
 import { RequiredDeep } from '~/core/typeUtils'
-import { mergeRight } from '~/core/utils/internal/mergeRight'
 import { handleRequest } from '~/core/utils/handleRequest'
 import { devUtils } from '~/core/utils/internal/devUtils'
-import { SetupServer, SetupServerInternalContext } from './glossary'
+import { mergeRight } from '~/core/utils/internal/mergeRight'
+import { SetupServer, SetupServerContext } from './glossary'
 import { isNodeExceptionLike } from './utils/isNodeExceptionLike'
 
 const DEFAULT_LISTEN_OPTIONS: RequiredDeep<SharedOptions> = {
@@ -23,7 +23,9 @@ export class SetupServerApi
   extends SetupApi<LifeCycleEventsMap>
   implements SetupServer
 {
-  private context: SetupServerInternalContext
+  private context: SetupServerContext = {
+    nodeEvents: undefined,
+  } as const
   protected readonly interceptor: BatchInterceptor<
     Array<Interceptor<HttpRequestEventMap>>,
     HttpRequestEventMap
@@ -38,7 +40,6 @@ export class SetupServerApi
   ) {
     super(...handlers)
 
-    this.context = this.createContext()
     this.interceptor = new BatchInterceptor({
       name: 'setup-server',
       interceptors: interceptors.map((Interceptor) => new Interceptor()),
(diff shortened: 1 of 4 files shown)


test/node/native/native.node.test.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { http } from 'msw'
+import { setupServer } from 'msw/native'
```
kettanaito, line 5: can you please tell me more about when this test would fail? Since we are in Node, events module will always exist so I think this test isn't testing anything.

test/node/native/native.node.test.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * @vitest-environment jsdom
```
kettanaito, line 2: there's vitest env react-native but i'm not sure whether it strips unsupported global node modules.

src/node/SetupServerApi.ts
```diff
@@ -48,14 +49,23 @@ export class SetupServerApi
...
+   * In a future major we might want to consider adding a
+   * new argument to the constructor to allow for this
+   * as a bit more streamlined setup, but this allows
+   * us to remove a dynamic `node:events` import, provide
+   * the same functionality, and remove a jest/jsdom/node bug
+   * where dynamically importing a node module could segfault
+   *
+   */
+  withContext(context: SetupServerContext) {
+    this.context = context
```
kettanaito, line 67: this is a bit unusual. the concept of context isn't strictly defined in MSW internally but the way we use it in the worker api is that api creates the context within. with this change, we now have setupworker that has internal, non-extendable context, and setupserver that exclusively relies on the context being provided by the consumer. i think that's a mix of intentions and we probably should drop the word "context" to prevent any confusion. wdyt about `.withModules()` or something? and use that exclusively for standard modules injection.

test/node/native/native.node.test.ts
kettanaito: we can also model the env by ourselves if vitest allows different vitest configs per environment. we could create an `alias` option to `node:events` to resolve it with a module that throws, thus emulating what happens in react native. that can be an approach to consider.

mattcosta7: This mostly ensures that the server works when we don't pass node events through. 

Since they're more of a dependency injection in this model, since we're using msw/native and not msw/node we don't even attempt to import them

I'd love a better way to test this, but not sure how best to set that up currently

test/node/native/native.node.test.ts
mattcosta7: I'll give that a shot - must have missed that in vitest docs!

src/node/SetupServerApi.ts
kettanaito: @mattcosta7 i also think we should make this method private and call it explicitly internally. i'm not comfortable with this api enough to make it public and thus prevent us from iterating on it in the future. this isn't something MSW users need anyway.

test/node/native/native.node.test.ts
kettanaito: the closer to the actual environment we model the test, the better that test will be. in our case, we need a test environment that fails when importing the `node:events` module. then we ensure that when using msw/native, we can still intercept and mock things.

src/node/SetupServerApi.ts
mattcosta7: > @mattcosta7 i also think we should make this method private and call it explicitly internally. i'm not comfortable with this api enough to make it public and thus prevent us from iterating on it in the future. this isn't something MSW users need anyway.

The issue with calling it internally, is that we need to import the node:events module in this module if we do that, vs doing it outside of this module. 

We could leave it private and just `@ts-expect-error` on the call site for now, avoiding exposing it but still allowing for `node:events` to be defined elsewhere?   Maybe the additional argument type in the alternative proposal would be better, and avoid consumers using this in the shorter term?

Definitely open to more discussion here - agreed I don't love it either, but causing jest to break is probably worse than an unloved api hack?

mattcosta7:
Closing in favor of https://github.com/mswjs/msw/pull/1885

PR #1880, opened 2023-11-22
fix: inject node:event dependencies through `SetupServerApi` first argument in constructor

mattcosta7:
An alternative take on https://github.com/mswjs/msw/pull/1879

This uses a modified first argument to SetupServerApi that optionally uses a config object instead of a straight array of interceptors. 

Ideally, i think an api like

`SetupServerApi(interceptors, config, ...handlers)` might be clearer, but that would be a breaking change I'm trying to avoid, and the pure additive approach here seems reasonable as an iterative step 

I don't have a strong preference between this version and the one in the other referenced pr, which provides a context setter

### DIFF
--- a/src/node/SetupServerApi.ts
+++ b/src/node/SetupServerApi.ts
@@ -1,18 +1,18 @@
-import { invariant } from 'outvariant'
 import {
   BatchInterceptor,
   HttpRequestEventMap,
   Interceptor,
   InterceptorReadyState,
 } from '@mswjs/interceptors'
+import { invariant } from 'outvariant'
 import { SetupApi } from '~/core/SetupApi'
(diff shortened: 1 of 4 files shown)


src/node/SetupServerApi.ts
```diff
@@ -1,44 +1,59 @@
...
     HttpRequestEventMap
   >
   private resolvedOptions: RequiredDeep<SharedOptions>
 
   constructor(
-    interceptors: Array<{
-      new (): Interceptor<HttpRequestEventMap>
-    }>,
+    interceptorsOrConfig:
+      | Interceptors
```
kettanaito, line 40: I'm not a fan of overly dynamic call signatures. Node has had enough of me. I think your approach to solve this problem in #1879 is better.

PR #1883, opened 2023-11-22
chore: remove ts-ignores. force ts-expect-error

mattcosta7:
This is very much stylistic/preference, and feel free to close

I noticed `transfer` was ts-ignored, even though the types are no longer incorrect in the version of typescript we develop with, so I thought we should move to using `ts-expect-error` generally instead of ignores to avoid scenarios where errors _stop happening_ still being ignored

I added small ambient declarations for some third party modules we use internally to avoid needing to ignore them.  I was not complete in those typings, but that seems fair

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -66,11 +66,11 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
-      - name: Install TypeScript ${{ matrix.ts }}
-        run: pnpm add typescript@${{ matrix.ts }}
-
       - name: Build
         run: pnpm build
+
+      - name: Install TypeScript ${{ matrix.ts }}
+        run: pnpm add typescript@${{ matrix.ts }}
 
       - name: Typings tests
         run: |
--- a/src/browser/setupWorker/start/utils/createMessageChannel.ts
+++ b/src/browser/setupWorker/start/utils/createMessageChannel.ts
@@ -27,6 +27,14 @@ export class WorkerChannel {
     ...rest: WorkerChannelEventsMap[Event]
   ): void {
     const [data, transfer] = rest
-    this.port.postMessage({ type: event, data }, { transfer })
+    this.port.postMessage(
+      { type: event, data },
+      {
+        // but TypeScript doesn't acknowledge that.
+        // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
+        // @ts-ignore ReadableStream can be transferred, but only in versions 4.8 and lower
+        transfer,
+      },
+    )
   }
 }
--- a/src/browser/setupWorker/start/utils/createMessageChannel.ts
+++ b/src/browser/setupWorker/start/utils/createMessageChannel.ts
@@ -27,13 +27,6 @@ export class WorkerChannel {
     ...rest: WorkerChannelEventsMap[Event]
   ): void {
     const [data, transfer] = rest
-    this.port.postMessage(
-      { type: event, data },
-      {
-        // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
-        // @ts-ignore ReadableStream can be transferred, but this isn't handled in typescript versions 4.8 and below
-        transfer,
-      },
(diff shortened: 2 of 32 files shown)


src/browser/setupWorker/start/utils/createMessageChannel.ts
```diff
@@ -30,8 +30,9 @@ export class WorkerChannel {
     this.port.postMessage(
       { type: event, data },
       {
-        // @ts-ignore ReadableStream can be transferred
         // but TypeScript doesn't acknowledge that.
+        // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
+        // @ts-ignore ReadableStream can be transferred, but only in versions 4.8 and lower
```
mattcosta7: this is the line I initially wanted to fix, but we can't, because it doesn't work in 4.8.

If we drop typescript 4.8, it works fine though

test/browser/msw-api/setup-worker/start/error.mocks.ts
```diff
@@ -7,7 +7,8 @@ const worker = setupWorker(
   }),
 )
 
-// @ts-ignore
-window.msw = {
-  worker,
-}
+Object.assign(window, {
```
mattcosta7, line 10: instead of this, we could (and maybe should) also extend the window types in our tests to account for the few additional properties, and `invariant` them to ensure they're defined

kettanaito (dismissed):
This is a good change. Thanks, @mattcosta7 👏

test/browser/msw-api/setup-worker/start/error.mocks.ts
kettanaito: We do that for test files themselves but not for these examples. We can use the same `declare` statements in those too but it's a bit repetitive.

.github/workflows/compat.yml
```diff
@@ -66,12 +66,12 @@ jobs:
...
 
-      - name: Install TypeScript ${{ matrix.ts }}
-        run: pnpm add typescript@${{ matrix.ts }}
-
       - name: Build
         run: pnpm build
 
+      - name: Install TypeScript ${{ matrix.ts }}
+        run: pnpm add typescript@${{ matrix.ts }}
+
```
mattcosta7, line 74: shifted the build and typescript version change steps

We always want to _build_ with the currently installed version of typescript, which is what we ship to end users, but we want to run the typing tests on different versions

mattcosta7: @kettanaito this change dismissed the previous review, but also removes the need to have any `ts-ignores` so I removed that final one

src/browser/setupWorker/start/utils/createMessageChannel.ts
```diff
@@ -27,13 +27,6 @@ export class WorkerChannel {
     ...rest: WorkerChannelEventsMap[Event]
   ): void {
     const [data, transfer] = rest
-    this.port.postMessage(
-      { type: event, data },
-      {
-        // @ts-ignore ReadableStream can be transferred
```
mattcosta7, line 33: typescript supports this post versions 4.8, and we _build_ with a newer version that that

kettanaito: What about 4.7 that we still support?

mattcosta7: This is internal to msw, which is _built_ with more modern typescript. 

Because of this, we don't need to ignore it in our build process.  That ignore won't propagate to consumers anyway.

Consumers can consume this either way, as is evident by the compat tests succeeding.  Consumer's never see this type, because it's internal to us.

if projects using versions less than 4.7 tried to write their own source with this line it would have a typescript error, but typescript doesn't look at every line of node_modules, just the definition files, and the definiton file for this doesn't keep internal implementation, just signatures for methods

kettanaito (approved):

kettanaito:
## Released: v2.0.9 🎉

---

PR #1885, opened 2023-11-23
fix: resolve segfault issue caused by "node:events" import

### LINKED ISSUE #1868: segfault in jest with msw@2.0.7

mattcosta7:
- Fixes https://github.com/mswjs/msw/issues/1868

A 3rd attempt which uses a subclass to implement the node specific behavior, instead of the methods in:

- https://github.com/mswjs/msw/pull/1879
- https://github.com/mswjs/msw/pull/1880

I think this is my favorite implementation, and by far the least leaky, while still allowing control of the imports outside of the class definition, which allows us to better utilize this in _only_ the node export

I still haven't managed to recreate a proper test environment that actually throws, but we're at least avoidng the codepath here


test/native/tsconfig.json
```diff
@@ -0,0 +1,48 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "target": "esnext",
```
mattcosta7, line 4: kind of copied this from react-native's codebase - still trying to setup a test suite that fails correctly when it should

test/native/vitest.config.ts
```diff
@@ -0,0 +1,28 @@
+import * as path from 'node:path'
+import { defineConfig } from 'vitest/config'
+
+export default defineConfig({
```
kettanaito, line 6: This seems to be the exact copy of the existing vitest config for node tests. Cannot we just reference to it for native tests, which I assume is the intention here?

src/node/SetupNodeServerApi.ts
```diff
@@ -0,0 +1,47 @@
+import { defaultMaxListeners, setMaxListeners } from 'node:events'
+import { SetupServerApi } from './SetupServerApi'
+import { SetupServer } from './glossary'
+import { isNodeExceptionLike } from './utils/isNodeExceptionLike'
+
+export class SetupNodeServerApi extends SetupServerApi implements SetupServer {
```
kettanaito: I like this approach. A couple of thoughts to make it stellar:

1. Don’t introduce a new class.
2. Add a protected `onRequest` method to the existing `SetupServerApi` class. Do nothing in that method.
3. In `src/node/index.ts`, create a new subclass from `SetupServerApi`. In that subclass, invoke the abort listener logic directly (no need to abstract to a method). This way, the import to `node:events` will be scoped to `node/index.ts` and won’t leak to `/native` on `SetupServerApi.ts` imports there.
4. No changes are necessary for the React Native setup API. 

So, this is basically identical to what you did just inlining the new derived class directly in` src/node/index.ts`. What do you think about this?

kettanaito: My main concern here is that `SetupApi` classes are meant to cover functionality. Here, we sort of address the extension issue that's import-specific (if we import `node:events` in the base `SetupApi`, any consumer automatically gets that import, which is not what we want). An internal issue.

test/native/vitest.config.ts
mattcosta7: yes we probably can.  I was separating this to try and get a test environment that worked properly for `react-native` but failed to do so, and kept this to provide a path to doing that in the future.  I think it's probably ok to duplicate here, but we could also nest the `native` test inside of the node folder and just run it there, since it's fairly small. 

No strong preferences, although a working native setup would be ideal.  I couldn't even get `vi.mock` to throw for builtins here, which I also tried

src/node/SetupNodeServerApi.ts
mattcosta7: yep, makes sense.  

https://github.com/mswjs/msw/pull/1885/commits/1e44a5b3bcc2b12f6129bac3c49db8dbc4f5a458 might be what you're suggesting (just keeping from a separate file?)

test/native/node-events.native.test.ts
```diff
@@ -0,0 +1,22 @@
...
+   *
+   * We need Vitest to load the ESM build here in order to
+   * use "vi.mock()" to mock the "node:events" imports to throw.
+   * Vitest doesn't support module mocking in CommonJS.
+   *
+   * But aliasing the build isn't enough for it to function.
+   * The root-level package.json is still CJS, which, I suspects,
+   * resolves any subsequent in-build imports to their CJS counterparts.
+   */
+  expect(() => setupServer()).not.toThrow()
```
kettanaito, line 21: I've added a very basic test that due to the lack of proper testing environment, tests implementation detail of the `node:events` import (or its lack) in the `msw/native` build.

test/native/vitest.config.ts
kettanaito: I made the `vi.mock()` work but it requires a module alias hack in Vitest.

test/node/msw-api/setup-server/input-validation.node.test.ts
```diff
@@ -14,6 +14,6 @@ test('throws an error given an Array of request handlers to setupServer', () =>
   }
 
   expect(createServer).toThrow(
-    `[MSW] Failed to construct "SetupServerApi" given an Array of request handlers. Make sure you spread the request handlers when calling the respective setup function.`,
+    `[MSW] Failed to construct "SetupNodeServerApi" given an Array of request handlers. Make sure you spread the request handlers when calling the respective setup function.`,
```
kettanaito: I don't like the subclass name leaking to the consumer. We should somehow keep this as `SetupServerApi`.

test/native/vitest.config.ts
```diff
@@ -0,0 +1,28 @@
...
+       * @note Force Vitest load ESM targets of MSW.
+       * If we run ESM in tests, we can use "vi.mock()" to
+       * emulate certain standard Node.js modules missing
+       * (like "node:events") in React Native.
+       *
+       * Vitest won't pick up the ESM targets because
+       * the root-level "package.json" is not "module".
+       */
+      'msw/node': path.resolve(LIB_DIR, 'node/index.mjs'),
+      'msw/native': path.resolve(LIB_DIR, 'native/index.mjs'),
```
kettanaito, line 23: @mattcosta7, this is the hack that makes `vi.mock()` work with overriding the standard node modules.

test/node/msw-api/setup-server/input-validation.node.test.ts
kettanaito: Will fix this by not relying on the constructor name in error messages and instead introducing a new `displayName` property on the setup class.

kettanaito (dismissed):
This looks good. Thanks for fixing this, @mattcosta7! I've addressed the remaining feedback here and it should be ready for merge.

kettanaito:
## Released: v2.0.9 🎉

---

PR #1894, opened 2023-11-27
fix: force response bodies for 304s to be null

### LINKED ISSUE #1893: Faulty 304 handling results in `Failed to construct 'Response'` error
### Prerequisites

### Environment check

### Browsers

- Chrome (119.0.6045.123)
- Firefox (120.0)

### Reproduction repository

https://github.com/halfbakedsneed/msw-bad-304-handling

### Reproduction steps

```
pnpm install
pnpm start
```

Browse to http://localhost:3000 and hit "Click Me"

### Current behavior

For "bypassed" requests, responses with a `304` status code will result in the following being thrown:

```
Failed to construct 'Response': Response with null body status cannot have body
TypeError: Failed to construct 'Response': Response with null body status cannot have body
    at http://localhost:3000/static/js/bundle.js:48332:69
    at ServiceWorkerContainer.<anonymous> (http://localhost:3000/static/js/bundle.js:48595:15)
```

### Expected behavior

The service worker script correctly treats responses with a `204` status code [differently](https://github.com/mswjs/msw/blob/main/src/mockServiceWorker.js#L124-L128) when handling responses. The same behaviour should be exhibited when handling `304` responses, such that an error is not thrown.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304

>The response must not contain a body...

halfbakedsneed:
- Fixes https://github.com/mswjs/msw/issues/1893

## Changes

- Forces the response body to be null for responses with `304` status codes.

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -122,10 +122,13 @@ async function handleRequest(event, requestId) {
     ;(async function () {
       const responseClone = response.clone()
       // When performing original requests, response body will
-      // always be a ReadableStream, even for 204 responses.
+      // always be a ReadableStream, even for 204 / 304 responses.
       // But when creating a new Response instance on the client,
-      // the body for a 204 response must be null.
-      const responseBody = response.status === 204 ? null : responseClone.body
+      // the body for a 204 / 304 response must be null.
+      const responseBody =
+        response.status === 204 || response.status === 304
+          ? null
+          : responseClone.body
 
       sendToClient(
         client,
--- a/test/browser/msw-api/regression/null-body.mocks.ts
+++ b/test/browser/msw-api/regression/null-body.mocks.ts
@@ -5,6 +5,9 @@ const worker = setupWorker(
   http.get('/api/books', () => {
     return new HttpResponse(null, { status: 204 })
   }),
+  http.get('/api/authors', () => {
+    return new HttpResponse(null, { status: 304 })
+  }),
 )
 
 worker.start()
--- a/test/browser/msw-api/regression/null-body.test.ts
+++ b/test/browser/msw-api/regression/null-body.test.ts
@@ -18,3 +18,21 @@ test('gracefully handles a 204 response null body during life-cycle events', asy
 
   expect(errors).toEqual([])
 })
+
+test('gracefully handles a 304 response null body during life-cycle events', async ({
+  loadExample,
+  fetch,
+  page,
+}) => {
+  await loadExample(require.resolve('./null-body.mocks.ts'))
+
+  const errors: Array<Error> = []
+  page.on('pageerror', (pageError) => {
+    errors.push(pageError)
+  })
+
+  await fetch('/api/authors')
+  await sleep(500)
+
+  expect(errors).toEqual([])
+})
--- a/test/browser/rest-api/null-body-responses.test.ts
+++ b/test/browser/rest-api/null-body-responses.test.ts
@@ -9,6 +9,10 @@ const httpServer = new HttpServer((app) => {
   app.get('/posts', (req, res) => {
     return res.status(204).end()
   })
+
+  app.get('/comments', (req, res) => {
+    return res.status(304).end()
+  })
 })
 
 test.beforeAll(async () => {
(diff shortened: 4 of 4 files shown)


src/mockServiceWorker.js
```diff
@@ -122,10 +122,13 @@ async function handleRequest(event, requestId) {
...
       const responseClone = response.clone()
       // When performing original requests, response body will
-      // always be a ReadableStream, even for 204 responses.
+      // always be a ReadableStream, even for 204 / 304 responses.
       // But when creating a new Response instance on the client,
-      // the body for a 204 response must be null.
-      const responseBody = response.status === 204 ? null : responseClone.body
+      // the body for a 204 / 304 response must be null.
+      const responseBody =
+        response.status === 204 || response.status === 304
```
kettanaito, line 129: When would a 304 response have body?

```js
new Response(null, { status: 304 }).body
// null
```

If you provide the body to the `Respnse`/`HttpResponse` constructor, it will throw in your response resolver directly (the standard behavior, not MSW). 

Can you tell me more about when this would happen that we need an explicit check? 

I think historically this was here to prevent a no-op responses declarations in response resolvers. But that's not possible now since we rely on the `Response` class directly, and it will throw when constructing an invalid response (if anything, perhaps we should remove this check altogether).

mattcosta7: In the issue linked there's an example, I was able. to get it running making a change to the pacakge.json, and it appears in 

https://github.com/mswjs/msw/blob/8100e6d6b7c114fa8c2f584888672c2e53808f21/src/browser/setupWorker/start/createResponseListener.ts#L31

we're constructing a new Response with a readable stream as a body instead of null - which should likely be getting set here to clear out that body before sending it to the client?  we probably need to ensure this for a few different status codes

This seems to happen for a passthrough request where the origin server returns a 304 (correctly with no content)

mattcosta7: A null body [status](https://fetch.spec.whatwg.org/#statuses) is a [status](https://fetch.spec.whatwg.org/#concept-status) that is 101, 103, 204, 205, or 304.

I think we probably want to check each of these statuses and ensure the response body is null for all of them?

mattcosta7: maybe something like

```ts
// outside of this scope
const nullBodyStatuses = new Set([101, 103, 204, 205, 304])

// instead of status === 204 ||| status === 304
nullBodyStatuses.has(response.status) ? null : responseJson.body
```

src/mockServiceWorker.js
```diff
@@ -122,10 +122,13 @@ async function handleRequest(event, requestId) {
     ;(async function () {
       const responseClone = response.clone()
       // When performing original requests, response body will
-      // always be a ReadableStream, even for 204 responses.
+      // always be a ReadableStream, even for 204 / 304 responses.
```
mattcosta7, line 125:
```suggestion
      // always be a ReadableStream, even for null body responses (204, 304 and similar).
```

src/mockServiceWorker.js
kettanaito: > I think we probably want to check each of these statuses and ensure the response body is null for all of them?

Yes, with the correction that the Fetch API spec won't allow you to construct a 1xx response. No need to check against that. 

> maybe something like

We are getting this kind of logic repeated across MSW/Interceptors. Perhaps it's time we unified it and reused it across the libraries. 

This hardly qualifies as a standalone package though. We may expose this as a public API from the Interceptors since they are the common dependency here. Wdyt?

kettanaito: I also realized this will be reproducible if the response has an empty string body. It will result in an empty readable stream, which the Response constructor still considers a present response body, thus throwing.

kettanaito: I've opened a PR in Interceptors: https://github.com/mswjs/interceptors/pull/488

kettanaito:
I've added the `isResponseWithoutBody` utility function as the public API to Interceptors since [0.25.13](https://github.com/mswjs/interceptors/releases/tag/v0.25.13). We should update the dependency here and use that function instead of manually checking the response status code.

src/mockServiceWorker.js
mattcosta7: > We are getting this kind of logic repeated across MSW/Interceptors. Perhaps it's time we unified it and reused it across the libraries.

Do we actually want to use that method here, since this is in the service worker, I don't believe we bundle imports directly (it has none currently)

kettanaito: @mattcosta7, no, you're right, the worker would have to repeat that logic.

mattcosta7:
I opened a PR based off of this one, that should fix the issues noted here: 
https://github.com/mswjs/msw/pull/1895

kettanaito:
Closing in favor of #1895. Thank you for kicking this off, @halfbakedsneed 👏

kettanaito:
## Released: v2.0.10 🎉

This has been released in v2.0.10!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.10)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.10)

---

PR #1895, opened 2023-11-28
fix: null-body responses enforced in createResponseListener instead o…

### LINKED ISSUE #1893: Faulty 304 handling results in `Failed to construct 'Response'` error

mattcosta7:
closes https://github.com/mswjs/msw/pull/1894
fixes #1893 

This should fix passthrough errors specifically, where the body was a stream and the status code cannot support a body.

### DIFF
--- a/src/browser/setupWorker/start/createResponseListener.ts
+++ b/src/browser/setupWorker/start/createResponseListener.ts
@@ -3,7 +3,7 @@ import {
   SetupWorkerInternalContext,
 } from '../glossary'
 import { ServiceWorkerMessage } from './utils/createMessageChannel'
-
+import { isResponseWithoutBody } from '@mswjs/interceptors'
 export function createResponseListener(context: SetupWorkerInternalContext) {
   return (
     _: MessageEvent,
@@ -25,10 +25,14 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
       return
     }
 
-    const response =
-      responseJson.status === 0
-        ? Response.error()
-        : new Response(responseJson.body, responseJson)
+    let response: Response
+    if (responseJson.status === 0) {
+      response = Response.error()
+    } else if (isResponseWithoutBody(responseJson.status)) {
+      response = new Response(null, responseJson)
+    } else {
+      response = new Response(responseJson.body, responseJson)
+    }
 
     context.emitter.emit(
       responseJson.isMockedResponse ? 'response:mocked' : 'response:bypass',
--- a/src/browser/setupWorker/start/createResponseListener.ts
+++ b/src/browser/setupWorker/start/createResponseListener.ts
@@ -4,6 +4,7 @@ import {
 } from '../glossary'
 import { ServiceWorkerMessage } from './utils/createMessageChannel'
 import { isResponseWithoutBody } from '@mswjs/interceptors'
+
 export function createResponseListener(context: SetupWorkerInternalContext) {
   return (
     _: MessageEvent,
--- a/src/browser/setupWorker/start/createResponseListener.ts
+++ b/src/browser/setupWorker/start/createResponseListener.ts
@@ -25,14 +25,21 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
       return
     }
 
-    let response: Response
-    if (responseJson.status === 0) {
-      response = Response.error()
-    } else if (isResponseWithoutBody(responseJson.status)) {
-      response = new Response(null, responseJson)
-    } else {
-      response = new Response(responseJson.body, responseJson)
-    }
+    const response =
+      responseJson.status === 0
+        ? Response.error()
+        : new Response(
+            /**
+             * Responses may be streams here, but when we create a response object
+             * with null-body status codes, like 204, 205, 304 Response will
+             * throw when passed a non-null body, so ensure it's null here
+             * for those codes
+             */
+            isResponseWithoutBody(responseJson.status)
+              ? null
+              : responseJson.body,
+            responseJson,
+          )
 
     context.emitter.emit(
       responseJson.isMockedResponse ? 'response:mocked' : 'response:bypass',
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -121,11 +121,6 @@ async function handleRequest(event, requestId) {
   if (client && activeClientIds.has(client.id)) {
     ;(async function () {
       const responseClone = response.clone()
-      // When performing original requests, response body will
-      // always be a ReadableStream, even for 204 responses.
-      // But when creating a new Response instance on the client,
-      // the body for a 204 response must be null.
-      const responseBody = response.status === 204 ? null : responseClone.body
 
       sendToClient(
         client,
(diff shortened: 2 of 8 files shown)


src/mockServiceWorker.js
```diff
@@ -121,11 +121,6 @@ async function handleRequest(event, requestId) {
   if (client && activeClientIds.has(client.id)) {
     ;(async function () {
       const responseClone = response.clone()
-      // When performing original requests, response body will
-      // always be a ReadableStream, even for 204 responses.
-      // But when creating a new Response instance on the client,
-      // the body for a 204 response must be null.
-      const responseBody = response.status === 204 ? null : responseClone.body
```
mattcosta7, line 128: I moved this transform into the `createResponseListener` method, which is where we attempted to construct the Response with a body incorrectly

src/browser/setupWorker/start/createResponseListener.ts
```diff
@@ -25,10 +25,14 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
...
-        ? Response.error()
-        : new Response(responseJson.body, responseJson)
+    let response: Response
+    if (responseJson.status === 0) {
+      response = Response.error()
+    } else if (isResponseWithoutBody(responseJson.status)) {
+      response = new Response(null, responseJson)
+    } else {
+      response = new Response(responseJson.body, responseJson)
+    }
```
mattcosta7: Handling this here instead of in the worker. this gives us a more integrated place to handle this, and makes it easy to reuse the helper from interceptors

mattcosta7: could extract to a small method, not sure it's worth jumping out of line here though

test/browser/msw-api/regression/null-body.test.ts
```diff
@@ -1,20 +1,22 @@
...
-  })
+    const errors: Array<Error> = []
+    page.on('pageerror', (pageError) => {
+      errors.push(pageError)
+    })
 
-  await fetch('/api/books')
-  await sleep(500)
+    await fetch(`/api/${code}`)
+    await sleep(500)
```
kettanaito: Why sleep? I think this is a good opportunity to remove it. I don't believe it's needed. There are no delays in the mocked responses, and if we want to await some side effect, we need to write a proper promise for that.

mattcosta7: just bad copy/paste - will drop that

test/node/msw-api/req/passthrough.node.test.ts
```diff
@@ -9,6 +9,15 @@ const httpServer = new HttpServer((app) => {
   app.post<never, ResponseBody>('/user', (req, res) => {
     res.json({ name: 'John' })
   })
+  app.post('/code/204', (req, res) => {
```
kettanaito: Nitpick: why not

```js
app.post('/code/:code', (req, res) => {
  res.status(req.params.code).end()
})
```

test/node/msw-api/req/passthrough.node.test.ts
```diff
@@ -88,3 +97,35 @@ it('performs a request as-is if nothing was returned from the resolver', async (
     name: 'John',
   })
 })
+
+it.each([{ code: 204 }, { code: 205 }, { code: 304 }])(
```
kettanaito: It'd be simpler to keep the array of status codes, the same you do in the other tests.

test/node/msw-api/req/passthrough.node.test.ts
```diff
@@ -88,3 +97,35 @@ it('performs a request as-is if nothing was returned from the resolver', async (
...
+  'performs a $code request as-is if nothing was returned from the resolver',
```
kettanaito: `$code` looks like a non-resolved template string value, does it not?

kettanaito:
Does this change negate #1894? If so, we can mark it as "Closes #1894".

src/browser/setupWorker/start/createResponseListener.ts
```diff
@@ -3,7 +3,7 @@ import {
   SetupWorkerInternalContext,
 } from '../glossary'
 import { ServiceWorkerMessage } from './utils/createMessageChannel'
-
+import { isResponseWithoutBody } from '@mswjs/interceptors'
```
kettanaito, line 6: Nitpick: a new line after the import would be nice. Are by any chance not running Prettier?

src/browser/setupWorker/start/createResponseListener.ts
```diff
@@ -25,10 +25,14 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
       return
     }
 
-    const response =
-      responseJson.status === 0
-        ? Response.error()
-        : new Response(responseJson.body, responseJson)
+    let response: Response
```
kettanaito: Would this be too confusing, in your opinion? 

```js
const response = responseJson.status === 0
  ? Response.error()
  : new Response(isResponseWithoutBody(responseJson.status) ? null : responseJson.body, responseJson)
```

kettanaito: But yes, overall, I prefer this being a standalone method that we can also unit test to perfection. That way the usage is clear and the implementation can use early returns to stay clear also.

mattcosta7: I think this style inline is probably the easiest to read, but not necessarily a big deal either way. 

Since we don't do this anywhere else, extracting it seems heavy, and more annoying to skim through, but no strong feeling

test/node/msw-api/req/passthrough.node.test.ts
mattcosta7: `.each` extracts that `$code` and inlines the value from the object. 

Playwright doesn't support it but vitest does.  we could do either way, but I was just doing this style here because the runner supports it.

test/node/msw-api/req/passthrough.node.test.ts
mattcosta7: no reason, really.  I went back and forth on this actually, and settled on the explicit handlers, since it's a bit easier to parse quickly that we're only defining a couple codes that we care about vs arbitrary ones (especially if we typod a non-numeric code)

src/browser/setupWorker/start/createResponseListener.ts
mattcosta7: I usually have it running, maybe it just lagged in vscode?

kettanaito (approved):
🚀

kettanaito:
## Released: v2.0.10 🎉

This has been released in v2.0.10!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.10)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.10)

---

PR #1896, opened 2023-11-29
chore: remove unused method and parameters

mattcosta7:
While looking through some of the codebase, I noticed that there are undocumented and unused properties/methods.  I've removed them here, which I guess _could_ be considered breaking, but since these are undocumented and unused in msw directly, I don't think we're actually breaking any external contracts here


src/core/utils/getResponse.ts
```diff
@@ -9,23 +9,18 @@ export interface ResponseLookupResult {
   response?: Response
 }
 
-export interface ResponseResolutionContext {
-  baseUrl?: string
-}
```
mattcosta7, line 14: We don't have any code that tests for something inclusive of this property, and I couldn't locate anything that defines it, so I'm assuming this became unused during the v1 -> 2 migration?

src/core/handlers/RequestHandler.ts
```diff
@@ -149,32 +147,10 @@ export abstract class RequestHandler<
...
-    resolutionContext?: ResponseResolutionContext
-  }): Promise<ParsedResult> {
+  async parse(_args: { request: Request }): Promise<ParsedResult> {
     return {} as ParsedResult
   }
 
-  /**
-   * Test if this handler matches the given request.
-   */
-  public async test(args: {
```
mattcosta7, line 162: We only called `test` in tests.  Not sure if we need these anymore, but maybe this is only public to assert some behavior works?

If so we can extend the documentation, and maybe make this `private` but ignore that privacy in tests

mattcosta7: We could also change the signature here and call it directly in `run`

```ts
public async test(args: {
    request: Request
  }): Promise<[boolean, parsedResult]> {
    const parsedResult = await this.parse({
      request: args.request,
    })

    return [
      this.predicate({
          request: args.request,
          parsedResult, 
      }),
      parsedResult
   ]
  }
```

kettanaito: Although we don't call this anywhere, the `.test()` method on the request handler is still a part of its public API. This becomes useful for custom request handlers as the developer can _test_ the handler against any request at any time. 

This is much more convenient than having to do parse + predicate manually (which is precisely what `.test()` abstracts).

I'd leave this method be, it's not a mistake. 

The action point here is to document the `.test()` method and give an example of usage in the [docs](https://mswjs.io/docs/api/request-handler).

src/core/utils/getResponse.ts
kettanaito: To be safe here, I recommend checking out the v1 tag in Git and taking a peek at what this option was used before, if at all.

src/core/utils/url/getAbsoluteUrl.test.ts
```diff
@@ -7,12 +7,6 @@ test('rebases a relative URL against the current "baseURI" (default)', () => {
   expect(getAbsoluteUrl('/reviews')).toEqual('http://localhost/reviews')
 })
 
-test('rebases a relative URL against a custom base URL', () => {
```
kettanaito, line 10: Why is this removed?

src/core/utils/matching/normalizePath.ts
```diff
@@ -6,16 +6,15 @@ import { getAbsoluteUrl } from '../url/getAbsoluteUrl'
  * Normalizes a given request handler path:
  * - Preserves RegExp.
  * - Removes query parameters and hashes.
- * - Rebases relative URLs against the "baseUrl" or the current location.
  * - Preserves relative URLs in Node.js, unless specified otherwise.
  */
-export function normalizePath(path: Path, baseUrl?: string): Path {
```
kettanaito, line 12: This is a breaking change of the public API.

While `normalizePath` isn't public, it's used in the `matchRequestUrl`, which _is_ public and does support the `baseUrl` optional argument. 

I think we should revert this.

src/core/utils/matching/matchRequestUrl.ts
```diff
@@ -55,8 +55,8 @@ export function coercePath(path: string): string {
 /**
  * Returns the result of matching given request URL against a mask.
  */
-export function matchRequestUrl(url: URL, path: Path, baseUrl?: string): Match {
-  const normalizedPath = normalizePath(path, baseUrl)
+export function matchRequestUrl(url: URL, path: Path): Match {
```
kettanaito, line 58: This is a public API change. We have to keep it for any consumers depending on this function. Perhaps it's even us.

kettanaito (changes requested):
Thanks for a bit of gardening, @mattcosta7! 

Overall, a good change. I left a few comments about the things we should not remove.

src/core/utils/getResponse.ts
mattcosta7: ooo - it looks like this is used here: https://github.com/mswjs/http-middleware/blob/dfd8c682696f47f3242c6f5a33ab9077a656c5b5/src/middleware.ts#L49C1-L55C11

which i came across, in which case maybe we need to do better at documenting that internally, even in test scenarios, to describe how it works and under what scenarios

mattcosta7:
I'll close this because it seems some of these are intentional, but not well documented and we can hopefully improve documentation

src/core/utils/getResponse.ts
kettanaito: Agree. A documentation issue.

PR #1905, opened 2023-12-03
[experiment] chore: performance updates

mattcosta7:
Expanding from: https://github.com/mswjs/msw/pull/1900

I was interested in trying to optimize the `handleRequest` pipeline without making substantial changes.  I hope this can help inform some future updates suggested in https://github.com/mswjs/msw/discussions/1901 and other related areas. 

I don't expect this branch to land directly at all, but solely to be used as discovery

## Takeaways

We can make significant improvements to performance and scalability with a few relatively small code changes. 

- Prior to executing handlers create a `requestUrl = new URL(request.url) and `mainRequestRef = request.clone()` passing them directly to each handler. We may not even need a `mainRequestRef` at all, as it doesn't seem used anywhere directly. We might not need to memoize `new URL` if we extract that outside of the handlers, but we'd still have this cost where we might not need it as frequently
- Cache `matchRequestUrl` lazily. Every request has to hit all paths, but once that occurs we'll be able to precompute the matches later, so we should be ok to re-use those after initial matching, giving a good cache key
- Cache `parseGraphQLRequest` results across handlers for a given request. Should we also cache the actual query parse here, which may be the same for multiple variables, where that input might currently not be identical - we should verify this (I did not try that yet)
- Cache `getAllRequestCookies` - this is actually cheap compared to everything else discussed above

## Test structure

Before: MSW v2.0.9 (with the `unshift` fix from this branch)
After: this branch

Results are described as a single run, and not a many run average, but they are very stable results so they're indicative of larger test windows

Using my m1 macbook pro, latest chrome. 2 tabs open (this issue and a vite sample project - code below)
100,000 handlers configured.

Fresh page load:

- Perform a single `http` request to the first handler.
- Perform a single `http` request to the last handler.
- Perform a second `http` request to the last handler

Fresh page load: 

- Perform a single `graphql` request to the first handler.
- Perform a single `graphql` request to the last handler.
- Perform a second `graphql` request to the last handler

### Http

1st handler: 5ms
Last handler (1x): 10.42s
Last handler (2x): 10.97s

Screen Shot 2023-12-03 at 8 46 18 AM

### Graphql

1st handler: 6ms
Last handler (1x): 13.20s
Last handler (2x): 13.15s

Screen Shot 2023-12-03 at 8 45 16 AM

### Http

1st handler: 5ms
Last handler (1x): 2.41s
Last handler (2x): 818ms

Screen Shot 2023-12-03 at 8 56 26 AM

### Graphql

1st handler: 6ms
Last handler (1x): 1.83s
Last handler (2x): 1.81s

Screen Shot 2023-12-03 at 8 56 45 AM

## Findings

1st handler results stay about identical.
Last handler results improve drastically
- HTTP 
    - last handler first request - 76.3%
    - last handler last request - 92.5%
    - average last handler improvement - 84.98%
- GraphQL
    - last handler first request - 86.1%
    - last handler last request - 86.2%
    - average last handler improvement - 86.15%

Obviously these results are extreme because of the number of handlers, but even for more moderate handler numbers these results are very promising

When a smaller, 50 handler set is used (raw numbers only)

Before:
- Http
    - first handler - 4ms
    - last handler (1x) - 13ms
    - last handler (2x) - 13ms
- GrapQL
    - first handler - 5ms
    - last handler (1x) - 15ms
    - last handler (2x) - 16ms

After:

- Http
    - first handler - 4ms
    - last handler (1x) - 6ms
    - last handler (2x) - 6ms
- GrapQL
    - first handler - 4ms
    - last handler (1x) - 6ms
    - last handler (2x) - 6ms

## Description of changes

Setup changes, equal across tests

* <a href="diffhunk://#diff-c9e56be02b046c738d10384a7b8490e0ea21fc397f6904d2b88dcc58fc745b34R62-R64">`src/core/SetupApi.ts`</a> and <a href="diffhunk://#diff-ea77d8a9f875c81840eb3b697aa5bcbda54cb96619c7a20a915313877626663bR7-R9">`src/core/utils/internal/requestHandlerUtils.ts`</a>: Instead if calling unshift with a spread, loop through the handlers in reverse and unshift them individually. This avoids maximum stack issues when a large number of handlers are 'use'd. Spreading forces the arguments all onto the stack, which is why it overflows. <a href="diffhunk://#diff-c9e56be02b046c738d10384a7b8490e0ea21fc397f6904d2b88dcc58fc745b34L62-R64">[1]</a> <a href="diffhunk://#diff-ea77d8a9f875c81840eb3b697aa5bcbda54cb96619c7a20a915313877626663bL7-R9">[2]</a>
* <a href="diffhunk://#diff-0b280a445be167f54cd916c0718c952b8e0d4ca9621903d05eec25efd4c6ed6dL4-R4">`tsconfig.base.json`</a>: Updated the TypeScript target version from ES6 to ES2020. These made debugging/testing simpler, since we don't create generators for promise resolution

Performance improvements:

* <a href="diffhunk://#diff-b6e49d193f2bfa26b1cbca804c9bc9636e72f980f5264116d49f0d3a43cc4a5fR1-R9">`src/core/utils/memoizedUrl.ts`</a>: Introduced a `memoizedUrl` function to create and cache URL objects, reducing the overhead of repeatedly creating new URL objects.
* <a href="diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6R192">`src/core/handlers/RequestHandler.ts`</a>: Reduced unnecessary request cloning by introducing a `mainRefCache` to store a reference to the original request. We probably don't need to clone this at all, but minimizing changes to call signatures meant we needed to provide a clone to the execution result for the one handler that matches. We don't seem to read this, so maybe we don't need it.  If we _do_ can we do this prior to starting the resolution pipeline?  <a href="diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6R192">[1]</a> <a href="diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6L201-R223">[2]</a> <a href="diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6L241-L242">[3]</a>
* <a href="diffhunk://#diff-c604d2555e89b3bedcb745c2e268a3f04e2b45f300061e31a7e1782a2000a6baR170-R182">`src/core/utils/internal/parseGraphQLRequest.ts`</a>: Added a caching mechanism to the `parseGraphQLRequest` function to store and reuse the parsed result of a GraphQL request. <a href="diffhunk://#diff-c604d2555e89b3bedcb745c2e268a3f04e2b45f300061e31a7e1782a2000a6baR170-R182">[1]</a> <a href="diffhunk://#diff-c604d2555e89b3bedcb745c2e268a3f04e2b45f300061e31a7e1782a2000a6baL188-R191">[2]</a> <a href="diffhunk://#diff-c604d2555e89b3bedcb745c2e268a3f04e2b45f300061e31a7e1782a2000a6baL199-R209">[3]</a>
* <a href="diffhunk://#diff-bb49a1e12580b5a8cd56124a5e43fc4681fe563544546886d6d8445bfaf523a8R55-R61">`src/core/utils/matching/matchRequestUrl.ts`</a>: Introduced a cache in the `matchRequestUrl` function to store and reuse the result of URL-path matching. <a href="diffhunk://#diff-bb49a1e12580b5a8cd56124a5e43fc4681fe563544546886d6d8445bfaf523a8R55-R61">[1]</a> <a href="diffhunk://#diff-bb49a1e12580b5a8cd56124a5e43fc4681fe563544546886d6d8445bfaf523a8L69-R77">[2]</a>
* <a href="diffhunk://#diff-3abcaf3e23405a6ac736486689634065f5c5b6ef27f366f1ea6c1abe997bf1f1R41-R43">`src/core/utils/request/getRequestCookies.ts`</a>: Added a cache to the `getAllRequestCookies` function to store and reuse the parsed cookies from a request. <a href="diffhunk://#diff-3abcaf3e23405a6ac736486689634065f5c5b6ef27f366f1ea6c1abe997bf1f1R41-R43">[1]</a> <a href="diffhunk://#diff-3abcaf3e23405a6ac736486689634065f5c5b6ef27f366f1ea6c1abe997bf1f1L71-R79">[2]</a>

## Code used in test setup:

```tsx
import { createRoot } from "react-dom/client";
import { setupWorker } from "msw/browser";
import { Outlet, RouterProvider, createBrowserRouter } from "react-router-dom";
import { HttpResponse, graphql, http } from "msw";

const worker = setupWorker();
const FIRST = 1;
const LAST = 100_000;
const Layout = () => {
  return (
    <main>
      <nav>
        <ul>
          <li>
            <a href="/graphql">GraphQL</a>
          </li>
          <li>
            <a href="/http">HTTP</a>
          </li>
        </ul>
      </nav>
      <Outlet />
    </main>
  );
};

function GraphQL() {
  return (
    <>
      <h1>GraphQL</h1>
      <button
        onClick={() => {
          makeGraphqlRequest(FIRST);
        }}
      >
        {FIRST}
      </button>
      <button
        onClick={() => {
          makeGraphqlRequest(LAST);
        }}
      >
        {LAST}
      </button>
    </>
  );
}

function Http() {
  return (
    <>
      <h1>Http</h1>
      <button
        onClick={() => {
          makeHttpRequest(FIRST);
        }}
      >
        {FIRST}
      </button>
      <button
        onClick={() => {
          makeHttpRequest(LAST);
        }}
      >
        {LAST}
      </button>
    </>
  );
}

worker.start().then(() => {
  const router = createBrowserRouter([
    {
      path: "/",
      element: <Layout />,
      children: [
        {
          path: "graphql",
          element: <GraphQL />,
          loader() {
            worker.restoreHandlers();
            worker.use(
              ...Array.from({ length: LAST }, (_, i) => {
                return graphql.query(`GetUser${i + 1}`, () => {
                  return HttpResponse.json({
                    data: {
                      user: {
                        id: i + 1,
                      },
                    },
                  });
                });
              })
            );
            return {};
          },
        },
        {
          path: "http",
          element: <Http />,
          loader() {
            worker.restoreHandlers();
            worker.use(
              ...Array.from({ length: LAST }, (_, i) => {
                return http.post(`/http/${i + 1}`, () => {
                  return HttpResponse.json({
                    data: {
                      user: {
                        id: i + 1,
                      },
                    },
                  });
                });
              })
            );
            return {};
          },
        },
      ],
    },
  ]);
  createRoot(document.getElementById("app")!).render(
    <RouterProvider router={router} />
  );
});

// share identical request shape and data to minimze graphql<-> http differences
function toRequestInit(id: number): RequestInit {
  return {
    method: "POST",
    headers: {
      "Content-Type": "application/json",
    },
    body: JSON.stringify({
      query: `
            query GetUser${id} {
              user {
                id
              }
            }
          `,
    }),
  };
}

function makeGraphqlRequest(id: number) {
  return fetch("/graphql", toRequestInit(id));
}

function makeHttpRequest(id: number) {
  return fetch(`/http/${id}`, toRequestInit(id));
}
```

### DIFF
--- a/src/core/SetupApi.ts
+++ b/src/core/SetupApi.ts
@@ -59,7 +59,9 @@ export abstract class SetupApi<EventsMap extends EventMap> extends Disposable {
       ),
     )
 
-    this.currentHandlers.unshift(...runtimeHandlers)
+    for (let i = runtimeHandlers.length - 1; i >= 0; i--) {
+      this.currentHandlers.unshift(runtimeHandlers[i])
+    }
   }
 
   public restoreHandlers(): void {
--- a/src/core/handlers/GraphQLHandler.ts
+++ b/src/core/handlers/GraphQLHandler.ts
@@ -20,6 +20,7 @@ import {
 import { getPublicUrlFromRequest } from '../utils/request/getPublicUrlFromRequest'
 import { devUtils } from '../utils/internal/devUtils'
 import { getAllRequestCookies } from '../utils/request/getRequestCookies'
+import { memoizedUrl } from '../utils/memoizedUrl'
 
 export type ExpectedOperationTypeNode = OperationTypeNode | 'all'
 export type GraphQLHandlerNameSelector = DocumentNode | RegExp | string
@@ -138,7 +139,7 @@ export class GraphQLHandler extends RequestHandler<
      * If the request doesn't match a specified endpoint, there's no
      * need to parse it since there's no case where we would handle this
      */
-    const match = matchRequestUrl(new URL(args.request.url), this.endpoint)
+    const match = matchRequestUrl(memoizedUrl(args.request.url), this.endpoint)
     if (!match.matches) return { match }
 
     const parsedResult = await parseGraphQLRequest(args.request).catch(
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -202,6 +202,15 @@ export abstract class RequestHandler<
       return null
     }
 
+    /**
+     * @deprecated do we need this request?
+     * Or, can we pass it into run instead,
+     * since it's _only_ used in the execution result,
+     * but then never read or utilized.
+     *
+     * We don't want to copy this for _every_ handler, as it
+     * is expensive to do so.
+     */
     const mainRequestRef = (() => {
       if (RequestHandler.mainRefCache.has(args.request)) {
         return RequestHandler.mainRefCache.get(
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -87,6 +87,10 @@ export interface RequestHandlerExecutionResult<
 > {
   handler: RequestHandler
   parsedResult?: ParsedResult
+  /**
+   * @deprecated do we need this request?
+   */
+  request: Request
   response?: Response
 }
 
@@ -185,6 +189,7 @@ export abstract class RequestHandler<
     return {} as ResolverExtras
   }
 
+  static mainRefCache = new WeakMap<Request, Request>()
   /**
    * Execute this request handler and produce a mocked response
    * using the given resolver function.
@@ -196,6 +201,17 @@ export abstract class RequestHandler<
     if (this.isUsed && this.options?.once) {
       return null
     }
+
+    const mainRequestRef = (() => {
+      if (RequestHandler.mainRefCache.has(args.request)) {
+        return RequestHandler.mainRefCache.get(
+          args.request,
+        ) as StrictRequest<any>
+      }
+      const clone = args.request.clone()
+      RequestHandler.mainRefCache.set(args.request, clone)
+      return clone
+    })()
 
     const parsedResult = await this.parse({
       request: args.request,
@@ -233,6 +249,7 @@ export abstract class RequestHandler<
     })) as Response
 
     const executionResult = this.createExecutionResult({
+      request: mainRequestRef,
       response: mockedResponse,
       parsedResult,
     })
@@ -290,11 +307,13 @@ export abstract class RequestHandler<
   }
 
   private createExecutionResult(args: {
+    request: Request
     parsedResult: ParsedResult
     response?: Response
   }): RequestHandlerExecutionResult<ParsedResult> {
     return {
       handler: this,
+      request: args.request,
       response: args.response,
       parsedResult: args.parsedResult,
     }
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -87,7 +87,6 @@ export interface RequestHandlerExecutionResult<
 > {
   handler: RequestHandler
   parsedResult?: ParsedResult
-  request: Request
   response?: Response
 }
 
@@ -198,10 +197,6 @@ export abstract class RequestHandler<
       return null
     }
 
-    // Clone the request instance before it's passed to the handler phases
-    // and the response resolver so we can always read it for logging.
-    const mainRequestRef = args.request.clone()
-
     const parsedResult = await this.parse({
       request: args.request,
       resolutionContext: args.resolutionContext,
@@ -238,9 +233,6 @@ export abstract class RequestHandler<
     })) as Response
 
     const executionResult = this.createExecutionResult({
-      // Pass the cloned request to the result so that logging
-      // and other consumers could read its body once more.
-      request: mainRequestRef,
       response: mockedResponse,
       parsedResult,
     })
@@ -298,13 +290,11 @@ export abstract class RequestHandler<
   }
 
   private createExecutionResult(args: {
-    request: Request
     parsedResult: ParsedResult
     response?: Response
   }): RequestHandlerExecutionResult<ParsedResult> {
     return {
       handler: this,
-      request: args.request,
       response: args.response,
       parsedResult: args.parsedResult,
     }
--- a/src/core/utils/internal/parseGraphQLRequest.ts
+++ b/src/core/utils/internal/parseGraphQLRequest.ts
@@ -167,6 +167,7 @@ async function getGraphQLInput(request: Request): Promise<GraphQLInput | null> {
   }
 }
 
+const cache = new WeakMap<Request, ParsedGraphQLRequest>()
 /**
  * Determines if a given request can be considered a GraphQL request.
  * Does not parse the query and does not guarantee its validity.
@@ -174,9 +175,11 @@ async function getGraphQLInput(request: Request): Promise<GraphQLInput | null> {
 export async function parseGraphQLRequest(
   request: Request,
 ): Promise<ParsedGraphQLRequest> {
+  if (cache.has(request)) return cache.get(request)
   const input = await getGraphQLInput(request)
 
   if (!input || !input.query) {
+    cache.set(request, undefined)
     return
   }
 
@@ -185,7 +188,7 @@ export async function parseGraphQLRequest(
 
   if (parsedResult instanceof Error) {
     const requestPublicUrl = getPublicUrlFromRequest(request)
-
+    cache.set(request, undefined)
     throw new Error(
       devUtils.formatMessage(
         'Failed to intercept a GraphQL request to "%s %s": cannot parse query. See the error message from the parser below.\n\n%s',
@@ -196,10 +199,12 @@ export async function parseGraphQLRequest(
     )
   }
 
-  return {
+  const data = {
     query: input.query,
     operationType: parsedResult.operationType,
     operationName: parsedResult.operationName,
     variables,
   }
+  cache.set(request, data)
+  return data
 }
--- a/src/core/utils/matching/matchRequestUrl.ts
+++ b/src/core/utils/matching/matchRequestUrl.ts
@@ -52,10 +52,13 @@ export function coercePath(path: string): string {
   )
 }
 
+const cache = new Map<string, Match>()
 /**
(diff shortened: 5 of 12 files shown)


src/core/SetupApi.ts
```diff
@@ -59,7 +59,9 @@ export abstract class SetupApi<EventsMap extends EventMap> extends Disposable {
       ),
     )
 
-    this.currentHandlers.unshift(...runtimeHandlers)
```
mattcosta7, line 62: I tested a case of applying a milion handlers, which failed due to the stack size, since spread places all of the array items onto the stack

src/core/handlers/GraphQLHandler.ts
```diff
@@ -138,7 +139,7 @@ export class GraphQLHandler extends RequestHandler<
      * If the request doesn't match a specified endpoint, there's no
      * need to parse it since there's no case where we would handle this
      */
-    const match = matchRequestUrl(new URL(args.request.url), this.endpoint)
```
mattcosta7, line 141: URL parsing is kind of expensive, and we can memoize this easily

src/core/handlers/RequestHandler.ts
```diff
@@ -198,9 +202,25 @@ export abstract class RequestHandler<
...
+    /**
+     * @deprecated do we need this request?
+     * Or, can we pass it into run instead,
+     * since it's _only_ used in the execution result,
+     * but then never read or utilized.
+     *
+     * We don't want to copy this for _every_ handler, as it
+     * is expensive to do so.
+     */
+    const mainRequestRef = (() => {
```
mattcosta7, line 214: ugly for now, but cloning on _every handler_ is actually quite expensive (one of the most expensive operations, attributing to the growth of both event listeners (for aborts), memory and time

src/core/utils/internal/parseGraphQLRequest.ts
```diff
@@ -167,16 +167,19 @@ async function getGraphQLInput(request: Request): Promise<GraphQLInput | null> {
...
 
+const cache = new WeakMap<Request, ParsedGraphQLRequest>()
 /**
  * Determines if a given request can be considered a GraphQL request.
  * Does not parse the query and does not guarantee its validity.
  */
 export async function parseGraphQLRequest(
   request: Request,
 ): Promise<ParsedGraphQLRequest> {
+  if (cache.has(request)) return cache.get(request)
```
mattcosta7, line 178: graphql parsing is shared across all graphql handlers, and `onUnhandledRequest`, so doing this once per requst is a great optimization

src/core/utils/matching/matchRequestUrl.ts
```diff
@@ -52,10 +52,13 @@ export function coercePath(path: string): string {
   )
 }
 
+const cache = new Map<string, Match>()
 /**
  * Returns the result of matching given request URL against a mask.
  */
 export function matchRequestUrl(url: URL, path: Path, baseUrl?: string): Match {
+  const key = `${url}|${path}|${baseUrl}`
```
mattcosta7, line 60: request matching is expensive, but for any url/path/base triplet, the result is always identical, so we can eat this cost once and cache it

This cache works across handlers, since handlers can fallthrough to the same urls

src/core/utils/memoizedUrl.ts
```diff
@@ -0,0 +1,9 @@
+const cache = new Map<string, URL>()
+
+export function memoizedUrl(url: string, base?: string): URL {
```
mattcosta7, line 3: new URL is costly, but we have the same urls frequently since every handler calls this

We could probably generate this before executing handlers, and pass it around instead of caching it (my preference)

src/core/utils/request/getRequestCookies.ts
```diff
@@ -37,7 +38,9 @@ export function getRequestCookies(request: Request): Record<string, string> {
   }
 }
 
+const cache = new WeakMap<Request, Record<string, string>>()
 export function getAllRequestCookies(request: Request): Record<string, string> {
```
mattcosta7, line 42: reading cookies isn't that expensive but it's idempotent to a single request, so we should be good to cache this

src/core/utils/matching/matchRequestUrl.ts
mattcosta7: We also would want to use a better key than this, since we'll strip url params/hashes/etc before mathching.  Maybe we can extract some logic here and only do this 'url preparation' once per request vs per attempt to match

mattcosta7:
What areas would be next for improvements?


Currently potential handlers are evaluated linearly based on when they were defined. This puts a minimum cost bottleneck that scales with the number of handlers (and the relative location in those handlers of the one that finally matches). 

What does this mean?

the first active handler is always going to resolve almost immediately. The last handler is always going to resolve in at minimum `O(n)` because each handler has to evaluate _something_ to get there. 


Can we re-structure this to be more efficient?

Express and similar tools utilize trie structures to handle matching.

A `Trie` is also an `O(n)` evaluation model, but the evaluation is done against the _length of segments of the url being matched_ instead of the number of handlers being defined. 

This would be a signficiant improvement in some cases, and likely a minimal cost in others, especially since the evaluation cost should be lower than just path extraction.    

We'll have some slightly interesting considerations here to still maintain the `use` type overriding, but I think something like this would be the next step to get the time for _any_ handler to be hit to become stable (vs taking longer if they were defined first)

mattcosta7:
closing in favor of https://github.com/mswjs/msw/pull/1953
closing after https://github.com/mswjs/msw/pull/1914

PR #1906, opened 2023-12-04
fix(deferNetworkUntil): enable mocking when no pendingWorker

mattcosta7:
Makes a change to https://github.com/mswjs/msw/pull/1865 which gets most browser tests passing, with a few stragglers.  

I'm not sure if this is the correct fix here, but seems like the area where tests were failing

easier to view diff without whitespace: https://github.com/mswjs/msw/pull/1906/files?diff=split&w=1

### DIFF
--- a/src/browser/setupWorker/start/createStartHandler.ts
+++ b/src/browser/setupWorker/start/createStartHandler.ts
@@ -118,7 +118,7 @@ If this message still persists after updating, please report an issue: https://g
         return registration
       }
 
-    registerWorker().then((pendingWorkerRegistration) => {
+    registerWorker().then(async (pendingWorkerRegistration) => {
       // If the activation was cancelled (e.g. by calling "worker.stop()"),
       // short-circuit: nothing to return.
       if (pendingWorkerRegistration == null) {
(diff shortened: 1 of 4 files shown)


src/browser/setupWorker/start/createStartHandler.ts
```diff
@@ -118,35 +118,42 @@ If this message still persists after updating, please report an issue: https://g
...
+            }
+
+            pendingWorker.addEventListener('statechange', stateChangeListener)
+          })
         }
 
-        pendingWorker.addEventListener('statechange', stateChangeListener)
-      }
-    })
+        workerActivationPromise.resolve(pendingWorkerRegistration)
```
mattcosta7, line 152: I think we need to resolve this once pending is activated, or if there wasn't a pending/installting state?

src/browser/setupWorker/start/createStartHandler.ts
```diff
@@ -118,35 +118,42 @@ If this message still persists after updating, please report an issue: https://g
...
+      })
+      .catch((error) => {
+        workerActivationPromise.reject(error)
```
mattcosta7, line 155: when registering the worker throws, we also need to reject the worker activation, I think?

test/browser/msw-api/setup-worker/stop.test.ts
```diff
@@ -109,6 +109,6 @@ test('prints a warning on multiple "worker.stop()" calls', async ({
 
   // Prints a warning so the user knows something is not right.
   expect(consoleSpy.get('warning')).toEqual([
-    `[MSW] Found a redundant "worker.stop()" call. Note that stopping the worker while mocking already stopped has no effect. Consider removing this "worker.stop()" call.`,
+    '[MSW] Found a redundant "worker.stop()" call. Note that stopping an idle worker has no effect. Consider removing this "worker.stop()" call.',
```
mattcosta7: update message

mattcosta7:
3 tests still failing here, but that's pretty good compared to 0 passing in the base branch

mattcosta7:
maybe some race conditions here, there's a few tests that seem sporadically failing

kettanaito:
Thank you so much for this, @mattcosta7! 

I was thinking about this issue the other day and I wonder if we should remove `deferNetworkUntil` altogether. It was introduced in the past to account for use cases when people didn't defer their application rendering and ran MSW and their app in parallel, causing race conditions:

```js
worker.start()

render(<App />)
```

So we added the concept of "deferring" all the network from the moment you call `worker.start()`.

Right now, we officially recommend deferring the application rendering as the default setup to follow. I believe that eliminates the need for this option altogether.

kettanaito:
## Released: v2.1.0 🎉

---

PR #1907, opened 2023-12-04
ci: run ci build related tests in parallel

mattcosta7:
Updates the `ci build` job to run each of the test suites in parallel instead of serial

This should save some wall clock time when running the tests in ci, since browser tests can start immediately after build instead of only after unit, node, and native tests complete

We'd likely want to extend the required jobs configuration once this lands to require each of the test jobs to pass

In main, the current `build` job is almost 10 minutes walltime
in this branch it's closer to 7minutes, because each of the node/unit/native/browser suites run in parallel and avoid blocking eachother.  the browser suite takes most of that wall-time still, but at least test suites won't block eachother

the big value, is that we can re-run one of these separate from the others


.github/workflows/ci.yml
```diff
@@ -7,45 +7,166 @@ on:
...
+
+      - name: Cache build output
+        uses: actions/cache@v3
+        with:
+          path: ./lib
+          key: ${{ runner.os }}-build-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-build-
+
+  test_unit:
```
kettanaito, line 44: Nitpick: let's use more human-friendly display name for this granular test jobs. I believe GitHub Actions support the `name` property on individual jobs that will be used as the display name in the GitHub UI.

```yml
test_unit:
  name: Unit tests
```

Wdyt?

.github/workflows/ci.yml
```diff
@@ -7,45 +7,166 @@ on:
...
+          restore-keys: |
+            ${{ runner.os }}-build-
+
+      - name: Install dependencies
+        run: pnpm install
 
       - name: Node.js tests
         run: pnpm test:node
 
+  test_browser:
```
kettanaito: It's worth noting that ~85% of the test runtime is occupied by Playwright tests. We have around 200 tests there, and GitHub machines only has 2 cores, which means we are running all of them sequentially. This is the main bottleneck of CI performance for us right now.

mattcosta7: yep, totally. 
In some cases though, being able to re-run the playwright tests a few times would be nice even if that was 7minutes each, without needing to rebuild, and run the other suites first.  3 minutes each x 5 runs is 15 minutes saved wall clock

kettanaito:
Thank you so much for improving the CI, @mattcosta7 🙏 I've added the Node.js matrix to cover tests across supported versions of Node.js, polished things, and now it's ready for merge.

kettanaito:
## Released: v2.6.7 🎉

This has been released in v2.6.7!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.6.7)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.6.7)

---

PR #1910, opened 2023-12-06
test: reproduction of MaxListenersExceededWarning in `msw/node`

mattcosta7:
This test is expected to fail.  It shows a scenario where the current logic for `setMaxListeners` fails. 

relates to https://github.com/mswjs/msw/issues/1911

We actually have 2 separate places where this fails currently.

1 - some handlers will call multiple `Request.prototype.clone()`
2 - onUnhandledRequest behaviors call another handful of `Request.prototype.clone()`.
3 - Some of our clones are of clones, which will append listeners to the intermediate request (it seems)

The caching proposed in https://github.com/mswjs/msw/pull/1905 should actually help here as well, since we'll be more resource conscious we won't need to make nearly as many clones and therefore many fewer objects extended

### DIFF
--- a/src/node/setupServer.ts
+++ b/src/node/setupServer.ts
@@ -19,10 +19,12 @@ class SetupServerApi extends BaseSetupServerApi implements SetupServer {
    */
   protected override onRequest(request: Request): void {
     try {
-      setMaxListeners(
-        Math.max(defaultMaxListeners, this.currentHandlers.length),
-        request.signal,
-      )
+      if (typeof setMaxListeners === 'function') {
+        setMaxListeners(
+          Math.max(defaultMaxListeners, this.currentHandlers.length),
+          request.signal,
+        )
+      }
     } catch (error: unknown) {
       /**
        * @note Mock environments (JSDOM, ...) are not able to implement an internal
--- a/test/node/regressions/many-request-handlers.test.ts
+++ b/test/node/regressions/many-request-handlers.test.ts
@@ -1,45 +1,68 @@
 /**
  * @vitest-environment node
  */
+import { HttpServer } from '@open-draft/test-server/http'
 import { graphql, http, HttpResponse } from 'msw'
 import { setupServer } from 'msw/node'
+import * as nodeEvents from 'node:events'
 
-// Create a large number of request handlers.
-const restHandlers = new Array(100).fill(null).map((_, index) => {
-  return http.post(
-    `https://example.com/resource/${index}`,
-    async ({ request }) => {
-      const text = await request.text()
-      return HttpResponse.text(text + index.toString())
-    },
-  )
-})
-
-const graphqlHanlers = new Array(100).fill(null).map((_, index) => {
-  return graphql.query(`Get${index}`, () => {
-    return HttpResponse.json({ data: { index } })
+const httpServer = new HttpServer((app) => {
+  app.post('/graphql', (_, res) => {
+    return res.status(500).send('original-response')
   })
 })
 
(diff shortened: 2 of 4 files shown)


mattcosta7:
The reproduction here showcases 2 separate failure modes:

> Received: 

  1st spy call:

    Array [
      "(node:4602) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 201 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
    (Use `node --trace-warnings ...` to show where the warning was created)
    ",
      [Function anonymous],
    ]

  2nd spy call:

    Array [
      "(node:4602) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 303 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
    ",
      [Function anonymous],
    ]


How does this happen

Case 1 - we register 200 handlers, and have a graphql  post request  to the last one. This results in 201 listeners, not the 200 we expected. This is correctly registering these listeners through the extended `onRequest` method

Case 2 - we register 200 handlers, and match none of them, but they are a grapqhl request. We copy and parse the request 1x per each hander, then we do it again for the onUnhandedRequest handler for each graphql request, resulting in 303 listeners applied to the _cloned request_ that gets passed to that method, and not the _original request_

test/node/regressions/many-request-handlers.test.ts
```diff
@@ -1,58 +1,98 @@
...
     headers: {
       'Content-Type': 'application/json',
     },
     body: JSON.stringify({
-      query: `query Get42 { index }`,
+      query: `query Get99 { index }`,
     }),
   }).then((response) => response.json())
 
+  expect(spy).toHaveBeenNthCalledWith(2, 200, expect.any(AbortSignal))
```
mattcosta7: we fail here with just requesting 99 because we're always at a minimum `n+1` listeners.  1 per handler + 1 per request. 

then we also add one per parse of grapqhl, and we're way over the limit!

Note - the HTTP call above works because order matters. the errors here would flip if we defined the gql handlers first

src/node/setupServer.ts
```diff
@@ -19,10 +19,12 @@ class SetupServerApi extends BaseSetupServerApi implements SetupServer {
    */
   protected override onRequest(request: Request): void {
     try {
-      setMaxListeners(
-        Math.max(defaultMaxListeners, this.currentHandlers.length),
-        request.signal,
-      )
+      if (typeof setMaxListeners === 'function') {
```
mattcosta7, line 22: we need this check in jsdom with the mjs file. We don't with cjs, apparently.  Not sure why that is offhand

test/node/regressions/many-request-handlers.test.ts
```diff
@@ -1,58 +1,111 @@
...
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      query: `query NotDefinedAtAll { index }`,
+    }),
+  })
+
+  expect(unhandledResponse.status).toEqual(500)
+
+  expect(requestCloneSpy).toHaveBeenCalledTimes(301)
```
mattcosta7: These requestClone spies are just to help showcase how many copies we're making, it's a lot!

kettanaito:
@mattcosta7, these look very similar to the tests you've added in #1914. Should we close this pull request?

mattcosta7:
> @mattcosta7, these look very similar to the tests you've added in #1914. Should we close this pull request? 

Yep, this commit was included in that other branch, only left it open to site the failing case

kettanaito:
## Released: v2.0.12 🎉

---

PR #1914, opened 2023-12-07
fix: cache request cloning and request parsing

### LINKED ISSUE #1911: MaxListenersExceededWarning in `msw/node`
### Prerequisites

### Environment check

- [x] I'm using Node.js version 14 or higher

### Browsers

_No response_

### Reproduction repository

https://github.com/mswjs/msw/pull/1910

### Reproduction steps

see test setup - https://github.com/mswjs/msw/pull/1910

### Current behavior

When using some collection of handlers, especially when responding to the last sets of handlers, some warnings may be triggered in node

> (node:26291) MaxListenersExceededWarning: Possible EventTarget memory leak detected. <number> abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit

Some notes:

When starting to respond to a request, we set the number of max listeners to the number of handlers currently registered.  This assumes each handler makes a single request clone, which is true for http handlers.

GraphQL handlers however, make a request clone _per parse_ because we parse the graphql request body by first cloning it. 

When `onUnhandledRequest` is called we also re-parse and re-clone the request body to check for grapqhl requests again, leading to a minimum of 

2x handlers + 2 request clones. 

Additionally, some of these parses occur against a clone of a clone, so the abort listeners are defined on both the intiial request _and the cloned request_ leading to potentially an unhandled case of a request clone.

Possible fixes:

1 - Let's make fewer clones, especially of nested requests. This may mean a couple extra clones up front, that get managed from a single location, which at least allows for a less dynamic number of clones to occur (handle this outside of the handlers themselves, and maybe dissallow handlers from 'clone()' calls directly to avoid regression
2 - increase the number of `maxListeners` allowed.  Do we really need to be strict or can we just `Infinity` the count, and avoid this warning in most cases?

### Expected behavior

No warnings occur

mattcosta7:
fixes https://github.com/mswjs/msw/issues/1911
closes https://github.com/mswjs/msw/pull/1910

I originally based this on the tests in #1910, but set the base to main  so we can land it easily

This pull request mainly focuses on improving the performance of the application by introducing caching mechanisms and removing unnecessary functionality. The changes involve the `GraphQLHandler.ts`, `RequestHandler.ts`, `SetupServerApi.ts`, `setupServer.ts` and several test files. The most significant changes include the introduction of caching for parsed GraphQL requests and cloned requests, the removal of the `onRequest` method from the `SetupServerApi` class, and the removal of the `setMaxListeners` functionality from the `SetupServerApi` class and related test cases.

Performance improvements:

* <a href="diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L90-R93">`src/core/handlers/GraphQLHandler.ts`</a>: Introduced a cache for parsed GraphQL requests to avoid parsing the same request multiple times. This cache is implemented as a WeakMap where the keys are the request objects and the values are the parsed requests. <a href="diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L90-R93">[1]</a> <a href="diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L144-R157">[2]</a>
* <a href="diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6R189-R192">`src/core/handlers/RequestHandler.ts`</a>: Introduced a cache for cloned requests to avoid cloning the same request multiple times. This cache is also implemented as a WeakMap where the keys are the original request objects and the values are the cloned requests. <a href="diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6R189-R192">[1]</a> <a href="diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6L203-R216">[2]</a>

Codebase simplification:

* <a href="diffhunk://#diff-f219a1426c2ea0fd1152caf70c2cdbbe12b1a4d74634a41414b5a8ebd8d227f4L53-L54">`src/node/SetupServerApi.ts`</a>: Removed the `onRequest` method, which was not doing anything and was only meant to be overridden by subclasses. <a href="diffhunk://#diff-f219a1426c2ea0fd1152caf70c2cdbbe12b1a4d74634a41414b5a8ebd8d227f4L53-L54">[1]</a> <a href="diffhunk://#diff-f219a1426c2ea0fd1152caf70c2cdbbe12b1a4d74634a41414b5a8ebd8d227f4L115-L121">[2]</a>
* <a href="diffhunk://#diff-aca86f4ed7af5af9ae924d0bb8d24731ef91f9f44dd7bb964d93bf6c11147399L4-L46">`src/node/setupServer.ts`</a>: Removed the `setMaxListeners` functionality from the `SetupServerApi` class, which was used to suppress memory leak warnings in Node.js. This functionality was removed because it is no longer necessary. Since we're being more memory efficient with request clones, we don't need to worry about the number of aborts

Test changes:

* `test/node/regressions/many-request-handlers-jsdom.test.ts`, <a href="diffhunk://#diff-ef027506b7bb39164d5d8449267df4538ac5c93b25536d5aac7c1606c2ffc94cL7-L15">`test/node/regressions/many-request-handlers.test.ts`</a>: Updated tests to reflect the removal of the `setMaxListeners` functionality and the introduction of the request cloning cache. The tests now expect each request to be cloned only once, regardless of the number of request handlers, and no longer expect `setMaxListeners` to be called. <a href="diffhunk://#diff-ef027506b7bb39164d5d8449267df4538ac5c93b25536d5aac7c1606c2ffc94cL7-L15">[1]</a> <a href="diffhunk://#diff-ef027506b7bb39164d5d8449267df4538ac5c93b25536d5aac7c1606c2ffc94cL29">[2]</a> <a href="diffhunk://#diff-ef027506b7bb39164d5d8449267df4538ac5c93b25536d5aac7c1606c2ffc94cL73-R64">[3]</a> <a href="diffhunk://#diff-ef027506b7bb39164d5d8449267df4538ac5c93b25536d5aac7c1606c2ffc94cL88-R78">[4]</a> <a href="diffhunk://#diff-ef027506b7bb39164d5d8449267df4538ac5c93b25536d5aac7c1606c2ffc94cL119-R105">[5]</a> <a href="diffhunk://#diff-ef027506b7bb39164d5d8449267df4538ac5c93b25536d5aac7c1606c2ffc94cL141-R123">[6]</a> <a href="diffhunk://#diff-0dacea7051dc23b683e76c5e69fe0aa17e8cf212e83acaaaa362d13dd914daafL7-L15">[7]</a> <a href="diffhunk://#diff-0dacea7051dc23b683e76c5e69fe0aa17e8cf212e83acaaaa362d13dd914daafL29">[8]</a> <a href="diffhunk://#diff-0dacea7051dc23b683e76c5e69fe0aa17e8cf212e83acaaaa362d13dd914daafL73-R64">[9]</a> <a href="diffhunk://#diff-0dacea7051dc23b683e76c5e69fe0aa17e8cf212e83acaaaa362d13dd914daafL91-R78">[10]</a> <a href="diffhunk://#diff-0dacea7051dc23b683e76c5e69fe0aa17e8cf212e83acaaaa362d13dd914daafL125-R106">[11]</a> <a href="diffhunk://#diff-0dacea7051dc23b683e76c5e69fe0aa17e8cf212e83acaaaa362d13dd914daafL150-R124">[12]</a>


src/core/handlers/RequestHandler.ts
```diff
@@ -200,7 +204,16 @@ export abstract class RequestHandler<
...
+    if (!RequestHandler.#mainRequestRefCache.has(args.request)) {
+      RequestHandler.#mainRequestRefCache.set(
+        args.request,
+        args.request.clone(),
+      )
+    }
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const mainRequestRef = RequestHandler.#mainRequestRefCache.get(
+      args.request,
+    )!
```
mattcosta7: not the ideal way to do this, only showing how we can remove the max listener warnings by removing almost all cloning. this is also a big perf win

I think we would actually want to generate this clone before looping handlers. 

If we implemented another style of resolution, like a `trie` path match and a non-linear match for graphql query names, I think we could do this here without issue, since we would only need to run matching handlers, rather than tesitng every handler

src/core/handlers/GraphQLHandler.ts
```diff
@@ -141,11 +144,17 @@ export class GraphQLHandler extends RequestHandler<
     const match = matchRequestUrl(new URL(args.request.url), this.endpoint)
     if (!match.matches) return { match }
 
-    const parsedResult = await parseGraphQLRequest(args.request).catch(
-      (error) => {
-        console.error(error)
-        return undefined
-      },
+    if (!GraphQLHandler.#parseGraphQLRequestCache.has(args.request)) {
+      GraphQLHandler.#parseGraphQLRequestCache.set(
```
mattcosta7: not the best place to do this, since we don't pass to onUnhandledRequest well, but gets rid of the `O(n)` cloning at least

src/node/setupServer.ts
```diff
@@ -1,49 +1,9 @@
...
 import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/XMLHttpRequest'
 import { FetchInterceptor } from '@mswjs/interceptors/fetch'
-import { defaultMaxListeners, setMaxListeners } from 'node:events'
 import { RequestHandler } from '~/core/handlers/RequestHandler'
-import { SetupServerApi as BaseSetupServerApi } from './SetupServerApi'
+import { SetupServerApi } from './SetupServerApi'
 import { SetupServer } from './glossary'
-import { isNodeExceptionLike } from './utils/isNodeExceptionLike'
-
-class SetupServerApi extends BaseSetupServerApi implements SetupServer {
```
mattcosta7, line 10: we no longer need this hack, because we aren't exposing `O(handlers * c)` request clones

src/core/handlers/RequestHandler.ts
```diff
@@ -200,7 +204,16 @@ export abstract class RequestHandler<
 
     // Clone the request instance before it's passed to the handler phases
     // and the response resolver so we can always read it for logging.
-    const mainRequestRef = args.request.clone()
+    if (!RequestHandler.#mainRequestRefCache.has(args.request)) {
```
kettanaito: This is great but a bit verbose. What do you think if we abstract it to a protected method on the `RequestHandler` class itself? Something like `.cloneRequest()`? 

> A good name would be really nice here. `.cloneUnlessCached(request)`. Heh. We may consider implementing a tiny store around this like `RequestHandler.cache.get(request)` that would return a cached request if it exists or clone and cache if it doesn't. We can use Proxies on the `cache` Map to achieve it as well so we don't need any custom abstractions.

src/core/handlers/GraphQLHandler.ts
```diff
@@ -141,11 +144,17 @@ export class GraphQLHandler extends RequestHandler<
     const match = matchRequestUrl(new URL(args.request.url), this.endpoint)
     if (!match.matches) return { match }
 
-    const parsedResult = await parseGraphQLRequest(args.request).catch(
-      (error) => {
-        console.error(error)
-        return undefined
-      },
+    if (!GraphQLHandler.#parseGraphQLRequestCache.has(args.request)) {
```
kettanaito: I think caching the request parsing result is something all request handlers would want to benefit from. Does it make sense to move this logic to the `RequestHandler` base class so the child classes don't have to manage this manually? 

The cache can also be a static Map, achieving caching across individual request handler instances.

kettanaito (commented):
This is a phenomenal work. I love it. Left a few comments to discuss whenever you got a minute. Thank you!

src/core/handlers/GraphQLHandler.ts
mattcosta7: I think we should instead pass the request clone to the handlers from higher up the tree, but I wasn't sure if that's a reasonable option, since it changes the args to `handler.run()` slightly?  This was mostly to show how many fewer clones can be made with a few changes

kettanaito: What would be the main reason behind passing it from the higher loop? The simplified implementation (we can drop this static cache)?

I think it's a good approach to think of request handlers as standalone primitives. There may not be a loop iterating over them, and request handlers should still make sense. If the implementation cost is the only concern, I'm fine with it. We can move it to the `RequestHandler` base class as much as possible, making individual handlers unaware of the internal caching.

src/node/setupServer.ts
kettanaito: I am very much happy to get rid of this.

src/core/handlers/GraphQLHandler.ts
kettanaito: I don't think `onUnhandledRequest` reads the request body to require a clone, does it? Maybe we shouldn't concern ourselves about it. If the consumer wants to read the request/response body, they can clone it in the `onUnhandledRequest` callback. Unless there's a timing issue, of course, and that clone should've been made prior to that callback being called.

src/core/handlers/GraphQLHandler.ts
mattcosta7: Actually, a few things.  I was thinking of the other cache point at first

This graphql parsing has 2 callsites. 1 per handler and 1 again when no call is matched.

Ideally, we'd only parse once per request and not in both cases, where it might be more reasonable to keep the version of `parseGrapqhlRequestHandler` internal to msw as a version that caches results (we'd probably expose the version that doesn't cache - but could expose a caching version). 


I'd like to have parse be cached by default, but haven't quite figured out the right api to make that ergononmic as a general feature for `RequestHandler` because there's 2 separate types of request parsing that happens - one that uses handler specific info (endpoint) and one that doesn't (request body).

In the intermediate, handling something along these lines is probably good enough, and keeping them private should allow us a bit of internal freedom to optimize/improve over time before exposing something more generally?

src/core/handlers/GraphQLHandler.ts
mattcosta7: > I don't think onUnhandledRequest reads the request body to require a clone, does it

yep!  it does this to determine which handlers might have matched - https://github.com/mswjs/msw/blob/ea28bd9ad2efe0918660cd37054f21883caa3d30/src/core/utils/request/onUnhandledRequest.ts#L144-L146

it's not a big deal, as this is only 1 extra clone per request cycle, but we will have already done this parsing once, potentially, so it's also possibly a wasted clone, where a cached `parseGraphQLRequest` will avoid that

mattcosta7:
@kettanaito let me know what you think about landing this or some version as is, and then considering how to be a bit more generic on the cache side.   I'm still not sure how easily a shared cache will function, and this at least gives a significant benefit without exposing any new apis

src/core/handlers/GraphQLHandler.ts
mattcosta7: We could break parsing into `parse endpoint` and `parse body` type calls, combining their outputs into parse, but this expands the surface area in a way we might not want to?

only body parsing would likely be cached in that split

kettanaito:
@mattcosta7, let me clone this and look around, if I don't find anything to suggest we will merge this as it is. This is a perfect beginning to address the problem.

src/core/handlers/GraphQLHandler.ts
kettanaito: Agreed. Let's keep the changes surface to a minimum. This is already a grand improvement as it is.

kettanaito:
I'm getting a consistent test failure after I merged a few stylistic changes:

```
 FAIL  test/node/regressions/many-request-handlers-jsdom.test.ts > graphql handlers > does not print a memory leak warning for onUnhandledRequest
TypeError: The "emitter" argument must be an instance of EventEmitter or EventTarget. Received an instance of AbortSignal
 ❯ test/node/regressions/many-request-handlers-jsdom.test.ts:114:31
    112| 
    113|   it('does not print a memory leak warning for onUnhandledRequest', async () => {
    114|     const unhandledResponse = await fetch(httpServer.http.url('/graphql'), {
       |                               ^
    115|       method: 'POST',
    116|       headers: {
```

I must have screwed something up. Looking into it.

_Edit:_ Seems to be Node.js-related. Fails on `v18.14`, passes on `v18.18`.

kettanaito (approved):
Simply fantastic. Let's have this released.

kettanaito:
## Released: v2.0.12 🎉

---

PR #1922, opened 2023-12-11
fix: expose mockServiceWorker.js file in package exports

thepassle:
Exposes the `mockServiceWorker.js` file so that it may be `require`d or `import`ed

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
       "import": "./lib/native/index.mjs",
       "default": "./lib/native/index.js"
     },
+    "./mockServiceWorker.js": "./lib/mockServiceWorker.js",
     "./package.json": "./package.json"
   },
   "bin": {


package.json
```diff
@@ -34,6 +34,7 @@
       "import": "./lib/native/index.mjs",
       "default": "./lib/native/index.js"
     },
+    "./mockServiceWorker.js": "./lib/mockServiceWorker.js",
```
kettanaito, line 37: Can we technically make this a `fix:`? 

It cannot break a behavior that wasn't promised before. This change cannot break any existing behavior (i.e. exports) in any way either. 

Making it a patch release would be easier. I want to release 2.1 with a few other things together. Unless you are not time-constrained and can wait a few weeks, I suggest we mark this as a `fix:`.

thepassle: changed it to fix 👍

kettanaito (approved):
Fantastic.

kettanaito:
## Released: v2.0.11 🎉

This has been released in v2.0.11!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.11)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.11)

---

PR #1941, opened 2023-12-22
fix: drop "node" and "browser" null export conditions

### LINKED ISSUE #1809: Jest : Cannot find module 'msw/node' from ...
### Prerequisites

### Environment check

### Node.js version

18.18

### Reproduction repository

https://github.com/openfun/joanie/pull/445

### Reproduction steps

yarn test numbers

### Current behavior

 ● Test suite failed to run

    Cannot find module 'msw/node' from 'mocks/server.ts'

    Require stack:
      mocks/server.ts
      jest.setup.js

      at Resolver._throwModNotFoundError (node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.<anonymous> (mocks/server.ts:11:15)
      at Object.<anonymous> (jest.setup.js:8:17)

### Expected behavior

There were no issues before upgrading to 2.0.1. The tests should run correctly.

jd-carroll:
Fixes #1809

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,21 +14,18 @@
       "default": "./lib/core/index.js"
     },
     "./browser": {
-      "node": null,
       "types": "./lib/browser/index.d.ts",
       "require": "./lib/browser/index.js",
       "import": "./lib/browser/index.mjs",
       "default": "./lib/browser/index.js"
     },
     "./node": {
-      "browser": null,
       "types": "./lib/node/index.d.ts",
       "require": "./lib/node/index.js",
       "import": "./lib/node/index.mjs",
       "default": "./lib/node/index.mjs"
     },
     "./native": {
-      "browser": null,
       "types": "./lib/native/index.d.ts",
       "require": "./lib/native/index.js",
       "import": "./lib/native/index.mjs",


kettanaito:
Hi, @jd-carroll. Thanks for opening this pull request. 

It's not that simple though. This change on its own doesn't fix anything because it removes a valid export condition for those particular export paths. You aren't meant to use `msw/node` in the browser and `msw/browser` in Node.js. Those export conditions state precisely that. 

I highly recommend following that GitHub issue for more context. The root cause for this issue is the tooling (Jest, Next, etc) is unable to handle export conditions properly. Those conditions are handled by the bundler but not all tools are configured to support multi-environment input, even if it's not explicitly restricted (like JSDOM pretending it runs in a browser, forcing browser module resolution onto you, which is a terrible mistake on their end). 

We have discussed this internally with the team and are considering dropping the `null` export conditions in favor of placeholder modules that will simply throw in their stead. Similar to what [client-only](https://www.npmjs.com/package/client-only) does for the `use client` directive in React. I am a bit skeptical about that approach though. Fixing the resolution error isn't the point. The point is to let the developers use the tools correctly. If something isn't meant to be used in a particular environment, it must fail fast. Failing on build time is faster than failing on runtime (which happens if we drop the export conditions and use placeholder modules). 

Any input is welcome on this.

jd-carroll:
Closing for now as the PR is old. Take a look at my comments in associated issue. If the updated patch fixes things then I'll open a new PR

PR #1942, opened 2023-12-27
feat: add "requestId" as response resolver argument

### LINKED ISSUE #1933: Pass "requestId" as a response resolver argument
### Scope

Improves an existing behavior

### Compatibility

### Feature description

In MSW v1 I was able to access `req.id` in a handler function which corresponded with the requestId in events.

For example
```ts
worker.events.on('response:mocked', ({requestId}) => {
// and
rest.post('my-url', (req) => { req.id 
```

were the same

In v2, the request object doesn't have an id attribute.  From what I can tell the requestId is unique to events

This seems like a nice to have addition to the handler args that wouldn't be hard to add

Here's what I had in mind as a diff - https://github.com/mswjs/msw/compare/main...skvale:msw:main

skvale:
Fixes: https://github.com/mswjs/msw/issues/1933

In MSW v1 I was able to access `req.id` in a handler function which corresponded with the requestId in events.

For example
```ts
worker.events.on('response:mocked', ({requestId}) => {
// and
rest.post('my-url', (req) => { req.id 
```

were the same value

In v2, the request object doesn't have an id attribute.  From what I can tell the requestId is unique to events

This seems like a nice to have addition to the handler args that wouldn't be hard to add


kettanaito:
Thanks for submitting this so quickly, @skvale! I will come back to review this in January.

src/core/handlers/RequestHandler.ts
```diff
@@ -216,6 +217,7 @@ export abstract class RequestHandler<
   public async run(args: {
     request: StrictRequest<any>
     resolutionContext?: ResponseResolutionContext
+    requestId?: string
```
mattcosta7: should this be required instead of optional?  are we leaving it optional to avoid making it required for non-standard usage?  just curious if we can accidentally miss this if it's optional?

skvale: I'm wondering if this is considered exposed of it's OK to change this function signature to get the required arg before the optional one

So from this
```ts
/**
 * Returns a mocked response for a given request using following request handlers.
 */
export const getResponse = async <Handler extends Array<RequestHandler>>(
  request: Request,
  handlers: Handler,
  resolutionContext?: ResponseResolutionContext,
  requestId?: string,
): Promise<ResponseLookupResult | null> => {
```
to this
```ts
/**
 * Returns a mocked response for a given request using following request handlers.
 */
export const getResponse = async <Handler extends Array<RequestHandler>>(
  request: Request,
  handlers: Handler,
  requestId: string,
  resolutionContext?: ResponseResolutionContext,
): Promise<ResponseLookupResult | null> => {
```

src/core/utils/getResponse.ts
```diff
@@ -16,16 +16,22 @@ export interface ResponseResolutionContext {
...
-  resolutionContext?: ResponseResolutionContext,
-): Promise<ResponseLookupResult | null> => {
+export const getResponse = async <Handler extends Array<RequestHandler>>({
+  request,
+  requestId,
+  handlers,
+  resolutionContext,
+}: {
+  request: Request
+  requestId: string
```
kettanaito, line 26: The only thing I'm not sure about here is whether requiring `requestId` for the `.run()` method is right. 

While this is not widely used, you can call a request handler directly using its `.run()` method. Just accepting `request` makes sense. Demanding `requestId` means the consumer has to contrive that ID. Hmm.

I think we can make this required and suggest using `crypto.randomUUID()` as a placeholder value if someone has to call a handler directly. If you want to—you must provide a request ID.

src/core/handlers/RequestHandler.ts
kettanaito: @skvale, I believe your last question is resolved since the `run()` method now accepts a single object argument where the argument position is irrelevant. Thanks for updating the PR!

kettanaito (approved):
This looks fantastic! Great work, @skvale 👏

kettanaito:
Welcome to contributors, @skvale! 🎉

kettanaito:
## Released: v2.1.0 🎉

---

PR #1780, opened 2023-10-20
feat: up tsconfig target for more modern syntax

thepassle:

PR #1953, opened 2024-01-06
fix: additional caching for improved handler performance

mattcosta7:
The intent of this PR is to help discover if there's a good general purpose cache api we can expose from the handlers, and if not we can see where caching improves performance further and how much.

These are mostly changes first looked at in https://github.com/mswjs/msw/pull/1905

This is a majority of the low-hanging request cycle optimizations I think we can do with good benefit and little overhead.  
Further performance could be gained with a slightly modified storage structure, potentially (instead of solely linear request parsing, we could order them at registration into a more treelike structure, however this seems fairly complex compared to the low-hanging performance we can unlock from the small changes in this pr 

This pull request introduces changes primarily aimed at improving the efficiency of URL and cookie parsing in the codebase. The changes involve the creation of a caching mechanism to store parsed URLs and cookies from requests, reducing the need for repeated parsing of the same data. The changes affect several files, with the most significant changes in `src/core/handlers/GraphQLHandler.ts`, `src/core/handlers/HttpHandler.ts`, and `src/core/handlers/RequestHandler.ts`.

URL and Cookie Parsing Improvements:

* [`src/core/handlers/GraphQLHandler.ts`](diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L13-R13): The `matchRequestUrl` function has been replaced with a new method `matchRequestURLOrGetMatchFromCache` that retrieves the match from the cache if it exists, otherwise it matches the request URL and stores the result in the cache. The `getAllRequestCookies` function has been replaced with a new method `parseAllRequestCookiesOrGetFromCache` that retrieves the cookies from the cache if they exist, otherwise it parses the request cookies and stores the result in the cache. [[1]](diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L13-R13) [[2]](diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L22-R22) [[3]](diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6R36) [[4]](diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L167-R188) [[5]](diffhunk://#diff-4f6e30ee2bc5cda860683f2a951c396d0748ca4103bc5011406af10452da15c6L227-R238)

* [`src/core/handlers/HttpHandler.ts`](diffhunk://#diff-93473d9ff9b2ac402a677635ffd0b18446d403d3f0fdbc3208878821a6f7621eL8-R10): Similar changes to `GraphQLHandler.ts` have been made, with the `matchRequestUrl` function replaced with `matchRequestURLOrGetMatchFromCache` and the `getAllRequestCookies` function replaced with `parseAllRequestCookiesOrGetFromCache`. [[1]](diffhunk://#diff-93473d9ff9b2ac402a677635ffd0b18446d403d3f0fdbc3208878821a6f7621eL8-R10) [[2]](diffhunk://#diff-93473d9ff9b2ac402a677635ffd0b18446d403d3f0fdbc3208878821a6f7621eL113-R113)

* [`src/core/handlers/RequestHandler.ts`](diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6R7-R8): Added two new methods `parseAllRequestCookiesOrGetFromCache` and `matchRequestURLOrGetMatchFromCache` to handle the caching of parsed cookies and matched URLs. [[1]](diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6R7-R8) [[2]](diffhunk://#diff-b2f9353d793b757713c58e3398de4e1f6975c46c5279c9595e1f80a8a6eefca6R189-R218)

* [`src/core/utils/internal/parseGraphQLRequest.ts`](diffhunk://#diff-c604d2555e89b3bedcb745c2e268a3f04e2b45f300061e31a7e1782a2000a6baR12): Replaced the direct creation of a new URL object from the request URL with a call to the `urlFromRequestOrCache` function. [[1]](diffhunk://#diff-c604d2555e89b3bedcb745c2e268a3f04e2b45f300061e31a7e1782a2000a6baR12) [[2]](diffhunk://#diff-c604d2555e89b3bedcb745c2e268a3f04e2b45f300061e31a7e1782a2000a6baL95-R96)

* [`src/core/utils/logging/serializeRequest.ts`](diffhunk://#diff-3d2523c54fda96d4289815b08d6ebb64046974137718b3ff6e6f2b929775b0e6R1-R2): Replaced the direct creation of a new URL object from the request URL with a call to the `urlFromRequestOrCache` function. [[1]](diffhunk://#diff-3d2523c54fda96d4289815b08d6ebb64046974137718b3ff6e6f2b929775b0e6R1-R2) [[2]](diffhunk://#diff-3d2523c54fda96d4289815b08d6ebb64046974137718b3ff6e6f2b929775b0e6L18-R20)

* [`src/core/utils/request/getPublicUrlFromRequest.ts`](diffhunk://#diff-d6fe2ee7b22908e609c9fabd5c44b0a46f289cb361e2c6613743aa41146879e6R1-R2): Replaced the direct creation of a new URL object from the request URL with a call to the `urlFromRequestOrCache` function. [[1]](diffhunk://#diff-d6fe2ee7b22908e609c9fabd5c44b0a46f289cb361e2c6613743aa41146879e6R1-R2) [[2]](diffhunk://#diff-d6fe2ee7b22908e609c9fabd5c44b0a46f289cb361e2c6613743aa41146879e6L10-R12)

* [`src/core/utils/request/getRequestCookies.ts`](diffhunk://#diff-3abcaf3e23405a6ac736486689634065f5c5b6ef27f366f1ea6c1abe997bf1f1R3): Replaced the direct creation of a new URL object from the request URL with a call to the `urlFromRequestOrCache` function. [[1]](diffhunk://#diff-3abcaf3e23405a6ac736486689634065f5c5b6ef27f366f1ea6c1abe997bf1f1R3) [[2]](diffhunk://#diff-3abcaf3e23405a6ac736486689634065f5c5b6ef27f366f1ea6c1abe997bf1f1L22-R23)

* [`src/core/utils/request/urlFromRequestOrCache.ts`](diffhunk://#diff-64f9b90225316396831f4860f42c924a2e713b7b4becde1315deb03d5c759a37R1-R17): Added a new utility function `urlFromRequestOrCache` that retrieves a URL object from a cache if it exists, otherwise it creates a new URL object from the request URL and stores it in the cache.

* [`src/core/utils/request/urlFromRequestOrCache.test.ts`](diffhunk://#diff-10691b760a41e2b3e265226621b2f924c96bad8bc56b527f3663d152075acd20R1-R28): Added new tests to verify the functionality of the `urlFromRequestOrCache` function.

### DIFF
--- a/src/core/handlers/GraphQLHandler.ts
+++ b/src/core/handlers/GraphQLHandler.ts
@@ -144,7 +144,7 @@ export class GraphQLHandler extends RequestHandler<
    * GraphQL handlers. This is done to avoid multiple parsing of the
    * request body, which each requires a clone of the request.
    */
-  private static async parseGraphQLRequestOrGetFromCache(
+  async parseGraphQLRequestOrGetFromCache(
     request: Request,
   ): Promise<ParsedGraphQLRequest<GraphQLVariables>> {
     if (!GraphQLHandler.parsedRequestCache.has(request)) {
@@ -165,19 +165,17 @@ export class GraphQLHandler extends RequestHandler<
      * If the request doesn't match a specified endpoint, there's no
      * need to parse it since there's no case where we would handle this
      */
-    const match = RequestHandler.matchRequestURLOrGetMatchFromCache(
+    const match = this.matchRequestURLOrGetMatchFromCache(
       urlFromRequestOrCache(args.request),
       this.endpoint,
     )
-    const cookies = RequestHandler.parseAllRequestCookiesOrGetFromCache(
-      args.request,
-    )
+    const cookies = this.parseAllRequestCookiesOrGetFromCache(args.request)
 
     if (!match.matches) {
       return { match, cookies }
     }
 
-    const parsedResult = await GraphQLHandler.parseGraphQLRequestOrGetFromCache(
+    const parsedResult = await this.parseGraphQLRequestOrGetFromCache(
       args.request,
     )
 
--- a/src/core/handlers/GraphQLHandler.ts
+++ b/src/core/handlers/GraphQLHandler.ts
@@ -144,7 +144,7 @@ export class GraphQLHandler extends RequestHandler<
    * GraphQL handlers. This is done to avoid multiple parsing of the
    * request body, which each requires a clone of the request.
    */
-  async parseGraphQLRequestOrGetFromCache(
+  private static async parseGraphQLRequestOrGetFromCache(
     request: Request,
   ): Promise<ParsedGraphQLRequest<GraphQLVariables>> {
     if (!GraphQLHandler.parsedRequestCache.has(request)) {
@@ -165,17 +165,19 @@ export class GraphQLHandler extends RequestHandler<
      * If the request doesn't match a specified endpoint, there's no
      * need to parse it since there's no case where we would handle this
      */
-    const match = this.matchRequestURLOrGetMatchFromCache(
+    const match = RequestHandler.matchRequestURLOrGetMatchFromCache(
       urlFromRequestOrCache(args.request),
       this.endpoint,
     )
-    const cookies = this.parseAllRequestCookiesOrGetFromCache(args.request)
+    const cookies = RequestHandler.parseAllRequestCookiesOrGetFromCache(
+      args.request,
+    )
 
     if (!match.matches) {
       return { match, cookies }
     }
 
-    const parsedResult = await this.parseGraphQLRequestOrGetFromCache(
+    const parsedResult = await GraphQLHandler.parseGraphQLRequestOrGetFromCache(
       args.request,
     )
 
@@ -185,11 +187,11 @@ export class GraphQLHandler extends RequestHandler<
 
     return {
       match,
+      cookies,
       query: parsedResult.query,
       operationType: parsedResult.operationType,
       operationName: parsedResult.operationName,
       variables: parsedResult.variables,
-      cookies,
     }
   }
 
--- a/src/core/handlers/GraphQLHandler.ts
+++ b/src/core/handlers/GraphQLHandler.ts
@@ -10,7 +10,7 @@ import { getTimestamp } from '../utils/logging/getTimestamp'
 import { getStatusCodeColor } from '../utils/logging/getStatusCodeColor'
 import { serializeRequest } from '../utils/logging/serializeRequest'
 import { serializeResponse } from '../utils/logging/serializeResponse'
-import { Match, matchRequestUrl, Path } from '../utils/matching/matchRequestUrl'
+import { Match, Path } from '../utils/matching/matchRequestUrl'
 import {
   ParsedGraphQLRequest,
   GraphQLMultipartRequestBody,
@@ -19,7 +19,7 @@ import {
 } from '../utils/internal/parseGraphQLRequest'
 import { getPublicUrlFromRequest } from '../utils/request/getPublicUrlFromRequest'
 import { devUtils } from '../utils/internal/devUtils'
-import { getAllRequestCookies } from '../utils/request/getRequestCookies'
+import { urlFromRequestOrCache } from '../utils/request/urlFromRequestOrCache'
 
 export type ExpectedOperationTypeNode = OperationTypeNode | 'all'
 export type GraphQLHandlerNameSelector = DocumentNode | RegExp | string
@@ -33,6 +33,7 @@ export interface GraphQLHandlerInfo extends RequestHandlerDefaultInfo {
 
 export type GraphQLRequestParsedResult = {
   match: Match
+  cookies: Record<string, string>
 } & (
   | ParsedGraphQLRequest<GraphQLVariables>
   /**
@@ -164,9 +165,14 @@ export class GraphQLHandler extends RequestHandler<
      * If the request doesn't match a specified endpoint, there's no
      * need to parse it since there's no case where we would handle this
      */
-    const match = matchRequestUrl(new URL(args.request.url), this.endpoint)
+    const match = this.matchRequestURLOrGetMatchFromCache(
+      urlFromRequestOrCache(args.request),
+      this.endpoint,
+    )
+    const cookies = this.parseAllRequestCookiesOrGetFromCache(args.request)
+
     if (!match.matches) {
-      return { match }
+      return { match, cookies }
     }
 
     const parsedResult = await this.parseGraphQLRequestOrGetFromCache(
@@ -174,7 +180,7 @@ export class GraphQLHandler extends RequestHandler<
     )
 
     if (typeof parsedResult === 'undefined') {
-      return { match }
+      return { match, cookies }
     }
 
     return {
@@ -183,6 +189,7 @@ export class GraphQLHandler extends RequestHandler<
       operationType: parsedResult.operationType,
       operationName: parsedResult.operationName,
       variables: parsedResult.variables,
+      cookies,
     }
   }
 
@@ -224,13 +231,11 @@ Consider naming this operation or using "graphql.operation()" request handler to
     request: Request
     parsedResult: GraphQLRequestParsedResult
   }) {
-    const cookies = getAllRequestCookies(args.request)
-
     return {
       query: args.parsedResult.query || '',
       operationName: args.parsedResult.operationName || '',
       variables: args.parsedResult.variables || {},
-      cookies,
+      cookies: args.parsedResult.cookies,
     }
   }
 
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -190,7 +190,7 @@ export abstract class RequestHandler<
     Request,
     Record<string, string>
   >()
-  protected static parseAllRequestCookiesOrGetFromCache(
+  protected parseAllRequestCookiesOrGetFromCache(
     request: Request,
   ): Record<string, string> {
     let cookies = RequestHandler.cookiesForRequestCache.get(request)
@@ -202,7 +202,7 @@ export abstract class RequestHandler<
   }
 
   private static matchRequestUrlCache = new Map<string, Match>()
-  protected static matchRequestURLOrGetMatchFromCache(
+  protected matchRequestURLOrGetMatchFromCache(
     url: URL,
     path: Path,
     baseUrl?: string | undefined,
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -4,6 +4,8 @@ import { isIterable } from '../utils/internal/isIterable'
 import type { ResponseResolutionContext } from '../utils/getResponse'
 import type { MaybePromise } from '../typeUtils'
 import { StrictRequest, StrictResponse } from '..//HttpResponse'
+import { getAllRequestCookies } from '../utils/request/getRequestCookies'
+import { Match, Path, matchRequestUrl } from '../utils/matching/matchRequestUrl'
 
 export type DefaultRequestMultipartBody = Record<
   string,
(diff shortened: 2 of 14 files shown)


src/core/handlers/RequestHandler.ts
```diff
@@ -184,6 +186,36 @@ export abstract class RequestHandler<
     })
   }
 
+  private static cookiesForRequestCache = new WeakMap<
+    Request,
+    Record<string, string>
+  >()
+  protected static parseAllRequestCookiesOrGetFromCache(
```
mattcosta7: cookies can be parsed once per request, not required once per handler

src/core/handlers/RequestHandler.ts
```diff
@@ -184,6 +186,36 @@ export abstract class RequestHandler<
...
+    let cookies = RequestHandler.cookiesForRequestCache.get(request)
+    if (!cookies) {
+      cookies = getAllRequestCookies(request)
+      RequestHandler.cookiesForRequestCache.set(request, cookies)
+    }
+    return cookies
+  }
+
+  private static matchRequestUrlCache = new Map<string, Match>()
+  protected static matchRequestURLOrGetMatchFromCache(
```
mattcosta7: URL matching can be computed _once_ per string comparison of URL/match/baseUrl - and reused for subsequent requests across all handlers.  There's some memory cost to this map, but even at very large sizes it should be fine

src/core/utils/request/urlFromRequestOrCache.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Returns the URL object for the given request. The URL object is cached so that
+ * subsequent calls to this function with the same request or requests to the same
+ * url will return the same URL object, to avoid multiple URL object allocations and
+ * parsing the same url multiple times.
+ */
+export function urlFromRequestOrCache(request: Request): URL {
+  const requestUrl = request.url
+  let url = urlCache.get(requestUrl)
+  if (!url) {
+    url = new URL(requestUrl)
```
mattcosta7, line 13: URL allocations aren't cheap as they do expensive string parsing to create a URL object.  we can cache this per string url, and avoid those object allocations and parses, which may occur multiple times for a single request lookup

src/core/handlers/GraphQLHandler.ts
```diff
@@ -33,6 +33,7 @@ export interface GraphQLHandlerInfo extends RequestHandlerDefaultInfo {
 
 export type GraphQLRequestParsedResult = {
   match: Match
+  cookies: Record<string, string>
```
mattcosta7, line 36: unrelated to this directly, however I noticed that we parsed cookies in the http handler in the parse function, but in graphql we only parsed them in the extendedArgs.  we should probably be consistent?

tsconfig.base.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "strict": true,
-    "target": "es6",
+    "target": "es2020",
```
mattcosta7, line 4: this is a perf fix, because async/await no longer become generators
see https://github.com/mswjs/msw/pull/1780

src/core/handlers/GraphQLHandler.ts
kettanaito: Agree. An oversight on my end, most likely. The `cookies` have been added to GraphQL handler recently.

src/core/handlers/GraphQLHandler.ts
```diff
@@ -164,21 +165,27 @@ export class GraphQLHandler extends RequestHandler<
      * If the request doesn't match a specified endpoint, there's no
      * need to parse it since there's no case where we would handle this
      */
-    const match = matchRequestUrl(new URL(args.request.url), this.endpoint)
+    const match = this.matchRequestURLOrGetMatchFromCache(
```
kettanaito, line 168:
```suggestion
    const match = this.matchRequestOrGetFromCache(
```

src/core/handlers/GraphQLHandler.ts
```diff
@@ -164,21 +165,27 @@ export class GraphQLHandler extends RequestHandler<
...
+      urlFromRequestOrCache(args.request),
+      this.endpoint,
+    )
+    const cookies = this.parseAllRequestCookiesOrGetFromCache(args.request)
```
kettanaito, line 172:
```suggestion
    const cookies = this.parseRequestCookiesOrGetFromCache(args.request)
```

src/core/handlers/RequestHandler.ts
```diff
@@ -184,6 +186,36 @@ export abstract class RequestHandler<
     })
   }
 
+  private static cookiesForRequestCache = new WeakMap<
```
kettanaito, line 189:
```suggestion
  private static requestCookiesCache = new WeakMap<
```

src/core/handlers/RequestHandler.ts
```diff
@@ -184,6 +186,36 @@ export abstract class RequestHandler<
...
+    Request,
+    Record<string, string>
+  >()
+  protected parseAllRequestCookiesOrGetFromCache(
```
kettanaito, line 193:
```suggestion
  protected parseRequestCookiesOrGetFromCache(
```

src/core/handlers/RequestHandler.ts
```diff
@@ -184,6 +186,36 @@ export abstract class RequestHandler<
...
+  ): Record<string, string> {
+    let cookies = RequestHandler.cookiesForRequestCache.get(request)
+    if (!cookies) {
+      cookies = getAllRequestCookies(request)
+      RequestHandler.cookiesForRequestCache.set(request, cookies)
+    }
+    return cookies
+  }
+
+  private static matchRequestUrlCache = new Map<string, Match>()
```
kettanaito, line 204: Why not `WeakMap` here also? Are we planning on freeing these resources manually?

src/core/handlers/RequestHandler.ts
```diff
@@ -184,6 +186,36 @@ export abstract class RequestHandler<
...
+  protected matchRequestURLOrGetMatchFromCache(
```
kettanaito, line 205:
```suggestion
  protected matchRequestOrGetFromCache(
```

src/core/utils/internal/parseGraphQLRequest.ts
```diff
@@ -92,7 +93,7 @@ function extractMultipartVariables<VariablesType extends GraphQLVariables>(
 async function getGraphQLInput(request: Request): Promise<GraphQLInput | null> {
   switch (request.method) {
     case 'GET': {
-      const url = new URL(request.url)
+      const url = urlFromRequestOrCache(request)
```
kettanaito, line 96: Are URL constructors really that costly? I suspect this optimization isn't worth the cost of readability and debugging.

src/core/utils/internal/requestHandlerUtils.ts
```diff
@@ -4,7 +4,10 @@ export function use(
   currentHandlers: Array<RequestHandler>,
   ...handlers: Array<RequestHandler>
 ): void {
-  currentHandlers.unshift(...handlers)
+  // we don't spread the handlers to avoid maximum stack errors on very large sets of handlers
```
kettanaito, line 7: Can we please add a test for this use case?

src/core/utils/internal/requestHandlerUtils.ts
```diff
@@ -4,7 +4,10 @@ export function use(
...
+  for (let i = handlers.length - 1; i >= 0; i--) {
```
kettanaito, line 8: I wonder if `.reduceRight` wouldn't be more readable here?

```js
handlers.reduceRight((_, handler) => {
  currentHandlers.unshift(handler)
})
```

src/core/utils/request/urlFromRequestOrCache.ts
```diff
@@ -0,0 +1,17 @@
+const urlCache = new Map<string, URL>()
```
kettanaito, line 1: Why not `WeakMap<Request, URL>`?

src/core/utils/request/urlFromRequestOrCache.ts
kettanaito: Okay, I see. Thanks for the brief explanation!

tsconfig.base.json
kettanaito: Can you please move this to a separate pull request?

src/core/utils/request/urlFromRequestOrCache.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Returns the URL object for the given request. The URL object is cached so that
+ * subsequent calls to this function with the same request or requests to the same
+ * url will return the same URL object, to avoid multiple URL object allocations and
+ * parsing the same url multiple times.
+ */
+export function urlFromRequestOrCache(request: Request): URL {
+  const requestUrl = request.url
```
kettanaito, line 10: We seem to be adopting this "look-in-cache-or-create" pattern quite extensively over these two pull requests. I think it may be worth exploring a tiny abstraction to make these actions more readable. 

```ts
function cached<V, K extends object>(producer: () => V, key: K, cache: WeakMap<K, V>): V {
  if (!cache.has(key)) {
    cache.set(key, producer())
  }
  
  return cache.get(key)!
}

// Usage.
const urlCache = new WeakMap<Request, URL>
export function cachedUrlFromRequest(request: Request) {
  return cached(() => new URL(request.url), request, urlCache)
}
```

src/core/handlers/RequestHandler.ts
mattcosta7: WeakMap can't accept string keys, but mostly because this cache is actually more valuable if it persists between requests.  Since multiple requests can have the same URL string, this maintains the cache across requests, avoiding the new URL overhead for multiple requests.

It's a slight extra optimization, but these shouldn't be so huge where they become an issue since it should be relatively small

We could later consider something more like an lru here if we need to evict things

src/core/utils/request/urlFromRequestOrCache.ts
mattcosta7: Mostly to keep URL objects between requests, likely a slight over optimization, which keeps subsequent requests slightly more efficient at the expense of memory. 

Both should be generally helpful

src/core/handlers/RequestHandler.ts
kettanaito: Got it. The choice of `Map` is clear.

kettanaito: My main concern is that these changes wouldn't hurt the readability of the code. The benefit they provide is tangible but rather minimal. I'm fine with having them as long as the code doesn't become a jungle. I suggested a `cached` helper to make this common cached logic reused, I'd like to know what you think about it.

src/core/utils/request/urlFromRequestOrCache.ts
mattcosta7: They're also not super expensive, so definitely a micro-optimization we could totally skip 

It's definitely the least important change.  on the order of 10ms for 10,000 requests of overhead.  doubtful that it'll matter to most!

Screen Shot 2024-01-07 at 10 57 55 PM

Screen Shot 2024-01-07 at 10 56 13 PM


I was mostly just trying to get a "what's left that is low hanging" branch together.

src/core/handlers/GraphQLHandler.ts
mattcosta7: @kettanaito I think it should be reasonable to move this outside of this pr, i'll open one with just that change separate from this pr!

tsconfig.base.json
mattcosta7: > Can you please move this to a separate pull request?

yep - it's part of 1780, so I think we can just let that land (it sets esnext, which i think should be ok?)

src/core/handlers/GraphQLHandler.ts
mattcosta7: https://github.com/mswjs/msw/pull/1957

opened this one to land that change separate from any other changes here

tsconfig.base.json
kettanaito: I believe so!

src/core/handlers/GraphQLHandler.ts
kettanaito: Amazing! It's been merged.

src/core/utils/request/urlFromRequestOrCache.ts
kettanaito: I'd say 10ms save on 10,000 requests is not worth the price we are paying here. Would you agree that we can skip this pull request as of now?

mattcosta7: this was _just the url part_ but I agree we don't need to ship this right now as they're generally small improvements.  Very much just a "what other things can be optimized easily" experiment

kettanaito:
@mattcosta7, I still appreciate the effort on this. Thank you so much for making MSW better!

PR #1957, opened 2024-01-08
fix(GraphQLHandler): move cookie parsing to parse phase

mattcosta7:
Moves `getAllRequestCookies` parsing in `GraphQLHandler` from `extendResolverArgs` to `parse` step.

This normalizes its usage with that of `HttpHandler`.

Also updates tests to account for this


src/core/handlers/GraphQLHandler.ts
```diff
@@ -165,20 +166,23 @@ export class GraphQLHandler extends RequestHandler<
      * need to parse it since there's no case where we would handle this
      */
     const match = matchRequestUrl(new URL(args.request.url), this.endpoint)
+    const cookies = getAllRequestCookies(args.request)
```
mattcosta7, line 170: We probably don't _need_ to parse cookies here, because we don't have a match, so this shouldn't be useful anywhere, but we do it in the `HttpHandler` so we should probably match that behavior here.
By here, I mean in cases where we aren't sure a mach exists

we could return `cookies: {}` in each of the cases of `!match.matches` or `parsedResult === undefined`, and only parse matches - in which case we should follow that pattern in both handlers.  This shouldn't really matter though in practice since it isn't super heavy, especially if we start caching the match and cookie parsing

we could mirror that check before parsing in `HttpHandler` too I think, but maybe we don't need to bother with that

kettanaito: It's okay to keep it here. When we follow through with custom predicate functions (#1804), it would be useful to allow developers to implement GraphQL request matching based on cookies as well. For that, we'd have to always parse cookies and then pass them to the predicate function that determines the match. 

Until we get there, let's leave it as it is right now after your change.

kettanaito (approved):
Looks great! Thank you 🚀

kettanaito:
## Released: v2.0.14 🎉

This has been released in v2.0.14!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.14)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.14)

---

PR #1961, opened 2024-01-11
fix: improve custom request handler and resolver annotations

### LINKED ISSUE #1955: Custom response resolver annotations
### Scope

Adds a new behavior

### Compatibility

### Feature description

There's currently no way to annotate a custom response resolver. The `ResponseResolver` type we expose depends on `ResolverExtraInfo`, which is not exposed for HTTP and GraphQL request handlers.

Ideally, export _new_ types:

- `HttpResponseResolver<Params, RequestBody, ResponseBody>`
- `GraphQLResponseResolver<Variables>`

The user mustn't be concerned with our internals like `ResolverExtraInfo` and stuff. That doesn't matter to them. It would be nice for the resolver types to keep the same generic signature as the respective request handlers they are attached to. 

### Alternative

Or, simply expose `HttpResolverExtraInfo` and `GraphQLResolverExtraInfo` to keep the experience of annotating custom resolvers consistent via `ResponseResolver` type. This is certainly a more flexible approach.

kettanaito:
- Fixes #1955
- Solves #1802


src/core/http.ts
```diff
@@ -10,22 +10,35 @@ import {
 } from './handlers/HttpHandler'
 import type { Path, PathParams } from './utils/matching/matchRequestUrl'
 
+export type HttpRequestHandler = <
```
kettanaito, line 13: Confusing terminology here. 

- `HttpRequestHandler` is when people want to create custom `http.get()` function.
- `HttpHandler` is _not_ `http.get()` interface at all. It's its own thing with `.parse()`, `.run()` and friends. 

I think in the context of the userland, it's fine to call this type a request handler type. It's precisely how we refer to the `http.get()` functions in the [docs](https://mswjs.io/docs/concepts/request-handler).

kettanaito:
Hi, @mattcosta7 👋 Please, would you have a minute to give this change a brief look? The aim here is to improve the way people can annotate custom request handlers and response resolvers in TypeScript by shipping a few utility types.

mattcosta7:
> Hi, @mattcosta7 👋 Please, would you have a minute to give this change a brief look? The aim here is to improve the way people can annotate custom request handlers and response resolvers in TypeScript by shipping a few utility types.

Will do, might not have time to look much until this weekend, but i'll try to get eyes on it sooner!

src/core/handlers/RequestHandler.ts
```diff
@@ -36,23 +36,28 @@ export interface RequestHandlerInternalInfo {
...
-  BodyType extends DefaultBodyType,
-> = MaybePromise<ResponseResolverReturnType<BodyType>>
+  ResponseBodyType extends DefaultBodyType,
+> = MaybePromise<ResponseResolverReturnType<ResponseBodyType>>
 
-export type AsyncResponseResolverReturnType<BodyType extends DefaultBodyType> =
-  | MaybeAsyncResponseResolverReturnType<BodyType>
+export type AsyncResponseResolverReturnType<
+  ResponseBodyType extends DefaultBodyType,
+> = MaybePromise<
```
kettanaito, line 53: This is a fantastic change suggested by @Andarist to enable proper return type inference when annotating custom higher-order resolvers. Gold.

src/core/http.ts
kettanaito, line 13: ## Todo

We don't have any tests for this type right now. Need to add it to make sure it compiles without issues.

kettanaito: Added the tests.

mattcosta7 (approved):
Nice, this looks like a great addition that should simply some use-cases

kettanaito:
## Released: v2.0.14 🎉

This has been released in v2.0.14!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.0.14)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.0.14)

---

PR #1963, opened 2024-01-13
feat: support throwing responses in response resolvers

### LINKED ISSUE #1690: Support throwing Response in resolvers
### Scope

Improves an existing behavior

### Compatibility

### Feature description

In Remix, I have this really handy utility:

```tsx
/**
 * Provide a condition and if that condition is falsey, this throws a 400
 * Response with the given message.
 *
 * inspired by invariant from 'tiny-invariant'
 *
 * @example
 * invariantResponse(typeof value === 'string', `value must be a string`)
 *
 * @param condition The condition to check
 * @param message The message to throw (or a callback to generate the message)
 * @param responseInit Additional response init options if a response is thrown
 *
 * @throws {Response} if condition is falsey
 */
export function invariantResponse(
	condition: any,
	message: string | (() => string),
	responseInit?: ResponseInit,
): asserts condition {
	if (!condition) {
		throw new Response(typeof message === 'function' ? message() : message, {
			status: 400,
			...responseInit,
		})
	}
}
```

```tsx
const post = await getPost(params.postId)
// post: Post | undefined
invariant(post, 'Post not found', { status: 404 })
// post: Post
```

This works because remix catches errors I throw in my actions and loaders and checks whether it's a response and uses that response if so.

I would love to be able to reuse this utility in my MSW handlers. I think it's a great way to allow folks to build abstractions.

If accepted, could you please get ahead of Remix on [this particular issue](https://github.com/remix-run/remix/discussions/7043) 😅?

kettanaito:
- Closes #1690

### DIFF
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -263,10 +263,20 @@ export abstract class RequestHandler<
       request: args.request,
       parsedResult,
     })
-    const mockedResponse = (await executeResolver({
-      ...resolverExtras,
-      request: args.request,
-    })) as Response
+
+    const mockedResponsePromise = (
+      executeResolver({
+        ...resolverExtras,
+        request: args.request,
+      }) as Promise<Response>
+    ).catch((errorOrResponse) => {
+      // Allow throwing a Response instance in a response resolver.
+      if (errorOrResponse instanceof Response) {
+        return errorOrResponse
+      }
+    })
+
+    const mockedResponse = await mockedResponsePromise
 
     const executionResult = this.createExecutionResult({
       // Pass the cloned request to the result so that logging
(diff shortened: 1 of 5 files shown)


src/core/handlers/RequestHandler.ts
```diff
@@ -263,10 +263,20 @@ export abstract class RequestHandler<
...
-      ...resolverExtras,
-      request: args.request,
-    })) as Response
+
+    const mockedResponsePromise = (
+      executeResolver({
+        ...resolverExtras,
+        request: args.request,
+      }) as Promise<Response>
+    ).catch((errorOrResponse) => {
```
kettanaito, line 272: I'm adding this on the lowest level possible for two reasons:

1. This way all handlers get this behavior, no matter if the `handleRequest` is run in the browser or Node context;
2. This way even the direct users of the `RequestHandler` get this behavior, which is essential.

kettanaito:
## Released: v2.1.0 🎉

---

PR #1964, opened 2024-01-14
feat: deprecate "waitUntilReady" (deferNetworkUntil) worker option

### LINKED ISSUE #1573: MSW doesn't restore fetch is the worker is stopped before "deferNetworkRequestUntil" resolves

kettanaito:
- Fixes #1573
- Closes #1865
- Closes #1906

This removes the `deferNetworkUntil` functionality altogether. It's been introduced in the past to support use cases when developers ran `worker.start()` and their application rendering _in parallel_. This has since been discouraged. You must defer your application's rendering to eliminate any race conditions between the worker registration and any requests your app makes on the initial render. The default recommended browser integration also encourages that in the docs.


kettanaito:
Marking this as a minor version bump to be safe. There are no breaking changes for those not using the `waitUntilReady` options. For those using it, the deprecation warning suggests how to migrate to the deferred application rendering scenario by linking to the docs. The `deferNetworkUntil` functionality itself introduced a bunch of issues that went unnoticed, which makes me believe there's a small percentage of people actually using that option.

kettanaito:
## Released: v2.1.0 🎉

---

PR #1982, opened 2024-01-20
fix: export "getResponse" for batched GraphQL queries

### LINKED ISSUE #510: Support batched GraphQL queries

kettanaito:
- Closes #510 
- Closes #513 
- Closes #1585 
- Documented in https://github.com/mswjs/mswjs.io/pull/351

Provides two integration test examples of how to intercept and mock a batched GraphQL query that's created by Apollo and batched-execute.

## Roadmap

- [x] Provide the way to call `handleRequest` so it doesn't require tapping into `server.emitter` and `server.resolvedOptions` internals. 
- [x] Consider if `getResponse` is a good public API. Rename it to `resolveResponse`? Usage-wise, it's good, just wondering about the naming. 
  - Renamed to `executeHandlers`. It's precisely what this function does. 
- [x] Add tests that combine some mocked and some bypassed queries within a single batched query response. 
- [x] Add the "Batched GraphQL queries" recipe to the documentation. 
- [x] Change `getResponse` signature to `(handlers, request)`. No need for an object, its contract is unlikely to change (its intention is simple). 
- [x] Add unit tests for `getResponse`. 
- [x] Rename this pull request as a `fix` that gets `getResponse`

### DIFF
--- a/test/node/graphql-api/batched-queries.apollo.test.ts
+++ b/test/node/graphql-api/batched-queries.apollo.test.ts
@@ -4,30 +4,23 @@
  * @see https://github.com/mswjs/msw/issues/510
  * @see https://www.apollographql.com/docs/router/executing-operations/query-batching
  */
-import { http, graphql, HttpResponse, executeHandlers } from 'msw'
+import {
+  http,
+  graphql,
+  HttpResponse,
+  executeHandlers,
+  RequestHandler,
+  HttpHandler,
+} from 'msw'
 import { setupServer } from 'msw/node'
 
-const graphqlHandlers = [
-  graphql.query('GetUser', () => {
-    return HttpResponse.json({
-      data: {
-        user: { id: 1 },
-      },
-    })
-  }),
-  graphql.query('GetProduct', () => {
-    return HttpResponse.json({
-      data: {
-        product: { name: 'Hoover 2000' },
-      },
-    })
-  }),
-]
-
-export const handlers = [
-  // Let's create a request handler for batched GraphQL requests
-  // made using Apollo.
-  http.post('http://localhost/graphql', async ({ request, requestId }) => {
+/**
+ * A higher-order request handler function that resolves any
+ * batched GraphQL queries to the given URL against the list
+ * of request handlers.
+ */
+function batchedGraphQLQuery(url: string, handlers: Array<RequestHandler>) {
+  return http.post(url, async ({ request, requestId }) => {
     const data = await request.json()
 
     // Ignore GraphQL queries that are not batched queries.
@@ -46,7 +39,7 @@ export const handlers = [
         const result = await executeHandlers({
           request: scopedRequest,
           requestId,
-          handlers: graphqlHandlers,
+          handlers,
         })
 
         return result?.response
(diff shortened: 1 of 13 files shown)


test/node/graphql-api/batched-queries.apollo.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const responses = await Promise.all(
+      // Resolve each individual query against the existing
+      // array of GraphQL request handlers.
+      data.map(async (operation) => {
+        const scopedRequest = new Request(request, {
+          body: JSON.stringify(operation),
+        })
+
+        const result = await getResponse({
```
kettanaito: This is a much cleaner public API than `handleRequest`, which is relying heavily on the surrounding context (needs `emitter`, `options`).

kettanaito:
@mattcosta7, may I ask your opinion on exporting the `executeHandlers` function as public now?

mattcosta7:
> @mattcosta7, may I ask your opinion on exporting the `executeHandlers` function as public now?

Oh nice, I like this.  It does a nice job of encapsulating these operations and just returning the results!  Makes sense to me!

kettanaito:
@mattcosta7, I have a few thoughts on that API:

- I somewhat dislike that `parsedResult` is present in the execution result object. This looks like an internal detail that should not leak to the user. But once this is public, we won't be able to remove it nicely.
- The whole `executeHandlers` brings me to our discussion around redesigning how request lookup works internally. I'd like it to become sort of a class instance with a queue where you push requests and get them resolved. It doesn't stop us from refactoring internally and wrapping the new API in `executeHandlers` for backward-compatibility so it's relatively a low priority concern.

mattcosta7:
> @mattcosta7, I have a few thoughts on that API:
> 
> - I somewhat dislike that `parsedResult` is present in the execution result object. This looks like an internal detail that should not leak to the user. But once this is public, we won't be able to remove it nicely.
> - The whole `executeHandlers` brings me to our discussion around redesigning how request lookup works internally. I'd like it to become sort of a class instance with a queue where you push requests and get them resolved. It doesn't stop us from refactoring internally and wrapping the new API in `executeHandlers` for backward-compatibility so it's relatively a low priority concern. 

Yeah that makes sense.  As far as users should know just requests and responses exist and the rest is detail (maybe though, matching is also something transparent?)

kettanaito:
I'm thinking whether it'd make more sense to make `executeHandlers` as a new function, with the input only needed for public use and the return picking the `handler` and `response` (or even just `response`) and returning it. Basically, a wrapped around our internal `executeHandlers`. Easier to maintain and we don't change the function that works right for internal needs because it makes a bit lesser sense when used publicly.

kettanaito:
I think my latest commit addressed all the issues I have with the function! It's now `getResponse`, and it returns a mocked `Response` instance only.

kettanaito:
## Released: v2.1.3 🎉

---

PR #1988, opened 2024-01-21
fix: bundle browser build

thepassle:


thepassle:
Weird, looks like this breaks ts compat for some reason, and some failing tests. Weird, because locally the output seems to work fine in my unit tests and local storybook. Ill have to dive into that later, if someone else wants to continu this feel free to pick it up :)

tsup.config.ts
```diff
@@ -55,13 +56,14 @@ const browserConfig: Options = {
...
   entry: ['./src/browser/index.ts'],
-  external: [mswCore, ecosystemDependencies],
+  external: [ecosystemDependencies],
   format: ['esm', 'cjs'],
   outDir: './lib/browser',
   bundle: true,
   splitting: false,
   sourcemap: true,
   dts: true,
+  noExternal: Object.keys(packageJson.dependencies),
```
kettanaito: This introduces an impossible state. `noExternal`, I imagine, takes precedence over `external`, and so we end up bundling ecosystem dependencies. Is that the case?

tsup.config.ts
```diff
@@ -55,13 +56,14 @@ const browserConfig: Options = {
   name: 'browser',
   platform: 'browser',
   entry: ['./src/browser/index.ts'],
-  external: [mswCore, ecosystemDependencies],
+  external: [ecosystemDependencies],
```
kettanaito: Types are likely failing because we excluded `mswCore` here. We mustn't do that. If we do, we end up in `browser.d.ts` having _their own_ `RequestHandler` class than `msw/lib/core.d.ts`, and so this code won't compile:

```ts
import { http } from 'msw'
import { setupWorker } from 'msw/browser'
// These will think of different "RequestHandler" class!
```

kettanaito: If we keep `mswCore` in `external`, it will compile correctly and the CI should pass. All while giving us the bundling for the _rest_ of the dependencies we do use also. Wouldn't that be the goal here?

thepassle: > All while giving us the bundling for the rest of the dependencies we do use also.

Yes, thats what I had initially, but I noticed it left internal imports in-tact, which seemed odd to me, and so I bundled those as well. I think for now bundling the external dependencies is fine indeed 👍

kettanaito:
I've excluded the core modules from being bundled for the browser and browser tests pass locally. I believe that's the right direction here if we want the right TypeScript (and perhaps even runtime) behavior. I'd love to see if this change as-is gives any improvements on one of those fancy graphs of yours, @thepassle 🙏

kettanaito:
@thepassle, if you have a moment to look at this, I'd be thankful. I'd love to release this alongside #1990. Thanks!

kettanaito (approved):

kettanaito:
## Released: v2.1.4 🎉

This has been released in v2.1.4!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.1.4)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.1.4)

---

PR #1993, opened 2024-01-24
fix: add "request" to response events

### LINKED ISSUE #1992: "request" in worker.events.on('response:mocked') is always null
### Discussed in https://github.com/mswjs/msw/discussions/1991

<sup>Originally posted by **canrozanes** January 22, 2024</sup>
Hi, 

I was following the docs on [life-cycle events](https://mswjs.io/docs/api/life-cycle-events/) and I was trying to assert on a response after making an API call.

I have the following piece of code that saves all the requests so that I can assert on the responses later.

```
// QUnit  setup
hooks.beforeEach(async function (this: TestContextWithMsw) {
    this.worker = setupWorker(...handlers);
    this.worker.events.on('response:mocked', ({ request, requestId, response }) => {
      this.handledRequests.push({ request, requestId, response });
    });
});
```

I can clearly see that MSW sends a response in the browser network tab but when I add a debugger inside the 'response:mocked' callback, I see that all the request are null. 

Screenshot 2024-01-22 at 10 39 35 AM

What could I be missing?</div>

kettanaito:
- Fixes #1992

This is actually a feature. I've added the `request` argument to the response events by mistake, trying to keep things consistent. But I've never implemented the thing.


src/browser/setupWorker/start/createRequestListener.ts
```diff
@@ -30,6 +30,8 @@ export const createRequestListener = (
     const request = parseWorkerRequest(message.payload)
     const requestCloneForLogs = request.clone()
 
+    context.requests.set(requestId, request.clone())
```
kettanaito: I'm not sure what I feel in particular about this approach but I've been meaning to provide in-memory requestId/request storage for some time now. This will simplify the usage of life-cycle events.

src/browser/setupWorker/glossary.ts
```diff
@@ -103,6 +103,7 @@ export interface SetupWorkerInternalContext {
   worker: ServiceWorker | null
   registration: ServiceWorkerRegistration | null
   requestHandlers: Array<RequestHandler>
+  requests: Map<string, Request>
```
kettanaito, line 106: I would very much like this to be a `WeakMap` so we wouldn't have to clean it up but then to access the request instance it'd have to be a `WeakMap<Request, string>`, iterating over which would be rather inefficient. 

@mattcosta7, what do you think about this? Is it worth rewriting this as `WeakMap`?

src/browser/setupWorker/start/createRequestListener.ts
kettanaito: This also means +1 request clone per request.

kettanaito: A few important notes:

1. This is always called _once_ per 1 request. 
2. Request resolution hasn't started yet so there's no cache. 

We can technically make this the earliest clone to prevent the first request handler from cloning this request _again_. I think this makes sense. Wdyt, @mattcosta7?

kettanaito: I've added this first request clone to `RequestHandler.cache` so no request handlers would have to clone the request. 

It's still a bit of an overhead since now we are _always_ making 1 clone for every outgoing request (in the browser). Previously, only the first request handler would make the clone. I'd say this is not a concern since request cloning was done in the parsing phase, and the only scenario when this actually changes anything is if you have 0 request handlers (than, indeed, no cloning is needed). 

But we still have to clone the request even in that scenario so you can access it in the `response:*` life-cycle events.

mattcosta7: Yes, I think this is definitely worth it, and the cost seems low enough that it shouldn't be a concern

src/browser/setupWorker/glossary.ts
mattcosta7: Hmm

I think we could, but it might be a bit complicated, and not worth the effort?  

I think we would need to `WeakMap<Request, {requestId, RequestClone>` to really make that change, which might be ok, but it's pretty straightforward to cleanup, and should generally be small enough at any point in time to not be an issue?

src/browser/setupWorker/start/createRequestListener.ts
kettanaito: Now request handlers won't clone requests at all, I think. Any outgoing request will be cloned here, and then request handlers will reuse the clone from cache.

kettanaito:
## Released: v2.1.5 🎉

This has been released in v2.1.5!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.1.5)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.1.5)

---

PR #1996, opened 2024-01-25
feat: automatically set "Content-Length" on text/json responses

### LINKED ISSUE #1658: Set Content-Length header when using context.body or context.json
### Scope

Improves an existing behavior

### Compatibility

### Feature description

Using Context.json() sets the body and the Content-Type header, but not the Content-Length header. This means I need to manually set that header, because my implementation code checks it to determine whether or not there is a body. This in turn makes the Context.json() function a bit pointless for me, as I would need to stringify my mock body in order to set an accurate content-length, so then I may as well use Context.body() and set both headers.

Workarounds (for my use case):
- test code can set any non-zero value for Content-Length, and use Context.json()
- as fetch() returns a ReadableStream as body, implementation code can read from that and checks if it's empty, rather than checking the content-length header (feels marginally less efficient)

Can you foresee any problems in setting this header based on the length of the generated body string? I suppose technically it could be a breaking change? For consistency you'd also want to set it with the other body setters, so the proposal scope is wider than just the Content.json() function.

kettanaito:
- Closes #1658


mattcosta7 (approved):
👍

kettanaito:
## Released: v2.2.0 🎉

---

PR #2000, opened 2024-01-26
feat: support concurrent test runs via "server.boundary"

### LINKED ISSUE #474: Support concurrent runs in Node.js

kettanaito:
- Fixes #474 
- Closes #1370
- A bit about how this works: https://github.com/mswjs/msw/issues/474#issuecomment-1912327731
- Documented in https://github.com/mswjs/mswjs.io/pull/356

## Changes

The `server.boundary()` API is a Node.js-only API that provides encapsulation to whichever given callback function by executing it within an `AsyncLocalStorage` context. This API is designed to support using MSW in concurrent test runs since no changes to request handlers (e.g. using `server.use()`) will ever leave the boundary function.  

`server.boundary()` has its practical use outside of testing as well. You can use it in any concurrent logic that performs network request, if you wish to handle those requests differently on a per-boundary basis. 

## Usage

You must wrap each test function in a concurrent test suite in the `server.boundary()` function to ensure proper isolation.

```js
it.concurrent('one', server.boundary(async () => {
  await fetch('/resource') // 200 OK
}))

it.concurrent('two', server.boundary(async () => {
  // This request handler override will only affect the network
  // calls that happen within this server boundary. Neat!
  server.use(http.get('/resource', () => HttpResponse.error()))
  await fetch('/resource') // Failed to fetch
}))
```

Server boundary can also be used anywhere else in your application to affect the network only within the given function scope. For example, you can debug what network requests a subtree of your application makes:

```js
server.boundary(() => {
  http.all('*', ({ request }) => console.log(request.method, request.url))
  subtree()
})()
```

## Todos

- [x] **Test `server.boundary()`**. Every behavior must be verified in tests. 
- [x] Ensure React Native build isn't broken. I doubt there's `node:async_events` there, so a dependency on it in `SetupServerApi` may throw on runtime. 4a1c9833dbe4bca0fc98f1d8923ae8f329e024b9
- [x] Document `server.boundary()`

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "check:exports": "node \"./config/scripts/validate-esm.js\"",
     "test": "pnpm test:unit && pnpm test:node && pnpm test:browser && pnpm test:native",
     "test:unit": "vitest",
-    "test:node": "vitest --config=./test/node/vitest.config.ts",
+    "test:node": "vitest run --config=./test/node/vitest.config.ts",
     "test:native": "vitest --config=./test/native/vitest.config.ts",
     "test:browser": "playwright test -c ./test/browser/playwright.config.ts",
     "test:modules:node": "vitest --config=./test/modules/node/vitest.config.ts",
--- a/src/browser/setupWorker/setupWorker.ts
+++ b/src/browser/setupWorker/setupWorker.ts
@@ -58,8 +58,10 @@ export class SetupWorkerApi
       isMockingEnabled: false,
       startOptions: null as any,
       worker: null,
+      currentHandlers: () => {
+        return this.handlersController.currentHandlers()
+      },
       registration: null,
-      requestHandlers: this.currentHandlers,
       requests: new Map(),
       emitter: this.emitter,
       workerChannel: {
@@ -150,16 +152,6 @@ export class SetupWorkerApi
         readableStreamTransfer: supportsReadableStreamTransfer(),
       },
     }
-
-    /**
-     * @todo Not sure I like this but "this.currentHandlers"
-     * updates never bubble to "this.context.requestHandlers".
-     */
-    Object.defineProperties(context, {
-      requestHandlers: {
-        get: () => this.currentHandlers,
-      },
-    })
 
     this.startHandler = context.supports.serviceWorkerApi
       ? createFallbackStart(context)
--- a/src/browser/setupWorker/setupWorker.ts
+++ b/src/browser/setupWorker/setupWorker.ts
@@ -58,7 +58,7 @@ export class SetupWorkerApi
       isMockingEnabled: false,
       startOptions: null as any,
       worker: null,
-      currentHandlers: () => {
+      getRequestHandlers: () => {
         return this.handlersController.currentHandlers()
       },
       registration: null,
--- a/src/core/SetupApi.ts
+++ b/src/core/SetupApi.ts
@@ -10,12 +10,38 @@ import { pipeEvents } from './utils/internal/pipeEvents'
 import { toReadonlyArray } from './utils/internal/toReadonlyArray'
 import { Disposable } from './utils/internal/Disposable'
 
+export abstract class HandlersController {
+  abstract prepend(runtimeHandlers: Array<RequestHandler>): void
+  abstract reset(nextHandles: Array<RequestHandler>): void
+  abstract currentHandlers(): Array<RequestHandler>
+}
+
+export class InMemoryHandlersController implements HandlersController {
+  private handlers: Array<RequestHandler>
+
+  constructor(private initialHandlers: Array<RequestHandler>) {
+    this.handlers = [...initialHandlers]
+  }
+
+  public prepend(runtimeHandles: Array<RequestHandler>): void {
+    this.handlers.unshift(...runtimeHandles)
+  }
+
+  public reset(nextHandlers: Array<RequestHandler>): void {
+    this.handlers =
+      nextHandlers.length > 0 ? [...nextHandlers] : [...this.initialHandlers]
+  }
+
+  public currentHandlers(): Array<RequestHandler> {
+    return this.handlers
+  }
+}
+
 /**
  * Generic class for the mock API setup.
  */
 export abstract class SetupApi<EventsMap extends EventMap> extends Disposable {
-  protected initialHandlers: ReadonlyArray<RequestHandler>
-  protected currentHandlers: Array<RequestHandler>
+  protected handlersController: HandlersController
   protected readonly emitter: Emitter<EventsMap>
   protected readonly publicEmitter: Emitter<EventsMap>
 
@@ -31,8 +57,7 @@ export abstract class SetupApi<EventsMap extends EventMap> extends Disposable {
       ),
     )
 
-    this.initialHandlers = toReadonlyArray(initialHandlers)
-    this.currentHandlers = [...initialHandlers]
+    this.handlersController = new InMemoryHandlersController(initialHandlers)
 
     this.emitter = new Emitter<EventsMap>()
     this.publicEmitter = new Emitter<EventsMap>()
@@ -59,24 +84,23 @@ export abstract class SetupApi<EventsMap extends EventMap> extends Disposable {
       ),
     )
 
-    this.currentHandlers.unshift(...runtimeHandlers)
+    this.handlersController.prepend(runtimeHandlers)
   }
 
   public restoreHandlers(): void {
-    this.currentHandlers.forEach((handler) => {
+    this.handlersController.currentHandlers().forEach((handler) => {
       handler.isUsed = false
     })
   }
 
   public resetHandlers(...nextHandlers: Array<RequestHandler>): void {
-    this.currentHandlers =
-      nextHandlers.length > 0 ? [...nextHandlers] : [...this.initialHandlers]
+    this.handlersController.reset(nextHandlers)
   }
 
   public listHandlers(): ReadonlyArray<
     RequestHandler<RequestHandlerDefaultInfo, any, any>
   > {
-    return toReadonlyArray(this.currentHandlers)
+    return toReadonlyArray(this.handlersController.currentHandlers())
   }
 
   private createLifeCycleEvents(): LifeCycleEventEmitter<EventsMap> {
--- a/src/node/SetupServerApi.ts
+++ b/src/node/SetupServerApi.ts
@@ -1,3 +1,4 @@
+import { AsyncLocalStorage } from 'node:async_hooks'
 import {
   BatchInterceptor,
   HttpRequestEventMap,
@@ -5,8 +6,12 @@ import {
   InterceptorReadyState,
 } from '@mswjs/interceptors'
 import { invariant } from 'outvariant'
-import { SetupApi } from '~/core/SetupApi'
-import { RequestHandler } from '~/core/handlers/RequestHandler'
+import { HandlersController, SetupApi } from '~/core/SetupApi'
+import {
+  RequestHandler,
+  RequestHandlerDefaultInfo,
+  RequestHandlerOptions,
+} from '~/core/handlers/RequestHandler'
 import { LifeCycleEventsMap, SharedOptions } from '~/core/sharedOptions'
 import { RequiredDeep } from '~/core/typeUtils'
 import { handleRequest } from '~/core/utils/handleRequest'
@@ -16,6 +21,30 @@ import { SetupServer } from './glossary'
 
 const DEFAULT_LISTEN_OPTIONS: RequiredDeep<SharedOptions> = {
   onUnhandledRequest: 'warn',
+}
+
+const store = new AsyncLocalStorage<{ handlers: Array<RequestHandler> }>()
+
+class AsyncHandlersController implements HandlersController {
+  constructor(private readonly initialHandlers: Array<RequestHandler>) {
+    store.enterWith({ handlers: initialHandlers })
+  }
+
+  public prepend(runtimeHandlers: Array<RequestHandler>) {
+    store.enterWith({
+      handlers: this.currentHandlers().concat(runtimeHandlers),
+    })
+  }
+
+  public reset(nextHandlers: Array<RequestHandler>) {
+    store.enterWith({
+      handlers: nextHandlers.length > 0 ? nextHandlers : this.initialHandlers,
+    })
+  }
+
+  public currentHandlers(): Array<RequestHandler> {
+    return store.getStore()?.handlers || this.initialHandlers
+  }
 }
 
 export class SetupServerApi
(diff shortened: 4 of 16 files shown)


src/node/SetupServerApi.ts
```diff
@@ -107,7 +150,29 @@ export class SetupServerApi
     )
   }
 
+  /**
+   * Wraps the given function in a boundary. Any changes to the
+   * network behavior (e.g. adding runtime request handlers via
+   * `server.use()`) will be scoped to this boundary only.
+   * @param callback A function to run (e.g. a test)
+   */
+  public boundary<Fn extends (...args: Array<unknown>) => unknown>(
```
mattcosta7: neat!  

should every call to `use` create some kind of `boundary` internally - can we avoid leaking that to the end user? (sorry if this is still early)

kettanaito: Yes and no. We can use `.enterWith()` from `AsyncLocalStorage` on each `server.use()`, and while it would solve _some_ use cases, it would break the other. 

The problem is that relying on async storage internally makes it hard to support sequential tests. Here's an example:

```js
const server = setupServer()

beforeAll(() => {
  // This will have a different execution id...
  server.use(...someHandlers)
  server.listen()
})

it('test', () => {
  // ...than this.
  server.use(...anotherHandlers)

  // So the request handling logic that needs the list
  // of all handlers won't be able to access the context
  // (and the handlers) introduced in "beforeAll".
  fetch()
})
```

The issue here is that the context created by the `server.use()` call in `beforeAll` will be different and inaccessible by the context created by `sever.use()` in the test ( they have different execution async ids). 

> I also have concerns that using `.enterWith()` on each `server.use()` call creates nested execution contexts (imagine you have multiple `.use()` calls in a single test), which may have negative impact on runtime performance. 

It is for this reason I'm proposing an explicit `server.boundary()` API. This API is effectively a wrapper around `.run()` function from async storage and it _doesn't create a new execution context_ (at least as far as I understand). In any case, since it's not bound to individual `.use()` calls, you can have as many calls as you wish within a single boundary and they will remain within the same execution context.

package.json
```diff
@@ -52,7 +52,7 @@
     "check:exports": "node \"./config/scripts/validate-esm.js\"",
     "test": "pnpm test:unit && pnpm test:node && pnpm test:browser && pnpm test:native",
     "test:unit": "vitest",
-    "test:node": "vitest --config=./test/node/vitest.config.ts",
+    "test:node": "vitest run --config=./test/node/vitest.config.ts",
```
kettanaito, line 55: Note: our integration tests in Node.js cannot be run in watch mode because we have one test that builds the library. It triggers the entire test run anew.

src/browser/setupWorker/setupWorker.ts
```diff
@@ -58,8 +58,10 @@ export class SetupWorkerApi
       isMockingEnabled: false,
       startOptions: null as any,
       worker: null,
+      currentHandlers: () => {
```
kettanaito: I'm making this a getter the way it was always supposed to be.

src/core/SetupApi.ts
```diff
@@ -10,12 +10,38 @@ import { pipeEvents } from './utils/internal/pipeEvents'
 import { toReadonlyArray } from './utils/internal/toReadonlyArray'
 import { Disposable } from './utils/internal/Disposable'
 
+export abstract class HandlersController {
```
kettanaito, line 13: This may be over-architecturing on my part so I'm open to feedback. The idea is that the request handlers list management is kind of its own thing. Overriding `.use()` and `.reset()` methods in child classes also require them to re-implement additional logic, like the input validation, and it feels a bit redundant. 

@mattcosta7, what is your opinion on this?

src/node/SetupServerApi.ts
mattcosta7: that makes sense.  it's too bad that this leaks though, but I don't think that's a big deal.

src/core/SetupApi.ts
mattcosta7: I don't mind this explicit abstract class.  Agreed it might be heavier than necessary, but it handles interface definition and provides a nice wrapper around the fixes 'array' storage for handlers, which I also kind of like, so I'm for it

src/node/SetupServerApi.ts
kettanaito: I actually think this is very good that it leaks. It makes a server boundary a conscious choice. Just as running tests in parallel is a conscious choice (no testing framework does it by default). 

Server boundary is also incredibly powerful outside of testing too! Imagine you want to quickly fail all requests coming from a sub-tree in your app:

```js
server.boundary(() => {
  http.all('*', () => HttpResponse.error())
  subtree()
})()
```

This will fail all network requests within `subtree()` but have no effect on the network outside of this boundary. Looks like a neat debugging tool to have! (And since handlers are "transparent" by default, it makes it twice as useful)

kettanaito:
If this API makes it, I will write a blog post about it. The more I throw different use cases at it, the more I love it!

src/core/SetupApi.ts
kettanaito: Thank you, @mattcosta7!

christoph-fricke:
This looks really interesting! Recently, I have been thinking about use-cases for `AsyncLocalStorage`, mostly using it as a dependency context mechanism. `server.boundary` looks like a really good use case for it. I like it!

Personally, I mostly use MSW to mock network requests in Playwright tests. My setups use either test or worker [fixtures](https://playwright.dev/docs/test-fixtures) to init and reset MSW, depending on requirements. Workers are Playwright's concept for concurrent tests, with each [worker being a different process](https://playwright.dev/docs/test-parallel#worker-processes). Since tests and their handlers are already isolated through workers, I have been wondering whether or not `server.boundary` is relevant when using MSW in Playwright. Don't get me wrong - I like it and think it is useful for MSW in general. But is it needed when using MSW in Playwright? :)

kettanaito:
@christoph-fricke, thank you so much for sharing your feedback!

> Since tests and their handlers are already isolated through workers, I have been wondering whether or not server.boundary is relevant when using MSW in Playwright

As far as I remember, Playwright uses workers to parallelize _test files_ but workers have no effect on the actual browser runtime. When you use MSW with Playwright, you enable it as a part of the browser's runtime, either in your application or in the browser directly via Playwright fixtures. 

With Playwright, you achieve network isolation because _every tab is a separate runtime_. This means that `worker.use()` in Tab 1 has no effect on the network in Tab 2 because those two runtimes of the same app has registered the same worker, and keep _their own `currentHandlers` list_ in-memory (which is not shared). 

This is why `server.boundary()` is a Node.js-specific API. The browser runtime has no such problem and there's also no `AsyncLocalStorage` in the browser.

mattcosta7:
> @christoph-fricke, thank you so much for sharing your feedback!
> 
> > Since tests and their handlers are already isolated through workers, I have been wondering whether or not server.boundary is relevant when using MSW in Playwright
> 
> As far as I remember, Playwright uses workers to parallelize _test files_ but workers have no effect on the actual browser runtime. When you use MSW with Playwright, you enable it as a part of the browser's runtime, either in your application or in the browser directly via Playwright fixtures.
> 
> With Playwright, you achieve network isolation because _every tab is a separate runtime_. This means that `worker.use()` in Tab 1 has no effect on the network in Tab 2 because those two runtimes of the same app has registered the same worker, and keep _their own `currentHandlers` list_ in-memory (which is not shared).
> 
> This is why `server.boundary()` is a Node.js-specific API. The browser runtime has no such problem and there's also no `AsyncLocalStorage` in the browser.

I think _it depends_?

You can use playwright, and mock the network from the browser with service workers (which I agree shouldn't need this).

But you can also intercept network requests from the playwright node process, in which case this could be used in playwright tests too - so it depends how/where you're applying mocks in your playwright configuration

mattcosta7 (commented):
I'm really liking this.  What testing/next steps  do we need to push it forward?

kettanaito:
> But you can also intercept network requests from the playwright node process

I'm missing context on this. What is "playwright node process"? I believe Playwright has two parts:

- Node.js process, which is your test. It establishes a message channel with the browser runtime.
- Browser process, the actual running browser runtime. 

If I understand you right, you mean you can use `setupServer` in the Playwright test to intercept requests you make in that test... But I fail to see when you would do those requests. If you use Playwright, your tested code runs _in the browser process_. Otherwise, you don't really need Playwright.

kettanaito:
> I'm really liking this. What testing/next steps do we need to push it forward?

It's fully ready. I've written a blog post on this as well to introduce this change and dive a bit into its technical implementation. The next step is to wait for my wife to finish the illustration for the blog post and we can get it released! This is likely to happen next Monday. Exciting! 

Until then, I can use some feedback on the [Server Boundary API docs](https://website-git-server-boundary-mswjs.vercel.app/docs/api/setup-server/boundary). If you have a minute, Matt, could you share your opinion on that page? Is everything explained clearly?

mattcosta7:
> > But you can also intercept network requests from the playwright node process
> 
> I'm missing context on this. What is "playwright node process"? I believe Playwright has two parts:
> 
> * Node.js process, which is your test. It establishes a message channel with the browser runtime.
> * Browser process, the actual running browser runtime.
> 
> If I understand you right, you mean you can use `setupServer` in the Playwright test to intercept requests you make in that test... But I fail to see when you would do those requests. If you use Playwright, your tested code runs _in the browser process_. Otherwise, you don't really need Playwright.

Yeah - you can use playwright's request interception to intercept the request from the node side instead of from the browser side.  I'm not really sure _why_ you would want to do that instead of just letting the service worker handle it - but you could do that https://playwright.dev/docs/mock

kettanaito:
@mattcosta7, but that's just API semantics. Regardless of what you choose to use in Playwright, the only environment where those requests actually happen is still the browser. Even with the `.route()` API in Playwright, _no requests will happen in the Node.js process_. It's just a binding that reports requests Playwright intercepts in the browser and forwards them to that binding. You wouldn't need the server boundary in this scenario.

mattcosta7:
> @mattcosta7, but that's just API semantics. Regardless of what you choose to use in Playwright, the only environment where those requests actually happen is still the browser. Even with the `.route()` API in Playwright, _no requests will happen in the Node.js process_. It's just a binding that reports requests Playwright intercepts in the browser and forwards them to that binding. You wouldn't need the server boundary in this scenario.

Ahh - interesting, this could have been a misunderstanding on my part!

mattcosta7:
> > I'm really liking this. What testing/next steps do we need to push it forward?
> 
> It's fully ready. I've written a blog post on this as well to introduce this change and dive a bit into its technical implementation. The next step is to wait for my wife to finish the illustration for the blog post and we can get it released! This is likely to happen next Monday. Exciting!
> 
> Until then, I can use some feedback on the [Server Boundary API docs](https://website-git-server-boundary-mswjs.vercel.app/docs/api/setup-server/boundary). If you have a minute, Matt, could you share your opinion on that page? Is everything explained clearly?

Will do - mostly likely tomorrow morning

kettanaito:
> Ahh - interesting, this could have been a misunderstanding on my part!

No worries! The node/browser bindings can get quite confusing. Afaik, everything you access from Playwright is created just for you. All of those are bindings that communicate with the browser process.

mattcosta7:
@kettanaito I think the docs looks good!

this is a minor nit/question:

I wonder if we should / need to call out that while the handlers are scoped to the boundary, any user objects defined outside of the boundary are not.   This might be not-necessary, since it's basically just how scope works in the language, but I wonder if we'll receive issues like

```ts
const someThing = new SomeThing()
const server = setupServer()

test('', () => {
   server.boundary(() => {
       server.use(
           http.post('https://example.com/login', () => {
               sometThing.field = 1
               return new HttpResponse.json(someThing, { status: 500 })
            })
       )
   })
   expect(someThing.field).toBe(1)
})

test('', () => {
   server.boundary(() => {
       server.use(
           http.get('https://example.com/login', () => {
               return new HttpResponse.json(someThing, { status: 500 })
            })
       )
   })
   expect(someThing.field).toBeUndefined()
})
```

super trivial example, but run concurrently these can race easily it seems likely state could leak into boundaries this way, and it might be good to call out that users should avoid leaking state when using boundaries and concurrency

I think this is generally just an indicator of how they should be setting up their tests to handle concurrency, but it might be worth warning about too.

kettanaito:
@mattcosta7, that's a good point. Let's omit this on the ground of being common knowledge but if we ever get an issue about someone expecting that from the server boundary, we can add a notice to the docs. 

The post is ready, planned to be released on Monday. 🚀

mattcosta7 (approved):

christoph-fricke:
> @mattcosta7, but that's just API semantics. Regardless of what you choose to use in Playwright, the only environment where those requests actually happen is still the browser. Even with the `.route()` API in Playwright, _no requests will happen in the Node.js process_. It's just a binding that reports requests Playwright intercepts in the browser and forwards them to that binding. You wouldn't need the server boundary in this scenario.  

@kettanaito That only right for CSR apps. I prefer using Playwright's network mocking API and delegate requests to MSW as described by you both. Kinda what [playwright-msw](https://www.npmjs.com/package/playwright-msw) is doing, as I prefer to test against a production build with zero test related code in it.

However, testing SSR apps requires  mocking requests that happen in a Node.js process. A colleague and I found quite a nice solution for testing Remix apps, which fully isolates tests with a mocked network thanks to Playwright's and MSW's flexible APIs. The trick is to use MSW Node.js API to intercept requests in worker processes and create the remix server in the same worker. Playwright fixtures are ideal for achieving this. Thanks to the strong worker isolation of Playwright (OS processes), this pattern still does not need `server.boundary()`. A rough POC implementation of this approach can be found here: https://github.com/luchsamapparat/remix-playwright-msw/blob/master/tests/fixture.ts

kettanaito:
@christoph-fricke, makes sense. We are also exploring `setupRemoteServer` for native server-side network interception in MSW (see #1617).

kettanaito:
## Released: v2.2.0 🎉

---

HristoKolev:
Just FYI this fixes a bug where `server.resetHandlers` would act as `server.use` and not reset handlers. I recently updated msw (node) from `2.0.11` to `2.2.13` and that caused a bunch of tests to fail because `resetHandlers` started to work correctly. I tracked it down to `2.2.0` and probably this PR.

Issue #1984, opened 2024-01-21
Avoid barrel file exports


thepassle:
### Scope

Improves an existing behavior

### Compatibility

- [ ] This is a breaking change

### Feature description

I've mentioned this on discord, but figured I'd create some issues to have this public as well and to better keep track of it. Currently MSW uses barrel files, Marvin Hagemeister wrote a good blog on why they are problematic for dev dependencies: https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/

It would be good to create a better separation of entrypoints rather than exposing a single barrel file. We should also look into this for any dependent packages under the `@mswjs/*` namespace.

Removing barrel files and a better separation of entrypoints (and avoiding internal modules _importing_ from barrel files) can lead to some pretty significant performance wins

kettanaito:
Thanks for proposing this, @thepassle! Totally support this, let's improve the way we import (and export, if talking about our ecosystem packages) things.

thepassle:
Here's the PR that removes imports to a barrel file by an internally used module https://github.com/mswjs/msw/pull/1987

Leads to a pretty nice change already:
Before:
<img width="1131" alt="Screenshot 2024-01-21 at 10 10 12" src="https://github.com/mswjs/msw/assets/17054057/4716c67b-7ee2-4db6-9b39-ef0541ed8152">
After:
<img width="1260" alt="Screenshot 2024-01-21 at 09 53 55" src="https://github.com/mswjs/msw/assets/17054057/67e91a68-a7b7-4dad-ab5a-9c939b8985b1">

kettanaito:
Fixed by #1987. 👏

thepassle:
Sorry, that PR just addresses one of the problems with barrel files; internal modules importing from a barrel file.

The issue itself still exists: the barrel file. Importing `http` from `"msw"` leads you to importing _everything_ in MSW, most of which largely goes unused, so we should still consider splitting up imports into separate(/grouped) entrypoints

kettanaito:
@thepassle, do you have the internal build entrypoints in mind (i.e. `tsup` config) or actually changing the way MSW exports things? Like `msw/http`? I have some concerns against the latter.

thepassle:
The latter; the way MSW exposes exports to users. E.g.: currently if you import `http` from `"msw"`, you import _everything_ in MSW. I recommend giving this a good read for the reasoning why: https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/

[...]

PR #2004, opened 2024-01-30
feat: add "/http" and "/graphql" export paths

### LINKED ISSUE #2038: Add "http" and "graphql" exports
### Scope

Adds a new behavior

### Compatibility

### Feature description

- Originally discussed in https://github.com/mswjs/msw/discussions/2037

thepassle:
… file

- Discussed in #2037 (#1984)
- Fixes #2038


package.json
```diff
@@ -20,6 +20,18 @@
...
+      "types": "./lib/core/http.d.ts",
+      "require": "./lib/core/http.js",
+      "import": "./lib/core/http.mjs",
+      "default": "./lib/core/http.mjs"
+    },
+    "./graphql": {
+      "types": "./lib/core/graphql.d.ts",
+      "require": "./lib/core/graphql.js",
+      "import": "./lib/core/graphql.mjs",
+      "default": "./lib/core/graphql.mjs"
```
kettanaito: We seem to have inconsistent `default` condition across exports:

- `/http`, `/graphql` and `/node` list the ESM build as the default;
- `.`, `/browser` and `/native` list the CJS build as the default. 

Is this something we should unify as a part of this change as well?

thepassle: we can change the `default` for `./http` and `./graphql`, but changing the defaults for other keys may be a breaking change

kettanaito: I was mainly suggesting the former. As we are in this weird CJS/ESM situation, the default must remain CJS. Would you please make the change?

thepassle: done

kettanaito (dismissed):
This looks great! Let's have this released in the next minor version. **Just don't merge this yet**, I'd like to release a few bug fixes before the next version bump. Thanks. 

@thepassle, how do you recommend us mention these exports in the docs? I still think that importing from the root should be the default. I wonder where these export paths would be proper to mention.

pleunv:
The annoying thing is that autocompletion will most likely favor the shortest path, and I don't think there's a way to tell it otherwise. What I generally end up doing is set up eslint's [no-restricted-imports](https://eslint.org/docs/latest/rules/no-restricted-imports) rule with a custom message to reroute users to the right import. Not the most user-friendly approach, though.

Something along the lines of:

```json
"rules": {
  "no-restricted-imports": [
    "error",
    {
      "paths": [
        {
          "name": "msw",
          "importNames": ["graphql"],
          "message": "Please import 'graphql' from 'msw/graphql' instead."
        },
      ]
    }
  ]
}
```

kettanaito:
@pleunv, autocomplete right now is pretty broken for me anyways. VS Code always tries to import things where their _type definitions are_, not what the export says in package.json. For example, when I autoimport `http` I get this:

```ts
import { http } from 'msw/lib/http'
```

Which is incorrect; you shouldn't be importing things from `/lib/`. This looks like a problem the end developer has to configure themselves.

pleunv:
Of all the TypeScript issues I've ever run into, fortunately that's not one of them 😄. That would honestly drive me mad. Did you override any of the "Import Module" settings?

kettanaito:
@pleunv, no, that's the default behavior.

thepassle:
> This looks great! Let's have this released in the next minor version. Just don't merge this yet, I'd like to release a few bug fixes before the next version bump. Thanks.

Okidoki, any idea on when the next minor will be released?

> I still think that importing from the root should be the default. I wonder where these export paths would be proper to mention.

Not really sure why you wouldnt change this:
image

to:
image

I hardly think this impacts DX for the largest usecase, even if you use both graphql and http, but I'm also not really interested in discussing that any longer; I think the perf difference outweighs DX, but it seems I haven't been able to convince you of that, so with peace and love, I won't go into that any further and leave it at this :)

> @thepassle, how do you recommend us mention these exports in the docs? 

Ill leave that up to your discretion

kettanaito:
> Okidoki, any idea on when the next minor will be released?

I'm releasing next 2 bug fixes related to React Native, and then it's the minor release time!

> Not really sure why you wouldnt change this: ... to:

You are omitting the experience implications here. All the tutorials, videos, and articles feature `from 'msw'`. If the person goes to the docs and, suddenly, it's `from 'msw/http'`, a red flag goes in their head that someone is wrong. Either the materials they read are wrong or the docs. 

That's why performance improvements like these deserve a separate page, I think. Before (and if) we drop the root-level export altogether, I won't recommend this by default for one simple reason: nobody has reported performance issues due to barrel imports in the past, so nobody had performance issues due to barrel imports, it's as simple as that. 

Those who need this fix should be able to find it. 

> but it seems I haven't been able to convince you of that

You've convinced me. From my side, I'm trying to show you the area of effect this change has on the end developer. I will not discard educational experience and simplicity of use. The `msw/http` and `msw/graphql` imports will be _opt-in_ for those who experience issues. Those who don't can keep using `msw` root-level export as they did in the past.

kettanaito:
I have one piece of feedback. The format we have now is `msw/<environment>`: `msw/node`, `msw/browser`, `msw/native`. I feel that `msw/http` breaks this rule. 

What do you think about `msw/core/http`/`msw/core/graphql`?

kettanaito:
## Released: v2.2.0 🎉

---

PR #2005, opened 2024-01-30
fix: use specific import over barrel file import

thepassle:
Reduces the module graph from importing 179 modules total to 74

Before:
```
  [ERROR]: "/Users/au87xu/msw/lib/core/index.mjs" leads to a module graph of 179 modules, which is more than the allowed maxModuleGraphSize of 10.
```

After:
```
  [ERROR]: "/Users/au87xu/msw/lib/core/index.mjs" leads to a module graph of 74 modules, which is more than the allowed maxModuleGraphSize of 10.
```

as discussed in https://github.com/mswjs/msw/issues/1984

### DIFF
--- a/src/core/utils/internal/parseGraphQLRequest.ts
+++ b/src/core/utils/internal/parseGraphQLRequest.ts
@@ -3,7 +3,7 @@ import type {
   OperationDefinitionNode,
   OperationTypeNode,
 } from 'graphql'
-import { parse } from 'graphql'
+import { parse } from 'graphql/language/parser'
 import type { GraphQLVariables } from '../../handlers/GraphQLHandler'
 import { toPublicUrl } from '../request/toPublicUrl'
 import { devUtils } from './devUtils'
--- a/src/core/utils/internal/parseGraphQLRequest.ts
+++ b/src/core/utils/internal/parseGraphQLRequest.ts
@@ -3,7 +3,7 @@ import type {
   OperationDefinitionNode,
   OperationTypeNode,
 } from 'graphql'
-import { parse } from 'graphql/language/parser'
+import { parse } from 'graphql/language/parser.js'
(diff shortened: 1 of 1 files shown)


thepassle:
~~I don't understand why the test is failing:~~
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/runner/work/msw/msw/test/modules/node/node-esm-tests/node_modules/.pnpm/file+..+..+..+..+msw-2.1.5.tgz/node_modules/graphql/language/parser' imported from /home/runner/work/msw/msw/test/modules/node/node-esm-tests/node_modules/.pnpm/file+..+..+..+..+msw-2.1.5.tgz/node_modules/msw/lib/core/utils/internal/parseGraphQLRequest.mjs
Did you mean to import graphql@16.8.1/node_modules/graphql/language/parser.js?
```

~~`graphql/language/parser` seems to be a valid import and this import change works fine in the browser, there's no package exports in graphql that prevent this import either, so it feels like this is just some tooling getting in the way or something? Does the error make sense to you @kettanaito ?~~

Nvm, think I fixed it

thepassle:
Ah, hehe. So looks like this is a dual CJS/ESM issue :) If I use the import `graphql/language/parser.js`, it works in the one test, and fails in the other, and if I use the import `graphql/language/parser.mjs` it fails in the one test and works in the other (and also it cant resolve the types).

So I guess we can't make this change, and I guess this only works for the `graphql` module specifier is because of the `"main"` and `"module"` they have setup in the graphql package.

We might be able to change the module specifier during the build, but the problem with that is that we then have to apply that transform to all the other tooling (like tests etc) as well, which is probably not a good idea.

So it's very unfortunate, but I guess I'll have to close this PR for now. A decrease of 105 modules in the module graph sure would have been nice.

kettanaito:
@thepassle, something to consider in 3.0 when (and I hope) MSW be ESM-only.

Issue #1921, opened 2023-12-11
Support URLPattern as request predicate


kettanaito:
### Scope

Adds a new behavior

### Compatibility

- [ ] This is a breaking change

### Feature description

As soon as [`URLPattern`](https://caniuse.com/mdn-api_urlpattern) gets globally supported, we should support it as the input to the request handlers.

```js
http.get(new URLPattern('https://example.com/:resource'), resolver)
```

We can also migrate to represent the request matching using the `URLPattern` instances instead of relying on `path-to-regexp` and a custom `matchRequestUrl` function.

kettanaito:
Closing this because the global `URLPattern` is unlikely to be supported. It has serious performance issues with large sets of route handlers, which makes it unsuitable for the way we are using route matching in MSW.

We might explore alternatives, like `@remix-run/route-pattern`, which is similar but more performant.

PR #2011, opened 2024-02-03
feat: support mocking WebSocket APIs

### LINKED ISSUE #156: Support mocking WebSocket APIs

kettanaito:
- Closes #156 
- Discussed in #2010 

## Roadmap

- [x] Support runtime overrides for event handlers.
- [x] ~~Extend `Handler` in `RequestHandler`. This way `HttpHandler` and `GraphQLHandler` will be assignable to type `Handler`, and `Array<Handler>` can annotate both HTTP and WebSocket handlers~~. 
  - Bad idea. Even abstracted, `RequestHandler` relies on caching to clone requests, parsed result, etc. Not enough common ground to reuse between it and `WebSocketHandler`. WS handler is better off as a standalone class (doesn't support `once` anyway). 
- [x] Call `handleWebSocketEvent()` in `SetupWorkerApi` and `SetupServerApi`.
- [x] handleRequest: Check that the handler is `instanceof RequestHandler` to skip WebSocket handlers.
- [x] Update to `@mswjs/interceptors` that ships `WebSocketInterceptor`.
- [x] **Improve types**. The `data` argument of message listeners must be annotated across the board (includes annotating the `MessageEvent` in the Interceptors). 
  - Not a good idea. The `data` you receive is still `string | Blob | ArrayBuffer`. You cannot just cast it to something else. Instead, introduce a custom parser that does that casting, as well as the runtime `typeof` checks. That's the way.
  - [x] Document this approach. 
- [x] Add links to jsdoc blocks of the new API.
- [x] Bug: look into why `pnpm test:modules:node` hangs forever on `setupServer()` without the explicit `process.exit()` in the generated `runtime.*` scripts. 
- [x] Add Node.js tests.
- [x] Add browser tests.
- [x] Add `.use()` override tests (i.e. `event.stopImmediatePropagation()`). 
- [x] Defer the interceptor application to the `.start()`/`.listen()` calls.
- [x] Add `server.close()` API (a task for Interceptors, here just add tests). 
- [x] Invoke `onUnhandledRequest()` (or similar) for unhandled WebSocket connections. Technically, there are no requests, so the existing method reads weird. https://github.com/mswjs/msw/discussions/2010#discussioncomment-8970823
  - https://github.com/mswjs/interceptors/pull/609
- [x] #2112 
- [x] https://github.com/mswjs/interceptors/issues/539
- [x] Forward client events to the server by default (https://github.com/mswjs/msw/discussions/2010#discussioncomment-9024640). 
- [x] Fix `.broadcast()` not broadcasting from N(1) -> N+ but working from N+ -> N+. (discovered in https://github.com/mswjs/examples/pull/111) (https://github.com/mswjs/msw/pull/2127). 
- [x] Annotate `this: WebSocket` in the `client`/`server` event listeners. 
- [x] Refreshing the page keeps adding the same client under a new ID to the `localStorage` (fix: https://github.com/mswjs/msw/pull/2133)

Screenshot 2024-04-12 at 13 09 35

- [x] Remove `console.log` (https://github.com/mswjs/msw/blob/f6922228f3ff47e059eb29480daa5dc6172316bd/src/core/ws/WebSocketClientManager.ts#L71)
- [x] Update the "[Client-to-server forwarding](https://website-git-docs-ws-mswjs.vercel.app/docs/basics/handling-websocket-events#client-to-server-forwarding-1)" docs. Both forwarding is _automatic_ now, and can be opt-out by calling `event.preventDefault()`. ([proof](https://github.com/mswjs/interceptors/blob/883a1bb37a63ad60a3de5f1639e917ec7e348066/src/interceptors/WebSocket/WebSocketServerConnection.ts#L43-L63))
- [x] Docs: document how runtime handlers are supposed to work (they are event-based) https://github.com/mswjs/msw/discussions/2010#discussioncomment-10315180
- [x] `WebSocketClientManager` still has the "first-tab-only" issue with multiple tabs.
- [ ] `wss://localhost:*` (`*` as port throws on `path-to-regexp`). Quite likely fixed by #1921 
- [x] Fix an issue where having multiple event handlers on the same page leads to the `DOMException: Key already exists in the object store.` error from IndexedDB (concurrent writes to the db).
- [x] Fix `(node:15076) MaxListenersExceededWarning: Possible EventTarget memory leak detected` in `pnpm test:unit src/core/ws/WebSocketClientManager.test.ts`


kettanaito:
## 📦 `msw@2.3.0-ws.rc-1`

🎉

Bram-Zijp:
I'm getting the following warning in jest (even without using any WebSockets)

```
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  MESSAGEPORT

    > 1 | import { http, HttpResponse } from "msw";
        | ^

      at Object.<anonymous> (../../node_modules/.pnpm/msw@2.3.0-ws.rc-1_typescript@5.2.2/node_modules/msw/src/core/ws/ws.ts:14:28)
      at Object.<anonymous> (../../node_modules/.pnpm/msw@2.3.0-ws.rc-1_typescript@5.2.2/node_modules/msw/src/core/index.ts:13:20)
```

kettanaito:
@Bram-Zijp, thanks for letting me know. This is a bug in Node.js (see https://github.com/nodejs/node/issues/52167#issuecomment-2011751095). This is unrelated to MSW.

kettanaito:
## Logging

Logging WebSocket connections and events is a bit more complex than I initially thought. Mostly, due to the nature of how you can handle them in MSW.

You can have multiple event handlers for the same connection. They can even send or modify outgoing/incoming events differently. Ideally, this is how logging should work:

1. You see a connection log for each handler that matched that connection. This way you know, okay, I have 2 handlers listening to the `ws://example.com` right now. 
2. Each handler must only log the events it sends to the client (`client.send()`). Tapping into `client.socket` is a bad idea because the same client/socket instance is shared across _all handlers_. 
3. For the original server connection, it only matters _once_. So it's enough to attach a spy on `server.connect` and add a `message` event listener _once_. 
4. It may be useful to log the `server.connect()` event to see which handler initiated the actual server connection.

kettanaito:
Intentionally pushed failing tests as verification of https://github.com/mswjs/msw/discussions/2010#discussioncomment-8894470. 

_Edit:_ Removed these. More context in the discussion.

kettanaito:
## 📦 `v2.3.0-ws.rc-2`

- Logging.
- New server and client events.

kettanaito:
## 📦 `v2.3.0-ws.rc-3`

- TypeScript improvements.

kettanaito:
## 📦 `v2.3.0-ws.rc-6`

- Uses `localStorage` for persisting all WebSocket clients (`.clients`).

kettanaito:
## 📦 `v2.3.0-ws.rc-7`

- Add `info` argument to the `connection` event (@mswjs/interceptors@0.30.0).

kettanaito:
Node tests on `HttpResponse.arrayBuffer()` are failing due to https://github.com/mswjs/msw/pull/2228.

kettanaito:
## Supporting `onUnhandledRequest`

I've decided for the unhandled WebSocket connections to also trigger the `onUnhandledRequest` calls. WebSocket connection is often preceded with an Upgrade handshake request, so it's technically an HTTP request first. You will see the URL, including the WebSocket protocol, giving you enough indication of what you are not handling. 

There's a difficulty in handling the `error` unhandled request strategy though. That strategy throws in MSW, and since MSW runs within `queueMicrotask(() => emitter.emit(here))`, for some reason, those exceptions are not caught by Interceptors (it's not the `queueMicrotask`/`process.nextTick` issue, I checked, and Interceptors already have internal error handling in the queued task). 

I think I'd have to expose some `.errorWith()` control utility to allow any layer to error the WebSocket connection (non-graraceful closure). That way, `onUnhandledRequest()` throws can be caught and translated to that closure:

```js
// msw
await onUnhandledRequest().catch((error) => {
  connection.client.errorWith(error)
})
```

Will have to think on this API.

kettanaito:
## 📦 `v2.3.0-ws.rc-8`

- Fixes the `WebSocketClientManager` not keeping track of cross-tabs clients (a refactor to use `IndexedDB` in the browser).

src/browser/setupWorker/setupWorker.ts
```diff
@@ -18,9 +18,13 @@ import { createFallbackStop } from './stop/createFallbackStop'
 import { devUtils } from '~/core/utils/internal/devUtils'
 import { SetupApi } from '~/core/SetupApi'
 import { mergeRight } from '~/core/utils/internal/mergeRight'
-import { LifeCycleEventsMap } from '~/core/sharedOptions'
+import type { LifeCycleEventsMap } from '~/core/sharedOptions'
+import type { WebSocketHandler } from '~/core/handlers/WebSocketHandler'
```
mknijnenberg, line 22: question: Do you want to bundle the handlers together or should these handler placed in the `~/core/ws` ?

kettanaito:
## 📦 `v2.3.0-ws.rc-9`

- Renames the `.on()` method on the WebSocket link to `.addEventListener`:
```diff
-api.on('connection', listener)
+api.addEventListener('connection', listener)
```
- Exports `WebSocketData` and `WebSocketEventListener` types from `msw`.

src/core/utils/handleRequest.ts
```diff
@@ -45,7 +44,7 @@ export interface HandleRequestOptions {
 export async function handleRequest(
   request: Request,
   requestId: string,
-  handlers: Array<RequestHandler>,
+  handlers: Array<unknown>,
```
mknijnenberg: Isn't this: `Array<RequestHandler | WebSocketHandler>`

src/core/ws/WebSocketClientManager.ts
```diff
@@ -0,0 +1,211 @@
...
+ * compatibility to the `WebSocketClientConnection` but relies
+ * on the given `BroadcastChannel` to communicate instructions
+ * with the client connections from other runtimes.
+ */
+export class WebSocketRemoteClientConnection
+  implements WebSocketClientConnectionProtocol
+{
+  constructor(
+    public readonly id: string,
+    public readonly url: URL,
```
mknijnenberg, line 187: question: Why is the `url` necessary? This is not used.

src/browser/setupWorker/setupWorker.ts
kettanaito: Sorry, I don't understand what you mean. Can you elaborate?

src/core/ws/WebSocketClientManager.ts
kettanaito: I've got this removed already. Likely forgot to push :)

src/core/utils/handleRequest.ts
kettanaito: Yes, this is a good point!

src/browser/setupWorker/setupWorker.ts
mknijnenberg: the related imports to WebSockets are coming from the `core/ws`. So i'm only suggesting that the `WebSocketHandler` should we imported from this path as well.

Something like this;
```
import type { WebSocketHandler } from '~/core/ws/handlers/WebSocketHandler'
...
import { webSocketInterceptor } from '~/core/ws/webSocketInterceptor'
import { handleWebSocketEvent } from '~/core/ws/handleWebSocketEvent'
import { attachWebSocketLogger } from '~/core/ws/utils/attachWebSocketLogger'
```

Or do you want to separate these functionalities and types?

kettanaito:
## 📦 `v2.3.0-ws.rc-10`

- Expose the server WebSocket instance on `server.socket`.
- Make the `close` server event cancelable.

kettanaito:
## 📦 `2.3.0-ws.rc-11`

- Simplified logging events.

kettanaito:
## 📦 `v2.3.0-ws.rc-12`

- Supports `event.stopPropagation()` and `event.stopImmediatePropagation()` for client and server events. You can now manage the event flow across different listeners and event handlers with better granularity.

kesupile:
> ## 📦 `v2.3.0-ws.rc-12`
> * Supports `event.stopPropagation()` and `event.stopImmediatePropagation()` for client and server events. You can now manage the event flow across different listeners and event handlers with better granularity.

This works really well for our needs; thanks for all the hard work! 

Is there a timeline for when it might be included in the main release branch? I noticed that none of this functionality is included in release `2.4.11`

kettanaito:
@kesupile, happy to hear that!

The WebSocket support is a release candidate right now. I'm planning on publishing it this year after I wrap up the remaining bits. You can use the release candidate for now, it will land on `main` as pretty much 1-1. I also try to keep this branch up-to-date to have the latest changes from `main` included in the release candidate. 

> Also, if you're using this at work, please consider [becoming MSW sponsor](https://github.com/sponsors/mswjs) 🙏 I could use your support to keep improving and maintaining the project. 

Let me know if you have any feedback while using the new API!

kettanaito:
## Released: v2.6.0 🎉

This has been released in v2.6.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.6.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.6.0)

---

AlexDroll:
Hi @kettanaito 
I fear that this PR introduced a breaking change in e.g. the `listHandlers()` method, since the return types changed (see [here](https://github.com/mswjs/msw/commit/ae786f59eb86b3d08c0fa5183ffca90e32af6448#diff-3245523a8f2195cdfa640858d62e07c49105d28a2d86f0190976205536826d0cR241)]. 
This is also not reflected yet in the documentation see for example [here](https://mswjs.io/docs/api/setup-server/list-handlers)

kettanaito:
Hi, @AlexDroll. Thanks for sharing this.

Adding a new type to the union isn't a breaking change. I will update the docs to mention that the method now returns all handlers. PR: https://github.com/mswjs/mswjs.io/pull/427

spawnrider:
Hi,

I just updated from MSW 2.3.5 to 2.6.5. and there is effectively a breaking change on the listHandlers() method. 
Now, you need to cast the handler object to access info attribute as below : 

```
for (const handler of handlers) {
  console.debug('Handler : ', (handler as RequestHandler).info.header);
}
```

Otherwise, you'll have this issue : 
```
error TS2339: Property 'info' does not exist on type 'RequestHandler<RequestHandlerDefaultInfo, any, any, RequestHandlerOptions> | WebSocketHandler'.
  Property 'info' does not exist on type 'WebSocketHandler'.

21       console.debug('Handler : ', handler.info.header);
```

pleunv:
You can avoid a cast by narrowing it:

```ts
worker.listHandlers().forEach(handler => {
  if ('isUsed' in handler) {
    // Narrowed to a `RequestHandler`
  } else {
    // Narrowed to a `WebSocketHandler`
  }
});
```
Or in a helper:

```tsx
function isRequestHandler(handler: RequestHandler | WebSocketHandler): handler is RequestHandler {
  return 'isUsed' in handler;
}
```

This works because `WebSocketHandler` has no `isUsed` field, at least at the moment.

kettanaito:
The answer from @pleunv is almost there. Don't rely on properties, check the class instance:

```js
import { RequestHandler } from 'msw'

const httpHandlers = worker.listHandlers().filter((handler) => {
  return handler instanceof RequestHandler
})

console.log(httpHandlers)
```

pleunv:
I generally avoid importing from `msw` because it pulls in the whole kitchen sink 🙂

spawnrider:
Could you update the documentation regarding this issue ?

kettanaito:
@pleunv, unless you are running a build-less app, I'd expect your bundler to tree-shake anything you don't use from `msw`. There may be exceptions (#2185), but those are on us and rather non-trivial to get right. 

@spawnrider, where do you think this update belongs best? It's hard to assume everyone needs to do the `instanceof` check. `.listHandlers()` is primarily intended as `console.log(worker.listHandlers())`. If you want more, we expect you to know what you are doing. 

That being said, I agree this needs to be mentioned. I can propose a new section to `worker.listHandlers()`/`server.listHandlers()` that shows how to list handlers only of a particular type (http/ws)? Wdyt?

spawnrider:
That's seems pretty good to me. 

I just managed to print the handlers list using this code (and not using a cast) : 
```
for (const handler of handlers) {
  if (handler instanceof RequestHandler) {
    console.debug('Handler : ', handler.info.header);
  }
}
```

kettanaito:
@spawnrider, please review https://github.com/mswjs/mswjs.io/pull/431.

Issue #1926, opened 2023-12-13
React Native Integration Issues


Keysamm:
Hi guys,

I have been trying to integrate **Mock Service Worker** into my application for several days. Next, I will describe the problems that arose, which I was able to fix and at which i have stuck.

First of all I followed setup provided here: https://mswjs.io/docs/integrations/react-native 

**_index.js:_**

```
// imports
async function enableMocking() {
  if (!__DEV__) {
    return;
  }

  await import('./msw.polyfills');
  const {server} = await import('./src/mocks/server');
  server.listen();
}

enableMocking().then(() => {
  AppRegistry.registerComponent(appName, () => App);
});
```

**_server.js_**

```
import {http, HttpResponse} from 'msw';
import {setupServer} from 'msw/native';

export const server = setupServer(
  http.post('http://localhost:3000/api/addData', () => {
    console.log('[ SERVER INTERCEPTOR ]');
    return HttpResponse.json({
      lastName: 'Steve',
      firstName: 'John',
    });
  }),
  http.get('http://localhost:3000/api/data', () => {
    return HttpResponse.json({
      id: '15d42a4d-1948-4de4-ba78-b8a893feaf45',
      firstName: 'John',
    });
  }),
);
```

Im providing errors from my console, as You can see its related to **_Class private methods_** , and missing some files with following path: _'/Users/distiller/react-native/packages/react-native/sdks/hermes/build_iphonesimulator/lib/InternalBytecode/InternalBytecode.js'_

<img width="854" alt="image" src="https://github.com/mswjs/msw/assets/62395388/0381feb2-561b-4e74-93da-675c8a79de9b">

After long research i changed my _index.js_ and _babel.config.js_ with following content: 

**index.js**

```
import {App} from '@app';
import {AppRegistry, Text, TextInput} from 'react-native';
import {name as appName} from './app.json';

if (__DEV__) {
  require('./msw.polyfills');

  const {server} = require('./src/mocks/server');
  server.listen();
}
AppRegistry.registerComponent(appName, () => App);
```

**babel.config.js**

```
plugins: [
      '@babel/plugin-transform-runtime',
      ['@babel/plugin-proposal-decorators', {legacy: true}],
      '@babel/plugin-transform-flow-strip-types',  // added this line
      ['@babel/plugin-transform-private-methods', {loose: true}],  // added this line
    ],
```

After this a new error occurred : 

[...]

PR #2017, opened 2024-02-06
fix: support intercepting fetch requests in React Native

sairus2k:
Switch to fetch interceptor to make MSW works in react-native. Relates to https://github.com/mswjs/interceptors/pull/503

Thanks @willdawsonme for [the solution](https://github.com/mswjs/msw/issues/1926#issuecomment-1905351237)

### DIFF
--- a/src/native/index.ts
+++ b/src/native/index.ts
@@ -1,4 +1,4 @@
-import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/XMLHttpRequest'
+import { FetchInterceptor } from '@mswjs/interceptors/fetch'
 import { RequestHandler } from '~/core/handlers/RequestHandler'
 import { SetupServerApi } from '../node/SetupServerApi'
 
@@ -13,5 +13,5 @@ export function setupServer(
 ): SetupServerApi {
   // Provision request interception via patching the `XMLHttpRequest` class only
   // in React Native. There is no `http`/`https` modules in that environment.
-  return new SetupServerApi([XMLHttpRequestInterceptor], ...handlers)
+  return new SetupServerApi([FetchInterceptor], ...handlers)
 }
--- a/src/native/index.ts
+++ b/src/native/index.ts
@@ -1,4 +1,5 @@
 import { FetchInterceptor } from '@mswjs/interceptors/fetch'
+import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/XMLHttpRequest'
 import { RequestHandler } from '~/core/handlers/RequestHandler'
 import { SetupServerApi } from '../node/SetupServerApi'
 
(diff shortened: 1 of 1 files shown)


src/native/index.ts
```diff
@@ -13,5 +13,5 @@ export function setupServer(
 ): SetupServerApi {
   // Provision request interception via patching the `XMLHttpRequest` class only
   // in React Native. There is no `http`/`https` modules in that environment.
-  return new SetupServerApi([XMLHttpRequestInterceptor], ...handlers)
+  return new SetupServerApi([FetchInterceptor], ...handlers)
```
kettanaito: We still need support XMLHttpRequest, do we not? Has React Native officially dropped XHR as the means to make requests? If so, I'd like to see that in their changelog. 

Otherwise, we better to _add_ `FetchInterceptor` to the list here.

kettanaito (changes requested):
One comment regarding dropping the XHR support in React Native.

src/native/index.ts
sairus2k: React native does not have its own implementation of fetch. [It uses whatwg-fetch](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Network/fetch.js) above their XMLHttpRequest implementation. Because we [fixed the error](https://github.com/mswjs/interceptors/pull/503) in fetch interceptor, xhr interceptor will still have the error. Is it OK to intercept both xhr and fetch if xhr will throw the error?

kettanaito: Can you point me to the place where XHR interceptor would still have the error? 

The problem as in accessing the `request.signal` that doesn't always exist. But XHR interceptor doesn't access that property, afaik. Happy to be proven wrong.

sairus2k: Never mind, I've just tried with both interceptors and it works

kettanaito: @sairus2k, still a question from my side: can I write this code in React Native without any explicit polyfills on my end:

```js
fetch('https://example.com')
```

sairus2k: Yes, fetch is the first class citizen in RN https://reactnative.dev/docs/network

kettanaito: Awesome! Then what you are proposing in this pull request definitely makes sense. Let's just _add_ the fetch interceptor to the list, keeping XHR.

kettanaito (approved):
Fantastic! 🚀 Yet another awesome contribution.

kettanaito:
## Released: v2.1.7 🎉

This has been released in v2.1.7!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.1.7)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.1.7)

---

PR #2028, opened 2024-02-09
Feat/ws sync handlers

chrisb2244:
Possible implementation for the feat/ws-sync-handlers branch.

A new dummyResponse is sent by the remote handler when the request is not handled by the remote handler - this allows the calling code (via `executeHandlers`) to identify that the request was not handled (even though it has been '(not-)handled' by the remote handler).

It's very possible I used the wrong commit tags/types, so am happy to rewrite them as required.

The remotePort is set to a static value here - guidance on how you'd prefer this to be set would be welcome (and I can rebase as needed before merging) - however, I wasn't sure the current intention because the `remote` flag mentioned in part of the upstream draft PR seems to have changed to the `remotePort` value, and its existence is used to determine remoteness (so defaulting seems like it would cause problems).
If the intent is for the user to specify a port (static or potentially via environment variable, but within their code) then I _think_ leaving a static value in the tests is perhaps(?) acceptable.

I'll add a comment to part of the changed code with further questions/thoughts on an alternative in case it isn't what/where you wanted changes.


src/node/SetupServerApi.ts
```diff
@@ -29,6 +34,10 @@ export type InterceptorsList = Array<{
   new (): Interceptor<HttpRequestEventMap>
 }>
 
+interface InteractiveRequest extends Request {
+  respondWith(response: Response): void
+}
```
chrisb2244: Here I created an interface to allow extracting a (private) function for the response. I couldn't find any exported type, but if you wanted to export from the interceptors package this could be removed and replaced with an import?

Alternatively, the private function can be removed and the code duplicated, or there may be alternatives for bypassing.

src/node/SetupServerApi.ts
```diff
@@ -63,13 +72,24 @@ export class SetupServerApi
...
-         * Playwright runs the app first (this part) and THEN it runs
-         * the tests (setupRemoteServer part). This causes the "/"
-         * request to pend forever on this line since the app has to
-         * load successfully before the tests run.
-         */
+        const remoteResolutionUrl = createWebSocketServerUrl(
+          this.resolvedOptions.remotePort,
+        )
+        // Build a pattern to match the socket.io connection
+        remoteResolutionUrl.pathname = '/socket.io'
```
chrisb2244: The pathname here depends on socket-io's default value, because MSW doesn't set the value in `opts.path` (see `setupRemoteServer.ts`, line 210, second arg). 
If you preferred, then that location could be changed to set the path, and then this wouldn't be dependent on their potential changes to default values.

src/node/SetupServerApi.ts
```diff
@@ -115,6 +129,27 @@ export class SetupServerApi
...
+      options,
+      emitter,
+    )
+
+    if (response) {
+      request.respondWith(response)
+    }
+    return
+  }
+
```
chrisb2244: I think this could be tidier, but I'm not sure how to best place/refactor it. It's used twice above, and requires `this` to be up-to-date for the handlers etc.
This requires a type for request, but could be changed to `any` if you preferred to avoid defining the interface (with comment above).

src/node/SetupServerApi.ts
chrisb2244: Removed in https://github.com/mswjs/msw/pull/2028/commits/64751b3184830834af988b61695b5cdf8a234479 since the same result can be achieved without the private function or duplication (early return produces desired behaviour).

src/node/SetupServerApi.ts
chrisb2244: Moved to L89 in SetupServerCommonApi.ts, resolving here and copying to new location (changed by merging v2.2.0).

src/node/SetupServerCommonApi.ts
```diff
@@ -0,0 +1,214 @@
...
+          this.resolvedOptions.remotePort,
+        )
+        // Build a pattern to match the socket.io connection
+        remoteResolutionUrl.pathname = '/socket.io'
+        const matcher = new RegExp(`${remoteResolutionUrl.href}/.*`)
+        const isRequestForSocket = matcher.test(request.url)
+        // Bypass handling if the request is for the socket.io connection
+        if (isRequestForSocket) {
+          return
+        }
```
chrisb2244: The pathname here depends on socket-io's default value, because MSW doesn't set the value in opts.path (see setupRemoteServer.ts, line 210, second arg).
If you preferred, then that location could be changed to set the path, and then this wouldn't be dependent on their potential changes to default values.

chrisb2244:
I'm not sure if retargeting this to main rather than feat/ws-sync-handlers will clear up a lot of the commits (most are from merging v2.2.0).

chrisb2244:
I'm not quite sure how to fix the CI checks here - it seems to be complaining about indentation in a file that I guess was changed in `main` between where the msw/feat/ws-sync-handlers branch started (which I used as the upstream branch to start) and now?

I tried resaving the file (which gave two extra spaces via prettier) but that didn't change the outcome of the action check as far as I can tell.

chrisb2244:
#2041 is a much cleaner (although arguably more fake) set of commits to the same destination, and may be easier to review/merge if desired.

PR #2039, opened 2024-02-13
style: remove unnecessary semicolon from IIFE in worker script

saul-atomrigs:
This PR removes the secomicolon in the yellow circle:

screenshot


kettanaito (commented):
Thanks for suggesting this!

Note that semicolon has been added automatically by Prettier. It's a common pattern of formatting IIFE to prevent them from joining with the preceding lines (since IIFE start with parenthesis and can be confused with a function call). 

I don't believe we need to fix anything here.

saul-atomrigs:
Wow I knew it was an IIFE, but didn't know the semicolon there was a common pattern 😅
Thanks for your feedback!

PR #2054, opened 2024-02-23
fix: add guard for missing context requests

### LINKED ISSUE #2053: Browser: Cannot read properties of undefined (reading 'url')
### Prerequisites

### Environment check

### Browsers

### Reproduction repository

### Reproduction steps

Run `yarn install`
Run `yarn start`

If server starts normally, kill the server and run `yarn start` again. Within ~5 server restarts, you run into the error.

It looks like the following line (line 21, createResponseListener.ts) is the problem:

`const request = context.requests.get(requestId)!`

Whereby `request` is undefined.

### Current behavior

`caught TypeError: Cannot read properties of undefined (reading 'url')
    at createResponseListener.ts:58:1
    at ServiceWorkerContainer.<anonymous> (setupWorker.ts:84:1)`

image

### Expected behavior

Server to start without errors or crashing every time.

lisatassone:
- Fixes https://github.com/mswjs/msw/issues/2053

The current code for `createResponseListener` assumes the `request` is always present in `context` but when `bypass` and `passthrough` functions are used with `fetch`, requests never get added to the `context`. Example [usage](https://mswjs.io/docs/recipes/response-patching/#proxying-requests) from the docs for a proxy request. 

This is the most straightforward fix that doesn't change the original logic, but something else may be desired if the intent of the code should be that those types of requests are added to context and handled specifically. Its unclear why they do not go through the normal flow if there are provisions to deal with these in `core/utils/handleRequest.ts` which is used in `createRequestListener`. 

The tests are TypeScript. If the non-null assertion was not there, the type system would have demanded undefined be handled.

Further, to ensure the Response URL is always set as intended in `createResponseListener`, perhaps the original request can be passed explicitly to the `RESPONSE` event to use when its not in context.

### DIFF
--- a/src/browser/setupWorker/start/createResponseListener.ts
+++ b/src/browser/setupWorker/start/createResponseListener.ts
@@ -15,9 +15,11 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
   ) => {
     const { payload: responseJson } = message
 
-    // Get the Request instance reference stored in the
-    // request listener for intercepted requests
-    // (bypass and passthrough requests will not be present here).
+    /**
+     * Get the Request instance reference stored in the
+     * request listener for intercepted requests
+     * (bypass and passthrough requests will not be present here).
+     */
     const { requestId } = responseJson
     const request = context.requests.get(requestId)
     context.requests.delete(requestId)
--- a/src/browser/setupWorker/start/createResponseListener.ts
+++ b/src/browser/setupWorker/start/createResponseListener.ts
@@ -16,9 +16,10 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
     const { payload: responseJson } = message
 
     // Get the Request instance reference stored in the
-    // request listener.
+    // request listener for intercepted requests
+    // (bypass and passthrough requests will not be present here).
     const { requestId } = responseJson
-    const request = context.requests.get(requestId)!
+    const request = context.requests.get(requestId)
     context.requests.delete(requestId)
 
     /**
@@ -53,7 +54,7 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
      * @see https://github.com/mswjs/msw/issues/2030
      * @see https://developer.mozilla.org/en-US/docs/Web/API/Response/url
      */
-    if (!response.url) {
+    if (!response.url && request) {
       Object.defineProperty(response, 'url', {
         value: request.url,
         enumerable: true,
--- a/src/core/sharedOptions.ts
+++ b/src/core/sharedOptions.ts
@@ -48,7 +48,7 @@ export type LifeCycleEventsMap = {
   'response:bypass': [
     args: {
       response: Response
-      request: Request
+      request: Request | undefined
       requestId: string
     },
   ]


src/core/sharedOptions.ts
```diff
@@ -48,7 +48,7 @@ export type LifeCycleEventsMap = {
   'response:bypass': [
     args: {
       response: Response
-      request: Request
+      request: Request | undefined
```
kettanaito, line 51: If that's the case, it's a bug. We shouldn't accommodate it. I will look into this once I have a moment. Thanks for bringing this to my attention!

ymarios-int: Hi,

What do we do until you fix this bug. Should we roll back to an older version?

I am on v2.2.2

src/browser/setupWorker/start/createResponseListener.ts
```diff
@@ -15,10 +15,13 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
...
-    // Get the Request instance reference stored in the
-    // request listener.
+    /**
+     * Get the Request instance reference stored in the
+     * request listener for intercepted requests
+     * (bypass and passthrough requests will not be present here).
+     */
     const { requestId } = responseJson
-    const request = context.requests.get(requestId)!
+    const request = context.requests.get(requestId)
```
cstephens-cni, line 24: The "non null assertion" going is definitely the right call.  it does need to be handled.

src/core/sharedOptions.ts
cstephens-cni: @ymarios-int this is a comment chain for a line of code.  You can ask this on the issue or commit.  I am not sure when or if this was introduced.  So you can look at the history to see what may have happened.   It's only been a few days so I would just wait a bit.  you can subscribe to the issue,

cstephens-cni: @ymarios-int looks to have been introduced in 2.1.5 so I just downgraded to "msw": "2.1.4" for now.  I am not sure if there will be other issue but I figured I'll share.

It was fixing a pretty important fixme it seems

https://github.com/mswjs/msw/commit/bad537f1fce7511f0a7258c0216a9ad49fbc27b8

thepassle:
I also ran into this issue :) Seems like this would fix it

kettanaito:
> The current code for createResponseListener assumes the request is always present in context but when bypass and passthrough functions are used with fetch, requests never get added to the context. Example [usage](https://mswjs.io/docs/recipes/response-patching/#proxying-requests) from the docs for a proxy request.

This is interesting. So the issue is when the response listener is called by the worker on _bypassed_ request, right? This needs an extremely careful examination, we don't want to jump to the wrong conclusion and introduce a wrong fix.

kettanaito:
Here's what I think happens now to a bypassed request (can be wrong, feel free to correct):

- It's caught by the worker, the worker sends the REQUEST message to the client;
- The request listener generates the request ID and sets that request in `context.requests`. This happens _before_ the request resolution logic that takes the bypassed state into account:

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/browser/setupWorker/start/createRequestListener.ts#L40

The resolution then happens after:

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/browser/setupWorker/start/createRequestListener.ts#L43

At this point, `context.request.get(requestId)` _must_ return the reference to that request object, even if the request has been bypassed. 

The bypassing logic itself lives in `handleRequest()`, which is called after the request has been set on the context:

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/core/utils/handleRequest.ts#L55-L60

**However**, I can see that the worker has its own logical branch for bypassing requests, and in that case it won't even send the REQUEST message to the client:

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/mockServiceWorker.js#L205-L210

I'm trying to remember why this duplication is necessary and my hunch tells me that `handleRequest()` is environment-agnostic. That `x-msw-intention` request header handling is there to support bypassed requests in Node.js. For the browser usage, it seems that the intention was to never go to the client since a bypassed request cannot be affected by MSW by design. 

What happens next is that the worker receives the original response and proceeds to message the client with the RESPONSE event:

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/mockServiceWorker.js#L116

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/mockServiceWorker.js#L125-L140

While this message will contain the `requestId`, the request listener has never been invoked on the client so `context.requests.has(requestId)` will be false. 

This means that bypassed requests cannot be persisted in the context while it's promised in the contract that even original responses will have the `request` object reference. 

> Further, to ensure the Response URL is always set as intended in createResponseListener, perhaps the original request can be passed explicitly to the RESPONSE event to use when its not in context.

The thing is, the worker doesn't know what's in the context. This suggestion would imply to always send the request alongside the RESPONSE event, which can be an overkill for the cases when the request has been stored in the context. Besides, we'd have to clone the request once in order to preserve its body so it'd be transferable in the RESPONSE event too. 

I'm considering the change that _all_ requests trigger the REQUEST event and the request listener, and then we decide whether to go into the request resolution or not, or let this handle the bypassed requests also to stay consistent across environments:

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/core/utils/handleRequest.ts#L55-L60

This proposal has an implication of involving an additional message roundtrip:

- Before: request -> worker -- is bypassed -> fetch() -> RESPONSE
- After: request -> worker -- is bypassed -> **REQUEST -> (client) NOT_FOUND/other ->** (worker) fetch() -> RESPONSE

Even with this considered, I think I like for the worker to consistently send REQUEST/RESPONSE events to the client, no matter the request type (bypassed/mocked). The worker should really only be used as the interception mechanism, with no logic that affects the request resolution. If we make it this way, it's precisely what it'll become. The surface that decides what to do with the request will always be the client, and it's the client's responsibility to take bypassed requests into account and send the worker the right instruction message on how to proceed. 

@thepassle @mattcosta7, what do you think about this, folks?

kettanaito:
## Proposal

Briefly summarizing my proposal from above.

### Step 0: Add a failing test for a bypassed request

We need to catch this issue in an automated test before even considering a fix. It sounds like it's a straightforward issue to solve. 

### Step 1: Remove the bypassed requests handling from the worker

This has to go:

https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/mockServiceWorker.js#L205-L210

### Step 2: Ensure that `handleRequest` will behave correctly with the bypassed scenarios in the browser. 

I believe that's the case already but verifying won't hurt.

mattcosta7:
> Even with this considered, I think I like for the worker to consistently send REQUEST/RESPONSE events to the client, no matter the request type (bypassed/mocked). The worker should really only be used as the interception mechanism, with no logic that affects the request resolution. If we make it this way, it's precisely what it'll become. The surface that decides what to do with the request will always be the client, and it's the client's responsibility to take bypassed requests into account and send the worker the right instruction message on how to proceed.

That makes sense to me, I can take a bit of time later to understand this a bit deeper, but on the surface that sounds like a good path, and the consistency between request handling sounds better than having multiple paths to doing that

lisatassone:
> > The current code for createResponseListener assumes the request is always present in context but when bypass and passthrough functions are used with fetch, requests never get added to the context. Example [usage](https://mswjs.io/docs/recipes/response-patching/#proxying-requests) from the docs for a proxy request.
> 
> This is interesting. So the issue is when the response listener is called by the worker on _bypassed_ request, right? This needs an extremely careful examination, we don't want to jump to the wrong conclusion and introduce a wrong fix.

This is correct and why I wanted to draw your attention to it. It felt like a larger decision needed to be made on whether the context should always contain bypassed requests - which would require a more extensive solution.

lisatassone:
> **However**, I can see that the worker has its own logical branch for bypassing requests, and in that case it won't even send the REQUEST message to the client:
> 
> https://github.com/mswjs/msw/blob/b3e47f71d3a0333aaeab3da5839cb9068a30d7ef/src/mockServiceWorker.js#L205-L210

> I'm trying to remember why this duplication is necessary and my hunch tells me that `handleRequest()` is environment-agnostic. That `x-msw-intention` request header handling is there to support bypassed requests in Node.js. For the browser usage, it seems that the intention was to never go to the client since a bypassed request cannot be affected by MSW by design.

This is what I understood as well and mentioned it in the original Issue tagged 👍 . Like you, I couldn't reason why that specific branch of logic was there and not in the dedicated request handlers.

> While this message will contain the `requestId`, the request listener has never been invoked on the client so `context.requests.has(requestId)` will be false.
> 
> This means that bypassed requests cannot be persisted in the context while it's promised in the contract that even original responses will have the `request` object reference.

Yes, this is what is causing the error to occur at the moment.

> I'm considering the change that _all_ requests trigger the REQUEST event and the request listener, and then we decide whether to go into the request resolution or not, or let this handle the bypassed requests also to stay consistent across environments.

Much prefer this solution!

kettanaito:
Merged a few preliminary changes to make this one easier:

- #2090 
- #2091 

Now we can move on with updating the worker script as I described above. @lisatassone, does this sound interesting to you?

kettanaito:
I'm tackling this issue in https://github.com/mswjs/msw/pull/2094. Also discovered that we are missing #2093. Nice find.

kettanaito:
## Released: v2.2.6 🎉

This has been released in v2.2.6!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.2.6)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.2.6)

---

lisatassone:
Nice one! You got through these quickly :D Thanks for this.

PR #2095, opened 2024-03-17
fix: resolve "~/core" import for .d.mts modules

kettanaito:
We use a custom script that resolves the `~/core` in the import paths in `.d.ts` files (since those won't resolve). But we miss to do that for `.d.mts` files, which leaves them broken. 

Also added two layers of protection to prevent this issue in the future:

- `grep` for any `~/core` in the resolved `.d.{ts,mts}` modules;
- `tsc` run on all the `.d.{ts,mts}` modules to make sure they compile.

### DIFF
--- a/config/replaceCoreImports.js
+++ b/config/replaceCoreImports.js
@@ -1,10 +1,17 @@
-function replaceCoreImports(fileContents, isEsm) {
-  const importPattern = isEsm
-    ? /from ["'](~\/core(.*))["'](;)?$/gm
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v2.2.7 🎉

This has been released in v2.2.7!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.2.7)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.2.7)

---

PR #2101, opened 2024-03-18
fix: properly infer parameter and return types in  `server.boundary()`

### LINKED ISSUE #2069: Infering the `boundary` callback arguments
### Scope

Improves an existing behavior

### Compatibility

### Feature description

Currently, when using the MSW boundaries, the callback provided to the boundary has no inferred types. 

This is a slight issue when using boundaries and concurrency, as when using concurrency in Vitest, [in certain cases you are required to use the local `expect` from the `context` instead of the imported `expect`](https://vitest.dev/guide/features#running-tests-concurrently).

Screenshot 2024-03-06 at 14 11 18

So I went digging a bit as to what the root cause of this is, and eventually I ran into [this issue on the TypeScript repository](https://github.com/microsoft/TypeScript/issues/50719). I believe this issue also describes the basic interaction between Vitest's `test` typings and MSW `boundary` typings.

I applied Daniel's suggested alternative to `boundary` locally

```diff
declare class SetupServerApi extends SetupServerCommonApi implements SetupServer {
    constructor(handlers: Array<RequestHandler>);
-   boundary<Fn extends (...args: Array<any>) => unknown>(callback: Fn): (...args: Parameters<Fn>) => ReturnType<Fn>;
+   boundary<Fn>(callback: Fn & Function): Fn & Function;
    close(): void;
}
```

and indeed the `boundary` callback's arguments became correctly inferred

Screenshot 2024-03-06 at 14 12 26

and I could use the expect from the context with inferred types for it.

Screenshot 2024-03-06 at 14 29 01

I'm unsure if this change would bring something undesireable or if there's something I missed. Correct me if I'm wrong, but I believe that the proposed change's outcome is an improvement as what the current types are trying to do anyway.

Andarist:
- Fixes https://github.com/mswjs/msw/issues/2069
- Closes https://github.com/mswjs/msw/pull/2100

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,15 @@ module.exports = {
     '@typescript-eslint/prefer-ts-expect-error': 2,
     '@typescript-eslint/no-explicit-any': 0,
     '@typescript-eslint/explicit-module-boundary-types': 0,
+    '@typescript-eslint/ban-types': [
+      'error',
+      {
+        types: {
+          Function: false,
+        },
+        extendDefaults: true,
+      },
+    ],
     '@typescript-eslint/ban-ts-comment': 0,
     '@typescript-eslint/no-namespace': [
       2,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,15 +19,6 @@ module.exports = {
     '@typescript-eslint/prefer-ts-expect-error': 2,
     '@typescript-eslint/no-explicit-any': 0,
     '@typescript-eslint/explicit-module-boundary-types': 0,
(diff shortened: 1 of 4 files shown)


kettanaito (approved):
Thank you so much, @Andarist ❤️

kettanaito:
## Released: v2.2.9 🎉

This has been released in v2.2.9!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.2.9)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.2.9)

---

PR #2107, opened 2024-03-25
fix: accept a narrower response body type by default

### LINKED ISSUE #2105: Narrowing the response body type in `HttpResponse.json`
### Scope

Improves an existing behavior

### Compatibility

### Reproduction steps
Versions:

"msw": "2.2.10"

### Feature description

Following the documentation of [Using with TypeScript](https://mswjs.io/docs/best-practices/typescript#higher-order-request-handlers), it works almost well.
But, it allows widing type of `HttpResponse.json`.

Here is an modified [Higher-order request handlers](https://mswjs.io/docs/best-practices/typescript#higher-order-request-handlers) example of adding non-existence `foo` field to `HttpResponse.json`.

```ts
import { http, HttpResponse, HttpResponseResolver } from 'msw';

type SdkRequest = {
  transactionId: string
}
 
type SdkResponse = {
  transactionId: string
  data: { ok: boolean }
}
 
function handleSdkRequest(
  resolver: HttpResponseResolver<never, SdkRequest, SdkResponse>
) {
  return http.post('https://some-sdk.com/internal/request', resolver)
}
 
export const handlers = [
  handleSdkRequest(async ({ request }) => {
    const data = await request.json()
    return HttpResponse.json({
      transactionId: data.transactionId,
      // `foo` should be error
      data: { ok: true, foo: 'bar' },
    })
  }),
]

### PR DESCRIPTION
**Pull request by kettanaito**
- Fixes #2105 

## Changes

- `HttpResponse.text()` and `HttpResponse.json()` _do not_ infer the given body type anymore. This allows them to correctly infer whichever response body type is set on the wrapping request handler. As a consequence, the response body type will now respect the narrower type provided to the handler:

```ts
http.get<never, never, { a: number }>('/', () => {
  // Error: Unknown property "b".
  return HttpResponse.json({ a: 1, b: 2 })
})
```

> Previously, this mocked response would satisfy the `{ a: number }` argument.

- When GraphQL handlers are used without explicit type arguments for the query/variables, those generic parameters become `never` even though they have fallback types. To prevent that, we are now explicitly checking for never: `GraphQLResponseBody<[Query] extends [never] ? GraphQLQuery : Query>`. This makes the GraphQL handlers compatible with the `NoInfer` changes to `HttpResponse.json()`. 
- Default GraphQL response body now supports `null | undefined`. It's still technically possible to respond to a GraphQL response with those. 
- Polishes a few type tests to take advantage of type asserting utilities instead of relying on `x.toUpperCase()` and such.

> [!WARNING]
> With the `NoInfer` changes to `HttpResponse.json()`, extra keys in the response body object _will cause type errors_. You have to explicitly allow (and describe) any extra keys your object may have:
>
> ```ts
> http.get<never, never, { [key: string]: string, a: number }>('/', () => {
>   return HttpResponse.json({ a: 1, b: 'hello' })
> })
> 
> I find this to be a good default because allowing extra keys is far easier than forbidding them in TypeScript. 

- As a nice side effect of this change, the type errors related to the response body are now displayed in the right place (now next to the actual properties at question; previously, next to the entire response resolver function):

```diff
   http.get<{ id: string }, never, 'hello'>(
     '/user/:id',
-    // Previously, errored here.
     withDelay(250, ({ params }) => {
       expectTypeOf(params).toEqualTypeOf<{ id: string }>()
-      return HttpResponse.text('non-matching')
+      return HttpResponse.text(
+        // Now, errors here.
+        'non-matching',
+      )
     }),
```

### DIFF
--- a/src/core/HttpResponse.ts
+++ b/src/core/HttpResponse.ts
@@ -1,4 +1,5 @@
 import type { DefaultBodyType, JsonBodyType } from './handlers/RequestHandler'
+import type { NoInfer } from './typeUtils'
 import {
   decorateResponse,
   normalizeResponseInit,
@@ -48,7 +49,7 @@ export class HttpResponse extends Response {
    * HttpResponse.text('Error', { status: 500 })
    */
   static text<BodyType extends string>(
-    body?: BodyType | null,
+    body?: NoInfer<BodyType> | null,
     init?: HttpResponseInit,
   ): StrictResponse<BodyType> {
     const responseInit = normalizeResponseInit(init)
@@ -77,7 +78,7 @@ export class HttpResponse extends Response {
    * HttpResponse.json({ error: 'Not Authorized' }, { status: 401 })
    */
   static json<BodyType extends JsonBodyType>(
-    body?: BodyType | null,
+    body?: NoInfer<BodyType> | null,
     init?: HttpResponseInit,
   ): StrictResponse<BodyType> {
     const responseInit = normalizeResponseInit(init)
--- a/src/core/graphql.ts
+++ b/src/core/graphql.ts
@@ -31,7 +31,10 @@ export type GraphQLRequestHandler = <
     | GraphQLHandlerNameSelector
     | DocumentNode
     | TypedDocumentNode<Query, Variables>,
-  resolver: GraphQLResponseResolver<Query, Variables>,
+  resolver: GraphQLResponseResolver<
+    [Query] extends [never] ? GraphQLQuery : Query,
+    Variables
+  >,
   options?: RequestHandlerOptions,
 ) => GraphQLHandler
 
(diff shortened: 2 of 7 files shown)


kettanaito:
## Released: v2.2.11 🎉

This has been released in v2.2.11!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.2.11)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.2.11)

---

VanTanev:
Something super tangential if someone else reads the code above:

Avoid `@ts-expect-error`, it's a very bad practice. Using `as any` is preferable in every situation. `@ts-expect-error` will destroy errors you actually care about (eg, mistyped method name, wrong number of arguments, etc etc).

PR #2108, opened 2024-03-25
fix: support async generators as response resolvers

### LINKED ISSUE #2109: "Cannot read properties of undefined (reading 'get')" when using generators/async generators as resolvers
### Prerequisites

### Environment check

### Node.js version

20.11

### Reproduction repository

https://github.com/juliusmarminge/msw-async-generator

### Reproduction steps

`bun i && bun run test`

### Current behavior

CleanShot 2024-03-25 at 13 01 38@2x

### Expected behavior

According to the docs here: https://mswjs.io/docs/recipes/polling async generator should be supported

kettanaito:
- Fixes #2109 
- Adds failing integration tests for using a generator (and async generator) function as response resolvers. 
- Supports async generators as response resolvers (generally, supports any iterators/async iterators!).
- Adds missing type tests for generators and async generators as response resolvers (needs #2107 to properly pass).


src/core/handlers/RequestHandler.ts
```diff
@@ -256,6 +266,9 @@ export abstract class RequestHandler<
       return null
     }
 
+    // Preemptively mark the handler as used.
+    // Generators will undo this because only when the resolver reaches the
+    // "done" state of the generator that it considers the handler used.
     this.isUsed = true
```
kettanaito, line 276: It's important to mark the handler as used immediately, even if using generators then opts-out from this behavior. This is what we promise right now, so let's keep that promise.

src/core/handlers/RequestHandler.ts
```diff
@@ -57,6 +56,11 @@ export type AsyncResponseResolverReturnType<
       MaybeAsyncResponseResolverReturnType<ResponseBodyType>,
       MaybeAsyncResponseResolverReturnType<ResponseBodyType>
     >
+  | AsyncGenerator<
```
kettanaito: Older versions of TypeScript (4.7 - 5.2) have trouble digesting this type:

```
 FAIL  resolver-generator.test-d.ts > supports async generator function as response resolver
TypeCheckError: Argument of type '() => AsyncGenerator<StrictResponse<JsonBodyType>, StrictResponse<JsonBodyType>, any>' is not assignable to parameter of type 'HttpResponseResolver<never, never, { value string; }>'.
  Type 'AsyncGenerator<StrictResponse<JsonBodyType>, StrictResponse<JsonBodyType>, any>' is not assignable to type 'AsyncResponseResolverReturnType<{ value string; }>'.
    Type 'AsyncGenerator<StrictResponse<JsonBodyType>, StrictResponse<JsonBodyType>, any>' is not assignable to type 'AsyncGenerator<ResponseResolverReturnType<{ value string; }>, ResponseResolverReturnType<{ value string; }>, ResponseResolverReturnType<{ value string; }>>'.
      The types returned by 'next(...)' are incompatible between these types.
        Type 'Promise<IteratorResult<StrictResponse<JsonBodyType>, StrictResponse<JsonBodyType>>>' is not assignable to type 'Promise<IteratorResult<ResponseResolverReturnType<{ value string; }>, ResponseResolverReturnType<{ value string; }>>>'.
          Type 'IteratorResult<StrictResponse<JsonBodyType>, StrictResponse<JsonBodyType>>' is not assignable to type 'IteratorResult<ResponseResolverReturnType<{ value string; }>, ResponseResolverReturnType<{ value string; }>>'.
            Type 'IteratorYieldResult<StrictResponse<JsonBodyType>>' is not assignable to type 'IteratorResult<ResponseResolverReturnType<{ value string; }>, ResponseResolverReturnType<{ value string; }>>'.
              Type 'IteratorYieldResult<StrictResponse<JsonBodyType>>' is not assignable to type 'IteratorYieldResult<ResponseResolverReturnType<{ value string; }>>'.
                Type 'StrictResponse<JsonBodyType>' is not assignable to type 'ResponseResolverReturnType<{ value string; }>'.
```

## Reproduction steps

- TypeScript 4.7
- `pnpm test:ts`

kettanaito: Hi, @Andarist! Sorry for troubling you once again. If you have a minute, could you please lend me a hand with this type issue?

Andarist: I would need to see a slimmed-down repro. The one that I tried put together in a rush errors in the current TS version too: [TS playground](https://www.typescriptlang.org/play?target=99&lib=lib.esnext.d.ts&ts=5.3.3#code/PTAEAEBsEsCMC5QFMDOA7JAPALgWAFDRrZIBOAZgIYDGSoASqgA4D2aKdA3gL4EHYBPJnQCylAbCQAFUiwC20DgB4AKgD5QAXlArQAH1Az5ipKrUBuPvgAmSapEqk61NimyhYLawJVCkiAFc0aABHALoUATlPSEt8AiISCho6AGVsUmhqbEYUVnZTACEvHz8NLBI0axQGZlcuAlBQJ0prNkgBUABtT29fYQBdRGK+vzjeePxBYVq8+tyWSAA3MkZsANI0ftNc-I4R0uENTUb9UAAKLt36g+2B5BwkKpquoNtyIiRrAdOmpoB+WZ7JC-P6IdKZbLXApKaH7ErbNQASlOBjeSA+GGsqNASxY0GscX4flAYgkSAAgpE0NQ4UgFstVkh1pttrC6gVbmUtKTxJIjAplKc6QyVqQ1hstn52XNOQiygQLFZpnQqQIaSLUIsxRLWdK6VyjjyyfzZILTDjNShtUyWVLhDLgYakGocQBxJ5kSjYFikJSgpomynU2kcjii22StkG+VHAA0Ad55LVGrD9K1jPFzKj+rTzrUCfwfz+QZTodl4YzOuzeodMdGR1BrvwSsmKqB8yrZBlNtIAFEcKRKABJNDkFhx2phVDYZ2T+uHF0886nIjjxAR-uDkdjicEJFaDRlq293X2nZ52MuuLtgAS2GwTBPmaUUkclDkKHnSGnbjnHblBslxOIsAMra0XzfIdP2-X9Zyvb8KyQfMiXwcggmyaA2FAEg3BXUCnAgsVEHvR9nzFJQMDFScqLISdOFxShIHCRA3EyNAAHNQG4At91AHhlRnc5KBDUB0JpbAsLQAAqC4Dx4JE4iwVhSHcHg4iAA)

kettanaito: Thanks for giving it a try, @Andarist. Here's a minimal repro:

```sh
pnpm install
pnpm build
pnpm test:ts test/typings/resolver-generator.test-d.ts
```

Andarist: That is not exactly minimal ;p it still depends on everything in the MSW's codebase. I could really use a self-contained TS playground here.

jakebailey: I stumbled on this and tried to bisect (via https://www.npmjs.com/package/every-ts) when this would have changed between 5.2 and 5.3.

What I discovered is that it's the literal version bump from 5.2 to 5.3; something in this package or its dependencies must be doing some typesVersions shenanigans that is causing this error.

jakebailey: Hmmm, no, I just flipped back and forth between the commits and now the 5.3 version change is failing. This makes no sense.

jakebailey: Skipping using `test:ts` and just doing regular compilation shows that https://github.com/microsoft/TypeScript/pull/51196 made this code legal. I'm not sure that there's really anything that can be done if you're hitting a limitation that was fixed.

kettanaito: Thank you so much for investigating this, @jakebailey! 

I'm glad to hear the team has fixed that (what a specific issue though!). Sadly, I don't expect they will backport the fix to all downstream versions that are not EOL. As CI reports, TS v5.3+ is fine with this change, but there's a huge chunk of versions MSW still supports before that. 

But one thing I'd like to clarify from your findings: is that a fundamentally TypeScript issue? Can we overcome it in any way on our end? Because if not, then this issue, transitively, also becomes a TS-caused issue, and we can only fix it once v5.3 becomes the minimal supported one.

jakebailey: We in general don't backport anything except critical regression fixes to the recent version (i.e. we're patching 5.5 right now).

I'm honestly not an expert in these types or their intracacies, so I can't really offer any other solution up for this... At best I could play around with the PR to attempt to make something pass, though I think anyone could probably do just as good as me in that area 😅

kettanaito: If you have the time, I'd be grateful! I gave it my best but my knowledge here ran short.

mattcosta7: > We in general don't backport anything except critical regression fixes to the recent version (i.e. we're patching 5.5 right now).
> 
> I'm honestly not an expert in these types or their intracacies, so I can't really offer any other solution up for this... At best I could play around with the PR to attempt to make something pass, though I think anyone could probably do just as good as me in that area 😅

I was looking into this a bit too - happy to have assistance if you have ideas - it's been an interesting one, and also not super familiar with these types

jakebailey: I got some time to take a look and believe I have a solution which actually improves things further.

Firstly, the code actually only depends on the existence of `Symbol.iterator` (or in this PR, `Symbol.asyncIterator`). This means that it's not actually working with generators at all, it's working with `Iterable`/`AsyncIterable` (and in fact shows that `isIterable` is incorrectly defined).

By switching the types over to use types defined with _just_ those two props, inference works again.

Fixing the implementation to reflect those types takes a little bit more code, but the upshot is that the final result allows users to pass in any `Iterable`/`AsyncIterable` they want, not just generators. I think this would actually allow people to pass in lists/sets of responses if they wanted, though I don't know if that's desirable or not.

How can I send this change to you? This is just a branch on this repo, so should I just PR it over? Here's a patch you could apply as well: https://gist.github.com/jakebailey/9702b2e0f16e2c9ce59799e40aefc7e7

kettanaito: @jakebailey, that is incredible! 😮 I would love to give this a try. 

Please, could you open a PR to _this feature branch_ (`fix/generator-resolver`)? Let me know if you need any help. 

What you are proposing sounds fantastic. I think that's the way to go. Need to see whether it complies with the other expectations toward the resolvers (especially the `isUsed` stuff).

jakebailey: Yes, I can do that.

jakebailey: FWIW I did compile and test locally, and it all seemed to work, though I did not acutally test the full range of TS versions (but I can't imagine it wouldn't work, maybe I'm jinxing myself)

kettanaito: Happy to give this a try once it's here 👀 Installing different `typescript` dependency versions triggers the `test:ts` into using that version for testing. The CI will also cover all the ranges for us. So excited we may actually ship this!

jakebailey: Sent https://github.com/mswjs/msw/pull/2213.

FWIW if you're looking for multi-version testing and aren't too tied to vitest, https://github.com/tstyche/tstyche is one option I've heard of which automatically runs multiple versions of TS without having to install them explicitly.

kettanaito: That's nice! Something to consider in the setup.

kettanaito:
The fix from @jakebailey (#2213) seems to fix the compatibility with older TypeScript versions! 🎉 Super job here. Any objections to merge this?

jakebailey:
This technically now enables passing in arrays, maps, etc; is that desirable? Should it be tested?

kettanaito:
@jakebailey as the return value of the resolver? It would be nice to guard against that on runtime at least.

jakebailey:
Actually, the old code may have already allowed that, at least for like .values()...

jakebailey:
It's probably fine as is, I'm just paranoid given I know nothing about this codebase!

kettanaito:
@jakebailey, well, now you know a little bit :) I'm infinitely grateful for your work on this. Looks like I know another TypeScript wizard now. 

I will give that false matching types scenario a try before releasing this.

kettanaito:
Yeah, @jakebailey, you're right. As long as the return type satisfies the Iterator type, there's no type error:

```ts
http.get('/user/:a/:b/:b', ({ params }) => {
  return []
  // ^ This compiles OK while it must error. 
})
```

jakebailey:
I guess the question is whether or not that's bad or not; I can just as easily pass in a generator which never yields today, right?

kettanaito:
That's certainly agains the established contract. Response resolvers either return/yield a `Response` instance or nothing. Nothing is an accepted value. But anything save for a `Response` instance is not. We don't have runtime guards against that now because we rely on type-level validation to prevent developers from returning non-Response stuff from their resolvers.

kettanaito:
This only seems to be a problem when returning an empty array `() => []`. Arrays are iterable, and they will even respect the resolver return type (response type) so that `() => [123]` errors, but an empty array is okay. Exploring if I can tweak the types so that an empty iterator is not allowed but that's equivalent of having a generator that doesn't yield value, which is still okay in MSW world. 

Huh.

kettanaito:
Oddly enough, I'm beginning to like the array stuff. Since arrays are iterable, one can return an array of mocked responses at once (no need for a generator). If that happens, MSW will exhaust that array as it does a generator function, producing mocked responses in order, and stopping on the last response to then use it forever. I wouldn't document this behavior but I like it!

kettanaito:
## Released: v2.3.3 🎉

This has been released in v2.3.3!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.3.3)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.3.3)

---

PR #2127, opened 2024-04-07
fix(WebSocketClientManager): use localStorage for clients persistence

kettanaito:
## Changes

We have a bug that a new runtime doesn't really know about any other runtimes. It only keeps itself in the list of clients and notifies _other_ runtimes about itself. That's not right.

BroadcastChannel is not a good API to introduce a shared state because one has to constantly keep that state in-sync. Instead, I propose we use `localStorage` to keep the serialized list of all the WebSocket clients for the app. 

- Fixes the failing test from https://github.com/mswjs/examples/pull/111

## Roadmap

- [x] Rewrite `src/core/ws/WebSocketClientManager.test.ts` tests, the constructor signature and the behavior of `WebSocketClientManager` have changed. 
- [x] Clear the localStorage key at some point.

### DIFF
--- a/src/core/ws/WebSocketClientManager.ts
+++ b/src/core/ws/WebSocketClientManager.ts
@@ -6,7 +6,7 @@ import type {
 } from '@mswjs/interceptors/WebSocket'
 import { matchRequestUrl, type Path } from '../utils/matching/matchRequestUrl'
 
-const MSW_WEBSOCKET_CLIENTS_KEY = 'msw:ws:clients'
+export const MSW_WEBSOCKET_CLIENTS_KEY = 'msw:ws:clients'
 
 export type WebSocketBroadcastChannelMessage =
   | {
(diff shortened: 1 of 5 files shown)


src/core/ws/WebSocketClientManager.ts
```diff
@@ -28,33 +25,122 @@ export type WebSocketBroadcastChannelMessage =
...
+  constructor(
+    private channel: BroadcastChannel,
+    private url: Path,
+  ) {
+    this.inMemoryClients = new Set()
+
+    if (typeof localStorage !== 'undefined') {
+      // When the worker clears the local storage key in "worker.stop()",
+      // also clear the in-memory clients map.
+      localStorage.removeItem = new Proxy(localStorage.removeItem, {
```
kettanaito, line 49: Ideally, we would know when `worker.stop()` is called but the manager lives in `ws.link()` while the stop event is only ever available in `handleWebSocketEvent()`. 

This would be an insane context drilling. Instead, let's proxy the `localStorage.removeItem()` calls.

PR #2140, opened 2024-04-28
feat: add `HttpResponse.html()` static method

scruffymongrel:
Add the ability to return html responses, enabling users to mock hypermedia APIs.

It mimics `HttpResponse.xml()`, except it creates a `Response` instance with the `Content-Type: text/html`.


kettanaito:
Hey, @scruffymongrel! Thanks for proposing this.

I will have to think about this for a while. The `.html()` method is certainly useful, I just don't know what's the stance regarding custom helper methods (more like, I have to define one).

scruffymongrel:
Thanks for taking a look. I'm not sure what you mean by "the stance regarding custom helper methods", but let me know if there's anything I can do to help :)

kettanaito:
@scruffymongrel can you please share your use case for this? When would you want to respond with HTML to a fetch request?

scruffymongrel:
Sorry for the delay in getting back to you -- this fell off my list and I just remembered I owed you a response!

I was originally intending to use MSW to work with [htmx](https://htmx.org), mocking responses which would be chucks of HTML. Unfortunately htmx uses XHR rather than fetch, so I'm guessing that this would be a bad fit. There was discussion around htmx moving to fetch in future versions but it doesn't seem to a priority at the moment.

If I've missed something and there is a way for MWS to work with XHR, I'd be keen to know more. Otherwise, feel free to close this PR.

kettanaito:
> Unfortunately htmx uses XHR rather than fetch, so I'm guessing that this would be a bad fit. 

Why? MSW is client-agnostic and works with XHR as well as it does with `fetch` or `http.request` (speaking of Node.js). You should certainly give it a try, you don't need anything special to make MSW work with XHR. If it doesn't, please submit a reproduction repository, I will take a look at what's wrong. 

I think `HttpResponse.html()` has a place to be. Will look at the PR, check the tests, and let's have this one merged.

kettanaito (approved):
Looks fantastic. Thank you for your work on this, @scruffymongrel. Marking it as a release candidate. Expect it to land in the next minor version. I will look at pending bugfixes, release those first, then this.

scruffymongrel:
> MSW is client-agnostic and works with XHR

Excellent! Was disappointed when I thought MSW was restricted to just fetch, so this is great news. Glad to hear that `HttpResponse.html()` is an appropriate feature for MSW and that I'll be able use it with htmx. Thanks for taking a look and for setting me straight on applicability.

kettanaito:
Documentation added in https://github.com/mswjs/mswjs.io/pull/415.

kettanaito:
## Released: v2.4.0 🎉

---

PR #2143, opened 2024-05-07
fix: prevent error when request not present

Codermar:
I get the the following error all the time in the console
```
createResponseListener.ts:58 Uncaught TypeError: Cannot read properties of undefined (reading 'url')
    at createResponseListener.ts:58:24
    at ServiceWorkerContainer.<anonymous> (setupWorker.ts:84:15)
```
This is because the request object maybe undefined in some cases

image

This PR is a simple fix for that.

### DIFF
--- a/src/browser/setupWorker/start/createResponseListener.ts
+++ b/src/browser/setupWorker/start/createResponseListener.ts
@@ -53,7 +53,7 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
      * @see https://github.com/mswjs/msw/issues/2030
      * @see https://developer.mozilla.org/en-US/docs/Web/API/Response/url
      */
-    if (!response.url) {
+    if (!response.url && request && request.url) {
       Object.defineProperty(response, 'url', {
         value: request.url,
         enumerable: true,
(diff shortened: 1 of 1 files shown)


kettanaito:
Hi, @Codermar 👋 Please take a look at this issue https://github.com/mswjs/msw/issues/2053.

What you are proposing is not a fix but a workaround. If the `request` is undefined, something terrible happened and we should let that surface. In other words, `request` being undefined is not a possible state, it's an unexpected state, thus we shouldn't handle it. 

I've made some changes to the worker script in how we communicate requests to the main thread, so request should not be undefined anymore. 

Do you have a reproduction repository for your issue? I'd love to take a look.

Codermar:
> Hi, @Codermar 👋 Please take a look at this issue #2053.
> 
> What you are proposing is not a fix but a workaround. If the `request` is undefined, something terrible happened and we should let that surface. In other words, `request` being undefined is not a possible state, it's an unexpected state, thus we shouldn't handle it.
> 
> I've made some changes to the worker script in how we communicate requests to the main thread, so request should not be undefined anymore.
> 
> Do you have a reproduction repository for your issue? I'd love to take a look.

Thanks for the response. I am not seeing the error anymore after the update of `msw": "^2.3.0"`, so far so your changes seem to be working. I Will update you if I run into it again.

Codermar:
Thanks for the response. I am not seeing the error anymore after the update of `msw": "^2.3.0"`, so far so your changes seem to be working. I Will update you if I run into it again.

PR #2144, opened 2024-05-08
fix(Disposable): make the ".dispose()" method synchronous

kettanaito:
Fixing an oversight. The `.dispose()` method of `Disposable` is not meant to be async. In fact, we aren't using it that way anywhere in MSW. Looks like a leftover from 1.x -> 2.x migration. 

To prevent confusion, rewriting that method to do the following:

1. Be sync.
2. Actually shift the subscription functions so you cannot fire them multiple times.

### DIFF
--- a/src/core/utils/internal/Disposable.ts
+++ b/src/core/utils/internal/Disposable.ts
@@ -1,9 +1,12 @@
-export type DisposableSubscription = () => Promise<void> | void
+export type DisposableSubscription = () => void
 
 export class Disposable {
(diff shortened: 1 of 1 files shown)


kettanaito:
## Released: v2.3.0 🎉

This has been released in v2.3.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.3.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.3.0)

---

PR #2145, opened 2024-05-08
test: add .use() replacement regression tests

### LINKED ISSUE #2129: Calls to worker.use() remove handlers to same endpoint with different method
I'm trying to write dynamic handler builders, but I'm hitting this issue:

```javascript
worker.use(
  http.get("/v1/issues", () => HttpResponse.json(/* some json */))
)

// this will make the get handler ignored!
worker.use(
  http.post("/v1/issues", () => HttpResponse.json(/* some json */))
);
```
I'm aware that repeated calls to `worker.use` prepend handlers to the existing context, and therefore order matters. But then why does this code work?

```javascript
// both handlers work
worker.use(
  http.get("/v1/issues", () => HttpResponse.json(/* some json */)),
  http.post("/v1/issues", () => HttpResponse.json(/* some json */))
);
```
Since both handlers work when added in a single `worker.use` call, shouldn't this method take into account any methods previously added to the context so as not to remove them and provide the same end result?

kettanaito:
- Closes #2129

### DIFF
--- a/test/browser/msw-api/regression/2129-worker-use.test.ts
+++ b/test/browser/msw-api/regression/2129-worker-use.test.ts
@@ -1,0 +1,17 @@
+/**
+ * @see https://github.com/mswjs/msw/issues/2129
+ */
+import { test, expect } from '../../playwright.extend'
+
+test('handles a stream response without throwing a timeout error', async ({
+  loadExample,
+  fetch,
+}) => {
+  await loadExample(require.resolve('./2129-worker-use.mocks.ts'))
+
+  const getResponse = await fetch('/v1/issues')
+  expect(await getResponse.text()).toBe('get-body')
+
+  const postResponse = await fetch('/v1/issues', { method: 'POST' })
+  expect(await postResponse.text()).toBe('post-body')
+})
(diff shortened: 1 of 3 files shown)


test/browser/msw-api/regression/2129-worker-use.test.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @see https://github.com/mswjs/msw/issues/2129
+ */
+import { test, expect } from '../../playwright.extend'
+
+test('handles a stream response without throwing a timeout error', async ({
```
kettanaito, line 6: Test for the worker.

test/node/regressions/2129-server-use.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+afterEach(() => {
+  server.resetHandlers()
+})
+
+afterAll(async () => {
+  server.close()
+})
+
+it('does not override existing handlers when adding override for a different method', async () => {
```
kettanaito, line 22: Test for the server.

kettanaito:
## Released: v2.3.1 🎉

This has been released in v2.3.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.3.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.3.1)

---

Issue #2168, opened 2024-05-24
Optional path parameters do not match correctly


GeorgeToka:
### Scope

Improves an existing behavior

### Compatibility

- [ ] This is a breaking change

### Feature description

There is a [step](https://github.com/mswjs/msw/blob/d97a528914df6a656dfe7b1f63985c85845df533/src/core/utils/url/cleanUrl.ts#L7-L12) that drops query parameters and it looks like it breaks the ability to specify optional URL parameters, e.g `/api/user/:id?`.

The cleanup routine should be limited to removing the query parameters only.

kettanaito:
Hi, @GeorgeToka. Thanks for reporting this!

The `cleanUrl` function you are referencing is applied to the _actual_ request URL. That one cannot have path parameter annotations, only the literal values of those parameters. 

But the issue you are describing is possible, the root cause may just be in a different place. We need an automated test to prove that.

kettanaito:
I've added an automated test for this in https://github.com/mswjs/msw/pull/2169, and it seems to be failing, which is good! More investigation required.

kettanaito:
## Released: v2.3.1 🎉

This has been released in v2.3.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.3.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.3.1)

Make sure to always update to the latest version (`npm i msw@latest`) to get the newest features and bug fixes.

---

_Predictable release automation by [@ossjs/release](https://github.com/ossjs/release)_.

GeorgeToka:
@kettanaito Thanks for the quick turnaround on this one.

PR #2169, opened 2024-05-25
fix: preserve trailing optional path parameters

kettanaito:
- Fixes #2168

PR #2170, opened 2024-05-25
fix: optional path parameter matching

KaiSpencer:
Updates logic and adds some more tests to ensure optional path parameters are matched correctly
fixes: #2168
to be merged with: #2169

### DIFF
--- a/src/core/utils/url/cleanUrl.test.ts
+++ b/src/core/utils/url/cleanUrl.test.ts
@@ -1,17 +1,22 @@
 import { cleanUrl } from './cleanUrl'
 
-test('removes query parameters from a URL string', () => {
+it('removes query parameters from a URL string', () => {
   expect(cleanUrl('/user?id=123')).toEqual('/user')
   expect(cleanUrl('/user?id=123&id=456')).toEqual('/user')
   expect(cleanUrl('/user?id=123&role=admin')).toEqual('/user')
 })
 
-test('removes hashes from a URL string', () => {
+it('removes hashes from a URL string', () => {
   expect(cleanUrl('/user#hash')).toEqual('/user')
   expect(cleanUrl('/user#hash-with-dashes')).toEqual('/user')
 })
 
-test('removes both query parameters and hashes from a URL string', () => {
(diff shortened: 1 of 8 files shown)


src/core/utils/url/cleanUrl.ts
```diff
@@ -5,8 +5,16 @@ export function getSearchParams(path: string) {
...
 
 /**
- * Removes query parameters and hashes from a given URL string.
+ * Removes search parameters and the fragment
+ * from a given URL string.
  */
 export function cleanUrl(path: string): string {
+  // If the path ends with an optional path parameter,
+  // return it as-is.
+  if (path.endsWith('?')) {
```
kettanaito, line 14: Instead of analyzing the last segment, I've added a check if the `path` ends with `?`. It will only end with a question mark in the case when it ends with an optional path parameter that must be preserved. 

If the path ends with a search string, it will always have value after `?`, so this check won't pass. 

@KaiSpencer, what do you think about this approach?

src/core/utils/url/cleanUrl.test.ts
```diff
@@ -1,17 +1,22 @@
...
   expect(cleanUrl('/user#hash-with-dashes')).toEqual('/user')
 })
 
-test('removes both query parameters and hashes from a URL string', () => {
+it('removes both query parameters and hashes from a URL string', () => {
   expect(cleanUrl('/user?id=123#some')).toEqual('/user')
   expect(cleanUrl('/user?id=123&role=admin#some')).toEqual('/user')
 })
+
+it('preserves optional path segments', () => {
```
kettanaito: Added the missing unit test for `cleanUrl`.

src/core/utils/url/cleanUrl.ts
KaiSpencer: Yes, This is a much cleaner approach 👍

kettanaito: Awesome! I will wait for the CI and merge this. Then, merge the main feature branch and have this released. Thank you for helping me with this!

kettanaito (approved):

PR #2187, opened 2024-06-21
feat: list "graphql" as an optional peer dependency

### LINKED ISSUE #2185: List "graphql" as a peer dependency
### Prerequisites

### Environment check

### Browsers

_No response_

### Reproduction repository

https://github.com/apollographql/graphql-testing-library

### Reproduction steps

When installing `msw`, `graphql` is installed as a transitive dependency at v16. This goes against the expectation in the GraphQL ecosystem that `graphql` always be treated as a peer dependency: https://github.com/graphql/graphql-js/issues/594#issuecomment-1189850052

From the [`graphql` source](https://github.com/graphql/graphql-js/blob/75dca3dd18e5be3031416965049fbde48a5adb20/src/jsutils/instanceOf.ts#L33-L42):

```
Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.
```

This cannot be trivially overridden with all package managers (e.g. npm overrides has [open issues](https://github.com/npm/cli/issues/4942)), and causes surprising behavior.

### Current behavior

`graphql` is a dependency of MSW.

Here's an example of a project using a number of packages that rely on `graphql`. All of them list it as a peer dependency with the range of 15 || 16, except for MSW.

CleanShot 2024-06-18 at 15 10 56

The result is my MSW GraphQL handler returning errors about unsupported GraphQL features for a version of GraphQL I haven't installed and don't expect to be executing my queries:
CleanShot 2024-06-18 at 15 12 34

### Expected behavior

I'd expect MSW to treat it as a peer dependency.

kettanaito:
- Fixes #2185 

## Changes

- Removes `graphql` as a regular dependency.
- Adds `graphql` as a peer dependency of MSW.
- Adds `graphql` to the `peerDependenciesMeta` as an _optional_ peer dependency (similar to `typescript`).


mattcosta7:
nice!  this is safe since we removed the line that parsed graphql to make suggested handlers, right? (I think we did, but can't recall exactly)

mattcosta7 (approved):

kettanaito:
@mattcosta7, thanks for getting your eyes on this. Yes, this should be safe! 

There's an effort to improve the unhandled requests logging but I advocated not to rely on GraphQl there, only to parse the body (#2227). I would love to include this pull request in the next minor release. For now, will see if there are any issues we should fix first...

kettanaito:
## Released: v2.4.0 🎉

---

Lalem001:
I am not sure this was a safe change. In my project, vitest is erroring out because it can no longer import `graphql`. I don't use `graphql` in my setup at all, so I was surprised to see the error.

```text
2:54:25 PM [vite] Internal server error: Failed to resolve import "graphql" from "node_modules/msw/lib/core/utils/internal/parseGraphQLRequest.mjs?v=5460f706". Does the file exist?
  Plugin: vite:import-analysis
  File: /project/node_modules/msw/lib/core/utils/internal/parseGraphQLRequest.mjs?v=5460f706:6:22
  1  |  import { parse } from "graphql";
```

I do not know why `parseGraphQLRequest.mjs` is being imported.

Installing `graphql` on my own resolves the issue, but essentially makes it a required dependency in my setup. Not optional. Having looked at the original issue, I see why it should be a peer dep of MSW, but it still broke at least one setup.

Greg-Smulko:
It fails for us as well:
> node_modules/msw/lib/core/GraphQLHandler-Cu4Xvg4S.d.ts(1,63): error TS2307: Cannot find module 'graphql' or its corresponding type declarations.
> node_modules/msw/lib/core/graphql.d.ts(1,30): error TS2307: Cannot find module 'graphql' or its corresponding type declarations.

We also don't use graphql at all.

topaxi:
Same here, imported in the node process, maybe via bundler and tree-shaking this might not be an issue. But imports via node crash due to missing modules.

kettanaito:
Thanks for reporting this! For now, you can revert to the previous minor version to not deal with this error.

I believe `graphql` gets imported if you are using a single import:

```js
import { http } from 'msw'
```

This imports `graphql` because not all tooling you use does tree-shaking or even bundling. 

One way to fix this is to use granular imports, which we've shipped some time back:

```js
import { http } from 'msw/core/http'
```

Let me know if this helps!

Akallabet:
@kettanaito thanks for looking into it.
In my case this is a pretty big deal as this change is blocking our CI, we use msw in most of our tests.
I appreciate there's a potential workaround using granular imports, but it will require changes in quite few repositories.
Would it be possible to just revert this change?

kettanaito:
My comment above mentions **you should downgrade if this affects you**. Please do. That's handling open source issues 101 ;) 

The fix is already open at #2250. I will release it once the build is complete.

PR #2190, opened 2024-06-24
fix: support typescript@5.5 (deprecate v4.7)

KaiSpencer:
Build and tests pass locally for me 👌


.github/workflows/compat.yml
```diff
@@ -52,7 +52,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ts: ['4.7', '4.8', '4.9', '5.0', '5.1', '5.2', '5.3', '5.4']
+        ts: ['4.7', '4.8', '4.9', '5.0', '5.1', '5.2', '5.3', '5.4', '5.5']
```
kettanaito: As a rule of thumb, adding a new version to the matrix means dropping the left-most one (4.7). Let me check if 4.7 is EOL... 
image

Yeah, 4.7 is EOL now. @KaiSpencer, could you please drop 4.7 from the matrix? Thanks!

test/typings/tsconfig.5.5.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.4.7.json",
```
kettanaito: Once we drop 4.7, we need to adjust the name of this base config. Let's see if it's specific to 4.7 in any way. Maybe it's better off as `tsconfig.base.json` instead.

KaiSpencer:
Hey @kettanaito,

Thanks for the review.

I have updated the tsconfig files to all inherit from the typings `tsconfig.json` (which the old 4.7.json was extending itself).

I see we also don't have a file for 4.9, yet it is in the matrix. So I have added this file too.

test/typings/tsconfig.5.2.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "./tsconfig.4.7.json",
+  "extends": "./tsconfig.json",
```
kettanaito, line 2: Let's tidy this up a bit. The implementation is that if there's no designated config for a TS version it grabs the default one. I did it so there be no repetitive configs.

kettanaito (approved):

kettanaito:
I treat this as a fix because TS 4.7 is EOL, nobody should be using it. If they are, it doesn't mean MSW will break, just means we won't address any type issues on that version going forward.

kettanaito:
## Released: v2.3.2 🎉

This has been released in v2.3.2!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.3.2)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.3.2)

---

PR #2196, opened 2024-07-03
fix(ws): call BroadcastChannel.unref in Node.js environment

DanielleHuisman:
When using the WebSocket preview, the Node.js process doesn't exit properly. It does not matter if you actually create WebSocket links. This is a minimal reproduction:

```js
import { http, passthrough } from "msw";

http.get("/", () => passthrough());
```

I added `why-is-node-running` to the script:
```js
import why from "why-is-node-running";
import { http, passthrough } from "msw";

http.get("/", () => passthrough());

setTimeout(why, 1000);
```

This shows there is a `MessagePort` keeping the process running:
```
# MESSAGEPORT
node:internal/async_hooks:202                                                              
node:internal/worker/io:341                                                                
file://./node_modules/msw/lib/core/ws.mjs:8
node:internal/modules/esm/module_job:262                                                   
node:internal/modules/esm/loader:474                                                       
node:internal/modules/run_main:109 
```

It traces back to [this line](https://github.com/mswjs/msw/blob/feat/ws/src/core/ws.ts#L14):
```js
const wsBroadcastChannel = new BroadcastChannel("msw:ws-client-manager");
```

This PR fixes the problem by calling `.unref()` in Node.js environments ([docs](https://nodejs.org/api/worker_threads.html#broadcastchannelunref)). The typing is a bit ugly, but `unref` is only available in the type when importing from `node:working_threads` and not when using the global variable. The browser version of `BroadcastChannel` does not have an `unref` function, so the call has to be conditional.

### DIFF
--- a/src/core/ws.ts
+++ b/src/core/ws.ts
@@ -11,7 +11,19 @@ import {
 import { Path, isPath } from './utils/matching/matchRequestUrl'
 import { WebSocketClientManager } from './ws/WebSocketClientManager'
 
+function isBroadcastChannelWithUnref(
+  channel: BroadcastChannel,
+): channel is BroadcastChannel & NodeJS.RefCounted {
+  return typeof Reflect.get(channel, 'unref') !== 'undefined'
+}
+
 const wsBroadcastChannel = new BroadcastChannel('msw:ws-client-manager')
+
+if (isBroadcastChannelWithUnref(wsBroadcastChannel)) {
+  // Allows the Node.js thread to exit if it is the only active handle in the event system.
+  // https://nodejs.org/api/worker_threads.html#broadcastchannelunref
+  wsBroadcastChannel.unref()
+}
 
 export type WebSocketLink = {
   /**


src/core/ws.ts
```diff
@@ -13,6 +13,15 @@ import { WebSocketClientManager } from './ws/WebSocketClientManager'
 
 const wsBroadcastChannel = new BroadcastChannel('msw:ws-client-manager')
 
+// Allows the Node.js thread to exit if it is the only active handle in the event system.
+// https://nodejs.org/api/worker_threads.html#broadcastchannelunref
+if ('unref' in wsBroadcastChannel) {
+  const nodeBroadcastChannel = wsBroadcastChannel as BroadcastChannel & {
```
kettanaito: You can achieve a better typing and overall usage experience by creating a custom predicate function for `unref`:

```ts
function isBroadcastChannelWithUnrefMethod(channel: BroadcastChannel): channel is BroadcastChannel & { unref(): void } {
  return typeof Reflect.get(channel, 'unref') !== 'undefined'
}

const channel = new BroadcastChannel('')
if (isBroadcastChannelWithUnrefMethod(channel)) {
  channel.unref()
}
```
> Double-check the actual call signature of `unref()` in Node.js to make sure we are annotating it correctly in the predicate function.

kettanaito (changes requested):
Hey, @DanielleHuisman. Thank you for fixing this! It looks great. 

I've left one suggestion around making the `unref` check a bit more friendly. Let me know what you think.

src/core/ws.ts
DanielleHuisman: Nice suggestion, I implemented it.

kettanaito (approved):
This is pure gold. Thank you so much for fixing this, @DanielleHuisman 👏 You've just made WebSockets much better in MSW.

PR #2205, opened 2024-07-12
test: add isolated response test for iframes

kettanaito:
Adds an automated test for the isolated response scenario when using multiple iframes on the page. 

## How does this work?

Each iframe becomes the outgoing request's `referrer`. If the iframe provides a check in the request handlers to ignore a matching request if it has a `referrer` different from the current iframe's `location.href`, it will allow two iframes to receive different responses to the same request.

### DIFF
--- a/test/browser/msw-api/setup-worker/scenarios/iframe-isolated-response/one.html
+++ b/test/browser/msw-api/setup-worker/scenarios/iframe-isolated-response/one.html
@@ -1,0 +1,34 @@
+<script>
+  const { worker, http } = window.parent.msw
+
+  function isolatedResolver(resolver) {
+    return (info) => {
+      if (info.request.referrer !== location.href) {
+        return
+      }
+      return resolver(info)
+    }
+  }
+
+  worker.use(
+    http.get(
+      './resource',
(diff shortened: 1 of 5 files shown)


test/browser/msw-api/setup-worker/scenarios/iframe-isolated-response/one.html
```diff
@@ -0,0 +1,34 @@
+<script>
+  const { worker, http } = window.parent.msw
+
+  function isolatedResolver(resolver) {
```
kettanaito, line 4: @yannbf, hi 👋 Let's try using this utility in the Storybook add-on and see if it provides us with the context isolation when rendering multiple stories on the same page (in iframes). I hope it does! 

The important part here is so this utility runs in the iframe's context, not in the parent context.

kettanaito:
## Released: v2.3.2 🎉

This has been released in v2.3.2!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.3.2)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.3.2)

---

PR #980, opened 2021-11-11
Decode path components in "req.params"

VanTanev:
Closes #976

PR #2209, opened 2024-07-18
feat!: use `URLPattern` for request matching

### LINKED ISSUE #1921: Support URLPattern as request predicate

### LINKED ISSUE #2204: Using the wildcard token to try to match random ports doesn't work if preceded by a colon
### Prerequisites

### Environment check

### Node.js version

v20.15.1

### Reproduction repository

https://github.com/silvenon/repro/tree/msw-wildcard-port

### Reproduction steps

The reproduction branch contains a project with two test, both of them are passing, but the the first one is set to fail using [`test.fails`](https://vitest.dev/api/#test-fails).

In a situation where you have to match an unknown port, atttempt to match it in MSW with something like this:

```ts
import { http, HttpResponse } from "msw";

const server = setupServer(
  http.get("http://localhost:*", () => HttpResponse.text("Hello world!")),
);

server.listen();
```

Then attempt to fetch something from `localhost` endpoints with random ports:

```ts
const res = await fetch("http://localhost:3000");
const result = await res.text();
```

### Current behavior

Currently the library for matching the endpoint, `path-to-regexp`, seems to detect this as the parameter:

```
{"name":"TypeError","message":"Missing parameter name at 17","stack":"TypeError: Missing parameter name at 17
    at lexer (/Users/silvenon/Code/repro/node_modules/path-to-regexp/src/index.ts:72:24)
    at parse (/Users/silvenon/Code/repro/node_modules/path-to-regexp/src/index.ts:141:18)
    at stringToRegexp (/Users/silvenon/Code/repro/node_modules/path-to-regexp/src/index.ts:493:25)
    at pathToRegexp (/Users/silvenon/Code/repro/node_modules/path-to-regexp/src/index.ts:620:10)
    at match (/Users/silvenon/Code/repro/node_modules/path-to-regexp/src/index.ts:379:14)
    at matchRequestUrl (file:///Users/silvenon/Code/repro/node_modules/msw/src/core/utils/matching/matchRequestUrl.ts:66:18)
    at HttpHandler.parse (file:///Users/silvenon/Code/repro/node_modules/msw/src/core/handlers/HttpHandler.ts:114:19)
    at HttpHandler.run (file:///Users/silvenon/Code/repro/node_modules/msw/src/core/handlers/RequestHandler.ts:239:37)
    at executeHandlers (file:///Users/silvenon/Code/repro/node_modules/msw/src/core/utils/executeHandlers.ts:36:28)
    at file:///Users/silvenon/Code/repro/node_modules/msw/src/core/utils/handleRequest.ts:64:12"}
```

If I remove the colon it acts as expected.

### Expected behavior

Regardless of whether the wildcard is used with or without colon it should achieve a similar result:

- `localhost:*`
- `localhost*`

I.e. for this use case it should work. Except that in the latter pattern I would expect the hostname to not only match `localhost`, but also `localhostable` etc., so I consider it to be a workaround rather than a solution.

kettanaito:
> [!IMPORTANT]
> **Please consider [some of `URLPattern` flaws](https://adventures.nodeland.dev/archive/you-should-not-use-urlpattern-to-route-http/) before introducing it to MSW**. We might also consider [`@remix-run/route-pattern`](https://github.com/remix-run/remix/tree/8b557a421fe1aebaf7cb9903f96f20895bb7dd99/packages/route-pattern) instead as it's backed into Remix 3, provides great feature set, doesn't have the `URLPattern` shenanigans and overall looks extremely promising. 

> [!WARNING]
> As much as I hate this, this _is_ a breaking change. Some of your existing handlers may start matching differently if they contain one of the dropped features (path params merging or trailing slashes). Even if this is a change for the better, this should land in v3.0. 

- Closes #1921
- Fixes #2204

This also significantly improves the test coverage for path matching in MSW. 

This change **does not intend to support `URLPattern` as the input**. For now, we will use it as internal mechanism (thus a polyfill version). Once the API gains global browser support, we may consider shipping an input support. 

## Changes to document

- Multiple named groups are not supported: `/:a/foo/:a`. Not a thing in the HTTP spec. They are supported by `path-to-regexp` but `URLPattern` decided not to implement them and stick closer to the spec.
- Trailing slashes matter now. A `/foo` handler will not match a `/foo/` request, and vice versa. I think MSW should respect that `URLPattern` behavior too to bring wider adoption of the API and be spec-compliant. 

## Roadmap

- [x] `/foo/:a/bar/:a` syntax throws in `URLPattern`. https://github.com/whatwg/urlpattern/issues/226. Doesn't throw in `path-to-regexp` and current MSW. Don't want to make a breaking change out of this. 
  - Dropping this promise. 
- [x] Support encoding since `URLPattern` doesn't. 
  - Tests introduced in #980 are passing as-is.


kettanaito:
This failing test:

```
 FAIL  src/core/utils/matching/matchRequestUrl.test.ts > merges multiple same-named groups into an array of values
TypeError: Failed to construct 'URLPattern': invalid pathname pattern '/user/:segment/bar/:segment'.
 ❯ new me node_modules/.pnpm/urlpattern-polyfill@9.0.0/node_modules/urlpattern-polyfill/dist/urlpattern.js:1:14389
 ❯ Module.matchRequestUrl src/core/utils/matching/matchRequestUrl.ts:96:19
```

Is caused by https://github.com/whatwg/urlpattern/issues/226.

kettanaito:
## Conclusion

Considering an abysmal performance if URLPattern, we are not going to move forward with adopting it in MSW. The next major version will rely on an in-house matching library like we used to do (see https://github.com/mswjs/msw/pull/2678). 

If you want to use URLPattern for request matching, you can do so by providing a function as the handler predicate and using that API there:

```ts
http.get(({ request }) => {
  return new URLPattern({ pathname: '...' }).test(request.url)
}, resolver)
```

PR #2213, opened 2024-07-21
fix: use iterables instead of generators

jakebailey:
This applies https://github.com/mswjs/msw/pull/2108#discussion_r1685551423 to the `fix/generator-resolver` branch.

Best viewed without whitespace: https://github.com/mswjs/msw/pull/2213/files?w=1

### DIFF
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -1,5 +1,9 @@
 import { getCallFrame } from '../utils/internal/getCallFrame'
-import { isIterable } from '../utils/internal/isIterable'
+import {
+  AsyncIterable,
+  Iterable,
+  isIterable,
+} from '../utils/internal/isIterable'
 import type { ResponseResolutionContext } from '../utils/executeHandlers'
 import type { MaybePromise } from '../typeUtils'
 import { StrictRequest, StrictResponse } from '..//HttpResponse'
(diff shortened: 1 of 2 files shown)


src/core/handlers/RequestHandler.ts
```diff
@@ -314,37 +318,38 @@ export abstract class RequestHandler<
...
-        this.isUsed = false
-
-        if (!this.resolverGenerator) {
-          this.resolverGenerator = result
+      if (!this.resolverGenerator) {
+        const result = await resolver(info)
+        if (!isIterable(result)) {
+          return result
         }
+        this.resolverGenerator =
```
jakebailey: Technically, this isn't a generator anymore, so the variable name is a lie.

src/core/handlers/RequestHandler.ts
```diff
@@ -314,37 +318,38 @@ export abstract class RequestHandler<
...
+          Symbol.iterator in result
```
jakebailey, line 327: `Reflect.has` (used in `isIterable`) is the same operation as `in`, except that TS can acutally narrow using `in`.

src/core/handlers/RequestHandler.ts
kettanaito: Is something like `resolverIterator` a more fitting name? These are internal so we can rename them to our hearts' content.

src/core/utils/internal/isIterable.ts
```diff
@@ -1,11 +1,20 @@
+// Redefine these types; the three-parameter forms are only available in TypeScript 5.6 or later.
```
kettanaito: Thank you for putting this in! Exactly what I was wondering about.

kettanaito:
😱 

Screenshot 2024-07-21 at 19 35 24

src/core/utils/internal/isIterable.ts
jakebailey: I can also make these JSDoc comments so they're carried down; whatever you'd prefer.

kettanaito: I'd mark them as `@todo` to remove these types once TS 5.6 is the minimal supported version. Not going to happen any time soon but still.

jakebailey: Updated.

kettanaito (approved):
This looks incredible! 🎉 Thank you so much, @jakebailey. You've made the impossible possible. Local tests are passing, type-compatibility on CI is green.

PR #2227, opened 2024-07-30
feat: print request body in `onUnhandledRequest` message

### LINKED ISSUE #2161: Include GraphQL operation name in the unhandled request warning
### Scope

Adds a new behavior

### Compatibility

### Feature description

Hello! Previously, in unhandled graphql request errors, you could see the operationName of the query, but at the moment this is not possible. Could you provide an example or a function that would fix this?
image

bitttttten:
- Closes #2161

I came to file a feature request very similar https://github.com/mswjs/msw/issues/2161 and thought I'd open a PR.

This will hopefully change this message:

```sh
[MSW] Warning: intercepted a request without a matching request handler:

  • POST https://graphql.staging.int.creativefabrica.com/query

Read more: https://mswjs.io/docs/getting-started/mocks
```

into

```sh
[MSW] Warning: intercepted a request without a matching request handler:

  • POST https://graphql.staging.int.creativefabrica.com/query

  • Request body: { "query": "{ user { id name } }" }

Read more: https://mswjs.io/docs/getting-started/mocks
```

### DIFF
--- a/src/core/utils/request/onUnhandledRequest.test.ts
+++ b/src/core/utils/request/onUnhandledRequest.test.ts
@@ -64,11 +64,9 @@ test('supports the "warn" request strategy with request body', async () => {
   await onUnhandledRequest(
     new Request(new URL('http://localhost/api'), {
       method: 'POST',
-      credentials: 'include',
       headers: {
         'Content-Type': 'application/json',
       },
-      cache: 'no-store',
       body: JSON.stringify({
         variables: {
           id: 'abc-123',
--- a/src/core/utils/request/onUnhandledRequest.test.ts
+++ b/src/core/utils/request/onUnhandledRequest.test.ts
@@ -14,7 +14,7 @@ const fixtures = {
 
 If you still wish to intercept this unhandled request, please create a request handler for it.
 Read more: https://mswjs.io/docs/getting-started/mocks`,
-  warningWithResonseBody: (url = `/api`) => `\
+  warningWithResponseBody: (url = `/api`) => `\
 [MSW] Warning: intercepted a request without a matching request handler:
 
   • POST ${url}
@@ -78,7 +78,7 @@ test('supports the "warn" request strategy with request body', async () => {
     }),
   )
 
-  expect(console.warn).toHaveBeenCalledWith(fixtures.warningWithResonseBody())
+  expect(console.warn).toHaveBeenCalledWith(fixtures.warningWithResponseBody())
 })
 
 test('supports the "error" request strategy', async () => {
--- a/src/core/utils/request/onUnhandledRequest.test.ts
+++ b/src/core/utils/request/onUnhandledRequest.test.ts
@@ -19,7 +19,7 @@ Read more: https://mswjs.io/docs/getting-started/mocks`,
 
   • POST ${url}
 
-  • Request body: {\"variables\":{\"id\":\"abc-123\"},\"query\":\"query UserName($id: String!) { user(id: $id) { name } } }\"}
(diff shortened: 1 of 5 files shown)


bitttttten:
Happy to take suggestions for the formatting of the message. Played with a few formats although settled on this one 🫡

src/core/utils/request/onUnhandledRequest.ts
```diff
@@ -24,7 +24,9 @@ export async function onUnhandledRequest(
   const url = new URL(request.url)
   const publicUrl = toPublicUrl(url) + url.search
 
-  const unhandledRequestMessage = `intercepted a request without a matching request handler:\n\n  \u2022 ${request.method} ${publicUrl}\n\nIf you still wish to intercept this unhandled request, please create a request handler for it.\nRead more: https://mswjs.io/docs/getting-started/mocks`
+  const requestBody = await request.text()
```
bitttttten: I also didn't know if I'd have to clone this request or not.

i.e. is the text/body read somewhere else already? Would we need to do `const requestBody = await request.clone().text()`?

kettanaito: Yeah, when in doubt-always clone the request/response!

src/core/utils/request/onUnhandledRequest.test.ts
```diff
@@ -51,6 +60,27 @@ test('supports the "warn" request strategy', async () => {
...
 
+test('supports the "warn" request strategy with request body', async () => {
+  await onUnhandledRequest(
+    new Request(new URL('http://localhost/api'), {
+      method: 'POST',
+      credentials: 'include',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      cache: 'no-store',
```
kettanaito: Is `cache` and `credentials` crucial to reproduce this behavior? If not, I'd drop them.

src/core/utils/request/onUnhandledRequest.ts
kettanaito: We should clone the request before reading it.

```suggestion
  const requestBody = await request.clone().text()
```

src/core/utils/request/onUnhandledRequest.ts
```diff
@@ -24,7 +24,9 @@ export async function onUnhandledRequest(
   const url = new URL(request.url)
   const publicUrl = toPublicUrl(url) + url.search
 
-  const unhandledRequestMessage = `intercepted a request without a matching request handler:\n\n  \u2022 ${request.method} ${publicUrl}\n\nIf you still wish to intercept this unhandled request, please create a request handler for it.\nRead more: https://mswjs.io/docs/getting-started/mocks`
+  const requestBody = await request.clone().text()
```
kettanaito: Some requests don't have a body by design, like HEAD or GET requests. Printing `Request body: ` with an empty string looks broken. We should check if the request _can_ have body before even cloning and reading it.

kettanaito (dismissed):
Thanks for improving MSW, @bitttttten! 👏 

I've pushed a few commits to polish this up. It's ready to be released!

src/core/utils/request/onUnhandledRequest.ts
kettanaito: Okay, I was wrong on this one. We have some tests that show we are aiming to minimize the number of times a request is cloned due to performance reasons. Since this is (a) an internal representation of an intercepted request (i.e. _not_ the actual request your app made), and (b) you cannot affect it past this point, I think it's safe to just read its body. 

The only use case I wonder is the `request:unhandled` event, and whether I'd be able to read the request body there.

bitttttten:
I actually was using a local copy of msw where I already added the clone, so thanks so much for fixing and adding that!

kettanaito:
## Released: v2.4.0 🎉

---

PR #2228, opened 2024-08-01
fix(HttpResponse): skip setting "Content-Length" if it is already set

kettanaito:
When upgrading @mswjs/interceptors to 0.33 (in https://github.com/mswjs/msw/pull/2011), I've noticed a failing mocked binary response test: it was listing the `Content-Length` header value _twice_

```
Content-Length: 1044, 1044
```

> [!WARNING]
> This indicates a deeper issue, as `.set()` on `Headers` instance must not join values. Something goes off here. Raw `Headers` from Undici behaved correctly on that branch. 

We don't have any check if the `Content-Length` response header has already been set on `HttpResponse.arrayBuffer()`, which means MSW will disregard your custom `Content-Length`, which is a bug.

### DIFF
--- a/src/core/HttpResponse.ts
+++ b/src/core/HttpResponse.ts
@@ -137,7 +137,7 @@ export class HttpResponse extends Response {
   static arrayBuffer(body?: ArrayBuffer, init?: HttpResponseInit): Response {
     const responseInit = normalizeResponseInit(init)
 
-    if (body) {
+    if (body && !responseInit.headers.has('Content-Length')) {
       responseInit.headers.set('Content-Length', body.byteLength.toString())
     }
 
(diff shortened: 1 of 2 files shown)


src/core/HttpResponse.ts
```diff
@@ -137,7 +137,7 @@ export class HttpResponse extends Response {
   static arrayBuffer(body?: ArrayBuffer, init?: HttpResponseInit): Response {
     const responseInit = normalizeResponseInit(init)
 
-    if (body) {
+    if (body && !responseInit.headers.has('Content-Length')) {
```
kettanaito, line 140: Added the missing check.

kettanaito:
## Released: v2.3.5 🎉

This has been released in v2.3.5!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.3.5)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.3.5)

---

PR #2232, opened 2024-08-07
feat(postinstall): add support for npm workspaces in postinstall routine

SuperiorJT:
This adds support for npm workspaces by going through each workspace defined in the root `package.json` and executing the init script for each workspace with the msw configuration defined. Let me know if there are any issues with this, as I've only tested it on a private project. Also, if there is anything in the code that is unclear or needs reworded, I can update that as well.

### DIFF
--- a/config/scripts/postinstall.js
+++ b/config/scripts/postinstall.js
@@ -1,36 +1,54 @@
 const fs = require('fs')
 const path = require('path')
 const { execSync } = require('child_process')
+const { globSync } = require('glob')
 
 // When executing the "postinstall" script, the "process.cwd" equals
 // the package directory, not the parent project where the package is installed.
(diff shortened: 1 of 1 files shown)


kettanaito:
Hi, @SuperiorJT! Thanks for working on this one. In good faith, I think we shouldn't ship this. Instead, I recommend putting the `msw.workerDirectory` property in your root-level `package.json`, the same way you define your workspaces on the root. 

MSW already supports an array of worker directories so this is your setup for a monorepo:

```json
// ./package.json
{
  "msw": {
    "workerDirectory": [
      "./apps/one/public",
      "./aps/two/static",
    ]
  }
}
```

Since you are installing dependencies on the root, MSW will be able to see this property and act accordingly. We are using the very same setup in https://github.com/mswjs/examples, which is a monorepo.

Issue #2193, opened 2024-07-02
Cannot read properties of undefined (reading 'url')


nemonemi:
### Prerequisites

- [X] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [X] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [X] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [X] I'm using Node.js version 18 or higher

### Browsers

Chromium (Chrome, Brave, etc.)

### Reproduction repository

N/A

### Reproduction steps

**msw version: 2.3.0-ws.rc-7** - using it for the subscription mocking

I believe it's the same behavior and reason as described in this issue https://github.com/mswjs/msw/issues/2083.
![image](https://github.com/mswjs/msw/assets/77624664/f743717a-79bb-441f-892e-c29f4147b988)

### Current behavior

Almost every time I reload my screen, or it is auto-reloaded it throws this error.

```
Cannot read properties of undefined (reading 'url')
```

### Expected behavior

Does not throw this error.

ghost:
This is a nasty bug. We use an error handler of last resort. When there is an unhandled exception it will tear down the application, try to send diagnostics and present a generic error message to the user.

All we can do here is check if the error message is equal to `Cannot read properties of undefined (reading 'url')` and pray this rather generic error was really emitted by `msw`.

nemonemi:
In my case, it happens only in the development mode. When it happens I simply close the React error popup and continue working. It is not super intrusive, but it would be great if it didn't occur.

kettanaito:
Looks like a duplicate of #2146. Let's track it there.

PR #2240, opened 2024-08-25
fix: cannot read properties of undefined (reading 'url')

baegofda:
fix error `cannot read properties of undefined (reading 'url')` (ref. https://github.com/mswjs/msw/issues/2193)
image
This bug is said to be fixed in the latest version, but it's still an issue for some users, including myself (I'm currently using version 2.3.5).

In the `createResponseListener` function, sometimes `const request = context.requests.get(requestId)` is possible `undefined`. so reference error in `request.url` in the code below.
```
if (!response.url) {
  Object.defineProperty(response, 'url', {
    value: request.url,
    enumerable: true,
    writable: false,
  })
}
```
The same issue does not occur when using the changed version with this code.

### DIFF
--- a/src/browser/setupWorker/start/createResponseListener.ts
+++ b/src/browser/setupWorker/start/createResponseListener.ts
@@ -18,7 +18,12 @@ export function createResponseListener(context: SetupWorkerInternalContext) {
     // Get the Request instance reference stored in the
     // request listener.
     const { requestId } = responseJson
-    const request = context.requests.get(requestId)!
+    const request = context.requests.get(requestId)
+
+    if (!request) {
+      return
+    }
+
     context.requests.delete(requestId)
 
     /**


kettanaito:
Hi, @baegofda. Thanks for opening this!

_This_ is the issue:

> sometimes const request = context.requests.get(requestId) is possible undefined

The solution you propose is a workaround. The error doesn't occur because the logic never runs. But it has to. 

If you have a moment, would you look if you can spot any consistent reproduction of when `context.requests.get(requestId)` is undefined? I've addressed it in the past, but obviously there's more I haven't accounted for. Would be nice to track it down and fix.

baegofda:
Hello, @kettanaito. Sure!

This is my Environment

> Next.js 14.0.4
> react: 18.2.0
> msw: 2.3.5
> @tanstack/react-query: 5.14.2

This is happening in my local environment where I'm developing Next.js and I'm requesting msw handler via react-query as a CSR. 

In my case, the error occurs when I run more than one browser tab and do a page reload.
image

baegofda:
It looks like the `createResponseListener` return is executed before the `context.requests.set` in the `createRequestListener`.

image
image

kettanaito:
@baegofda, that looks interesting. Thanks for diving in! Do you have a reproduction repo where I an look at this. If I have a repro, we can track it down in no time, I believe.

econneely:
Hi can this get merged still encountering this issue on latest version

artsiomshushkevich:
Hi @kettanaito. msw v2.6.4 also has this issue. I was able to easily reproduce it while having multiple tabs of the same app opened.

kettanaito:
Closing this in favor of #2455, which provides a solution to the underlying problem.

kettanaito:
## Released: v2.9.0 🎉

This has been released in v2.9.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.9.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.9.0)

---

Issue #2248, opened 2024-08-29
Vitest tests fail with error "Cannot find package 'graphql' imported" in v2.4.0


am-ma:
### Prerequisites

- [X] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [X] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [X] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [X] I'm using Node.js version 18 or higher

### Node.js version

v20.10.0

### Reproduction repository

https://github.com/am-ma/msw2.4.0-examples-for-issue/tree/main/examples/with-vitest

### Reproduction steps

1. Upgrade nsw to v2.4.0
2. Run tests   <- error occurred

### Current behavior

All tests fail.
```bash
 FAIL  example-jsdom.test.ts [ example-jsdom.test.ts ]
 FAIL  example.test.ts [ example.test.ts ]
Error: Cannot find package 'graphql' imported from {userDir}/msw2.4.0-examples-for-issue/examples/with-vitest/node_modules/msw/lib/core/utils/internal/parseGraphQLRequest.mjs
```

### Expected behavior

Tests succeed without error.

nartc:
We ran into the same thing here #2247

CAcarSci:
I have the same issue here. Installing the graphql package `npm i graphql` solves the error. However, this means that the latest version of MSW does not make the graphql dependency optional.

ghost91-:
Also happens to us with jest:
```
  ● Test suite failed to run

    Cannot find module 'graphql' from '../../node_modules/msw/lib/core/utils/internal/parseGraphQLRequest.js'

    Require stack:
      /home/.../node_modules/msw/lib/core/utils/internal/parseGraphQLRequest.js
      /home/.../node_modules/msw/lib/core/handlers/GraphQLHandler.js
      /home/.../node_modules/msw/lib/core/graphql.js
      /home/.../node_modules/msw/lib/core/index.js
```

THETCR:
@kettanaito 
You only added `graphql` to `peerDependenciesMeta` and forget to actually add the dependency under `peerDependencies`.

kettanaito:
@THETCR, great catch! This, alongside the fix I described [here](https://github.com/mswjs/msw/issues/2247#issuecomment-2317874411) should resolve the issue.

[...]

Issue #2247, opened 2024-08-29
2.4.0 Main entry point still exports handlers (i.e: `GraphQLHandler`)


nartc:
### Prerequisites

- [X] I confirm my issue is not in the [opened issues](https://github.com/mswjs/msw/issues)
- [X] I confirm the [Frequently Asked Questions](https://mswjs.io/docs/faq) didn't contain the answer to my issue

### Environment check

- [X] I'm using the [latest](https://github.com/mswjs/msw/releases/latest) `msw` version
- [X] I'm using Node.js version 18 or higher

### Node.js version

20.14.0

### Reproduction repository

https://github.com/nartc/msw-entry-point-bug

### Reproduction steps

`npm install`
`npm test`

Script fails with `Cannot find module "graphql"` 

### Current behavior

Thank you for your effort on `msw`.

Before I say anything, I believe our issue is our own "user error". However, I also believe this issue can still be an overlook as current code would break at runtime with 2.4.0 and the release note does not mention anything about the new secondary entry points available in 2.4.0

```ts
import { http, HttpHandler } from 'msw';
```

This works in pre 2.4.0 but will break in 2.4.0 for projects that don't have `graphql` installed. This is due to `msw` core entry point still exports the handlers which makes `graphql` **NOT** an optional peer dependency.

The proper fix would be to import from `msw/core/http` but it seems like `msw/core/http` does not export `HttpHandler` either.

Once again, our project runs into this due to many errors/mistakes on our part but I want to raise this issue in case others run into the same thing (or have the same user errors/mistakes)

### Expected behavior

```ts
import { http } from 'msw';
```

This import should at least break at compilation time (type checking time) with something like: msw does not export `http`. And release note should mention migrating to `msw/core/http` (or respective entry point for the handlers).

ghost91-:
Unfortunately, using more specific exports doesn't always seem to be possible. For example, unless I am mistaken, `delay` is only exported via the main export.

Similarly, `HttpResponse` also is not exported from `msw/core/http`. Probably applies to quite a few more things.

If `msw/core/http` is really intended as the "don't need graphql" entrypoint, the exports should probably be identical to those of `msw`, except from the graphql related stuff.

nartc:
@ghost91- or `msw` entry point should be cleared of anything that requires optional peer dependencies so things like main utilities can still be exported from the main entry point.

kettanaito:
[...]

PR #2249, opened 2024-08-29
fix: add `graphql` to "peerDependencies"

THETCR:
- Related to #2247, #2248

### DIFF
--- a/package.json
+++ b/package.json
@@ -200,6 +200,7 @@
     "webpack-http-server": "^0.5.0"
   },
   "peerDependencies": {
+    "graphql": ">= 16.8.x",
     "typescript": ">= 4.7.x"
   },
   "peerDependenciesMeta": {


kettanaito (approved):
Thanks for fixing this! 👏

kettanaito:
## Released: v2.4.1 🎉

This has been released in v2.4.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.4.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.4.1)

---

pascalmann:
Not work for me.
grapql was added to devDependencies (so no installed in my repo), but is still imported (lazily), even if I not use it.

thepassle:
This should've been a breaking change (and it did cause breakage for some of our projects). If people import something from `'msw'`, it'll still hit the dynamic graphql import (even if we don't use anything graphql related from msw), which now doesnt get automatically installed because its only a peer dependency.

Thankfully I can work around it by rewriting our import from `'msw'` to `'msw/core/http'`, since we only use `http` and not `graphql`, but more people will start running into this issue as well.

kettanaito:
> If people import something from 'msw', it'll still hit the dynamic graphql import (even if we don't use anything graphql related from msw)

This is the part that's confusing. How? Dynamic imports are by definition a part of your runtime. I don't suppose the browser extracts that dynamic import, puts it on the root scope, and then evaluates it. That mustn't happen, otherwise I see no point in a dynamic import as a feature. 

Removing a graphql dependency is fix by design. You shouldn't be installing that package if you aren't planning on mocking anything GraphQL. As such, it was shipped as fix. The fact that the first few iterations of that fix are problematic _in certain environments_ is a separate discussion. 

I am welcoming reproduction repos at this point because none of my prior testing caught this. Here's where I've tested these fixes:

- Vitest ESM
- Vitest CJS
- Jest ESM
- Jest CJS
- webpack-based project

I hope that anybody reaching out understands these changes are done in good faith. I cannot physically test every existing combination of your tooling. Yes, things may break. You are more than welcome to pin the MSW version and go on with your life. If not, you should contribute with discussions, suggestions, and pull requests. Thanks.

PR #2251, opened 2024-08-29
fix: use typescript@4.8 as the minimal supported version

THETCR:
To fix the discrepancy between the dev dependency and the version installed by end users

### DIFF
--- a/package.json
+++ b/package.json
@@ -201,7 +201,7 @@
   },
   "peerDependencies": {
     "graphql": ">= 16.8.x",
-    "typescript": ">= 5.5.x"
+    "typescript": ">= 4.8.x"
   },
   "peerDependenciesMeta": {
     "graphql": {
--- a/package.json
+++ b/package.json
@@ -201,7 +201,7 @@
   },
   "peerDependencies": {
     "graphql": ">= 16.8.x",
-    "typescript": ">= 4.7.x"
+    "typescript": ">= 5.5.x"
   },
   "peerDependenciesMeta": {
     "graphql": {


THETCR:
@kettanaito here is another PR to resolve a potential conflict between the used version of `Typescript`

kettanaito:
This is incorrect. MSW still supports TS 4.7 and up. With your change, you are effectively removing that support. 

There's no discrepancy possible because what we install for development is irrelevant as long as it's in the supported TS version range. That's why each build runs an extensive type test suite on every supported version of TS to let us know if we break anything up or down the line. 

https://github.com/mswjs/msw/blob/8a9568ad4593b662a8c320bb1a02b380791a1696/.github/workflows/compat.yml#L55-L57

Give this a look. I will likely close this. 

> What this change did help to uncover is that we are okay with 4.7 in peer deps but have officially deprecated its support (it's EOL). _That_ is a good find and would be a nice change : 4.7.x -> 4.8.x.

THETCR:
@kettanaito 
So you have covered possible compatibility issues between major version 4 and 5, then this is indeed fine.

My advice would be though to set the SemVer range between 4.8.X and 5.X.X, because in the current setup it would also automatically resolve to higher major versions. Which can possibly break the package for end users.

kettanaito:
> So you have covered possible compatibility issues between major version 4 and 5, then this is indeed fine.

Yep! Exactly my point, we have. There's been a dedicated effort for 5.x support, and we've adjusted the types so they work fine across 4.x and 5.x 👌 The automation I linked as the proof of that. 

> My advice would be though to set the SemVer range between 4.8.X and 5.X.X, because in the current setup it would also automatically resolve to higher major versions.

We discussed this previously as well (sorry, cannot find that discussion right now). TL;DR MSW supports all TS versions by default, if there's a bug, you report it. That's a better experience than throwing errors on install whenever you upgrade a TS to a version not listed in MSW's peer dependencies. There's no guarantee that the version will break. 

Would you like to bump the minimal peer dependency version for TS to 4.8.x with this change? _That_ is indeed an issue we have right now.

kettanaito (approved):
I've fixed the typescript peer dependency to correspond to the minimal supported version (4.8).

kettanaito:
## Released: v2.4.2 🎉

This has been released in v2.4.2!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.4.2)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.4.2)

---

PR #2252, opened 2024-08-29
chore(iframe): await `window.request` to be set

kettanaito:
Attempt to address a flaky browser-side `iframe.test.ts`. 

- https://github.com/mswjs/msw/actions/runs/10617574987/job/29430753429?pr=2250


kettanaito:
Hm. Awaiting the function seems not to be enough. The tests now time out that it's _never_ defined. This may be the frame setup issue then. Needs deeper investigation.

kettanaito:
I believe this should be fixed by the recent changes that await the idle network state in the frame-related tests (#2324).

PR #2250, opened 2024-08-29
fix: import `graphql` lazily

kettanaito:
- Fixes #2247 
- Fixes #2248 

## Changes

MSW depends on `graphql` only when using the `graphql` request handlers, and only to import its `parse()` method (other imports are type imports and are not present on runtime, neither can they affect it). 

- This change makes `graphql` a _dynamic import_ where it's needed. This way, even though MSW exports `{ graphql } from './graphql'` in the root, it will never result in `import X from 'graphql'` unless you `import { graphql } from 'msw'` on runtime. 

In a long term, we should drop http/graphql handlers from the root entry and keep them only in `msw/core/http` and `msw/core/graphql`. 

- Also adds a custom error message if importing `graphql` fails for any reason. This will remind the developer that they have to install `graphql` if they want to intercept GraphQL requests.

PR #2298, opened 2024-09-30
fix: revert the lazy import of `graphql`

grebenyuksv:
This reverts the recent attempts to import `graphql` lazily, which we agreed to do in https://github.com/mswjs/msw/pull/2250#issuecomment-2382625016.


kettanaito:
Thanks for opening this, @grebenyuksv. Can you please remind me what issue are you facing with the dynamic `require`? 

I have a suspicion that merging this will cause errors for ESM consumers. Importing anything from the root results in an import to `parseGraphQL`, and then errors that the `graphql` package is not installed.

kettanaito (approved):
This looks great! Thank you, @grebenyuksv. 

I confirm this is how it used to be before the lazy-loading disaster ([ref](https://github.com/mswjs/msw/blob/78360c881ac4ff4d6de32a7ea344cdd65c0c2ca1/src/core/utils/internal/parseGraphQLRequest.ts)).

kettanaito:
I'm concerned that this change will make https://github.com/mswjs/msw/issues/2248 an issue again. 

Perhaps we should take a closer look at why your use case errors, @grebenyuksv.

grebenyuksv:
Thanks for your review, @kettanaito! Let's close it for now. I've parked that Storybook project where the dynamic import doesn't work and, as I said, I would rather bet that the problem is in other tooling, not in MSW.

PR #2299, opened 2024-09-30
feat: support server-sent events (SSE)

### LINKED ISSUE #2117: First-class Server-Sent Events (SSE) API
### Scope

Adds a new behavior

### Compatibility

### Feature description

SSE offer real-time communication over HTTP. You can already mock SSE with MSW even now but shipping a fine-tailored API for SSE may be beneficial. 

## References

- **[Mocking SSE with MSW](https://alexocallaghan.com/mock-sse-with-msw)**
- https://gist.github.com/Acetyld/74ce258def9ceb93b3a71960498feaf5
- https://github.com/epicweb-dev/kcdshop/blob/11208ee1c99e70bf41458411d95cf96936a84c14/packages/workshop-app/app/routes/login-sse.tsx#L28

kettanaito:
- Closes #2117 

## Todo

- [x] Accept an `OutgoingEvents` type argument at `sse()` to make type-safe `event` property on the mock payload. 
- [x] Support `passthrough()`. How do you intercept an EventSource, interact with it from the mock, and then passthrough after that?
- [x] Support event augmentation. How do you listen to the actual server's events, modify them, and send them back to the client?
- [x] Add test for `withCredentials` affecting the request cookies propagation. 
- [x] How to handle `passthrough()`? What if I want to let the `EventSource` receive whichever events the actual server sends from now on? Still manual? 
- [x] Add tests for `.use()` with `sse()`. 
- [x] Support [`retry`](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#retry)
- [x] Add an invariant for `typeof EventSource !== 'undefined'` to let the developer know when they are using `sse()` in the environment that doesn't support it. 
- [x] Fix the `[Error [TypeError]: Failed to fetch]` error during `test/browser/sse-api/sse.server.connect.test.ts` tests. It seems to be originating from the actual `fetch()` in the Service Worker (the scenarios call `server.connect()` on the SSE).
  - This is caused by the missing logger!
- [x] Finish `sse.client.send.test.ts` (the last test case is unfinished).
- [x] ⚠️  Consider exporting `sse` from `msw` root. While `EventSource` doesn't exist in Node.js, it exists in Deno. It might also be added to Node.js in the future. Afaik, there's nothing specific to browsers in our implementation. Keep the global check, export from the root for consistency. 
- [x] Support writing to the stream from extraneous sources (e.g. pubsub). Something similar to [GraphQL subscriptions](https://github.com/mswjs/msw/pull/2357#issuecomment-2833557716). `client.from()`?
  - Excessively complicated and not needed. DYI. Will show an example of that in the docs.
- [x] Documentation (https://github.com/mswjs/mswjs.io/pull/470)

## References

- https://github.com/Azure/fetch-event-source


kettanaito:
## Augmenting actual server-sent events

```ts
sse(url, ({ source, server }) => {
  // This creates a new EventSource request
  // and propagates all the events from here
  // to the underlying pending stream for `source`.
  server.connect()
})
```

Modifying the event is similar to that in WebSockets: prevent its default and send a new event:

```ts
sse(url, ({ source, server }) => {
  server.connect()

  server.addEventListener('message', (event) => {
    // Prevent this server event from reaching the client.
    event.preventDefault()
    const newEvent = modify(event.data)
    source.send(newEvent)
  })
})
```

May be a good idea to rename `source` to `client` and have a consistent SSE/WebSocket experience.

kettanaito:
The tests are failing likely due to Node.js bump to v20 in CI. 

_Edit:_ Looks like an issue specific to a particular version of Node.js. Reported here: https://github.com/nodejs/undici/issues/3676

kettanaito:
## Test update

I've resolved the infinite request issue that failed the majority of SSE tests (was using legacy passthrough header; migrating to `accept: msw/passthrough` + respecting that later on helped). 

Two tests are still failing, both seem to be related to errors. 

There's an odd rogue `TypeError: Network error` being thrown _right when the Playwright window closes_. Odd. The SSE request is not considered a network error from what I can tell. The error test case also fails. Maybe the two are related?

kettanaito:
## Test failures

Still having this error:

```
Error: require() of ES Module /Users/runner/work/msw/msw/node_modules/.pnpm/@epic-web+test-server@0.1.0/node_modules/@epic-web/test-server/lib/http.js from /Users/runner/work/msw/msw/test/browser/sse-api/sse.server.connect.test.ts not supported.
```

😬

kettanaito:
The require in tests issue should be resolved now that the repo is using ESM internally. This should unblock the future work on the SSE support.

src/core/utils/internal/isObject.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * Determines if the given value is an object.
  */
-export function isObject(value: any): boolean {
+export function isObject(value: any): value is Record<string, any> {
```
Andarist, line 4: fwiw, this isn't exactly type-safe in a way you'd assume: [TS playground](https://www.typescriptlang.org/play/?ts=5.9.0-dev.20250601#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXywGcB5AIwCtwMAKANygmRAC54pUBPASlfsYSLwASuBwxgAHkIYYWVAHMANG04A+ANwBYAFA7QkWAgggM8ALZQOpEAEEYMALKXrZcqwDeSHDlbTZC+ABfeAAfeFRkM2sYAG0AXS1dbSxEeGoiVypqCytbeydc1y4ueHcdeHNnPMcq10SKgHoG+AA9AH4dYJAIQgQy7Qqc6zsawop6+CbWju1AnSA)

src/browser/sse.ts
```diff
@@ -0,0 +1,764 @@
...
+      | {
+          id?: string
+          event?: 'message'
+          data?: EventMap['message']
+          retry?: never
+        }
+      | {
+          id?: string
+          event?: EventType
+          data?: EventMap[EventType]
```
Andarist: as I mentioned on Twitter, I don't think this is exactly correct because it can allow invalid combinations - using a discriminated union here would fit this better. If you want to still use an event map as an input, you can convert it to a discriminated union at type-level - I put a PR with that for u here: https://github.com/mswjs/msw/pull/2521

Andarist: Should this be a valid call?
```ts
sse<{ custom: string; }>('/stream', ({ client }) => {
  client.send({ type: "custom" })
})
```

IMHO, it shouldn't be because the user forgot to pass `data`. But currently it is valid because `data` is always optional here

src/core/utils/internal/isObject.ts
kettanaito: Ouch.

src/browser/sse.ts
kettanaito: It shouldn't. Events without data must be allowed but if you've described an event having a data type in your events map, then you have to pass `data` (otherwise use `{ custom: never }`).

kettanaito:
## `[Error [TypeError]: Failed to fetch]` error

### Root cause

The `log` method of the `HttpHandler` class that SSE is extending was consuming the response body in `await serializeResponse(args.response)`, which resulted in the said error. The body of the SSE cannot be consumed since it's being streamed and it incomplete.

### Solution

The SSE handler must implement its own logger to solve this issue and also be more contextual.

kettanaito:
## Update

The `sse` API is ready. You can install a preview release by following the instructions [here](https://github.com/mswjs/msw/pull/2299#issuecomment-2945347423). I will let the folks use it and share their feedback for a bit and then merge it.

zeroregard:
Just wanna prefix this by saying, thanks a lot for introducing this and we are enjoying using it!!

Tip/note to other potential PNPM users for when you are using `https://pkg.pr.new/msw@2299` (or other PR identifiers for other bleeding-edge MSW versions, it's not exclusive to this branch), since it's not versioned per se, you will run into an issue like
```javascript
ERR_PNPM_UNEXPECTED_PKG_CONTENT_IN_STORE  Package name mismatch found while reading {"tarball":"https://pkg.pr.new/msw@2299"} from the store.

This means that either the lockfile is broken or the package metadata (name and version) inside the package's package.json file doesn't match the metadata in the registry. 
Expected package: msw@2.11.0. 
Actual package in the store with the given integrity: msw@2.10.4.
```

The workaround is that you can use the specific commit version like `https://pkg.pr.new/mswjs/msw/msw@ad82c8e` to lock yourself in, otherwise your CI builds might fail suddenly.

kettanaito:
Thanks, @zeroregard. Let me know if you have some suggestions around this API or some use cases that you think it doesn't cover.

The feature is complete and tested, and I plan on releasing it later this month. It's a great chance to share early feedback until then so we could address potentially breaking changes before they are locked up until the next major release. Thanks!

kettanaito:
## Released: v2.12.0 🎉

This has been released in v2.12.0.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.0)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.0)

```
npm i msw@latest
```

---

PR #2324, opened 2024-10-22
feat: isolate parent and child frames when handling requests

kettanaito:
## The problem

If the page calls `worker.start()` and has a child frame that also calls `worker.start()`, the requests from the frame will only be resolved against the top-level parent's `worker` and its handlers. 

> This has been reported by the Vitest team. 

## The solution

During the `resolveMainFrame` client lookup, if the event `clientId` is in the active client ids of the worker, return it. Don't lookup the clients because the top-level client will always match first, ignoring any underlying otherwise matching nested client ids.

### DIFF
--- a/src/browser/setupWorker/start/utils/printStartMessage.ts
+++ b/src/browser/setupWorker/start/utils/printStartMessage.ts
@@ -1,3 +1,4 @@
+import type { ServiceWorkerIncomingEventsMap } from 'browser/setupWorker/glossary'
 import { devUtils } from '~/core/utils/internal/devUtils'
 
 export interface PrintStartMessageArgs {
@@ -5,6 +6,7 @@ export interface PrintStartMessageArgs {
   message?: string
   workerUrl?: string
   workerScope?: string
+  client?: ServiceWorkerIncomingEventsMap['MOCKING_ENABLED']['client']
 }
 
 /**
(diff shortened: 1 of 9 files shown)


src/mockServiceWorker.js
```diff
@@ -155,6 +160,10 @@ async function handleRequest(event, requestId) {
 async function resolveMainClient(event) {
   const client = await self.clients.get(event.clientId)
 
+  if (activeClientIds.has(event.clientId)) {
```
kettanaito, line 163: The fix.

src/browser/setupWorker/start/utils/printStartMessage.ts
```diff
@@ -41,6 +43,11 @@ export function printStartMessage(args: PrintStartMessageArgs = {}) {
     console.log('Worker scope:', args.workerScope)
   }
 
+  if (args.client) {
+    // eslint-disable-next-line no-console
+    console.log('Client ID: %s (%s)', args.client.id, args.client.frameType)
```
kettanaito, line 48: I'm also improving the activation message to include the client ID that prompted it, as well as the client frame type (top-level/nested). This helps during debugging.

kettanaito:
Also a good point that we still won't properly support this scenario:

```
- a. parent with MSW
  - a.1. child with MSW
    - a.1.1. child without MSW (request here!)
```

Since a.1.1 doesn't call `worker.start()`, it won't be in the `activeClientIds` in the worker script at all. That will trigger the regular client lookup in all visible frames, and that will still resolve with the top-level client (a). 

I'd say the expectation here is that the `worker` in a.1 should handle the requests coming from a.1.1. 

Unfortunately, there's no hierarchical relationship of the service worker clients. You also cannot know the actual client ID on the client to do any sort of logic on the library's side. See https://github.com/w3c/ServiceWorker/issues/1556.

sheremet-va (approved):
Thank you for the quick fix!

kettanaito:
## Released: v2.5.0 🎉

This has been released in v2.5.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.5.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.5.0)

---

PR #2331, opened 2024-10-26
Exclude SW file from deno-lint: require-await

muhrifqii:
When the project is using deno lint as a linter, this mock sw file will fall to the deno-lint rule https://lint.deno.land/rules/require-await on lines 132 and 277

### DIFF
--- a/src/mockServiceWorker.js
+++ b/src/mockServiceWorker.js
@@ -1,3 +1,4 @@
+// deno-lint-ignore-file
 /* eslint-disable */
 /* tslint:disable */
 


kettanaito:
Hi, @muhrifqii. Thanks for opening this. 

Please read [this document](https://github.com/mswjs/msw/blob/16c7c1020311c23815c9e00131edccaf717abbe5/decisions/linting-worker-script.md) on why adding the ignore comments to the worker script isn't a solution to the problem you're experiencing. Thanks.

PR #2353, opened 2024-11-10
fix: prevent infinite loop when bypassing `sendBeacon()` requests

### LINKED ISSUE #1216: `sendBeacon` causes infinite loop with response patching
### Prerequisites

### Environment check

### Browsers

Chromium (Chrome, Brave, etc.), Firefox, Safari

### Reproduction repository

https://github.com/hpohlmeyer/msw-send-beacon-loop

### Reproduction steps

1. Check out the repro: `git clone https://github.com/hpohlmeyer/msw-send-beacon-loop.git`
2. Run `npm install` in the root folder
3. Run `npm start`
4. Go to [http://localhost:3002/](http://localhost:3002/) in Chrome (FF just freezes without seeing anything in the network tab)
5. Open the network tab in the devtools
6. Click on the "Trigger request" button
    - You will see an infinite request loop

### Current behavior

It looks like `ctx.fetch(req)` does not seem to bypass msw and triggers the same request over and over again, causing an infinite request loop.

### Expected behavior

`ctx.fetch(req)` should bypass msw and not trigger the same route handler again.

kettanaito:
- Fixes #1216 

## Motivation

Requests issued via the `sendBeacon` function _have immutable headers_. Any header but for the known headers will be ignored. This also means that the `x-msw-intention: passthrough` request header set by the `bypass()` function will be ignored, causing the infinite loop when bypassing an intercepted `sendBeacon` request. 

## Changes

Move the `x-msw-intention: passthrough` to `accept: msw/passthrough`. The `accept` header is always preserved, and it should be safe to rely on it to provision bypassed requests identification. 

> Note that MSW will remove that `msw/passthrough` part of the request's `accept` header before actually performing the request. 

Also fixes the issue where the request bypass header was preserved for passthrough requests in Node.js only.

### DIFF
--- a/src/core/bypass.test.ts
+++ b/src/core/bypass.test.ts
@@ -67,3 +67,15 @@ it('allows modifying the bypassed request instance', async () => {
   expect(await request.text()).toBe('hello world')
   expect(original.bodyUsed).toBe(false)
 })
+
+it('supports bypassing "keepalive: true" requests', async () => {
+  const original = new Request('http://localhost/resource', {
+    method: 'POST',
+    keepalive: true,
+  })
+  const request = bypass(original)
+
+  expect(request.method).toBe('POST')
+  expect(request.url).toBe('http://localhost/resource')
+  expect(request.body).toBeNull()
+})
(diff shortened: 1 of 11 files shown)


kettanaito:
## Released: v2.6.4 🎉

This has been released in v2.6.4!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.6.4)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.6.4)

---

PR #2391, opened 2024-12-12
feat: add `baseUrl` option in `getResponse`

ukyoda:
I appreciate the availability of this library. Thanks to all the maintainers.

I have been using the `getResponse` function lately. It is a very useful API, but I thought it would be better if I could also use handlers created with relative paths.

So I would like to add a `baseUrl` argument to `getResponses`. I think this can be adapted without causing side-effects, since it just allows existing implementations to be used.

Usage:

```typescript
const response = await getResponse(
  [http.get('/user', () => Response.json({ name: 'John' }))],
  new Request('http://localhost/user'),
  'http://localhost',
)
```

If this is not a problem, I would love to have it merged. Please review it.

I am not good at English and I don't have much experience in Contribute to OSS, so I am sorry if I am rude.

### DIFF
--- a/src/core/getResponse.test.ts
+++ b/src/core/getResponse.test.ts
@@ -61,3 +61,39 @@ it('returns the response from the first matching handler if multiple match', asy
   expect(response?.headers.get('Content-Type')).toBe('application/json')
   expect(await response?.json()).toEqual({ name: 'John' })
 })
+
+it('returns the response from a handler that matchs path when set baseUrl', async () => {
+  const response = await getResponse(
+    [http.get('/user', () => Response.json({ name: 'John' }))],
+    new Request('http://localhost/user'),
+    'http://localhost',
+  )
+  expect(response?.status).toBe(200)
+  expect(response?.headers.get('Content-Type')).toBe('application/json')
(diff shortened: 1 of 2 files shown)


kettanaito:
Hi, @ukyoda. Thanks for proposing this. 

Actually, this hints at a different problem we have: MSW doesn't use `resolutionContext` anymore at all during the request handling. We should remove that option altogether. 

There has been numerous discussions on the base URL support in MSW, and the recommended way is for you to implement that support on your end. A simple utility function would work just fine. See https://github.com/mswjs/msw/issues/397#issuecomment-751230924. 

## Context

I've found some context in the tests, `resolutionContext.baseUrl` was primarily meant for extensions utilizing MSW to implement additional functionality. I need to think on the right way to approach this.

ukyoda:
@kettanaito 

Thank you for your response.

I now understand why there is no such thing as `baseUrl` in `getResponse`.
I will close this PR and use the solution you suggested :)

kettanaito:
@ukyoda, thanks. I've pushed the removal of `resolutionContext` to a feature branch but will think on it. I don't want to break existing third-party consumers of that property, although one should probably not be using it.

PR #2403, opened 2024-12-29
fix: support nullable GraphQL queries

huuyafwww:
Hello, thank you for developing this wonderful library called msw. I am using msw in my graphql project. There I have a Query with nullable return value.
And I generate that type using graphql-codegen.
Here is an example of the generated type.

```ts
export type Maybe<T> = T | null;

export type IQuery = {
  __typename?: 'Query';
  authUser: Maybe<IUser>;
  user: Maybe<IUser>;
};
```

And I tried to use this query type which may return null in graphql.query and graphql.mutation.

```ts
import { graphql } from 'msw';

graphql.query<IQuery['authUser'], /* variable type */>(queryName, () => {});
graphql.mutation</* nullable mutation */, /* variable type */>(queryName, () => {});
```

Then I got a type error that nullable queries and mutations are not supported.

I would like to have the actual value returned by a query or mutation set strictly as a type in mock as well.
Therefore, I have modified `GraphQLQuery` to allow null.
What do you think?


kettanaito:
## Released: v2.8.4 🎉

This has been released in v2.8.4!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.8.4)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.8.4)

---

PR #2439, opened 2025-02-24
fix(HttpResponse): set the default `bodyType` to `any`

kettanaito:
## Problem

When annotating the response body type of your handlers, MSW doesn't allow raw `HttpResponse` instances to be returned because they are missing the `[bodyType]` symbol in them:

```ts
http.get<never, never, { id: string }>('/resource', () => {
  // This errors. Expected one body type, got no [bodyType] symbol at all.
  return new HttpResponse(null, { status: 500 })
})
```

This is incorrect. You should always be able to tap into a type-unsafe response declaration, especially when returning error responses. 

## Solution

- Add the `[bodyType]` symbol to the `HttpResponse` class and set it to `any` by default. This will make `new HttpResponse()` instances type-compatible with any response body types while preserving stricter type matching when using shorthands like `HttpResponse.json()`.

### DIFF
--- a/src/core/HttpResponse.ts
+++ b/src/core/HttpResponse.ts
@@ -37,6 +37,8 @@ export interface StrictResponse<BodyType extends DefaultBodyType>
  * @see {@link https://mswjs.io/docs/api/http-response `HttpResponse` API reference}
  */
 export class HttpResponse extends FetchResponse {
+  [bodyType]: any
+
   constructor(body?: BodyInit | null, init?: HttpResponseInit) {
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v2.7.2 🎉

This has been released in v2.7.2!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.7.2)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.7.2)

---

PR #2440, opened 2025-02-24
fix: do not treat static asset requests as unhandled by default

kettanaito:
By default, MSW catches all requests that go through the Service Worker. This includes all the static assets too, like CSS or HTML or fonts. Treating them as unhandled rarely makes sense. 

```ts
import { isCommonAssetRequest } from 'msw'
```

## Changes

- Export the `isCommonAssetRequest` utility function from `msw`.
- Change the default behavior of `onUnhandledRequest` to _exclude common static assets_ before applying string behaviors like `'warn'`.
- Allow opting out from the default behavior if you provide a _function_ as the value of `onUnhandledRequest`. If you do, then you have to import and use `isCommonAssetRequest()` manually to get the default behavior back. This is intended for behavior customization.


kettanaito:
## Released: v2.7.3 🎉

This has been released in v2.7.3!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.7.3)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.7.3)

---

PR #2267, opened 2024-09-07
fix: revert "graphql" as optional peer dependency

kettanaito:
- Fixes #2264
- Fixes #2248
- Related to #2187 
- Closes #2265 

## Motivation

ESM imports of optional peer dependencies are virtually impossible. You have to opt-out from your bundler treating them as code splitting points because the dependency may not exist, and then as a result, _you_ become in charge of serving that dependency over HTTP + providing an `importmap` for the browser to understand where to look it up. 

This is an insane ask from a developer, and I'm not willing to make it. I was prematurely optimistic about the state of optional peer dependencies, causing the recent release to be a failure. Lessons learned. 

## Resolving the original issue

The original issue reported in #2185 states that the `graphql` dependency _has_ to be a peer dependency to prevent duplicate GraphQL versions installed in a single app (MSW installs v16.1, your app has v16.8, and that potentially causes issues).

PR #2442, opened 2025-02-25
fix(deps): make graphql an optional peer dependency

slinkardbrandon:
Working on an application that currently has vulnerabilities due to graphql but the application does not even use graphql, that eventually led us to this repo.

I updated the dependency chain so that graphql is an optional peer dependency which should enable apps to not install additional unwanted dependencies that are unused. Additionally, if you are working on an application with graphql, it probably makes sense to just use the host's graphql version anyways rather than having a duplicate from this library.

Edit: While in there I also moved some dependencies to dev deps that didn't need to be production level dependencies just to clean things up a bit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,16 +147,12 @@
     "@mswjs/interceptors": "^0.37.0",
     "@open-draft/deferred-promise": "^2.2.0",
     "@open-draft/until": "^2.1.0",
-    "@types/cookie": "^0.6.0",
-    "@types/statuses": "^2.0.4",
-    "graphql": "^16.8.1",
     "headers-polyfill": "^4.0.2",
     "is-node-process": "^1.2.0",
     "outvariant": "^1.4.3",
     "path-to-regexp": "^6.3.0",
     "picocolors": "^1.1.1",
     "strict-event-emitter": "^0.5.1",
-    "type-fest": "^4.26.1",
     "yargs": "^17.7.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -166,11 +162,13 @@
     "@open-draft/test-server": "^0.4.2",
     "@ossjs/release": "^0.8.1",
     "@playwright/test": "^1.48.0",
+    "@types/cookie": "^0.6.0",
```
kettanaito, line 165: These types have to be shipped as regular dependencies. For libraries that are written in TS, their TS dependencies are regular dependencies, never dev. Otherwise they may not get installed and start causing issues for consumers.

package.json
```diff
@@ -209,10 +208,10 @@
     "webpack-http-server": "^0.5.0"
   },
   "peerDependencies": {
-    "typescript": ">= 4.8.x"
```
kettanaito, line 212: We still need to keep TypeScript. I don't see how your proposal implies it has to be removed.

package.json
```diff
@@ -209,10 +208,10 @@
...
+    "graphql": ">=16.8.1"
```
kettanaito, line 211: Sadly, this isn't going to work. There isn't really a thing as optional peer dependency in JavaScript right now. You can think you defined one in `peerDependencies`, but that's just the installation part.

There's no way to _consume_ an optional peer dependency in the library's code. Dynamic imports aren't supported everywhere, and requires cause issues with ESM. 

We've tried this in the past to no avail, always meeting some edge case from our users. 

Sharing some context in case you will be more successful than me:

- #2187
- #2249
- #2267

kettanaito:
Hi, @slinkardbrandon. Thanks for opening this. I left some comments on the code, elaborating on why this won't really work in practice. 

But the most important thing, perhaps, is this:

> Working on an application that currently has vulnerabilities due to graphql but the application does not even use graphql, that eventually led us to this repo.

I strongly advise you configure your vulnerability report tooling to ignore dev dependencies, which is MSW (and its respective dev dependencies). Those do not pose any risk to your users and, by default, should be ignored. That will unblock you in your project. 

If you still want to pursue optional graphql, I'd be happy to help, but none of my attempts were successful (see the referenced issues).

kettanaito:
I am going to close this. Optional peer dependencies exist in JavaScript only on paper. Once you try shipping them to the mixed CJS/ESM ecosystem, CIs will start failing and people get mad. We would have to come up with a different approach to this, if any.

PR #2447, opened 2025-03-05
fix: properly handle concurrent requests while stopping the service worker

### LINKED ISSUE #2323: Request made immediately after calling stop on the service worker doesn't always passthrough properly
### Prerequisites

### Environment check

### Browsers

Chromium (Chrome, Brave, etc.), Safari

### Reproduction repository

https://github.com/spocke/mswjs-stop-bug

### Reproduction steps

1. Clone the attached repo
1. Install the npm dependencies using `npm i`
2. Start the server using `npm start`
3. Open the Safari or Chrome and go to go to `http://localhost:3000`
4. Re-load the page N times might take a few tries
5. Observe that sometimes the second image is not served (network tabs shows it in a pending state)
6. Observe that sometimes the second image is served though the service worker but not properly served it just spins and spins forever in the network tab

### Current behavior

Resources are still served though an active client rather than passed though the service worker even if you call `worker.stop()` sometimes that even fails to serve the request since likely because of the handlers being removed for it on the parent page side.
It's likely flakey like this because the post message from the browser to the service worker takes a while to process so it can't be a fire and forget it needs to be async and send a ACK back when it has properly intercepted the message to stop intercepting requests.

### Expected behavior

No requests should be processed though an active client on the service worker once you call stop on it. The stop function probably needs to be async to know when it's safe to do new requests and expect them to be passed though by https://github.com/mswjs/msw/blob/ca6cb7e8e95ad76c6fe6c44d7f4b80d36bc53478/src/mockServiceWorker.js#L109.

spocke:
- Fixes https://github.com/mswjs/msw/issues/2323

Not sure how to write tests for this since it requires precise timing you need to have requests in-flight while the `stop` function is called on the service worker and then some how detect that it hanged on those. So need some guidance there.

This change does these things:
1. It changes the `stop` function of the service worker from being sync to async so that you know that after it resolved the service worker is no longer passing requests over to the page handlers. So once `stop` resolved everything should be unmocked and just passthrough.
2. It alters the order of the deactivation so that the the service worker deactivation is ensured before we unregister any event handles on the parent page side.
3. It sends a new `MOCK_DEACTIVATE_RESPONSE` from the service worker to the page so that we can ensure that the message got received. Processing of messages are not instant so sending a `MOCK_DEACTIVATE` call and then assuming that it got it doesn't work.
4. It tracks any in flight HTTP requests that are processed while we are stopping the service worker mocking. If that happens then those in-flight requests just becomes passthrough requests.

### DIFF
--- a/src/browser/setupWorker/glossary.ts
+++ b/src/browser/setupWorker/glossary.ts
@@ -64,6 +64,7 @@ export interface ServiceWorkerIncomingEventsMap {
   KEEPALIVE_RESPONSE: never
   REQUEST: ServiceWorkerIncomingRequest
   RESPONSE: ServiceWorkerIncomingResponse
+  MOCK_DEACTIVATE_RESPONSE: never
 }
 
 /**
@@ -191,7 +192,7 @@ export type StartHandler = (
   options: RequiredDeep<StartOptions>,
   initialOptions: StartOptions,
 ) => StartReturnType
-export type StopHandler = () => void
+export type StopHandler = () => Promise<void>
 
 export interface SetupWorker {
   /**
--- a/src/browser/setupWorker/setupWorker.ts
+++ b/src/browser/setupWorker/setupWorker.ts
@@ -206,11 +206,13 @@ export class SetupWorkerApi
     return await this.startHandler(this.context.startOptions, options)
   }
 
-  public stop(): void {
+  public async stop(): Promise<void> {
+    await this.stopHandler()
+
     super.dispose()
+
     this.context.events.removeAllListeners()
     this.context.emitter.removeAllListeners()
-    this.stopHandler()
   }
 }
 
--- a/src/browser/setupWorker/stop/createStop.ts
+++ b/src/browser/setupWorker/stop/createStop.ts
@@ -5,7 +5,7 @@ import { printStopMessage } from './utils/printStopMessage'
 export const createStop = (
   context: SetupWorkerInternalContext,
 ): StopHandler => {
-  return function stop() {
+  return async function stop() {
     // Warn developers calling "worker.stop()" more times than necessary.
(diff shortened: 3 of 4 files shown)


kettanaito:
Hi, @spocke. Thank you for working on this!

This may be a naive question on my end, but wouldn't we achieved the same if we flipped the logic around? Whenever the service worker receives a mocked response from the client, check if the worker is "stopped" (received the stop event), and if it is passthrough the response. 

This will have a slight cost of handler lookup but when compared to introducing a breaking change (`await worker.stop()`) it seems insignificant.

spocke:
> This may be a naive question on my end, but wouldn't we achieved the same if we flipped the logic around? Whenever the service worker receives a mocked response from the client, check if the worker is "stopped" (received the stop event), and if it is passthrough the response.

@kettanaito I think we can make this sync. What if we register another event listener after we unregistered all listeners in the `stop` function. This event listener would simply passthrough any stray request that happens after the `stop` has been called. Then that also waits for the `MOCK_DEACTIVATE_RESPONSE` to come in once that has been returned from the service worker it's safe for us to remove this event listener and cleanup memory. Since when that happens then we know that the service worker will simply mark any in flight post message requests as `passthrough`.

But how do I register such a event listener in a clean way in the `stop` function. Or is it some form of catch all request handler. Basically we would need a thing that catches all request and returns them as passthrough and then a way to remove that catch all thing. I need some pointers there.

kettanaito:
## Released: v2.11.2 🎉

This has been released in v2.11.2!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.11.2)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.11.2)

---

PR #2448, opened 2025-03-05
refactor: use as const instead of enum in http methods

nayounsang:
# Abstract
- Use `as const` object in http methods.

# Description
- No effect on existing behavior
- However, It can improve DX. It can be used as an object, allowing iterable code to be written, and also operates as a union type.
- The bundle size will not change significantly. This is focused on DX. (only reduce 0.2KB, test with source-map-explorer)

Why not try `as const`?


kettanaito (commented):
I'm grateful for your proposal, @nayounsang, but I will probably close this one. My reasoning is this:

- enums do become regular objects post build so we are getting the behavior you're describing as it is right now;
- I don't really see any issues with our present usage or enums to warrant a change.

In the future, it would be best to move away from `HttpMethods` being an object and instead make it an array of strings representing the supported methods.

nayounsang:
> I'm grateful for your proposal, @nayounsang, but I will probably close this one. My reasoning is this:
> 
> * enums do become regular objects post build so we are getting the behavior you're describing as it is right now;
> * I don't really see any issues with our present usage or enums to warrant a change.
> 
> In the future, it would be best to move away from `HttpMethods` being an object and instead make it an array of strings representing the supported methods.

Oh, okay, this might seem a bit rude since I did it when I didn't know much about open source culture, but thanks for the comment.
If there is a place for me to contribute later, I will give it a try!

kettanaito:
Nothing was rude! I hope I haven't voiced any of my concerns in a rude way either. I highly appreciate your contribution and glad that we could discuss it.

PR #2453, opened 2025-03-13
fix: use `tough-cookie@6` directly

### LINKED ISSUE #2568: argument name is invalid (cookie name)
### Prerequisites

### Environment check

### Browsers

### Reproduction repository

https://github.com/edikdeisling/msw-cookie-issue

### Reproduction steps

Add cookie that contains `:` in the name

### Current behavior

All requests start to fall with this error

Image

Image

### Expected behavior

According to mdn `:` is a valid symbol for cookie name

> A \<cookie-name> can contain any US-ASCII characters except for: control characters ([ASCII](https://developer.mozilla.org/en-US/docs/Glossary/ASCII) characters 0 up to 31 and ASCII character 127) or separator characters (space, tab and the characters: ( ) < > @ , ; : \ " / [ ] ? = { })

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#cookie-namecookie-value

sampsonjoliver:
Mirror of https://github.com/mswjs/msw/pull/2396, updating tough-cookie from ^5.0 to ^5.1.

- Closes #2396 
- Related to #2356
- Fixes #2568

As per [that discussion](https://github.com/mswjs/msw/pull/2396#issuecomment-2611487422), it appears the only necessary change is to update `tough-cookie` to 5.1.x in order to effectively drop the `@bundled-es-modules/tough-cookie` clone.

`pnpm build` succeeds against 5.1.x where it fails against 5.0.x.


kettanaito:
Thanks for opening this one, @sampsonjoliver. Looks good to me, let's just make sure that this update also doesn't break any CJS consumers.

kettanaito:
Looks like ESM consumers are broken:

```
  1) esm-browser.test.ts:61:5 › runs in an ESM browser project ─────────────────────────────────────

    Error: expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 4

    - Array []
    + Array [
    +   "The requested module './../../../../../../tough-cookie@5.1.2/node_modules/tough-cookie/dist/cookie/index.js' does not provide an export named 'Cookie'
    + ",
    + ]
```

Reproduction:

```sh
pnpm test:modules:browser
```

sampsonjoliver:
Hm, I recall in #2356, it was said:
> In september with the release of v5 tough-cookie was rewritten in typescript and (I'm assuming here) is probably compatible with ES without this wrapper package.

I wonder if this assumption has been validated! Perhaps we still need the esm wrapper 🤔 

I may have time to look into this during my AM hours tomorrow, or else later next week.

sampsonjoliver:
Hey @kettanaito, I started looking into the `module:browser` tests but couldn't really make heads or tails of it, in part due to unfamiliarity with the test suite and the way it's bundling/generating the module code.

I can see when running the test suite with `--ui` that there is a network request to get the tough-cookie module, shortly before the error is thrown to the console:
```
URL: http://localhost:62644/node_modules/.pnpm/tough-cookie@5.1.2/node_modules/tough-cookie/dist/cookie/index.js
Method: GET
Status Code: 200
```

and the response body matches the file in `node_modules`, including an export definition for `Cookie` like so:
```js
var cookie_1 = require("./cookie");
Object.defineProperty(exports, "Cookie", { enumerable: true, get: function () { return cookie_1.Cookie; } });
```

What is _interesting_ is that I seem to be able to modify `cookieStore.ts` to completely remove the import of `tough-cookie` (replacing instances of its members with `{} as any`), then run `pnpm build`, run the tests again, and get the same error.

There should be no reference to tough-cookie in the code under test, yet the same error is produced. This is the question that's currently stumping me - _where_ is the attempted import coming from?

sampsonjoliver:
Also, this may be of some relevance, though it does not solve my question above: https://github.com/salesforce/tough-cookie/pull/506

kettanaito:
The error seems to indicate that the cookie package is CJS-only. To properly import it, we need to do a wrapper import like this:

```ts
import cookiePkg from 'tough-cookie'

cookiePkg.Cookie
cookiePkg.Store
// etc.
```

I believe we have one or two packages we import the same way in the code base. If I may ask you to look around and try this one locally, that would be highly appreciated! 

Thank you for putting so much work into this. 

_Edit_: Checked the code quickly and no, we've dropped such imports (likely migrated to ESM wrappers for that reason). There are references online though.

kettanaito:
I've polished the changes, updated to `tough-cookie@6`, which now natively supports ESM, which is freaking fantastic. Tests are passing locally. Waiting for the CI before releasing this.

kettanaito (approved):

kettanaito:
Thanks for making this happen, @sampsonjoliver 🚀

kettanaito:
## Released: v2.11.1 🎉

This has been released in v2.11.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.11.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.11.1)

---

sampsonjoliver:
Thank you @kettanaito! This one's all you 😀

PR #2455, opened 2025-03-16
fix: solve the "Cannot read properties of undefined (reading 'url')" error

### LINKED ISSUE #2146: 'Cannot read properties of undefined (reading 'url')' when multiple browser tabs are open
### Prerequisites

### Environment check

### Browsers

### Reproduction repository

### Reproduction steps

1. install nodejs v20.12.2
2. npm i
3. npm start
4. open 2 tabs 'http://localhost:3000/'
5. after opening the second tab, you'll see the following error:
image

### Current behavior

when the second tab opened, I see 'Cannot read properties of undefined (reading 'url')' error.

Affected versions: >2.1.7
Versions that work correctly: <= 2.1.7

### Expected behavior

application should continue working when it is opened in 2 tabs.

IrishBruse:
- Closes #2240
- Related to #2193
- Fixes #2146

fix error cannot read properties of undefined (reading 'url') 

This fixes the underlying issue with cross tab talk between the service worker

This also include a repoducable react app with this bug it uses cra as its the fastest way to get a webpack example that causes this bug but i have seen it in custom webpack builds too cant seem to repro with vite

The example uses the patched version of mockServiceWorker.js if you want to see the bug revert to 2.7.3


src/mockServiceWorker.js
```diff
@@ -68,6 +68,7 @@ self.addEventListener('message', async function (event) {
             frameType: client.frameType,
           },
         },
+        clientId,
```
kettanaito, line 71: Why forward `clientId` with this event if the library doesn't process `clientId` passed with this event? Does this have some hidden effect you noticed?

src/mockServiceWorker.js
```diff
@@ -108,6 +109,10 @@ self.addEventListener('fetch', function (event) {
     return
   }
 
+  if (!activeClientIds.has(clientId)) {
```
kettanaito, line 112: Does putting this check earlier than `handleRequest` also has some effect?

We do check it right here:
https://github.com/mswjs/msw/blob/827a5dc0552f0c07822c0b178c85d6bca3d3bd6c/src/mockServiceWorker.js#L130

kettanaito (commented):
Thanks for opening this one, @IrishBruse. I really appreciate good effort on it. One thing I struggle through is understanding how these changes exactly fix the issue.

Could you please elaborate what _caused_ the issue? I will get a better understanding what the fix does then. Thanks.

IrishBruse:
I cleared both consoles and click refresh on the right(Tab B) this is the logs of client ids

image

The service worker is global and shared between tabs and it seems like there was some cross tab talk and msw takes some time before being registering where the other tab thats not been refreshed and is active causes some issues with the service worker

My memory is a little hazy on this i went down alot of rabit holes investigating this but this is how i think the events occur now with the fix.

1. Tab A initializes msw
2. The service worker makes a list of tabs that said msw is activated
3. Tab B makes a request
4. The service worker checks: Is Tab B on the list - this was missing
5. If Tab B IS on the list, the service worker might mock the request

The order it fetches the js means msw is after the other js so those first two fetch are being intercepted by the worker on the other tab which seems to be the cause of the issues

I havent confirmed but i think vite might be fetching msw differently or in a different order hence why i couldnt repro this issue with vite

image

src/mockServiceWorker.js
IrishBruse: This still returns a response if the active client isnt registered doing it early stops the response entirely

kettanaito: Oh, I see! That's a good point. Thanks for elaborating on this one ❤️

kettanaito:
@IrishBruse, thanks for sharing more details. Immensely appreciate your effort to solve this issue.

I wonder how is that scenario possible given the application awaits MSW to be registered (which is the recommended approach)? For what you describe to happen, MSW registering and the page requests have to be in a race condition, which is impossible if your application's rendering is deferred until the worker registration is complete (i.e. `await worker.start()`). 

I will try to put your reproduction scenario to an automated test and see what that gives us.

kettanaito:
I am also a little worried since the fix itself just skips the request handling altogether. With that in mind, is the problematic request expected to be handled with MSW or not? Since it will be skipped, we may be trading one issue for another here.

kettanaito:
From what I can see, this issue has to do with concurrency. I can reproduce it in your example app and in tests, but as soon as I add _anything_ to the test (like a `console.error` listener), the issue is gone. This makes it tricky to create a test case from. Trying to see what I can do...

kettanaito:
Here's my attempts to write a test for this: https://github.com/mswjs/msw/pull/2473. Doesn't work as of now. Workers are truly magical in how unpredictable they are.

kettanaito:
Sorry, I messed up your branch while rebasing. Migrated your solution to #2510. Closing in favor of that.

PR #2464, opened 2025-04-05
feat(SetupServerApi): allow using custom interceptors

### LINKED ISSUE #2460: Expose a way to pass custom interceptors
### Scope

Adds a new behavior

### Compatibility

### Feature description

👋 It looks like only `SetupServerApi` is exposed on the exports, but that class sets specific interceptors since https://github.com/mswjs/msw/pull/2000.  `SetupServerCommonApi` doesn't look like it's exported.

Is there another way to pass custom interceptors, or would you be open to adding an optional param to `SetupServerApi` or exporting `SetupServerCommonApi`?

tastypackets:
- Make interceptors an optional 2nd parameter on `SetupServerApi`, if no interceptors are passed the default 3 node.js interceptors are used.
- Update contributing file to call out Playwright chromium install when running browser tests.

resolves #2460


CONTRIBUTING.md
```diff
@@ -193,6 +193,12 @@ test('returns a mocked response', async ({ loadExample, fetch }) => {
 
 ##### Running all browser tests
 
+Make sure Playwright chromium has been installed before running browser tests.
+
+```sh
+pnpm playwright install chromium
```
kettanaito: This is kind of implied by using `playwright` (it won't run without browsers) and isn't really specific to MSW. I would rather not include this step here. With it here, we need to make sure it reflects any possible future changes to `playwright`, its browser usage, or its CLI.

test/node/msw-api/setup-server/scenarios/custom-interceptors.node.test.ts
```diff
@@ -0,0 +1,37 @@
+import { HttpResponse, http } from 'msw'
+import { SetupServerApi } from 'msw/node'
+import { FetchInterceptor } from '@mswjs/interceptors/fetch'
+import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/XMLHttpRequest'
+
+const fetchInterceptorSpy = vi.spyOn(FetchInterceptor.prototype, 'apply')
```
kettanaito: Let's drop these spies and instead:

1. Declare a request handler as you normally would.
2. Assert that the matching request receives the mocked response. 

This will give us assurance that the custom interceptor indeed works.

src/node/SetupServerApi.ts
```diff
@@ -52,11 +53,15 @@ export class SetupServerApi
...
   implements SetupServer
 {
-  constructor(handlers: Array<RequestHandler | WebSocketHandler>) {
-    super(
-      [ClientRequestInterceptor, XMLHttpRequestInterceptor, FetchInterceptor],
-      handlers,
-    )
+  constructor(
+    handlers: Array<RequestHandler | WebSocketHandler>,
+    customInterceptors: Array<{ new (): Interceptor<HttpRequestEventMap> }> = [
```
kettanaito: Hmm, technically, we can accept interceptors of any event map, like WebSocket. I need to think on this one... I am tempted to say we should do `Interceptor<any>` here as the end event map is defined by the MSW usage of particular interceptors, but this may have unexpected repercussions.

kettanaito: But probably a better solution would be to limit the supported event maps as we only have two at the moment:

```ts
Interceptor<HttpRequestEventMap | WebSocketEventMap>
```

Both types are available from the `@mswjs/interceptors` package.

What do you think?

kettanaito (changes requested):
Thanks for working on this one, @tastypackets! 

I left a few suggestions, let me know your thoughts about them. Excited to see this landing in MSW soon!

src/node/SetupServerApi.ts
tastypackets: I think we'd have to update the `SetupServerCommonApi` type, this matches that one since it extends it. However could probably make a shared type or infer type, so it only needs to be updated in 1 place long term instead of both.

https://github.com/mswjs/msw/blob/main/src/node/SetupServerCommonApi.ts#L40

kettanaito: That's okay, let's update it. The way it's annotated right now is a bug, and adding a union means we are backward-compatible anyway.

kettanaito:
Thanks for addressing the feedback, @tastypackets. 

I've pushed a few changes, we should be ready to release this now. Short summary of my changes, if you are interested:

- Reverts the type changes. No need to abstract the constructor parameters type. It brings complexity while offering very little in return.
- Changed the type of the interceptors from a constructor type to be instance type. I'm not sure why we kept constructor types `{ new(): T }`. Perhaps historically interceptors could accept arguments. Not a thing anymore.
- Updated the test to drop assertions on interceptor internals and instead assert on the public thing—the received mocked response.

kettanaito:
This looks great! Thanks once again for making this happen. 

As this is a new feature, I will see if there are any other bugfixes I can squeeze into the current version range before releasing this.

kettanaito (approved):

kettanaito:
Thanks for your work on this one, @tastypackets! Welcome to contributors.

kettanaito:
## Released: v2.8.0 🎉

---

PR #2472, opened 2025-04-11
feat: do not exhaust one-time handlers with custom predicate in resolver

### LINKED ISSUE #2399: Do not exhaust one-time handlers with custom predicate in resolver
### Scope

Improves an existing behavior

### Compatibility

## Background

As stated on https://mswjs.io/docs/best-practices/custom-request-predicate, we can implement predicate filtering by wrapping the resolver.

```ts
export const handlers = [
  http.get(
    '/user',
    withPredicate(
      // Only match "GET /user" requests that have
      // the "userId" query parameter present.
      (params) => params.has('userId'),
      ({ request, params, cookies }) => {
        return HttpResponse.json({
          name: 'John Maverick',
        })
      }
    )
  ),
  { once: true }
]
```

However, if we want to use `options.once` at this point, it will result in the `once` being consumed even if the predicate is not passed.

This makes it hard for our implemented predicate to support both `once with predicate` simultaneously.

---

[`restoreHandlers()`](https://mswjs.io/docs/api/setup-server/restore-handlers) can restore once.
But it will also reset all the handlers.

---

## Solution

So I think `Custom request predicate function` is very important. It gives us more control.

> https://github.com/mswjs/msw/issues/1804
 
I would like to know when we can expect to have this feature available, or if there are any alternative solutions to address the current issue?

ytoshiki:
## Description
- Mark the request handler as used only after its resolver function has returned a response.

resolves #2399

## Concern

There might be an edge case where a user's internal resolver doesn't return anything but still expects the request to be consumed only once.
I'm not sure how realistic this scenario is, but I thought it was worth mentioning.
Thank you.

### DIFF
--- a/src/core/handlers/RequestHandler.ts
+++ b/src/core/handlers/RequestHandler.ts
@@ -325,6 +325,7 @@ export abstract class RequestHandler<
       if (!this.resolverIterator) {
         const result = await resolver(info)
         if (!isIterable(result)) {
+          this.isUsed = result instanceof Response
           return result
         }
         this.resolverIterator =
@@ -346,8 +347,7 @@ export abstract class RequestHandler<
       if (done) {
         // A one-time generator resolver stops affecting the network
         // only after it's been completely exhausted.
-        this.isUsed = true
-
+        this.isUsed = nextResponse instanceof Response
         // Clone the previously stored response so it can be read
         // when receiving it repeatedly from the "done" generator.
         return this.resolverIteratorResult?.clone()
--- a/test/node/msw-api/setup-server/use.node.test.ts
+++ b/test/node/msw-api/setup-server/use.node.test.ts
@@ -137,3 +137,34 @@ test('throws if provided the invalid handlers array', async () => {
     '[MSW] Failed to call "use()" with the given request handlers: invalid input. Did you forget to spread the array of request handlers?',
   )
 })
+
+test('does not consume "once" handler if it does not return a response', async () => {
+  let shouldHandle = false
+
+  server.use(
+    http.get(
+      httpServer.http.url('/book/:bookId'),
+      ({ request }) => {
+        if (!shouldHandle) {
+          return undefined
+        }
+        return HttpResponse.json({ title: 'One-time override' })
+      },
+      { once: true },
+    ),
+  )
+
+  const res1 = await fetch(httpServer.http.url('/book/abc-123'))
+  expect(res1.status).toBe(200)
+  expect(await res1.json()).toEqual({ title: 'Original title' })
+
+  shouldHandle = true
+
+  const res2 = await fetch(httpServer.http.url('/book/abc-123'))
+  expect(res2.status).toBe(200)
+  expect(await res2.json()).toEqual({ title: 'One-time override' })
+
+  const res3 = await fetch(httpServer.http.url('/book/abc-123'))
+  expect(res3.status).toBe(200)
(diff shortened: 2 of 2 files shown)


src/core/handlers/RequestHandler.ts
```diff
@@ -325,6 +325,7 @@ export abstract class RequestHandler<
       if (!this.resolverIterator) {
         const result = await resolver(info)
         if (!isIterable(result)) {
+          this.isUsed = result instanceof Response
```
ytoshiki, line 328: I wasn’t able to apply the change in the `run` method [as you suggested](https://github.com/mswjs/msw/issues/2399#issuecomment-2561762949), because we need to take generator use cases into consideration.

Also, I kept this line as is to ensure parallel processing works correctly.

https://github.com/mswjs/msw/blob/e3234fdce5a36f41c77719492e2eca0f77d7aaa6/src/core/handlers/RequestHandler.ts#L280

Please let me know if you have any suggestions or thoughts on this.

kettanaito:
Hi, @ytoshiki. Thank you so much for submitting this. 

Please give me some time to think about the best direction here. Changes like this are pretty fundamental, and I don't have enough focus right now to be the best judge for them. I will share my thoughts on this once I have proper time to spend on this issue. 

My initial impression is that I vaguely recall that request handlers were considered used as soon as they _matched_, even if they didn't return a mocked response. This may or may not make sense in the present, just something I think I remember.

ytoshiki:
> Hi, @ytoshiki. Thank you so much for submitting this.
> 
> Please give me some time to think about the best direction here. Changes like this are pretty fundamental, and I don't have enough focus right now to be the best judge for them. I will share my thoughts on this once I have proper time to spend on this issue.
> 
> My initial impression is that I vaguely recall that request handlers were considered used as soon as they _matched_, even if they didn't return a mocked response. This may or may not make sense in the present, just something I think I remember.

@kettanaito 
Thank you for your response.
No problem at all. I completely understand. I appreciate you taking the time.

Hardanish-Singh (approved):

kettanaito:
I don't think we can reliably solve this with the present API, unfortunately. 

The issue is that this solution breaks how request handlers work in general, and does that only for one-time handlers. Right now, as soon as any request handler _matches_ a request, that request is considered handled, even if no handlers returned a mocked response. With this change, that would no longer be the case as we are introducing an escape hatch to facilitate custom predicate within the response resolver (in the context of one-time handlers but that's less important here). 

Instead, the correct way forward is implementing #1804. That will allow developers to provide a _function_ as the first argument to a request handler, acting as a predicate function. It would works predictably and consistently for any handlers, one-time or not.

```ts
http.get(({ request }) => {
  return request.url.includes('foo')
}, resolver, { once: true })
```

> The predicate function taps into the predicate phase of a handler directly, _not exhausting it_ because the exhaustion logic lives in the resolver and fires when the resolver is ran.

The main difficulty of implementing function-as-predicate is cloning the `request` preemptively and correctly so the memory usage is adequate. Ideally, we should tell if the predicate argument is a function ahead of time and only if that's the case, clone the `request` and provide the cloned version to that predicate function. 

@ytoshiki, does this make sense? Would you be interested in giving the predicate implementation a try?

ytoshiki:
@kettanaito 

Thanks for the detailed explanation. That makes perfect sense.
I’m interested in giving [the predicate implementation](https://github.com/mswjs/msw/issues/1804) a try.
Thank you!

PR #2476, opened 2025-04-18
fix(ws): ignore `/socket.io/` path prefix during url matching

### LINKED ISSUE #2477: `TransportError: websocket error` when mocking Socket.IO WebSockets
### Discussed in https://github.com/mswjs/msw/discussions/2470

<sup>Originally posted by **colinbes** April 10, 2025</sup>
I am battling with setting up mocked support for sockets on vue client.

I have mixed mswjs handler of socket and http requests.

I am using socket.io-client 4.8.1 and am setting up socket using (I have also tried not setting transports).

```
export const socket = io(URL, {transports: ["websocket"], autoConnect: true}); 
```
where URL is `ws://localhost:3002`

For handler setup I have the following setup:

```
const mysocket = ws.link('ws://localhost:3002/socket.io')

export const handlers: Array<RequestHandler | WebSocketHandler> = [
  mysocket.addEventListener('connection', ({ client }) => {
    console.log("got client", client.url)
  }),
  
  http.get(getAppUsersUrl, () => {
    return HttpResponse.json(mockedUsers)
  }),
 ...
]
```
In browser I see error `websocket.js:119 WebSocket connection to 'ws://localhost:3002/socket.io/?EIO=4&transport=websocket' failed: `

In web inspector network tab I see repeated requests for`ws://localhost:3002/socket.io/?EIO=4&transport=websocket` hence my reasoning for using `ws.link('ws://localhost:3002/socket.io')` 

I never trigger breakpoint and never see message `got client xxx` as defined in handler addEventListener.

If I fire up backend express server with socket support on 3002 then my client app does connect to socket server so I believe my connection string/url to be correct which brings me to question the url 'ws://localhost:3002/socket.io' I am using for mysocket in define ws.link

Appreciate nudges in right direction.
</div>

kettanaito:
- Fixes #2477 

## Changes

- The `/socket.io/` suffix of the actual WebSocket client URL will be _ignored_ if present. User-defined `socket.io` part of the pathname will be preserved (if it's not a prefix).
- Adds missing unit tests for the `parse` method of `WebSocketHandler`.

## Motivation

Right now, MSW does a literal path matching between `connection.client.url` and the handlers'` this.url`. That means that it will never match SocketIO connection requests because they include `/socket.io/` in the path while the WebSocket link doesn't.

The users shouldn't leak such implementation details to their handlers so I propose to make an exception and drop the `/socket.io/` path prefix, if any, during the WebSocket handler parsing.

### DIFF
--- a/src/core/handlers/WebSocketHandler.test.ts
+++ b/src/core/handlers/WebSocketHandler.test.ts
@@ -1,0 +1,160 @@
+import type { WebSocketConnectionData } from '@mswjs/interceptors/WebSocket'
+import { WebSocketHandler } from './WebSocketHandler'
+
+describe('parse', () => {
+  it('matches an exact url', () => {
+    expect(
+      new WebSocketHandler('ws://localhost:3000').parse({
+        event: new MessageEvent('connection', {
+          data: {
+            client: {
+              url: new URL('ws://localhost:3000'),
+            },
+          } as WebSocketConnectionData,
+        }),
+      }),
+    ).toEqual({
+      match: {
+        matches: true,
+        params: {},
+      },
+    })
+  })
+
+  it('ignores trailing slash', () => {
+    expect(
+      new WebSocketHandler('ws://localhost:3000').parse({
+        event: new MessageEvent('connection', {
+          data: {
+            client: {
+              url: new URL('ws://localhost:3000/'),
+            },
+          } as WebSocketConnectionData,
+        }),
+      }),
+    ).toEqual({
+      match: {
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v2.7.5 🎉

This has been released in v2.7.5!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.7.5)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.7.5)

---

PR #2485, opened 2025-05-01
chore(first-attempt): migrate to ESM internally

kettanaito:
This aims to migrate to ESM for internal purposes only. This implies no public changes to the users of the library. This should also make the public-facing ESM-only change easier in the future. 

## TODO

- [x] @open-draft/test-server/lib/http.js (directory import not allowed; might be broken export conditions in that package)
  - Playwright just refuses to resolve these export conditions (tried on `@epic-web/test-server`). Will leave this alone. 
- [x] **Removing `main` will break React Native as it doesn't support `exports`.** 
  - RN has their own export conditions they seem to respect (see https://github.com/reduxjs/redux-toolkit/pull/4887/files)


kettanaito:
## Unknown compiler option `xyz`

✅ Fixed by setting `strict: false` in all TS configs used for type tests. This isn't ideal but that seems to be the only way to ignore unknown TS options, which happens because Vitest generates a temporary tsconfig that's fixed to a certain version of TS (doesn't respect the tsconfig path provided for that case).

kettanaito:
## Released: v2.8.0 🎉

---

PR #2490, opened 2025-05-05
feat: migrate to ESM internally, remain CJS-first

kettanaito:
- Closes #2485. Mostly the same but cleaner diff. 
- Closes #2492 

## Motivation

I want for everything happening in the MSW repo to be ESM. The includes the CLI, scripts, plugins, tests of all levels, etc. **This has no effect on the end consumer**. The package remains CJS-first, dual CJS/ESM publish. Having everything in ESM internally will ease the future migration to ESM-only.

## Roadmap

- [x] `pnpm test:browser`. Currently failing with Playwright resolving `msw` from the root directory as CJS (which it is per `type`) and fails on any named export (which is odd). Is there a way to force ESM version of MSW for Playwright tests?
  - It's the Node.js side (the tests themselves) that fails here. In-browser usage resolves `msw` via `webpack` and should be okay. 
- [x] `pnpm test:ts`. Failed to resolve `msw` alias from `tsconfig.json` at all. No idea why.
  - I suspect that the `"msw": [".."]` alias in `tsconfig.json` disregards any export conditions and only expects the root-level `types` be present. Since I deleted those, that alias fails and so do the type tests.


.github/workflows/compat.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       matrix:
         # As a general rule, when adding a new version, remove the oldest version
         # Check if the oldest version is EOL or not first.
-        ts: ['4.8', '4.9', '5.0', '5.1', '5.2', '5.3', '5.4', '5.5']
+        ts: ['5.0', '5.1', '5.2', '5.3', '5.4', '5.5']
```
wojtekmaj, line 57: Aren't you essentially breaking the rule stated above? Should the list be expanded forwards?

cli/init.js
```diff
@@ -33,6 +33,7 @@ module.exports = async function init(args) {
     // You can also provide the "--no-save" option, and then "args.save"
     // will equal to false.
     else if (args.save == null) {
+      // eslint-disable-next-line no-console
```
wojtekmaj, line 36: Nit: feels like you could minimize the diff by doing ESLint rule changes separately.

config/plugins/esbuild/resolveCoreImportsPlugin.ts
```diff
@@ -1,6 +1,6 @@
 import { Plugin } from 'esbuild'
-
-const { replaceCoreImports } = require('../../replaceCoreImports')
+import { replaceCoreImports } from '../../replaceCoreImports'
```
wojtekmaj: Perhaps you'd want to specify an extension here to make it to ESM spec?

test/browser/graphql-api/anonymous-operation.test.ts
```diff
@@ -1,4 +1,4 @@
-import { HttpServer } from '@open-draft/test-server/http'
+import { HttpServer } from '@open-draft/test-server/lib/http.js'
 import { test, expect } from '../playwright.extend'
 import { gql } from '../../support/graphql'
 import { waitFor } from '../../support/waitFor'
```
wojtekmaj, line 4: Missing extensions?

wojtekmaj: If yes then that would apply globally to all tests

test/browser/msw-api/integrity-check.test.ts
```diff
@@ -1,18 +1,19 @@
-import * as fs from 'fs'
-import * as path from 'path'
+import fs from 'node:fs'
 import { test, expect } from '../playwright.extend'
 import copyServiceWorker from '../../../config/copyServiceWorker'
-import { version } from '../../../package.json'
-
-const { SERVICE_WORKER_SOURCE_PATH } = require('../../../config/constants')
+import packageJson from '../../../package.json' assert { type: 'json' }
```
wojtekmaj, line 4: Wait wait wait. Dragons here?

AFAIK, import assertions were removed in Node.js v22.0.0, while import attributes became available in Node.js v18.20.5, v20.18.3, and v22.12.0.

test/browser/msw-api/setup-worker/stop/removes-all-listeners.test.ts
```diff
@@ -15,9 +15,12 @@ test('removes all listeners when the worker is stopped', async ({
   fetch,
 }) => {
   const consoleSpy = spyOnConsole()
-  await loadExample(require.resolve('./removes-all-listeners.mocks.ts'), {
-    skipActivation: true,
-  })
+  await loadExample(
+    new URL('./removes-all-listeners.mocks.ts', import.meta.url),
```
wojtekmaj, line 19: Not: How stupid it would be to just pass an example name here and resolve it INSIDE loadExample? Quite a lot of repetition here.

test/node/ws-api/ws.stop-propagation.test.ts
```diff
@@ -28,7 +28,7 @@ afterAll(async () => {
 })
 
 it('stops propagation for client "message" event', async () => {
-  const clientMessageListener = vi.fn<[number]>()
+  const clientMessageListener = vi.fn<(input: number) => void>()
```
wojtekmaj, line 31: Is this change actually related to ESM migration?

test/typings/vitest.config.ts
```diff
@@ -0,0 +1,53 @@
+import fs from 'node:fs'
+import url from 'node:url'
+import path from 'node:path'
+import { defineConfig } from 'vitest/config'
+import { invariant } from 'outvariant'
+import tsPackageJson from 'typescript/package.json' assert { type: 'json' }
```
wojtekmaj, line 6: My previous concern about type assertions applies here as well.

wojtekmaj (commented):
I thoroughly reviewed the whole PR, hopefully these comments will render useful :)

.github/workflows/compat.yml
kettanaito: It will be! I am keeping the change radius as small as possible with this change because it's already affects a lot of things. Once this is merged, I will follow up with extending the TS version support.

cli/init.js
kettanaito: I could but `no-console` is considered an error in our ruleset, which means it breaks the build. I have to inline these exceptions.

test/browser/msw-api/integrity-check.test.ts
kettanaito: They were but the new `with` assertion was never backported. So `assert` breaks Node.js v22+ while `with` breaks Node.js <v22. We don't support Node.js v22 officially, so we are staying with `assert` until we do. 

This is a test file anyway, which means the version of Node.js is our lowest supported version. As of now, that is Node.js v18.

test/browser/msw-api/setup-worker/stop/removes-all-listeners.test.ts
kettanaito: `import.meta.url` will point to a different path if it's called in `loadExample`. How would you get the path of the module that calls `loadExample()`? Sounds like much more logic than defining a simple URL... But I might be not seeing what you're seeing.

test/node/ws-api/ws.stop-propagation.test.ts
kettanaito: No but Vitest 2.0 has bugs related to executing ESM so a 3.0 bump is necessary. This is the only change that comes with that bump.

test/browser/graphql-api/anonymous-operation.test.ts
kettanaito: I don't believe extensions are a part of ESM spec per se. They are a part of module resolution strategies and it seems you can use imports without extensions while in ESM (like you see above, nothing errors). Based on the `moduleResolution` of the project, extension might or might not be required.

kettanaito:
## Vitest, ESM, and typechecking

- Vitest generates temporary TS configs during the typecheck run but those configs are fixed to the _modern version of TS_. This makes the type tests for older versions _fail_ with "Unknown compiler option XYZ". ✅ Set `strict: false`.
- `null` and `undefined` began satisfying type arguments, like `GraphQLVariables`. Caused by `strict: false` and solved by `strictNullChecks: true`.
- TypeScript 5.0 and 5.1 type tests were failing because `module: preserve` wasn't a thing in those versions. ✅ Set `module: esnext` for those two targets. 
- TypeScript was failing with "Unknown type WeakKey" and other errors from the Vitest source. ✅ Never validate third-party libs, set `skipLibCheck: true`.

kettanaito:
## Released: v2.8.0 🎉

---

PR #2491, opened 2025-05-06
chore(setup): fixed `test:browser`

Andarist:


test/tsconfig.json
```diff
@@ -13,11 +13,6 @@
     "esModuleInterop": true,
     "declaration": false,
     "noEmit": true,
-    "baseUrl": ".",
-    "paths": {
-      "msw": ["../"],
-      "msw/*": ["../*"]
-    },
```
Andarist, line 20: This prevents `package.json#exports` lookup, see: https://github.com/microsoft/playwright/blob/a84ba4db691e7d2a51ff5b2460c9946d406edfc0/packages/playwright/src/util.ts#L365-L375

Andarist: You might have to *build* before you run your tests now. If this is a problem let's figure out the solution to this once we make everything functional

Andarist: Oh, and I'd really just encourage you to get rid of the other `tsconfig.json#paths` you have. My rule of thumb when it comes to `tsconfig.json#paths` is simple: "just don't". I don't think it's worthwhile to save yourself from a few relative paths - especially in the context of a *library*

package.json
```diff
@@ -288,6 +288,7 @@
     "json-bigint": "^1.0.0",
     "knip": "^5.51.1",
     "lint-staged": "^15.2.10",
+    "msw": "workspace:*",
```
Andarist, line 291: this is a funny thing, normally "self resolve" would just work with a modern resolution algorithm (as in - importig/requiring msw from within the msw package). But that only works if your closest `package.json` has a matching `.name`.

But you are introducing a new `package.json` scope to override the module's type here:
https://github.com/mswjs/msw/blob/faed5c016c91e2f84f491d34c9ede92d43dc53e8/test/package.json

and that breaks the node's lookup algorithm for the self-resolve. So to work around this, I have just added an explicit self-devDependency so a normal node_modules lookup can find this.

Question: why are you trying to have the root as commonjs and then override specific directories with module type? I'd probably try the opposite - just make the root a module and then override with commonjs if needed anywhere or just use .cjs if needed

kettanaito: > Question: why are you trying to have the root as commonjs and then override specific directories with module type?

Because changing from no `type` (which is implied as CJS) to `type: module` is a breaking change that makes MSW ESM-first. It will break consumers requiring MSW in their apps. I don't want to do that right now. 

I am adding explicit `type: commonjs` simply to be explicit and keep the current behavior. The package must remains CJS-first. What I'm pursuing with this change is to make all of internals ESM.

kettanaito:
Thank you so much, @Andarist 🙏 From what I can see, the current `test:browser` failures are related to CJS leftovers. I will cherry-pick the changes from another PR, I've already solve these. Excited to see how this goes.

test/tsconfig.json
kettanaito: > You might have to build before you run your tests now.

That was always the case for browser tests. I believe it wasn't the case for node tests but that's a small price to pay for having aliases working.

package.json
Andarist: > Because changing from no type (which is implied as CJS) to type: module is a breaking change that makes MSW ESM-first. It will break consumers requiring MSW in their apps. I don't want to do that right now.

I don't fully see this argument. The `package.json#type` is just a default value that you can "override". If you override all places then your CJS consumers won't notice a thing. The same rule applies to the opposite situation, the one that you are trying to achieve here. You have a CJS-first repo, but you provide ESM entrypoints so your consumers won't even know that it's configured as CJS.

kettanaito:
Looks like `msw` imports are still broken in tests:

```
> msw@2.7.6 test:browser /Users/kettanaito/Documents/web/mswjs/msw
> playwright test -c ./test/browser/playwright.config.ts

(node:47745) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
Error: Cannot find package 'msw' imported from /Users/kettanaito/Documents/web/mswjs/msw/test/browser/rest-api/response-patching.test.ts
```

This _might_ be a TS-only issue. TS fails to resolve types for `msw` imports at all.

I'm getting this on Node.js v18.20.4 and TypeScript 5.5.2, if that makes any difference.  

_Edit:_ I don't quite understand how come the CI tests are not having this error. Changing to `type: module` has no effect on this error.

package.json
kettanaito: Perhaps you're right. I simply fear the cases where people do `require('msw')`. I don't want those to start failing out of a sudden because the root-level `type` has been set to `module`. 

Keep in mind that we are also talking about _really old_ tools. Tools that don't know what `exports` is. I know for a fact there are MSW consumers that rely on those tools. Changing the module `type` in that case is a breaking change in every sense of the word.

kettanaito: I suppose we do have module operability tests to keep us somewhat safe and validate if what you're proposing won't break CJS consumers. I can try that.

Andarist:
> Looks like msw imports are still broken in tests:

This is unexpected. `msw` resolved fine when running `test:browser` for me before I pushed this. The job run on my commit suggests the same (240 tests passed): https://github.com/mswjs/msw/actions/runs/14855375796/job/41707378564

> I'm getting this on Node.js v18.20.4 and TypeScript 5.5.2, if that makes any difference.

`pnpm test:browser test/browser/rest-api/response-patching.test.ts` runs just OK for me on node v18.20.3. But overall, ye - Playwright is using node itself (with customizations) to resolve and load those modules. So the results could be specific to a node version (I doubt the `0.0.1` difference is at fault here though)

---

Either way, I'll take a look at the current state of things a little bit later :p

kettanaito:
> I doubt the 0.0.1 difference is at fault here though

Hah! 0.0.1 _did_ make a difference. Now I have the tests running on v18.20.8. 

I think your changes fixed the browser tests! 🎉 I will wrap this up and merge it.

Andarist:
I was running on an older one that yours though. That's not a great sign :/

kettanaito:
Regressions happen. We should be running on the latest of any major anyway. I don't treat this as an issue.

Andarist:
Ah, so it works on one that is even newer. I agree, that's fine 👍

package.json
Andarist: > Perhaps you're right. I simply fear the cases where people do require('msw'). I don't want those to start failing out of a sudden because the root-level type has been set to module.

Whatever tool that respects `package.json#type` should also respect `package.json#exports`. In that situation, those tools will resolve through `exports` first and they will just find relevant `require` conditions there (or `module-sync`).

> Keep in mind that we are also talking about really old tools. Tools that don't know what exports is. I

For those just keep `package.json#main` pointing to the CJS build. If they are unaware of `package.json#exports` they are surely also not aware of `package.json#type`

kettanaito:
@Andarist, you, sir, are a wizard once again. Thank you so much!

PR #2492, opened 2025-05-06
chore: type module with require fallbacks

jacob-ebey:


package.json
```diff
@@ -2,10 +2,9 @@
   "name": "msw",
   "version": "2.7.6",
   "description": "Seamless REST/GraphQL API mocking library for browser and Node.js.",
-  "type": "commonjs",
-  "main": "./lib/core/index.js",
-  "module": "./lib/core/index.mjs",
-  "types": "./lib/core/index.d.ts",
+  "type": "module",
```
kettanaito, line 5: This is a breaking change even if we define `require` fallbacks. Bear in mind, MSW is also used with really old tools that won't even look at our export conditions. To them, `require('msw')` that has `type: module` will be a runtime exception.

package.json
```diff
@@ -2,10 +2,9 @@
...
+  "main": "./lib/core/index.mjs",
```
kettanaito, line 6: I heard the `main` field should be discouraged as it's effectively ignored by modern tooling.

package.json
```diff
@@ -17,50 +16,42 @@
...
-        "default": {
-          "types": "./lib/core/index.d.ts",
-          "default": "./lib/core/index.js"
-        }
-      },
       "import": {
         "types": "./lib/core/index.d.mts",
         "default": "./lib/core/index.mjs"
       },
-      "default": {
```
kettanaito, line 34: Removing the `default` condition means that if `msw` import is resolved and none of the above conditions match, the import will fail as opposed to falling back to the CJS output. I believe this is a mistake.

kettanaito:
## Released: v2.8.0 🎉

---

PR #2500, opened 2025-05-12
fix(HttpResponse): export `StrictResponse` type again with added deprecation notice

### LINKED ISSUE #2499: FIY: Release v2.8.0 with #2118 introduced a breaking change
I just wanted to inform you that the release of #2118 introduced a breaking change. It removed the export from the `StrictResponse<BodyType>` interface in favor of the new generic argument for `HttpResponse<BodyType>`.

I noticed it today when my automatic [compatibility test in OpenAPI-MSW](https://github.com/christoph-fricke/openapi-msw/actions/runs/14929502137/job/41941886879) broke. OpenAPI-MSW has been required to use `StrictResponse` for its "type magic". Surprisingly, OpenAPI-MSW still appears to be usable for consumers with the latest version of MSW.

Even though it somehow still works, I think it is better fixed by switching from `StrictResponse` to the now generic `HttpResponse`. Feels wrong to rely on an export that no longer exists and just "somehow still works"... Annoyingly, fixing this tiny change requires a major release of OpenAPI-MSW since I have to bump the peer dependency on MSW from "^2.7.0" to at least "^2.8.0".

I totally understand the intention of only exporting `HttpResponse` and I probably would have done the same. It feels wrong/unnecessary to export both things. Sadly, it just has introduced a breaking change.

How does this change effect the rest of the ecosystem? While it feels overkill, I am okay with updating and releasing a new major of [OpenAPI-MSW](https://github.com/christoph-fricke/openapi-msw). What do you think about this? Should I move forward with the update or do you want to reintroduce the export in MSW and remove it later with a v3 version?

christoph-fricke:
This PR implements the changes you (@kettanaito) proposed in #2499. I added an additional deprecation notice to the `StrictResponse` type to guide users to `HttpResponse` instead.

closes #2499

### DIFF
--- a/src/core/HttpResponse.ts
+++ b/src/core/HttpResponse.ts
@@ -19,7 +19,7 @@ export interface StrictRequest<BodyType extends JsonBodyType> extends Request {
 /**
  * Opaque `Response` type that supports strict body type.
  *
- * @deprecated Will be removed in v3. Please use {@link HttpResponse} instead.
+ * @deprecated Please use {@link HttpResponse} instead.
  */
 export type StrictResponse<BodyType extends DefaultBodyType> =
   HttpResponse<BodyType>
--- a/src/core/HttpResponse.ts
+++ b/src/core/HttpResponse.ts
@@ -38,7 +38,7 @@ export type StrictResponse<BodyType extends DefaultBodyType> =
 export class HttpResponse<
   BodyType extends DefaultBodyType,
 > extends FetchResponse {
-  [bodyType]: BodyType = null as any
+  readonly [bodyType]: BodyType = null as any
 
   constructor(body?: NoInfer<BodyType> | null, init?: HttpResponseInit) {
     const responseInit = normalizeResponseInit(init)
--- a/src/core/HttpResponse.ts
+++ b/src/core/HttpResponse.ts
@@ -18,10 +18,11 @@ export interface StrictRequest<BodyType extends JsonBodyType> extends Request {
 
 /**
  * Opaque `Response` type that supports strict body type.
+ *
+ * @deprecated Will be removed in v3. Please use {@link HttpResponse} instead.
  */
-interface StrictResponse<BodyType extends DefaultBodyType> extends Response {
-  readonly [bodyType]: BodyType
-}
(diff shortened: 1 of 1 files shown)


src/core/HttpResponse.ts
```diff
@@ -34,10 +35,9 @@ interface StrictResponse<BodyType extends DefaultBodyType> extends Response {
  *
  * @see {@link https://mswjs.io/docs/api/http-response `HttpResponse` API reference}
  */
-export class HttpResponse<BodyType extends DefaultBodyType>
-  extends FetchResponse
-  implements StrictResponse<BodyType>
```
christoph-fricke, line 39: With `StrictResponse` aliasing `HttpResponse`, it seems wrong to implement the type. While TypeScript does not complain about it, it would just express "`HttpResponse` implements `HttpResponse`". Seems unnecessary.

kettanaito: Yeah, good catch. As long as HttpResponse has the same body symbol type, it should be okay to drop `implements`.

src/core/HttpResponse.ts
```diff
@@ -18,10 +18,11 @@ export interface StrictRequest<BodyType extends JsonBodyType> extends Request {
 
 /**
  * Opaque `Response` type that supports strict body type.
+ *
+ * @deprecated Will be removed in v3. Please use {@link HttpResponse} instead.
```
kettanaito: Let's drop the v3 promise and just recommend using a different type instead.

src/core/HttpResponse.ts
christoph-fricke: Just to be sure, I marked the `bodyType` field on `HttpResponse` as `readonly` as well to match the previous interface field exactly.

kettanaito:
Thanks for fixing this, @christoph-fricke 👍

kettanaito:
## Released: v2.8.3 🎉

This has been released in v2.8.3!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.8.3)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.8.3)

---

PR #2508, opened 2025-05-27
fix: update `@mswjs/interceptors` to 0.38.7

### LINKED ISSUE #2405: `MockHttpSocket` and `MockHttpAgent` are not being garbage collected
### Prerequisites

### Environment check

### Node.js version

22.12.0

### Reproduction repository

https://github.com/rossipedia/msw-leak-test

### Reproduction steps

Instructions are in the reproduction repo, but they are essentially:

1. clone repro
2. `npm install`
3. `./demo.sh`

### Current behavior

> [!NOTE]
> This is a follow up to my colleague's issue #2390

Enabling MSW with no handlers and then load-testing using `ab` causes heap usage to balloon, even when manually invoking garbage collection.

As best I can tell, the largest percentage of retained objects are primarily:

- `MockHttpSocket`
- `MockHttpAgent`

I do not believe this is related to `express`, as only `ClientRequest` is intercepted, and I don't think `express` or `http.createServer` use `ClientRequest`, only `IncomingMessage` and `ServerResponse` (I could be wrong about that though).

### Expected behavior

Memory usage should be stable under load when simply enabling MSW in a node application.

kettanaito:
- Potentially fixes #2405

### DIFF
--- a/package.json
+++ b/package.json
@@ -299,7 +299,7 @@
     "typescript": "^5.5.2",
     "undici": "^6.20.0",
     "url-loader": "^4.1.1",
-    "vitest": "^3.1.3",
+    "vitest": "^3.1.4",
     "vitest-environment-miniflare": "^2.14.4",
     "webpack": "^5.95.0",
     "webpack-http-server": "^0.5.0"
--- a/package.json
+++ b/package.json
@@ -241,7 +241,7 @@
     "@bundled-es-modules/statuses": "^1.0.1",
     "@bundled-es-modules/tough-cookie": "^0.1.6",
     "@inquirer/confirm": "^5.0.0",
-    "@mswjs/interceptors": "^0.37.0",
+    "@mswjs/interceptors": "^0.38.7",
     "@open-draft/deferred-promise": "^2.2.0",
     "@open-draft/until": "^2.1.0",
     "@types/cookie": "^0.6.0",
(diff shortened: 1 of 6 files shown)


kettanaito:
## Failing CI

The tests are failing for two reasons:

1. Interceptors now perform one additional `request.clone()` so that the `response` events could read the request body, too.
2. The `request` reference in the `request` and `response` events does _not_ point to the same request instance anymore. That kind of make sense because if it did and you read the request body in your `request` listener, you wouldn't be able to read it in the `response` listener.

I am suspecting that these two issues might be solved if we don't clone the request preemptively and instead delegate that responsibility to the end consumer. I think this should still work:

```ts
.on('request', async ({ request }) => {
  const data = await request.clone().json()
})
```

I need to find the reasoning we've added that request clone to Interceptors. 

_Edit:_ That clone was added not due to the consumer reasons but because `pureFetch(request)` **will always read the request's stream**. And since you cannot clone a request object that's been read, we've introduced a preemptive clone so the consumers can still consume the request stream in the `response` listeners. 

I think that change makes sense and we need to adjust your tests to not care about the referential equality but instead check that the request object describes the same request across different life cycle events.

kettanaito:
## Released: v2.8.5 🎉

This has been released in v2.8.5!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.8.5)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.8.5)

---

PR #2510, opened 2025-05-27
feat: send `request` reference within the `RESPONSE` event

### LINKED ISSUE #2146: 'Cannot read properties of undefined (reading 'url')' when multiple browser tabs are open

kettanaito:
> [!WARNING]
> This change will require you to upgrade your worker script via `npm init`. If you've saved the worker script location in the past using the `--save` flag, the worker script will be updated automatically. 

- **Follow up to #2455**
- Related to #2193
- Fixes #2146
- Closes #2240

Test attempt at https://github.com/mswjs/msw/pull/2473. Highly flaky due to how the client registrations are resolved across tabs. I don't believe it's worth it to pursue that test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -266,6 +266,7 @@
     "@types/express": "^4.17.21",
     "@types/json-bigint": "^1.0.4",
     "@types/node": "~18.19.28",
+    "@types/serviceworker": "^0.0.136",
     "@typescript-eslint/eslint-plugin": "^8.8.1",
     "@typescript-eslint/parser": "^8.8.1",
     "@web/dev-server": "^0.4.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -90,6 +90,9 @@ importers:
       '@types/node':
         specifier: ~18.19.28
         version: 18.19.28
+      '@types/serviceworker':
+        specifier: ^0.0.136
+        version: 0.0.136
       '@typescript-eslint/eslint-plugin':
         specifier: ^8.8.1
         version: 8.16.0(@typescript-eslint/parser@8.16.0(eslint@8.57.0)(typescript@5.5.2))(eslint@8.57.0)(typescript@5.5.2)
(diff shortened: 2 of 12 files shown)


kettanaito:
Hi, @IrishBruse. I hope you don't mind me moving the discussion around the fix here.

## Update

Your solution does work but it doesn't account for nested clients (i.e. requests coming from iframes). In those cases, the `clientId` will point to the _child_ frame and not the frame that has registered the worker (the parent). Thus, the child ID is never be in the `activeClientIds`, resulting in requests originating from the child frames being unhandled.

That's a pickle because resolving the actual client id is an asynchronous action (we already have the `resolveMainClient()` utility in the worker) and you cannot cal asynchronous code before `event.respondWith()`, that's a no-op. Even reading the `Client` from `clientId` is an async action.

kettanaito:
## Update

Since the original proposal breaks the nested frames support, I'm trying to remove the need in `context.requests`, in the first place. The `RESPONSE` event from the worker script will now forward the entire serialized request associated with that response. It's a bit more expensive since now it would have to (1) clone the request one more time in order to (2) read its body for the request instance to be properly transferred to the client. 

That being said, I don't expect any performance degradation or impact, for that matter, from this change.

dmezei:
I updated my repo to use this build during our tests (vitest browser). This version seem to fix the `Cannot read properties of  undefined (reading 'url')` error. Thanks for the fix! 

Note: I've seen the comment about the worker script. In my case no such thing was required. I created this little vite plugin for msw running in the browser: 
```
export const mockServiceWorkerPlugin = (): PluginOption => ({
    name: 'add-mock-server',
    apply: 'serve',
    configureServer: (server) => {
        server.middlewares.use((req, res, next) => {
            if (!req.url?.endsWith('mockServiceWorker.js')) {
                return next();
            }
            res.setHeader('Content-Type', 'text/javascript');
            res.statusCode = 200;
            res.write(readFileSync(resolve('./node_modules/msw/lib/mockServiceWorker.js')));
            res.end();
        });
    },
});
```

kettanaito:
@dmezei, thank you so much for confirming!

I was just thinking about making a Vite plugin for MSW's worker yesterday! What a coincidence. Still weighing the pros and cons but thinking it would be worth it. We can have a simple plugin that appends a new route handler to Vite's dev server and make it serve `/mockServiceWorker.js` from node_modules. No more worker updates!

If you feel like it, you can suggest this as a feature so we can track it. Thanks.

kettanaito:
## Released: v2.9.0 🎉

This has been released in v2.9.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.9.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.9.0)

---

PR #2521, opened 2025-06-02
Use discriminated union more properly

Andarist:


Andarist:
TS 5.0 fails likely because it lacks https://github.com/microsoft/TypeScript/pull/53098 . This should be fixable and I could work on it, if you'd plan to merge this

kettanaito:
@Andarist, how is that you know all the secrets of the universe?

kettanaito:
Yes, I absolutely plan on merging this! If you have a moment to look into that 5.0 issue, I'd be immensely grateful!

Andarist:
Actually, this difference is caused by https://github.com/microsoft/TypeScript/pull/53709 . In this case, it changed **error locations**. From a quick look at the errors reported by 5.0 and 5.1+ - this works exactly the same. We get the same number of expected errors, but their locations slightly changed and `// @ts-expect-error` only works for the very next line.

A workaround for this would be to fold those calls into single lines and disable pretty printing. Thoughts? ;p

Andarist:
> @Andarist, how is that you know all the secrets of the universe?

I don't :P but I **do** feel comfortable in the TS corner of the universe 😉 just like you feel comfortable in the service worker/mocking/whatnot corner ;p

kettanaito:
@Andarist, I think folding and prettier-ignore sounds like a good solution as long as we put a `/** @note */` explaining that above and link to this conversation!

Andarist:
@kettanaito done

kettanaito:
Thank you so much, @Andarist ❤️

PR #2522, opened 2025-06-03
fix(sse): error on missing required data

Andarist:


kettanaito:
Thanks, @Andarist. I've moved the test case out, added one more for optional message `data` type. Let's keep type behaviors in their own test cases from now on. I will refactor the existing big test case to split it.

PR #2528, opened 2025-06-06
fix: update `@mswjs/interceptors` to support WebSocket server protocol

kettanaito:
> [!WARNING]
> This removes type annotations for `client.socket` and `server.socket`. **Those were there by mistake**. The raw `WebSocket` reference mustn't be exposed to the user and wasn't documented, to begin with. The raw reference is still there, just not annotated to emphasize that it's not a part of the public API (`client['socket']`/`server['socket']`).

- Follow up to https://github.com/mswjs/interceptors/pull/726

## Roadmap

- [x] https://github.com/mswjs/interceptors/pull/727. Previously, `client.socket` and `server.socket` were accessible due to us relying on `WebSocketClientConnection` and `WebSocketServerConnection`. The protocols for client and server _do not require you to implement the `socket` property_. So we never receive those properties from the `connection` (although they _are_ there). 
- [x] Adjust MSW's `connection` event data _not_ to mirror the Interceptor's one. Instead, the `client` and `server` objects are strict protocol implementations. That way, custom implementations of those objects don't have to provide values which they might not have (like the `socket` property).

### DIFF
--- a/package.json
+++ b/package.json
@@ -241,7 +241,7 @@
     "@bundled-es-modules/statuses": "^1.0.1",
     "@bundled-es-modules/tough-cookie": "^0.1.6",
     "@inquirer/confirm": "^5.0.0",
-    "@mswjs/interceptors": "^0.38.7",
+    "@mswjs/interceptors": "^0.39.0",
     "@open-draft/deferred-promise": "^2.2.0",
     "@open-draft/until": "^2.1.0",
     "@types/cookie": "^0.6.0",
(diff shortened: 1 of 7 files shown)


kettanaito:
## Released: v2.10.1 🎉

This has been released in v2.10.1!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.10.1)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.10.1)

---

PR #2541, opened 2025-07-01
feat: custom request predicate function

### LINKED ISSUE #1804: Custom request predicate function

### LINKED ISSUE #2399: Do not exhaust one-time handlers with custom predicate in resolver

### LINKED ISSUE #2482: Request considered handled if the handler returned nothing
### Scope

Improves an existing behavior

### Compatibility

### Feature description

I'm utilizing @mswjs/source to generate request handlers from .har files for a record-and-replay testing setup. I've observed that the generated handlers match requests based on the HTTP method and path, and they also consider query parameters. However, if a request is made to the same endpoint with different query parameters than those recorded in the HAR, the handler is skipped and the request is bypassed.

Use Case:

In my testing strategy, I aim to use .har files as exact snapshots of API interactions. This means that any deviation in query parameters from the recorded .har should result in an unhandled request, ensuring that tests only pass when the request precisely matches the recorded interaction.

Expected Behavior:

I would like the ability to configure fromTraffic to generate handlers that enforce strict matching, including query parameters. This would ensure that only requests with exact matches in method, path, and query parameters are handled, and any discrepancies result in unhandled requests.

ytoshiki:
- Closes #1804
- Fixes #2399
- Fixes #2482
- Closes #2573
- [x] Make sure redundant query params are checked only if `path` is a string/regexp. Do not check them when using a custom predicate (#2573)
- [x] Document this (done in https://github.com/mswjs/mswjs.io/pull/495)

## Overview

Support for custom predicate functions in both HTTP and GraphQL request handlers

```javascript
http.get(({ request }) => request.url.includes('foo'), resolver)
```

```javascript
http.post(async ({ request }) => {
  const body = await request.clone().json()
  return body.property === 'value'
}, resolver)
```

```javascript
graphql.query(
  ({ operationName, variables }) =>
    operationName === 'GetUserDetail' && variables.id === 'abc-123',
  () => {
    return HttpResponse.json({
      data: {
        user: {
          firstName: 'John',
          lastName: 'Maverick',
        },
      },
    })
  },
),
```

### Custom Predicate Support

- HTTP and GraphQL handlers accept a custom predicate function as the first argument.
- The predicate receives relevant request information and determines whether the handler should match the request.

### Type Definitions

- HTTP
    - The custom predicate receives `request` and `cookies` (not `params`, which was always {} and thus removed for clarity).
- GraphQL
    - The custom predicate receives `request`, `query`, `operationName`, `variables`, and `cookies`.


src/core/handlers/HttpHandler.ts
```diff
@@ -52,6 +52,13 @@ export type HttpRequestResolverExtras<Params extends PathParams> = {
   cookies: Record<string, string>
 }
 
+export type HttpCustomPredicate = (args: {
+  request: Request
+  cookies: Record<string, string>
+}) => boolean | Promise<boolean>
```
ytoshiki: `params` is always an empty object when using a custom predicate because `path` is a function, so users cannot actually use it for any matching logic.

src/core/handlers/HttpHandler.ts
```diff
@@ -124,7 +133,22 @@ export class HttpHandler extends RequestHandler<
...
+      // Clone the request to avoid locking its body stream.
+      // The body can only be read once, so cloning allows both
+      // the predicate and resolver to access it independently.
+      const clonedRequest = args.request.clone()
+
+      return await this.customPredicate({
+        request: clonedRequest,
+        cookies: args.parsedResult.cookies,
+      })
+    }
```
ytoshiki: pass cloned `request` to the predicate function rather than letting users to do it manually

src/core/handlers/HttpHandler.ts
kettanaito: A good point. I wonder if we can do something to help with this. I think exporting a function that parses any request URL against any path would be nice. That way, if the developer is using a custom predicate function, they can still parse out the request url into params. 

We have `matchRequestUrl()` exported from `msw` to do precisely that!

src/core/handlers/HttpHandler.ts
kettanaito: What if we delegate request cloning to the user? I wonder if them cloning the request in the predicate would be okay. A good area to try out in tests. 

Delegating this to the user means less request clones on our side. Especially handy if the user ends up with a predicate logic that doesn't need request's stream.

src/core/graphql.ts
```diff
@@ -30,7 +31,8 @@ export type GraphQLRequestHandler = <
   operationName:
     | GraphQLHandlerNameSelector
     | DocumentNode
-    | TypedDocumentNode<Query, Variables>,
+    | TypedDocumentNode<Query, Variables>
+    | GraphQLCustomPredicate<GraphQLVariables>,
```
kettanaito: We should pass `Query` type here too, no? The `{ query }` value should be accessible in the predicate function as by that point we've parsed out the query itself.

src/core/handlers/GraphQLHandler.ts
```diff
@@ -77,6 +77,14 @@ export type GraphQLResponseBody<BodyType extends DefaultBodyType> =
   | null
   | undefined
 
+export type GraphQLCustomPredicate<GraphQLVariables> = (args: {
+  request: Request
+  query: string
```
kettanaito, line 82: `query` can have a narrower type if we drill down the `Query` type argument here.

src/core/http.ts
```diff
@@ -39,6 +40,12 @@ function createHttpHandler<Method extends HttpMethods | RegExp>(
   method: Method,
 ): HttpRequestHandler {
   return (path, resolver, options = {}) => {
+    if (typeof path === 'function') {
```
kettanaito: What do you think if we handle this branching in the `HttpHandler` (if not `RequestHandler`) instead? 

1. Rename `path` argument to `predicate`.
2. Handle it being a function in the `.predicate()` method. 

This way, these namespace functions don't have to become more complex.

src/core/handlers/RequestHandler.ts
```diff
@@ -273,7 +273,7 @@ export abstract class RequestHandler<
       request: args.request,
       resolutionContext: args.resolutionContext,
     })
-    const shouldInterceptRequest = this.predicate({
+    const shouldInterceptRequest = await this.predicate({
```
kettanaito, line 276: Now that the predicate can potentially be more expensive, we should see if we can utilize the existing caching layer in `RequestHandler` to memoize custom predicates if they match the `request` key. In other words, when dealing with recurring requests (think polling), we don't have to re-run the predicate if the request reference didn't change (it will match anyway). 

We currently have a caching layer to minimize the number of request clones. We should be able to reuse it for predicate caching, too. This should be implemented in the base `RequestHandler` class, I think.

kettanaito (changes requested):
Hi, @ytoshiki. This looks amazing! I left a few comments, would love to hear your thoughts on them.

kettanaito:
One great way to reduce complexity in APIs that accept variable input (like our predicate) is to bring all that input to a single shared denominator. In our case, that would be the predicate function! 

Basically, translate `string` paths into `(path: string) => matchRequestUrl(path, ...)` function calls. This way, for the underlying `RequestHandler`, it only ever deals with a single predicate function described by the child handler class (HTTP/GraphQL handlers). 

Does this make sense?

src/core/handlers/HttpHandler.ts
ytoshiki: > What if we delegate request cloning to the user? I wonder if them cloning the request in the predicate would be okay. A good area to try out in tests.
> 
> Delegating this to the user means less request clones on our side. Especially handy if the user ends up with a predicate logic that doesn't need request's stream.

Thank you.
I agree with your approach. Delegating the request cloning to the user makes sense from a performance perspective because not all predicates need to read the request body as you mentioned.

src/core/http.ts
ytoshiki: I completely agree. That is a much cleaner approach.

src/core/handlers/HttpHandler.ts
kettanaito: The only gotcha here is to make sure that when predicate runs, it's not too late to clone a request. I don't think it should be, but testing will tell us for certain.

src/core/graphql.ts
ytoshiki: The `parsedResult.query` value is always a string (the raw GraphQL query string).
So, even if we add a generic Query type parameter, the actual value is still a string, and TypeScript cannot guarantee type safety without a cast.
Do you have any suggestions for making this truly type-safe without casting, given the current implementation?

https://github.com/mswjs/msw/blob/13e52aa154dbd1111c7b17cc123b60aff844e55f/src/core/handlers/GraphQLHandler.ts#L169-L197

ytoshiki:
> One great way to reduce complexity in APIs that accept variable input (like our predicate) is to bring all that input to a single shared denominator. In our case, that would be the predicate function!
> 
> Basically, translate `string` paths into `(path: string) => matchRequestUrl(path, ...)` function calls. This way, for the underlying `RequestHandler`, it only ever deals with a single predicate function described by the child handler class (HTTP/GraphQL handlers).
> 
> Does this make sense?

Are you suggesting that the HTTP Handler's constructor should always create a predicate function like below?

```javascript
this.predicateFunction = typeof predicate === 'function'
  ? predicate
  : (args) => {
      const hasMatchingMethod = this.matchMethod(args.request.method)
      const hasMatchingUrl = matchRequestUrl(new URL(args.request.url), predicate).matches
      return hasMatchingMethod && hasMatchingUrl
    }
```

So, the type check just moves from the predicate method to the constructor. the logic itself doesn't go away.
Is my understanding correct?

src/core/graphql.ts
kettanaito: I think it should be a string! The point of exposing `query` is so that the user can do custom matching based on it (e.g. provide it to a `graphql` function and see if it resolves).

src/core/handlers/RequestHandler.ts
ytoshiki: @kettanaito 
I tried reusing the existing request caching layer for predicate caching, but found some problems.

- The existing request cache uses 
```javascript
WeakMap<
    StrictRequest<DefaultBodyType>,
    StrictRequest<DefaultBodyType>
  >()
```
- `fetch()` creates new Request objects each time, even for identical requests. So we can't use Request objects as cache keys.

We'd need a different approach with content-based cache keys, but it would be complex and potentially slower than just re-running predicates.

ytoshiki:
> Hi, @ytoshiki. This looks amazing! I left a few comments, would love to hear your thoughts on them.

@kettanaito 
Thank you for your review! I fixed them below.

- https://github.com/mswjs/msw/pull/2541#discussion_r2178082190
- https://github.com/mswjs/msw/pull/2541#discussion_r2178072790

As for caching, I commented below. 
- https://github.com/mswjs/msw/pull/2541#discussion_r2188157198

kettanaito:
Thank you so much for your work, @ytoshiki. I will get to this as soon as I have a moment. Would love to get this released.

src/core/handlers/RequestHandler.ts
kettanaito: No need to cache or even clone the request here. Predicate should happen early enough for the user to clone the request themselves if they intend to read its stream!

src/core/handlers/HttpHandler.ts
kettanaito: I am making the custom predicate return type a union of boolean or `{ match: boolean, params: PathParams }` so the user can provide an extended predicate if they still wish to preserve path parameter parsing. 

They should be able to do that using the `matchRequestUrl` function from MSW:

```ts
import { matchRequestUrl } from 'msw'

http.get(({ request }) => {
  return {
    matches: myCustomLogic,
    params: matchRequestUrl(request.url, '/:foo').params,
  }
})
```

kettanaito (approved):
I've pushed a few minor changes here and there, otherwise this looks great!

src/core/handlers/GraphQLHandler.ts
kettanaito: I was wrong about this. At this point, query can be anything so it should be of a generic `Query` type.

kettanaito:
## Released: v2.11.0 🎉

This has been released in v2.11.0!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.11.0)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.11.0)

---

PR #2546, opened 2025-07-11
feat: add quiet cli option for suppressing informational output

rolule:
Hi, in our setup we use a pnpm prepare hook to run `msw init`, since we do not want to check the worker script into version control. This works fine, however the output gets annoying after a while, since every `pnpm install` prompts you with a whole block of messages on what happened and how to use msw:

msw output

In the Pipeline it clutters the logs even more:
Screenshot 2025-07-11 at 10 24 25

This is fine for new users, but we have been working with MSW for years (and think its great). So I thought a `--quiet` CLI option could be the way to suppress these informational messages for more advanced users. Let me know what you think!

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -25,9 +25,15 @@ yargs(process.argv.slice(2))
           description: 'Custom current worker directory',
           normalize: true,
         })
+        .option('quiet', {
+          alias: 'q',
+          type: 'boolean',
+          description: 'Suppress informational output messages',
+        })
         .example('msw init')
         .example('msw init ./public')
         .example('msw init ./static --save')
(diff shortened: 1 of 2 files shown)


kettanaito:
Hi, @rolule. Thanks for proposing this.

Have you tried `npx msw init >> /dev/null`? That can be a more elegant solution that requires no changed to the library as you can still pass all the other options and just flush the stdout into nothingness.

kettanaito:
I'm going to close this. Please see my [original suggestion](https://github.com/mswjs/msw/pull/2546#issuecomment-3065049938) as to how ignore the output of the init command.

PR #2553, opened 2025-07-17
test: add `range` mocked response test

### LINKED ISSUE #2552: `Range` request header missing for `no-cors` buffer requests
### Prerequisites

### Environment check

### Browsers

### Reproduction repository

https://github.com/ryankav/msw-video

### Reproduction steps

Start the dev server on the `partial-content-issue` branch and notice that when the full arrayBuffer is responded with then the video cannot be seeked (due to the handler not returning a partial content response). However, when the partial content handler is used the video doesn't get responded at all.

### Current behavior

I note that this issue was flagged before in #1972 and a test was added but I think that test might not cover some of the complexities surrounding the video and audio requests specifically. To highlight this the `main` branch of my reproduction repo doesn't use MSW and just serves the video directly from the public directory and it is possible to seek the video in this instance. Then the `parital-content-issue` branch shows the issue when MSW is added.

In the [webdev sw](https://web.dev/articles/sw-range-requests), mentioned in the previous issue, there is a link to [workbox docs](https://developer.chrome.com/docs/workbox/serving-cached-audio-and-video) on serving videos which references a fairly detailed  [workbox issue](https://github.com/GoogleChrome/workbox/issues/1663#issuecomment-448755945) comment of some extra complexity associated with handling serving audio and video. I'm able to get the range request header by adding the `crossorigin="anonymous"` to the video element but after I start using the range headers to respond with partial content I cannot get the video to respond.

### Expected behavior

Make the video seekable through msw. Any recommendation on potential workarounds would be appreciated.

kettanaito:
- Fixes #2552 

> [!IMPORTANT]
> This test case is not a 100% representation of the submitted use case. If you set `mode: 'no-cors` on the dispatched request in the test, the test will fail as described in the issue. That is caused by the browser behavior (stripping the `range` header from no-cors requests). Nothing we can do about this.

### DIFF
--- a/test/browser/rest-api/request/body/body-arraybuffer-range.mocks.ts
+++ b/test/browser/rest-api/request/body/body-arraybuffer-range.mocks.ts
@@ -1,0 +1,31 @@
+import { http, HttpResponse } from 'msw'
+import { setupWorker } from 'msw/browser'
+
+const buffer = new TextEncoder().encode('hello world')
+
+const worker = setupWorker(
+  http.get('/resource', async ({ request }) => {
+    const range = request.headers.get('range')
+
+    if (!range) {
+      throw new Response('Missing range', { status: 400 })
+    }
+
+    const ranges = range.replace(/bytes=/, '').split('-')
+    const start = +ranges[0]
+    const end = ranges[1] ? +ranges[1] : buffer.byteLength - 1
+    const content = buffer.slice(start, end)
+
+    return HttpResponse.arrayBuffer(content, {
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v2.10.5 🎉

This has been released in v2.10.5!

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.10.5)
- 📦 [npm package](https://www.npmjs.com/package/msw/v/2.10.5)

---

PR #2312, opened 2024-10-10
fix: update `@bundled-es-modules/cookie` to 2.0.1

kettanaito:
- Fixes #2308

## Blockers

- [x] https://github.com/bundled-es-modules/cookie/pull/3
- [ ] https://github.com/bundled-es-modules/cookie/pull/4
- [x] Bump the version of `@bundled-es-modules/cookie` once https://github.com/bundled-es-modules/cookie/pull/3 is released.

PR #2606, opened 2025-10-08
fix: use `cookie` directly via a shim

### LINKED ISSUE #2605: `@types/cookie` is not needed and deprecated
In https://github.com/mswjs/msw/blob/v2.11.3/package.json#L245

`@types/cookie` npm package is not needed and deprecated

See https://www.npmjs.com/package/@types/cookie

Image

<br><br>

One less dependency :-)

kettanaito:
## Changes

- Drops dependency on `@bundled-es-modules/cookie`.
- Drops `@types/cookie`.
- Uses the `cookie` package directly so we could get security fixes faster.
- Adds a shim to CJS-only `cookie` could run in ESM environments (e.g. browser).

## Issues

- Closes #2605 
- Related to #2312

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,7 +1,3 @@
-declare module '@bundled-es-modules/cookie' {
-  export * as default from 'cookie'
-}
-
 declare module '@bundled-es-modules/statuses' {
   import * as statuses from 'statuses'
   export default statuses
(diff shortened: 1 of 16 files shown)


kettanaito:
This shouldn't break any consumers. We've got both CJS and ESM module tests passing in Node.js/browser. If anything breaks, we shall treat that as a bug and use that as an opportunity to improve our module compatibility tests.

kettanaito:
## Released: v2.11.4 🎉

This has been released in v2.11.4.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.11.4)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.11.4)

```
npm i msw@latest
```

---

Issue #2550, opened 2025-07-14
Stop managing cookies in node.js


clovis1122:
### Scope

Adds a new behavior

### Compatibility

- [x] This is a breaking change

### Feature description

From the discussion at https://github.com/mswjs/msw/discussions/2467, because node.js does not manage cookies in the default implementation, msw shouldn't manage cookies in node.js either.

This will allow node.js users to decide whether they want to manage cookies via other userland solutions or stick to the default behavior.

clovis1122:
So, simply removing the cookie functionality may be a breaking change. I wonder if it can be added to 2x via a flag (`future_noStoreCookie`) or similar mechanism 🤔.

kettanaito:
## Research

As expected, Undici does not implement any sort of cookie store for requests/responses. Here are some other relevant materials. 

- Open proposal to add a `cookieStore` option to `fetch` in https://github.com/whatwg/fetch/issues/1384. 
- Similar topic brought up in Undici in https://github.com/nodejs/undici/issues/1640.

## Safe assumptions

Even if Undici starts handling cookies and `fetch` has a `cookieStore` option, that has no effect on MSW and its users. We still mustn't have any proprietary cookie handling in Node.js. If that handling will be added to the environment—good, MSW users will benefit from it automatically. Since it's not present as of now, we shouldn't interfere with the environment.

## Breaking change

I can see this breaking hybrid environment consumers, like JSDOM, but we do not support JSDOM (or Jest) so that is out of our concerns. Browser code must be tested in the browser, and the cookie handling behavior won't change there. 

@clovis1122, are there any other points that would make this a breaking change?

kettanaito:
This has been implemented and is coming in 3.0.

PR #2626, opened 2025-11-12
fix: compatibility issue with node.js 25

### LINKED ISSUE #2624: Node.js 25: localStorage.getItem is not a function in MSW
### Prerequisites

### Environment check

### Node.js version

v25.1.0

### Reproduction repository

https://github.com/qjatjs123123/reproduction-msw-localStorage-bug

### Reproduction steps

-  error test

```
clone repo

npm install

nvm use 25.1.0

npm run dev
```

- success test
```
clone repo

npm install

nvm use 18.18.2

npm run dev
```

It seems that the issue reported in [#2612](https://github.com/mswjs/msw/issues/2612)
 is still unresolved.

### Current behavior

```
unhandledRejection:  TypeError: localStorage.getItem is not a function
    at eval (webpack-internal:///(rsc)/./mocks/server.ts:6:61)
    at <unknown> (rsc)/./mocks/server.ts (C:\Users\Hong\Desktop\study\reproduction-msw-localStorage-bug\.next\server\_rsc_mocks_server_ts.js:20:1)
    at async eval (app\layout.tsx:7:24)
```

### Expected behavior

Should not use localStorage on Node.js 25

himself65:
Fixes: https://github.com/mswjs/msw/issues/2624
Upstream: https://github.com/nodejs/node/pull/60351
Related: https://github.com/better-auth/better-auth/pull/5918

### DIFF
--- a/src/core/utils/cookieStore.ts
+++ b/src/core/utils/cookieStore.ts
@@ -36,6 +36,11 @@ class CookieStore {
   }
 
   private getCookieStoreIndex(): MemoryCookieStoreIndex {
+    if (isNodeProcess()) {
+      // In Node.js 25, localStorage is available but throws an error when accessed without argv.
+      return {}
+    }
+
     if (typeof localStorage === 'undefined') {
       return {}
     }


kettanaito:
Hi, @himself65. Thanks for opening this.

I have two issues with this:

1. Node.js v25 is an experimental version by design. It means that (1) you shouldn't be using it in your app; (2) its behaviors and APIs might change, including within the same major version.
2. This isn't an issue to be addressed on MSW's side.

This change has gathered some negative feedback online and I've seen Node.js maintainers considering removing it. Why introduce a workaround for an experimental Node.js version (which no tool can promise supporting, including MSW) that we would have to then remember to rollback?

A proper solution here is to use Node.js v24 (or any other even version, which stands for stable) or wait for Node.js v25 to fix this issue, which is likely to happen in the future.

jnoordsij:
I ran into the same issues with NodeJS v25 and figured I could share some thoughts.

On the subject of "experimental" Node releases: as far as I can see and have understood, odd-numbered releases are in no way "experimental", just less tested and thus potentially less stable than LTS releases. However, they are still published *specifically* to allow the ecosystem to add support for new features. From https://nodejs.org/en/about/previous-releases:
> Major Node.js versions enter Current release status for six months, which gives library authors time to add support for them. After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to Active LTS status and are ready for general use.

So, although it indeed is not advisable to use Current releases for *production* environments due to potential stability issues, generally I think aiming to strive for compatibility with such releases is a proper goal for libraries. NodeJS still seems to strive towards non-breaking behavior within major release cycles, which is the main reason a secondary change w.r.t. the `localStorage` API was reverted (see https://github.com/nodejs/node/pull/60750).

What is experimental here is usage of the `localStorage` API of NodeJS, which is behind a feature-flag that is enabled by default as of NodeJS v25, which generally means that in future LTS versions this will be the case and moreover this change may be backported to existing LTS versions once it has been sufficiently tested. Therefore I expect this problem to become larger over time, rather than suddenly disappearing.

As for the proposed fix: I totally agree it feels a bit specific in the way it currently operates, which I agree is probably not the proper route to take given potential changes still to appear in NodeJS behavior. However, the constructor of this class is already assuming `localStorage` to not exist at all in NodeJS specifically, which is likely to become an incorrect assumption at some point in the future. Therefore it might make sense to have a slightly altered feature-detection system, that is more future-proof without being to specific to feel like a per-case hack.

The current feature-detection in constructor and `getCookieStoreIndex`/`persist` methods only checks for existence of `localStorage`, which works "fine" for browsers, but no longer for recent NodeJS versions. Maybe it could be slightly altered to actually detect availability of the employed methods? Looking into https://github.com/nodejs/node/commit/4fbb1ab1015614258c440700373148efa4da8074, it seems NodeJS in their behavior opted to throw an error even, which seems to be spec compliant, but (for now!) reverted it due to being too impactful. All in all, maybe introducing the following feature-detection method and applying it in all methods would be worthwhile? This can of course be somewhat simplified if preferable.

```js
const hasLocalStorage = (() => {
  try {
    return typeof localStorage?.getItem !== 'undefined' && typeof localStorage?.setItem !== 'undefined';
  } catch {
    return false;
  }
})();
```

kettanaito:
I wouldn't check method types, just that the `global.localStorage` property can be accessed and doesn't throw while doing so. We've got a similar [utility](https://github.com/mswjs/interceptors/blob/7513097134081492da40cf8f3c787c88ed7f12ba/src/utils/isPropertyAccessible.ts#L9) for Interceptors where certain global APIs also throw or something in React Native. We might adopt this approach in MSW.

kettanaito:
But as it is, there's no fix here because there's no bug here. Node.js throws _only_ when accessing the global without a feature flag. So once the need in the feature flag is dropped, it won't stop anymore. I don't want to break consumers through MSW though so I guess we'd have to introduce some sort of checking here. Weird.

jnoordsij:
> But as it is, there's no fix here because there's no bug here. Node.js throws _only_ when accessing the global 

The "unfortunate" part here is that the "throwing" behavior on accessing `localStorage` has actually be [reverted](https://github.com/nodejs/node/pull/60750) for now (and as such only exists on 25.2.0 specifically), so that rather than throwing it now (i.e. 25.0/1.x and >= 25.2.1) once again will return an empty object `{}`. Therefore `typeof localStorage` is not `"undefined"`, will not throw, but will not allow accessing `getItem` either.

> without a feature flag. 

To make it more confusing, there's actually two flags (see also https://github.com/mswjs/msw/issues/2624#issuecomment-3546368492): one for disabling the entire webstorage (i.e. both local and session) and one for configuring the localstorage.

jnoordsij:
Also: I do agree that current NodeJS behavior is heavily confusing and may be changing a bit more throughout the 25 major and then once more on v26 release (somewhere in April), so it might also make sense to just hang on a bit and point people towards the workarounds per https://github.com/mswjs/msw/issues/2624#issuecomment-3546368492 (or just using v24 for now) while waiting to see how things work out.

kettanaito:
Oh, those are great finds. Thank you, @jnoordsij.

IIRC MSW accesses local storage on multiple occasions but those have different intents. In the case of the cookie store, I doubt even having local storage would actually make the functionality correct in Node.js (see #2550). In this case, the `isNodeProcess()` check might be the best way forward.

But I believe we access the storage in other places, too, right?

jnoordsij:
I've only ran into issues with the cookie store and a [quick GitHub search](https://github.com/search?q=repo%3Amswjs%2Fmsw%20localStorage&type=code) for this repository only shows that, but there [seem to be more occurences](https://github.com/search?q=org%3Amswjs%20localStorage&type=code) throughout the entire organization.

For example there's this check:
https://github.com/mswjs/cookies/blob/24ba6ad6504ccae68a82a6d36d6b341e515e9bae/src/store.ts#L22
which (incorrectly?) compares to `null` rather than `undefined`, but then proceeds with some feature-like testing to check support.

kettanaito:
The `== null` comparison is valid as it unfolds to `=== undefined || === null`.

rvagg:
Alternative form of this fix that's a little more targeted and also complete: https://github.com/mswjs/msw/pull/2636

kettanaito:
## Released: v2.12.4 🎉

This has been released in v2.12.4.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.4)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.4)

```
npm i msw@latest
```

---

PR #2627, opened 2025-11-12
fix(node): move `browser` and `react-native` export conditions lower

### LINKED ISSUE #2545: "No known conditions for "./node" specifier in "msw/node" package" in Svelte Testing Library
We test our Svelte Components with vitest, jsdom, svelteTesting  and msw. And we love msw <3.
It worked pretty well until 2.7.6 
In 2.8.0  in package.json "browser": null moved to the top level of the node export. This leads to the missing export always if browser is present in your runtime condition. Even if you have node and browser. 

Are there known workarounds? 
Without browser our components fail to render, without node msw wont load node server. 

Since it's not mentioned in the release notes, neither as breaking change, can browser be moved below node again?

Thanks!

vidschofelix:
Moved the "node" and "react-native" export definitions lower in the exports map. 
This fixes the "No known conditions for './node' specifier" error.

- Fixes mswjs/msw#2545

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,6 @@
       }
     },
     "./node": {
-      "browser": null,
-      "react-native": null,
       "module-sync": {
         "types": "./lib/node/index.d.mts",
         "default": "./lib/node/index.mjs"
@@ -79,6 +77,8 @@
         "types": "./lib/node/index.d.mts",
         "default": "./lib/node/index.mjs"
       },
+      "browser": null,
+      "react-native": null,
       "default": {
         "types": "./lib/node/index.d.ts",
         "default": "./lib/node/index.js"


package.json
```diff
@@ -79,6 +77,8 @@
         "types": "./lib/node/index.d.mts",
         "default": "./lib/node/index.mjs"
       },
+      "browser": null,
```
kettanaito, line 80: I'm just worried that some browser environments might false-positively match some of the preceding conditions, like `module`. 

@markerikson, hi! Sorry for a direct ping. May I ask for your wisdom on this change? Do you see any unexpected behaviors this would cause for browsers importing the `/node` path?

markerikson: I'm actually not sure I've _seen_ use of a `null` value for a condition before, but then again I also stopped trying to think about this after we got the RTK packages in shape :)

I don't have a good answer here, but looking at the issue thread it sounds like this helps.  Worth a shot?

kettanaito: Thanks! Appreciate your input. I also no see any dangers with this, so release it is!

kettanaito (approved):
Thank you for fixing this, @vidschofelix 🚀

kettanaito:
## Released: v2.12.2 🎉

This has been released in v2.12.2.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.2)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.2)

```
npm i msw@latest
```

---

PR #2643, opened 2025-12-23
fix(sse): handle multi-line messages

BSteffaniak:
Fixes an issue where SSE messages containing newlines in the `data` field would be incorrectly parsed by the client.

Per the SSE spec, multi-line data must be sent as multiple `data:` lines that the EventSource parser concatenates. Previously, MSW sent the data as a single `data:` field, which caused messages containing `\n\n` to be truncated (since `\n\n` terminates an SSE message).

**Before:**
data:line1\n\nline2
Client receives: `"line1"` (truncated)

**After:**
data:line1
data:
data:line2

Client receives: `"line1\n\nline2"` (correct)

The fix handles all SSE line terminators per spec: LF (`\n`), CR (`\r`), and CRLF (`\r\n`).

Let me know your thoughts. Thanks!

### DIFF
--- a/src/core/sse.ts
+++ b/src/core/sse.ts
@@ -327,11 +327,12 @@ class ServerSentEventClient<
       frames.push(`event:${message.event?.toString()}`)
     }
 
-    // Split data on line terminators (LF, CR, or CRLF) per SSE spec
-    // and send each as a separate data: line.
-    const dataStr = message.data?.toString() ?? ''
-    for (const line of dataStr.split(/\r\n|\r|\n/)) {
-      frames.push(`data:${line}`)
+    if (message.data != null) {
+      // Split data on line terminators (LF, CR, or CRLF) per SSE spec
+      // and send each as a separate data: line.
+      for (const line of message.data.toString().split(/\r\n|\r|\n/)) {
+        frames.push(`data:${line}`)
+      }
     }
 
     frames.push('', '')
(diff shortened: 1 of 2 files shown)


kettanaito (dismissed):
Hi, @BSteffaniak. Thank you for fixing this! And an extra thanks for those test cases.

I've pushed a change related to how the message data is processed, making that processing conditional (no need to fallback to `''` to then split via RegExp into an empty array). 

Could you please share a link to the paragraph section that outlines how line terminators have to be handled in the SSE spec? We could leave it in the implementation for future references. Thanks.

BSteffaniak:
Here are some references in the spec:

https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

here's also some MDN docs around it too (a bit easier to read):

https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#data

BSteffaniak:
This is also relevant: https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream

kettanaito:
Thanks! This is great.

kettanaito:
## Released: v2.12.5 🎉

This has been released in v2.12.5.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.5)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.5)

```
npm i msw@latest
```

---

PR #2646, opened 2025-12-27
fix(graphql): support `TypedDocumentString` as request predicate

### LINKED ISSUE #1583: Support GraphQL `TypedDocumentString`
### Scope

Improves an existing behavior

### Compatibility

### Feature description

Hello, and thank you for all of your work here.

I ran across a case today that could use the suggested improvement.

With the release of both [`graphql-typed-document-node`](https://github.com/dotansimha/graphql-typed-document-node) and [`@graphql-codegen/typed-document-node`](https://npmjs.com/package/@graphql-codegen/typed-document-node), you can use the `documentMode: string` config to produce strings instead of AST document nodes.

The reason for this is the much smaller size of bundles produced with `documentMode: string`. Instead of AST document nodes, it just produces thin class wrappers around strings.

It looks something like this:

```ts
export class TypedDocumentString<TResult, TVariables>
  extends String
  implements DocumentTypeDecoration<TResult, TVariables>
{
  __apiType?: DocumentTypeDecoration<TResult, TVariables>['__apiType'];

  constructor(private value: string, public __meta__?: { hash: string }) {
    super(value);
  }

  toString(): string & DocumentTypeDecoration<TResult, TVariables> {
    return this.value;
  }
}

// insert the query there
export const SomeQuery = new TypedDocumentString(`{QUERY_STR}`);
```

So it looks like to support this, we need to support the `DocumentTypeDecoration` interface instead of the concrete `TypedDocumentNode`.

### Workaround

I have the following workaround in place that works well.

```ts
const query = createScopedGraphQLHandler('query', '*');
const mutation = createScopedGraphQLHandler('mutation', '*');

export function createScopedGraphQLHandler(operationType: ExpectedOperationTypeNode, url: string) {
  return <Query extends Record<string, any>, Variables extends GraphQLVariables = GraphQLVariables>(
    operationName: DocumentNode | TypedDocumentNode<Query, Variables> | DocumentTypeDecoration<Query, Variables>,
    resolver: ResponseResolver<GraphQLRequest<Variables>, GraphQLContext<Query>>
  ) => {
    let operation: DocumentNode | TypedDocumentNode<Query, Variables> | string;
    // If it is a typed document string, then extract the operation name
    if (typeof operationName === 'object' && operationName instanceof String) {
      operation = getOperationName(operationName.toString()) || '';
    } else {
      operation = operationName as TypedDocumentNode<Query, Variables>;
    }

    return new GraphQLHandler<GraphQLRequest<Variables>>(operationType as OperationTypeNode, operation, url, resolver);
  };
}

export { query, mutation };
```

I will happily PR this if you think it belongs in `msw`.

### More context

When using `TypedDocumentNode` which is the AST object of a given query, it produces a lot code compared to the original query size. If you are using Apollo on the BE or a Node.js server then it is all good since it can use that AST. Some other implementations in Ruby/PHP land do not so you need to print it to a string before you pass it. This is a waste of bandwidth, memory, and bundle size. 

Since in that case, you would be interested in the Typing support aspect only. You can use a `TypedDocumentString` which produces a much smaller footprint value.

P.S: I maintain a small library called [`villus`](https://github.com/logaretm/villus) that just supported this new typed document string.

kettanaito:
- Fixes #1583

## Changes

This widens the type and runtime support of structures generated by GraphQL Code Gen. Now, MSW checks for the extension of a generic `DocumentTypeDecoration`, which both `TypedDocumentNode` and `TypedDocumentString` extend. 

## Roadmap

- [x] Runtime support
- [x] Add type tests


kettanaito:
## Released: v2.12.7 🎉

This has been released in v2.12.7.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.7)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.7)

```
npm i msw@latest
```

---

PR #2651, opened 2026-01-13
chore: add support for windows builds

AndyMJones:
When trying to build the project on a windows machine there are a few errors relating to file paths and commands used within `tsup.config.ts` and `patch-ts.js`. I have updated the files to support both linux / mac and windows development but I have not created a new ci job to run on windows (happy to add that if you would like it).

**tsup failure:**
image

**patch failure:**
image

`tsup.config.ts` is an easy fix as there are simple params that can be added to `glob.sync` that solve the path issues.

`patch-ts.js` is slightly more involved, you cannot run `/bin/bash` and grep on windows. I have written a simple version using glob to find the file paths, then fs to read the file to search for references of "~/core" etc. This works for both windows and Linux but is significantly slower than using grep so I have kept the grep implementation for non windows builds. Happy to consolidate this if you care more about simplicity over performance (I think the functions are a little bit ugly but hopefully clear).

### DIFF
--- a/config/scripts/patch-ts.js
+++ b/config/scripts/patch-ts.js
@@ -8,11 +8,124 @@ import { hasCoreImports, replaceCoreImports } from '../replaceCoreImports.js'
 const execAsync = promisify(exec)
 
 const BUILD_DIR = new URL('../../lib/', import.meta.url)
+const TYPESCRIPT_FILE_MATCHER = '**/*.d.{ts,mts}'
+const DEFINITION_TYPESCRIPT_FILE_MATCHER = '**/*.d.ts'
+const IS_WINDOWS = process.platform === 'win32'
 
-async function patchTypeDefs() {
-  const typeDefsPaths = glob.sync('**/*.d.{ts,mts}', {
+async function findUnresolvedImportsLinux() {
+  const result = await execAsync(
+    `grep "~/core" ./${TYPESCRIPT_FILE_MATCHER} -R -l || exit 0`,
+    {
+      cwd: BUILD_DIR,
+      shell: '/bin/bash',
+    },
+  )
+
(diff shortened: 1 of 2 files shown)


kettanaito:
Hi, @AndyMJones. Thanks for looking into this, much appreciated. 

We can merge `glob` arguments additions straight away but otherwise this looks too big of a change to fix the issue. Perhaps there's a dependency we can use that gives us `grep` powers and has Windows/Linux support baked in. Or, even better, we find a way to validate file contents without relying on grep.

AndyMJones:
Hi @kettanaito, no problem. I've been using MSW for years and felt I should give back a little.

`glob` already does the heavy lifting any other library would do. I had a look into using `shelljs` or `fast-glob` but there wasn't really any major gain from what I could see.

If you are happy for grep to not be used, the flow for windows actually already worked for both linux and windows. I have refined this PR down to one solution instead of one for windows and one for linux. Let me know if you still think its too big.

kettanaito:
@AndyMJones, I've pushed a few minor changes to improve the windows support. Could you please try them out and confirm if they are working for you?

AndyMJones:
Yes this still works on my windows machine. Gave it a few runs and a clean install etc. Looks good to me

image

kettanaito:
Thanks for checking, @AndyMJones. Excited to have more Windows users on board :D

kettanaito:
## Released: v2.13.0 🎉

This has been released in v2.13.0.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.13.0)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.13.0)

```
npm i msw@latest
```

---

PR #2655, opened 2026-02-02
fix: translate unhandled connection handler errors to WebSocket error events

### LINKED ISSUE #2446: Bug: Unhandled errors aren't translated to WebSocket error event
The impact is that it's not possible to write tests for `websocket.onerror`.

### Prerequisites

### Environment check

- [x] I'm using Node.js version 18 or higher

### Node.js version

v22

### Reproduction repository

https://github.com/joshua-tj/msw-ws-onerror-test/

```
test("Connection should error", () =>
  new Promise((done, fail) => {
    const api = ws.link("wss://example.com/path");

    const server = setupServer(
      api.addEventListener("connection", () => {
        console.log("connecting");
        throw new Error("Connection error");
      })
    );
    server.listen();

    const connection = new WebSocket("wss://example.com/path");
    connection.onerror = (e) => {
      console.log("onerror", e);
      done();
    };

    connection.onclose = (e) => {
      console.log("onclose", e);
    };

    connection.onopen = () => {
      console.log("onopen");
      fail("Connection should not be opened");
    };
  }));
```

### Reproduction steps

`yarn`
`yarn test`

### Current behavior

Errors and fails tests as `onerror` handler is not called

### Expected behavior

Should not fail since we handle errors and call `done()` in test

JarvisInvestInsight:
Fixes #2446.

## Summary

When an error is thrown in a WebSocket connection handler, it now translates to a WebSocket error event on the client, making it possible to test \websocket.onerror\ scenarios.

## Changes

- Wrapped the \Promise.all(handlers.map(...))\ in \handleWebSocketEvent.ts\ with try/catch
- On error, dispatch an error event to \connection.client.socket\ with the original error as \cause\
- Added tests for both sync and async error cases

## Before

Errors thrown in connection handlers would bubble up unhandled, making it impossible to test \onerror\ behavior.

## After

Errors are caught and translated to WebSocket error events, consistent with how \onUnhandledRequest\ errors are already handled in the same file.

---
*Found this helpful? [☕ Buy me a coffee](https://ko-fi.com/jarvisdev)*

### DIFF
--- a/src/core/ws/handleWebSocketEvent.ts
+++ b/src/core/ws/handleWebSocketEvent.ts
@@ -23,14 +23,26 @@ export function handleWebSocketEvent(options: HandleWebSocketEventOptions) {
     if (handlers.length > 0) {
       options?.onMockedConnection(connection)
 
-      await Promise.all(
-        handlers.map((handler) => {
-          // Iterate over the handlers and forward the connection
-          // event to WebSocket event handlers. This is equivalent
-          // to dispatching that event onto multiple listeners.
-          return handler.run(connection)
-        }),
-      )
+      try {
+        await Promise.all(
+          handlers.map((handler) => {
+            // Iterate over the handlers and forward the connection
+            // event to WebSocket event handlers. This is equivalent
+            // to dispatching that event onto multiple listeners.
+            return handler.run(connection)
+          }),
+        )
+      } catch (error) {
+        // Translate unhandled errors in connection handlers
+        // to WebSocket error events on the client.
+        const errorEvent = new Event('error')
+        Object.defineProperty(errorEvent, 'cause', {
+          enumerable: true,
+          configurable: false,
+          value: error,
+        })
+        connection.client.socket.dispatchEvent(errorEvent)
+      }
 
       return
     }
--- a/test/node/ws-api/ws.connection-error.test.ts
+++ b/test/node/ws-api/ws.connection-error.test.ts
@@ -1,0 +1,73 @@
+// @vitest-environment node-websocket
+import { ws } from 'msw'
+import { setupServer } from 'msw/node'
+
+const server = setupServer()
+
+const service = ws.link('ws://example.com/*')
+
+beforeAll(() => {
+  server.listen()
+})
+
+afterEach(() => {
+  server.resetHandlers()
+})
+
+afterAll(() => {
+  server.close()
+})
+
+it('translates unhandled connection handler errors to WebSocket error events', async () => {
+  const errorListener = vi.fn()
+  const openListener = vi.fn()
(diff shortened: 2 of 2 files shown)


kettanaito:
Hi, @JarvisInvestInsight. Thank you for opening this pull request.

The right place to fix this issue would be [here](https://github.com/mswjs/interceptors/blob/9560baaacc7fb98864fc6ae6f474f1ae13dbc466/src/interceptors/WebSocket/index.ts#L147-L168). For some reason, exceptions during handlers (called within `emitAsync`) don't trigger that catch block. _That_ is the actual issue. 

Addressing this on MSW side is wrong, it shouldn't care about these things. 

The logic here is as follows:

- Interceptors emits `connection` on all listeners.
- MSW [listens](https://github.com/mswjs/msw/blob/e0a46431e0f78f1b89d906c3919540dc7947b5bf/src/core/ws/handleWebSocketEvent.ts#L19) to the `connection` event and calls all the relevant handlers.
- If [running a handler](https://github.com/mswjs/msw/blob/e0a46431e0f78f1b89d906c3919540dc7947b5bf/src/core/ws/handleWebSocketEvent.ts#L31) throws, it is expected to propagate to [this `catch` block](https://github.com/mswjs/interceptors/blob/9560baaacc7fb98864fc6ae6f474f1ae13dbc466/src/interceptors/WebSocket/index.ts#L147), which already translates exceptions to the appropriate WebSocket `error` events. 

Starting for a unit test for [`emitAsync`](https://github.com/mswjs/interceptors/blob/9560baaacc7fb98864fc6ae6f474f1ae13dbc466/src/utils/emitAsync.ts#L23) that ensures its promise rejects when any of the listeners throw might be a good idea. 

Meanwhile, I'm migrating the toolchain to use [`rettime`](https://github.com/kettanaito/rettime), which has `.emitAsPromise()`. I don't recall if I have unit tests for this behavior there, but I think it's worth mentioning just in case our custom implementations in MSW are incorrect.

kettanaito:
> [!CAUTION]
> The issue might also be fixed already. Interceptors uses `emitAsync` while some of the contributors point out that perviously it relied on `this.emitter.emit()`, which did not process async listeners correctly.

kettanaito:
I've added a test for this and it seems to be passing as-is. See https://github.com/mswjs/msw/pull/2656.

kettanaito:
## Released: v2.12.8 🎉

This has been released in v2.12.8.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.8)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.8)

```
npm i msw@latest
```

---

PR #2661, opened 2026-02-06
fix(ws): support relative urls in `ws.link`

kettanaito:
- Requires https://github.com/mswjs/interceptors/pull/763

Looks like `ws.link()` doesn't resolve relative URLs against the page's origin. The logic is there on MSW side, but Interceptors throw on relative WebSocket URLs because it never resolves them according to the specification. 

> Our logic didn't account that the `WebSocket` constructor will enforce the `ws`/`wss` scheme on the actual URL. Since a relative handler URL is resolved against `location.href`, it always produced a mismatch. 

```ts
ws.link('/api') // "http://localhost/api"
new WebSocket('/api') // "ws://localhost/api"
```

We are now correctly resolving relative paths in the `ws.link` before passing them to the URL matching. 

> [!CAUTION]
> Relative URLs are still not allowed in Node.js. However, if you have `baseUrl` defined in the resolution context, MSW will use it to resolve any relative WebSocket link URLs against it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -238,7 +238,7 @@
   "sideEffects": false,
   "dependencies": {
     "@inquirer/confirm": "^5.0.0",
-    "@mswjs/interceptors": "^0.41.0",
+    "@mswjs/interceptors": "^0.41.2",
     "@open-draft/deferred-promise": "^2.2.0",
     "@types/statuses": "^2.0.6",
     "cookie": "^1.0.2",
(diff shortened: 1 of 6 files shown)


kettanaito:
## Released: v2.12.9 🎉

This has been released in v2.12.9.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.9)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.9)

```
npm i msw@latest
```

---

PR #2668, opened 2026-03-08
test: prevent handling in-flight requests after worker was stopped

### LINKED ISSUE #2597: MSW sometimes still handles requests after worker is stopped
### Prerequisites

### Environment check

### Browsers

### Reproduction repository

https://github.com/diego-aquino/msw-worker-stop-bypass-issue

### Reproduction steps

1. `pnpm install`
2. `pnpm test`

### Current behavior

Starting in msw@2.11.2, I found out that workers sometimes handle requests even after `worker.stop()` is called.

In the reproduction repository, this expect sometimes passes, sometimes fails.

https://github.com/diego-aquino/msw-worker-stop-bypass-issue/blob/main/example.browser.test.ts#L22-L27

```ts
  worker.stop();

  // This expect sometimes passes, sometimes fails.
  await expect(async () => {
    await fetch(`${baseURL}/users`);
  }).rejects.toThrowError(TypeError);
```

From my tests, I did not see this issue happening with `server.close()`, neither with `worker.stop()` and `server.close()` before 2.11.2 (I tested 2.11.1).

### Expected behavior

I expected all requests made after `worker.stop()` was called to not be handled.

diego-aquino:
I was investigating https://github.com/mswjs/msw/issues/2597 and comparing the reproduction repository with MSW's test suite. The reproduction tests frequently fail due to requests still being handled after the worker was stopped.

However, the existing MSW test checking this behavior does not appear to be flaky:

https://github.com/mswjs/msw/blob/5ab5857bc1d67062d39627519f92fc88d1aa3f07/test/browser/msw-api/setup-worker/stop/in-flight-request.test.ts#L31

I've created two new test cases to better understand the problem:

1. `bypasses requests not immediately made after the worker was stopped, considering no default response`

   This test stops the worker and makes a request in **different `page.evaluate` calls**. It checks if the request was bypassed by expecting a `Cannot GET` response text. From my local executions, it always passes.

2. `bypasses requests immediately made after the worker was stopped, considering no default response`

   This test stops the worker and makes a request **in the same `page.evaluate` call**. Similarly to test 1, it expects a `Cannot GET` response text, but it shows the same flaky behavior as the reproduction tests, although less frequently.

  Here's an example of a failure:

  ```ts
  1) [chromium] › test/browser/msw-api/setup-worker/stop/in-flight-request.test.ts:67:3 ›
     bypasses requests immediately made after the worker was stopped, considering no default response (40) 

    Error: expect(received).toContain(expected) // indexOf

    Expected substring: "Cannot GET /resource"
    Received string:    "hello world"

       98 |     })
       99 |
    > 100 |     expect(data).toContain('Cannot GET /resource')
          |                  ^
      101 |   })
      102 | }
      103 |
  ```

> [!NOTE]
>
> I've added a for loop to repeat the new tests and make them more likely to fail. Still, not all test executions cause failures. Run them a couple of times if they continue to pass.
>
> ```bash
> pnpm test:browser in-flight-request.test.ts
> ```

Maybe the overhead of `page.evaluate` is masking the issue in the existing test? 

---

This pull request does not yet contain any changes to the source code. I've only added tests trying to simulate the issue.

@kettanaito, as I'm new to the codebase, I'm not yet sure about how to fix this. I'm willing to contribute after hearing your feedback and possible courses of action. Thanks!

Closes https://github.com/mswjs/msw/issues/2597.


kettanaito:
Hey, @diego-aquino. Thanks for adding these tests! 

Yes, `page.evaluate` definitely adds to the behaviors here. As any async logic, it makes things more/less reproducible depending on how you look. `worker.stop()` is synchronous, but it routes through the worker, which makes its _effect_ asynchronous. I suppose that is the problematic window during which in-flight requests might still get handled. 

I highly recommend put this on hold until #2650 is merged. It contains fundamental changes to the internal architecture and how the worker is handled, in particular. While it won't fix this issue, it would be easier to branch the fix from there.

kettanaito:
Closed by https://github.com/mswjs/msw/pull/2724.

PR #2669, opened 2026-03-14
fix(node): avoid open handles for infinite delays

### LINKED ISSUE #778: Using delay('infinite') leaves unresolved promises which make jest warn
## Describe the bug
⚠️ I'm not sure if this is the correct repo to report this.

When trying to use `delay('infinite')` to mock loading states, jest throws a warning and hangs forever.

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped
in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot
this issue.
```

This happens because the promises hang forever, but I can't find a way to cancel/reject/terminate them.
I'm not really sure what's the correct way of handling this.

## Environment

- `msw: 0.29.0`
- `nodejs: 14.17.0`
- `npm: 7.15.1`

## To Reproduce

```js
// sample.spec.js

const { rest } = require('msw');
const { setupServer } = require('msw/node');

const fetch = require('node-fetch');

it('hangs', () => {
  const server = setupServer();

  server.listen();

  server.use(rest.get('http://localhost/a-request', (_req, res, ctx) => res(ctx.delay('infinite'))));

  fetch('http://localhost/a-request').then(() => undefined).catch(console.error);

  expect(true).toBe(true);

  server.close();
});
```

Just to give some extra context, the real code is more like:

```tsx
it('shows a loading spinner', async () => {
	render(<Component />);

	const loadingBar = await findByRole('progressbar')

	expect(loadingBar).toBeInTheDocument();
});
```

## Expected behavior
Requests are aborted(?), the tests conclude and jest doesn't warn about opened handlers.

## Screenshots

image

felmonon:
## Summary
- unref the Node.js timer used by `delay('infinite')` so it no longer keeps the event loop alive on its own
- add a unit test covering the `unref()` behavior on infinite delays
- add a Node regression that spawns a child process and verifies `server.close()` can exit cleanly after an infinite delay request

Closes #778.

## Testing
- pnpm build
- pnpm exec vitest run src/core/delay.test.ts
- pnpm exec vitest run --config=./test/node/vitest.config.ts test/node/msw-api/context/delay.node.test.ts
- pnpm exec eslint src/core/delay.ts src/core/delay.test.ts test/node/msw-api/context/delay.node.test.ts


test/node/msw-api/context/delay.node.test.ts
```diff
@@ -71,3 +73,38 @@ test('uses realistic server response time when "real" mode is provided', async (
   expect(responseTime).toBeGreaterThan(5)
   expect(await res.text()).toBe('john')
 })
+
+test('does not keep the Node.js process alive when using "infinite" delay', () => {
+  const result = spawnSync(
```
kettanaito, line 75: I much prefer putting the MSW usage code into a separate file and spawning _that_ (see [this](https://github.com/mswjs/interceptors/blob/92b79585d5ee978f88ccd4501da4eb2173c43a88/test/features/remote/remote.test.ts#L8) as an example). That way, the test is actual code through-and-through, with type validation, intellisense, etc.

src/core/delay.ts
```diff
@@ -16,6 +20,12 @@ function getRealisticResponseTime(): number {
   )
 }
 
+function isTimeoutWithUnref(
+  timeoutId: ReturnType<typeof setTimeout>,
+): timeoutId is TimeoutWithUnref {
```
kettanaito: Let's drop the custom `TimeoutWithUnref` type and use an inline assertion type the same way we do for `isBroadcastChannelWithUnref`:


```suggestion
): timeoutId is NodeJS.Timeout & NodeJS.RefCounted {
```

src/core/delay.ts
```diff
@@ -16,6 +20,12 @@ function getRealisticResponseTime(): number {
   )
 }
 
+function isTimeoutWithUnref(
```
kettanaito: Given we also check for `isBroadcastChannelWithUnref` in the WebSocket implementation, do you think we should introduce a more generic utility for that or would that be an overkill?

Something like

```ts
function hasRefCounted<T extends object>(value: T): value is T & NodeJS.RefCounted {
  return (
    'ref' in value && typeof value.ref === 'function' &&
    'unref' in value && typeof value.unref === 'function'
  )
}
```

And then use it in the WebSocket implementation and here like:

```ts
if (hasRefCounted(timeoutId)) {
  timeoutId.unref()
}
```

kettanaito (changes requested):
Oh wow, this looks extremely promising! Thank you for working on this, @FelmonFekadu. I left a few comments, please let me know what you think. Excited to get this merged, this has been one of the long-standing issues with infinite delays.

src/core/delay.ts
felmonon: Dropped the custom timeout type. I extracted a shared hasRefCounted() utility and reused it here instead of keeping a delay-specific guard.

test/node/msw-api/context/delay.node.test.ts
felmonon: Moved the spawned scenario into a standalone fixture file and switched it to ESM imports from the built package, so this path is now regular code instead of an inline -e script.

src/core/delay.ts
felmonon: I went with the shared helper. ws.ts already had the same ref-counted check for BroadcastChannel, so extracting hasRefCounted() kept both call sites aligned instead of adding another one-off unref guard.

felmonon:
Addressed the review comments in b76a8a9c. Changes: moved the child-process scenario into a dedicated fixture, extracted a shared hasRefCounted() helper reused by delay and ws, and reran the focused build/test/lint set. Verified with: pnpm build; pnpm exec vitest run src/core/delay.test.ts src/core/utils/internal/hasRefCounted.test.ts; pnpm exec vitest run --config=./test/node/vitest.config.ts test/node/msw-api/context/delay.node.test.ts; pnpm exec eslint src/core/delay.ts src/core/ws.ts src/core/delay.test.ts src/core/utils/internal/hasRefCounted.ts src/core/utils/internal/hasRefCounted.test.ts test/node/msw-api/context/delay.node.test.ts.

felmonon:
@kettanaito Thanks for the review and the direction here. I addressed the inline comments in `b76a8a9c`: moved the spawned scenario into a dedicated fixture, extracted a shared `hasRefCounted()` helper reused by both `delay` and `ws`, and reran the focused build/test/lint set locally. Let me know if you want me to tighten anything else before merge.

kettanaito (dismissed):

kettanaito:
## Released: v2.12.11 🎉

This has been released in v2.12.11.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.11)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.11)

```
npm i msw@latest
```

---

PR #2676, opened 2026-03-18
fix: ensure `RequestHandler` type is accepted in setup functions

felmonon:
## Summary
- accept base `RequestHandler` instances in the shared `AnyHandler` type
- keep `setupWorker` and `setupServer` compatible with explicitly typed `RequestHandler[]` arrays
- add regression coverage for both setup APIs

## Verification
- pnpm build
- pnpm test:ts
- pnpm test:unit -- --run src/core/experimental/handlers-controller.test.ts test/node/msw-api/setup-server/listHandlers.node.test.ts
- pnpm lint

Fixes the compatibility gap reported while testing #2650, where arrays typed as `RequestHandler[]` no longer spread into `setupWorker`/`setupServer`.

### DIFF
--- a/src/core/experimental/handlers-controller.ts
+++ b/src/core/experimental/handlers-controller.ts
@@ -1,11 +1,10 @@
 import { invariant } from 'outvariant'
-import { type HttpHandler } from '../handlers/HttpHandler'
-import { type GraphQLHandler } from '../handlers/GraphQLHandler'
+import { type RequestHandler } from '../handlers/RequestHandler'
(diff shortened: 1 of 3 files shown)


kettanaito (approved):

kettanaito:
Thanks for raising this, @FelmonFekadu. I've improved the type tests to actually have assertions and cleaned up the manual `AnyHandler` type from the core. Looks good now.

felmonon:
Thanks for tightening that up and taking it the rest of the way. I appreciate the cleanup and the stronger type assertions.

PR #2677, opened 2026-03-19
fix: support wildcard ports in url matching

### LINKED ISSUE #2204: Using the wildcard token to try to match random ports doesn't work if preceded by a colon

trknhr:
Support absolute URLs that use wildcard ports in request matching.

MSW already coerces request masks before passing them to `path-to-regexp`. Numeric ports were escaped there, but wildcard ports like `http://localhost:*` were left as `:(.*)`, which `path-to-regexp` interprets as an invalid parameter declaration and throws.

This updates `coercePath()` to escape the port separator when the port is either numeric or a wildcard-expanded capture group. It also adds regression tests for HTTP and WebSocket URLs, including a pathname case.

This keeps the current `path-to-regexp`-based matching behavior intact instead of broadening the fix into a larger matcher change.

Fixes #2204

### DIFF
--- a/src/core/utils/matching/matchRequestUrl.test.ts
+++ b/src/core/utils/matching/matchRequestUrl.test.ts
@@ -93,7 +93,7 @@ describe('matchRequestUrl', () => {
       .toEqual({
         matches: true,
         params: {
-          '0': '3000',
+          '0': '3000/',
         },
       })
 
--- a/src/core/utils/matching/matchRequestUrl.test.ts
+++ b/src/core/utils/matching/matchRequestUrl.test.ts
@@ -130,6 +130,42 @@ describe('matchRequestUrl', () => {
       },
     })
   })
+
+  test('returns true when matching URLs with wildcard ports', () => {
+    expect(
+      matchRequestUrl(new URL('http://localhost:3000'), 'http://localhost:*'),
+    ).toEqual({
+      matches: true,
+      params: {
+        '0': '3000/',
+      },
+    })
+  })
+
+  test('returns true when matching URLs with wildcard ports and pathnames', () => {
+    expect(
+      matchRequestUrl(
+        new URL('http://localhost:3000/resource'),
+        'http://localhost:*/resource',
+      ),
+    ).toEqual({
+      matches: true,
+      params: {
+        '0': '3000',
+      },
+    })
+  })
+
+  test('returns true for matching WebSocket URLs with wildcard ports', () => {
+    expect(
+      matchRequestUrl(new URL('ws://localhost:3000'), 'ws://localhost:*'),
+    ).toEqual({
+      matches: true,
+      params: {
+        '0': '3000/',
+      },
+    })
+  })
 })
 
 describe('coercePath', () => {
@@ -156,6 +192,10 @@ describe('coercePath', () => {
     expect(coercePath('https://example.com:8080/:5678')).toEqual(
       'https\\://example.com\\:8080/:5678',
     )
+    expect(coercePath('http://localhost:*')).toEqual(
+      'http\\://localhost\\:(.*)',
+    )
+    expect(coercePath('ws://localhost:*')).toEqual('ws\\://localhost\\:(.*)')
   })
 
   test('replaces wildcard with an unnnamed capturing group', () => {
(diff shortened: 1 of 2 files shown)


kettanaito:
Hi, @trknhr. Thanks for opening this! 

I've added a few missing test cases and they are failing at the moment. I don't believe your suggestion has truly implemented the behavior that the users expect. I might take a look at this once I have a moment. Hopefully, shouldn't be a big change.

src/core/utils/matching/matchRequestUrl.test.ts
```diff
@@ -87,6 +85,77 @@ describe('matchRequestUrl', () => {
...
 
+  test('returns true when matching URLs with wildcard ports', () => {
+    expect
+      .soft(
+        matchRequestUrl(new URL('http://localhost:3000'), 'http://localhost:*'),
+      )
+      .toEqual({
+        matches: true,
+        params: {
+          '0': '3000',
```
kettanaito: This one is failing because the URL constructor always appends an implicit `/` pathname if it's not provided in the argument. So by the time coercing runs, `:*` matches against `:3000/` and returns `3000/` as the token match, whereas this isn't what the user typically expects.

kettanaito (approved):
This looks good and the expectations align with #2678.

kettanaito:
Thank you for fixing the issue! 👏

kettanaito:
## Released: v2.12.14 🎉

This has been released in v2.12.14.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.12.14)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.12.14)

```
npm i msw@latest
```

---

PR #2681, opened 2026-03-23
fix(browser): reject duplicate setupWorker calls

### LINKED ISSUE #2585: MSW is not handling requests after `setupWorker` is called more than once
### Prerequisites

### Environment check

### Browsers

### Reproduction repository

https://github.com/diego-aquino/msw-vitest-browser-issue

### Reproduction steps

1. `pnpm install`
2. `pnpm test`

### Current behavior

After upgrading from msw@2.11.1 to msw@2.11.2, many of our tests started to fail. After many hours of investigation, I found out that calling `setupWorker` more than once is causing the test requests to stop being handled.

I created a minimal reproduction simulating this issue: [example.browser.test.ts](https://github.com/diego-aquino/msw-vitest-browser-issue/blob/main/example.browser.test.ts). Restarting an existing worker seems fine:

```ts
it('should handle requests when an existing worker is restarted', async () => {
  let response = await fetch('http://localhost:3000/users');
  expect(response.status).toBe(200);
  expect(await response.json()).toEqual([]);

  worker.stop();
  await worker.start({ quiet: true });

  response = await fetch('http://localhost:3000/users');
  expect(response.status).toBe(200);
  expect(await response.json()).toEqual([]);
});
```

However, calling `setupWorker` a second time makes all requests stop being intercepted:

```ts
it('should handle requests when a new worker is created', async () => {
  let response = await fetch('http://localhost:3000/users');
  expect(response.status).toBe(200);
  expect(await response.json()).toEqual([]);

  worker.stop();
  worker = setupWorker(...handlers); // <-- Creating a new worker cases the problem
  await worker.start({ quiet: true });

  response = await fetch('http://localhost:3000/users'); // <-- Fails as if there was no worker
  expect(response.status).toBe(200);
  expect(await response.json()).toEqual([]);
});
```

```
 FAIL   browser (chromium)  example.browser.test.ts > Example tests (Browser) > should handle requests when a new worker is created
TypeError: Failed to fetch
 ❯ example.browser.test.ts:48:21
     46|     await worker.start({ quiet: true });
     47| 
     48|     response = await fetch('http://localhost:3000/users');
       |                     ^
     49|     expect(response.status).toBe(200);
     50|     expect(await response.json()).toEqual([]);
```

I can confirm that up until msw@2.11.1, calling `setupWorker` more than once caused no problems for us. Also, this issue does not happen at all with `setupServer` (see [example.node.test.ts](https://github.com/diego-aquino/msw-vitest-browser-issue/blob/main/example.node.test.ts)).

### Expected behavior

I saw that [calling `setupWorker` multiple times can lead to issues](https://github.com/mswjs/msw/issues/2581#issuecomment-3280695133). However, up to msw@2.11.1, this did not cause problems for us, and we always make sure to stop the previous worker before calling `setupWorker` again.

Do you consider this is a bug or a usage problem on our part? If it's not a bug, would it be possible to show a warning, throw an error, and/or explain the correct usage in the docs? Sorry if there's already a warning somewhere. I didn't find it.

Moreover, if it's not a bug, why does `setupServer` work differently? Could they behave similarly regarding multiple calls?

trknhr:
Reject duplicate `setupWorker()` calls in the same page by guarding initialization with a page-global symbol. This makes the failure explicit before multiple worker instances can attach conflicting browser-side listeners.

The browser worker state is shared per page, so creating multiple setupWorker instances is ambiguous and leads to clashing request handling. Failing fast here keeps `setupWorker()` as a single-construction API and points users toward reusing the existing worker with `worker.use()` or `worker.resetHandlers()`.

This also adds a browser regression test for the duplicate call path while asserting only on a stable error prefix so the developer-facing guidance can still evolve.

Fixes #2585

### DIFF
--- a/src/browser/setupWorker/setupWorker.ts
+++ b/src/browser/setupWorker/setupWorker.ts
@@ -24,6 +24,8 @@ import { printStartMessage } from './start/utils/printStartMessage'
 import { printStopMessage } from './stop/utils/printStopMessage'
 import { supportsServiceWorker } from '../utils/supports'
 
+const SETUP_WORKER_APPLIED = Symbol.for('msw.setupWorker')
+
 export class SetupWorkerApi
   extends SetupApi<LifeCycleEventsMap>
   implements SetupWorker
@@ -180,5 +182,15 @@ export class SetupWorkerApi
 export function setupWorker(
   ...handlers: Array<RequestHandler | WebSocketHandler>
 ): SetupWorker {
-  return new SetupWorkerApi(...handlers)
+  invariant(
+    Reflect.get(globalThis, SETUP_WORKER_APPLIED) == null,
+    devUtils.formatMessage(
+      'Failed to execute `setupWorker()` more than once in the same page. Create the worker once and reuse it for the lifetime of the page. If you wish to apply a different set of request handlers, use `worker.use()` or `worker.resetHandlers()` instead.',
+    ),
+  )
+
+  const worker = new SetupWorkerApi(...handlers)
+  Reflect.set(globalThis, SETUP_WORKER_APPLIED, true)
+
+  return worker
 }
--- a/src/browser/setupWorker/setupWorker.ts
+++ b/src/browser/setupWorker/setupWorker.ts
@@ -24,7 +24,7 @@ import { printStartMessage } from './start/utils/printStartMessage'
 import { printStopMessage } from './stop/utils/printStopMessage'
 import { supportsServiceWorker } from '../utils/supports'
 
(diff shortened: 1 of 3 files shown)


kettanaito:
Hi, @trknhr. Thanks for proposing this!

I believe the new `defineNetwork` architecture and the `ServiceWorkerSource` have eliminated this issue (see #2650, #2715). I've also added a few missing tests to confirm that. 

Please update to the latest version of the library and let me know if you're still experiencing this issue. Thanks.

PR #2682, opened 2026-03-24
fix(browser): restore setupWorker lifecycle

felmonon:
Follow-up for #2650.

## Summary
- reuse the browser HTTP and WebSocket sources across `worker.start()` / `worker.stop()` cycles
- restore the active `ServiceWorkerRegistration` return value on redundant `worker.start()` calls
- stop service-worker keepalive traffic on `worker.stop()` while preserving post-stop passthrough and restart behavior
- add browser regressions for redundant `worker.start()`, restart after stop, and keepalive cleanup

## Testing
- `pnpm build`
- `pnpm test:browser -- test/browser/msw-api/setup-worker/start/start.test.ts test/browser/msw-api/setup-worker/stop.test.ts test/browser/msw-api/setup-worker/stop/keepalive.test.ts`

### DIFF
--- a/src/browser/setup-worker.ts
+++ b/src/browser/setup-worker.ts
@@ -66,7 +66,50 @@ export function setupWorker(...handlers: Array<AnyHandler>): SetupWorkerApi {
     handlers,
   })
 
+  const webSocketSource = new InterceptorSource({
+    interceptors: [new WebSocketInterceptor() as any],
+  })
+
   let isStarted = false
+  let currentRegistration: ServiceWorkerRegistration | undefined
+  let currentStart: Promise<ServiceWorkerRegistration | undefined> | undefined
+  let currentStop: Promise<void> | undefined
+  let httpSource: ServiceWorkerSource | FallbackHttpSource | undefined
+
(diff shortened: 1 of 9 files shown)


kettanaito:
Hi! I am super confused what you are doing here. Are there any problems you were trying to solve with these?

Also, are you using LLMs to generate your pull requests?

felmonon:
Thanks for calling that out. You're right that I should have made the problem statement much clearer.

This PR was meant to address three browser lifecycle regressions I hit on `fix/define-network`, relative to the current `setupWorker` behavior:

- `worker.stop(); worker.start()` did not re-enable mocking.
- redundant `worker.start()` no longer returned the active `ServiceWorkerRegistration`.
- the service-worker keepalive interval continued after `worker.stop()`.

The added browser tests were meant to lock those behaviors down.

I do use LLM tooling sometimes to help draft and iterate, but I only push changes I can reproduce and validate locally. If these behaviors are intentional on this branch, or if I'm missing the intended direction, I'm happy to close this rather than add noise.

PR #2697, opened 2026-04-08
fix(delay): prevent `infinite` mode from throwing

### LINKED ISSUE #2696: `.delay('infinite')` rejects instead of hanging after 2.12.11
### Prerequisites

### Environment check

### Node.js version

v.24.13.1

### Reproduction repository

https://github.com/ksoni94/msw-reproduction/

### Reproduction steps

clone the reproduction repo and run `npm run test`. The tests should pass indicating that the `Reflect.get` logic throws. 

### Current behavior

Currently, using `delay('infinite')` in `jsdom` isn't working as expected. I think this is because in `jsdom` the `setTimeout` function returns a number instead of an object as its replicating a browser environment. 

I believe this is causing the promise to reject instead of hang which means the delay doesn't actually apply? 

### Expected behavior

`delay('infinite')` should cause the promise to hang. I think the fix might be to update the logic to something like: 

```typescript
  function hasRefCounted(value) {
    return typeof value === "object" && value !== null
      && typeof Reflect.get(value, "ref") === "function"
      && typeof Reflect.get(value, "unref") === "function";
  }
```

kettanaito:
- Fixes #2696

### DIFF
--- a/src/core/utils/internal/hasRefCounted.test.ts
+++ b/src/core/utils/internal/hasRefCounted.test.ts
@@ -14,3 +14,24 @@ it('returns false for a non-refcounted object', () => {
   expect(hasRefCounted({ ref() {} })).toBe(false)
   expect(hasRefCounted({ unref() {} })).toBe(false)
 })
+
+it('returns false for non-object values', () => {
+  expect(
+    hasRefCounted(
+      // @ts-expect-error Runtime value.
+      null,
+    ),
+  ).toBe(false)
+  expect(
+    hasRefCounted(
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v2.13.2 🎉

This has been released in v2.13.2.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.13.2)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.13.2)

```
npm i msw@latest
```

---

PR #2702, opened 2026-04-10
fix: use `verbatimModuleSyntax` consistently in exports

christoph-fricke:
The new network-source API is not usable as soon as a project tries to import from `msw/experimental`. **It crashes with a runtime error:**

```
SyntaxError: The requested module './define-network.mjs' does not provide an export named 'DefineNetworkOptions'
```

This PR fixes the problem by using a type export in _src/core/experimental/index.ts_ to ensure `DefineNetworkOptions` is stripped from the JS output.

---

I initially tried to prevent such problems in the future by enabling `verbatimModule` syntax in `tsconfig.base.json` and fixing multiple TS errors reported by `pnpm exec tsc --noEmit -p src/tsconfig.src.json`, but it caused strange problems in CI.

Besides this, there appear to be multiple TS errors in the project. Some picked in the editor, some only reported by checking with one of the various TS config files such as `pnpm exec tsc --noEmit -p ./tsconfig.test.unit.json`. Some source code errors I noticed:

- /src/browser/sources/fallback-http-source.ts
- /src/browser/utils/deserializeRequest.ts
- /src/browser/utils/pruneGetRequestBody.ts

It does not look like the CI finds and reports the errors. Are you aware of these existing TypeScript problems?


src/browser/utils/deserializeRequest.ts
```diff
@@ -1,12 +1,12 @@
 import { pruneGetRequestBody } from './pruneGetRequestBody'
-import type { ServiceWorkerIncomingRequest } from '../glossary'
```
christoph-fricke, line 2: This type does not exist in _glossary_...

src/core/experimental/index.ts
```diff
@@ -1,4 +1,5 @@
-export { defineNetwork, DefineNetworkOptions } from './define-network'
+export { defineNetwork } from './define-network'
+export type { DefineNetworkOptions } from './define-network'
```
christoph-fricke: For some reason it has to be this separate `export type {....}`. `export {..., type ...}` appears to not strip the interface from the output _/lib/core/experimental/index.mjs_ file (unless I was looking at some cached file...).

kettanaito:
Thanks for opening this, @christoph-fricke. I've discovered a bigger issue: we weren't using `verbatimModuleSyntax` at all so there was no restriction to export types properly from the public entrypoints. 

I've fixed that and updated _everything_ we export/import publicly and internally.

src/core/experimental/index.ts
kettanaito: That's no longer necessary. Not sure why, maybe because I've modified the ts config. Can you double-check please?

christoph-fricke:
@kettanaito Awesome! When I tried to `verbatimModuleSyntax` it caused weird, seemingly unrelated runtime errors in CI. I enabled it in `tsconfig.base.json` as well... Weird.

But I am glad it works now! Installed a local built into https://github.com/mswjs/playwright/pull/48 and the tests run without crashing. Looks good to go from my POV.

src/browser/utils/deserializeRequest.ts
christoph-fricke: @kettanaito Looks like your change added this line back. When I look at the file in VS Code (lokal TS SDK) I get an error `Module '"../glossary"' has no exported member 'ServiceWorkerIncomingRequest'.`. Is the error not reported for you?

kettanaito: Good catch! Patched in.

kettanaito:
## Released: v2.13.4 🎉

This has been released in v2.13.4.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.13.4)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.13.4)

```
npm i msw@latest
```

---

PR #2705, opened 2026-04-11
fix: hermes crash rettime

JavanPoirier:
v2.13.2 breaks RN mocking. Maybe freeze the `rettime` dep, as it is susceptible to breaking changes?

On React Native with Hermes, polyfilled Event doesn't define stopPropagation on its prototype. MSW's WebSocketHandler wraps it with new Proxy(event.stopPropagation, ...), which throws TypeError: new Proxy target must be an Object when stopPropagation is undefined. This is the same class of bug that was fixed in rettime@0.11.x.

### Changes
- src/core/handlers/WebSocketHandler.ts — Replace new Proxy(event.stopPropagation, { apply }) with a plain function wrapper using optional chaining:
```
// Before — crashes when stopPropagation is undefined
event.stopPropagation = new Proxy(event.stopPropagation, {
  apply: (target, thisArg, args) => { ... },
})

// After — safe on Hermes
const originalStopPropagation = event.stopPropagation
event.stopPropagation = function (...args) {
  Reflect.get(event, KOnStopPropagation)?.call(handler)
  return originalStopPropagation?.apply(this, args)
}
```
- test/native/hermes-event-polyfill.native.test.ts — Regression test that stubs a minimal Hermes-like Event/MessageEvent (no stopPropagation on prototype) and verifies setupServer() from msw/native doesn't crash. Runs in the existing test:native vitest suite with no new dependencies.

- src/core/handlers/WebSocketHandler.test.ts — Unit tests for createStopPropagationListener covering both the Hermes case (undefined stopPropagation) and standard Events.


kettanaito:
Hi, @JavanPoirier. Appreciate the detailed explanation of the issue. But I have a serious concern with the proposed solution. You are suggesting we tailor to a broken polyfill. There is nothing wrong relying on `Event.prototype.stopPropagation` to exist. A proper solution here is to open an issue for that polyfill and make them implement that method or at least create a dummy function like Node.js does if the implementation is inapplicable.

I am open to discussion on this, but at its current state, I cannot accept these suggestions.

JavanPoirier:
> Hi, @JavanPoirier. Appreciate the detailed explanation of the issue. But I have a serious concern with the proposed solution. You are suggesting we tailor to a broken polyfill. There is nothing wrong relying on `Event.prototype.stopPropagation` to exist. A proper solution here is to open an issue for that polyfill and make them implement that method or at least create a dummy function like Node.js does if the implementation is inapplicable.
> 
> I am open to discussion on this, but at its current state, I cannot accept these suggestions.

@kettanaito, valid concern. I have updated the PR to continue to rely on stopPropagation. Bumping rettime will support HTTP-only mocks.

I am not knowledgeable enough in this area to verify this, but if you could cross reference my AI agents response, it appears Websockets & SSE mocks will still be imapcted... 

"`rettime@0.11.7` changes the wrapping strategy from Proxy (which validates its target eagerly) to a plain function (which captures the value lazily). Since MSW's normal code path never invokes stopPropagation() on emitted events, the undefined is never dereferenced, and everything works.". 

"Polyfills alone cannot work around the new Proxy(undefined, ...) crash in rettime 0.10.x unless you can guarantee that Event.prototype.stopPropagation is a real function on every event instance before rettime ever calls emit() — and even then, the import ordering issue makes this extremely fragile."

image

Specifically: "To make WebSocket mocking work on Hermes, MSW would also need to change createStopPropagationListener in WebSocketHandler.ts to avoid new Proxy() — using the same pattern rettime 0.11.7 adopted (capture + plain function wrapper instead of Proxy). "

kettanaito:
I've opened https://github.com/mswjs/msw/pull/2711. Does this fix your issue? 

> To make WebSocket mocking work on Hermes, MSW would also need to change createStopPropagationListener in WebSocketHandler.ts to avoid new Proxy() — using the same pattern rettime 0.11.7 adopted (capture + plain function wrapper instead of Proxy). 

Nothing is using `new Proxy` anymore. Not sure what this comment means.

kettanaito:
## Released: v2.13.3 🎉

This has been released in v2.13.3.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.13.3)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.13.3)

```
npm i msw@latest
```

---

PR #2710, opened 2026-04-13
fix(WebSocketFrame): allow for broader connection types

### LINKED ISSUE #2707: [experimental] `WebSocketNetworkFrame` cannot be initialized with `WebSocket(Client|Server)ConnectionProtocol` implementations
_This issue has come up while [migrating `@msw/playwright`](https://github.com/mswjs/playwright/pull/48) to the new network source architecture._

## Description

When constructing an implementation of the `WebSocketNetworkFrame`, it requires a `WebSocketClientConnection` and `WebSocketServerConnection` connection. This results in errors when providing `WebSocket(Client|Server)ConnectionProtocol` implementations.

```ts
const frame = new PlaywrightWebSocketNetworkFrame({
  connection: {
    client: new PlaywrightWebSocketClientConnection(route),
    server: new PlaywrightWebSocketServerConnection(route),
    info: { protocols: [] },
  },
})
```

- **client:** Type 'PlaywrightWebSocketClientConnection' is missing the following properties from type 'WebSocketClientConnection': socket, transport, [kEmitter$1]
- **server:** Type 'PlaywrightWebSocketServerConnection' is missing the following properties from type 'WebSocketServerConnection': client, transport, createConnection, mockCloseController, and 7 more.

Furthermore, it results in errors when passing protocol implementations to [`webSockerHandler.run(...)`](https://github.com/christoph-fricke/msw-playwright/blob/feat/network-source-api/src/fixture.ts#L176).

## Proposed Solution

As pointed out [here](https://github.com/mswjs/playwright/pull/48#discussion_r3066890254), this seems to be a bug since `WebSocket(Client|Server)ConnectionProtocol` is the intended way for working with `WebSocketNetworkFrame`.

> I'm getting the WebSocketConnectionData from the Interceptors as-is, but that's a mistake. That type is a narrower type than what the frame should care about.
> 
> We should probably annotate that connection option as WebSocketHandlerConnection from WebSocketHandler.ts, which contains the broader WebSocketClientConnectionProtocol.

kettanaito:
- Fixes #2707 

I'm a bit tentative about this approach. 

- [ ] Find a proper name for `WebSocketConnectionFoo`.

### DIFF
--- a/src/core/experimental/frames/websocket-frame.ts
+++ b/src/core/experimental/frames/websocket-frame.ts
@@ -1,9 +1,9 @@
 import { TypedEvent } from 'rettime'
-import { type WebSocketConnectionData } from '@mswjs/interceptors/WebSocket'
 import {
   kConnect,
   kAutoConnect,
   type WebSocketHandler,
+  type WebSocketConnectionFoo,
 } from '../../handlers/WebSocketHandler'
 import {
   NetworkFrame,
@@ -16,8 +16,10 @@ import {
 import { devUtils } from '../../utils/internal/devUtils'
 import { HandlersController, AnyHandler } from '../handlers-controller'
 
-export interface WebSocketNetworkFrameOptions {
-  connection: WebSocketConnectionData
+export interface WebSocketNetworkFrameOptions<
+  Connection extends WebSocketConnectionFoo,
+> {
+  connection: Connection
 }
 
 export type WebSocketNetworkFrameEventMap = {
@@ -68,14 +70,16 @@ class UnhandledWebSocketExceptionEvent<
   }
 }
 
-export abstract class WebSocketNetworkFrame extends NetworkFrame<
+export abstract class WebSocketNetworkFrame<
+  Connection extends WebSocketConnectionFoo,
+> extends NetworkFrame<
   'ws',
   {
-    connection: WebSocketConnectionData
+    connection: Connection
   },
   WebSocketNetworkFrameEventMap
 > {
-  constructor(options: WebSocketNetworkFrameOptions) {
+  constructor(options: WebSocketNetworkFrameOptions<Connection>) {
     super('ws', {
       connection: options.connection,
     })
--- a/src/core/experimental/sources/interceptor-source.ts
+++ b/src/core/experimental/sources/interceptor-source.ts
@@ -154,7 +154,7 @@ class InterceptorHttpNetworkFrame extends HttpNetworkFrame {
   }
 }
 
-class InterceptorWebSocketNetworkFrame extends WebSocketNetworkFrame {
+class InterceptorWebSocketNetworkFrame extends WebSocketNetworkFrame<WebSocketConnectionData> {
   constructor(args: { connection: WebSocketConnectionData }) {
     super({ connection: args.connection })
   }
--- a/src/core/handlers/WebSocketHandler.ts
+++ b/src/core/handlers/WebSocketHandler.ts
@@ -22,10 +22,13 @@ export type WebSocketHandlerEventMap = {
   connection: [args: WebSocketHandlerConnection]
 }
 
-export interface WebSocketHandlerConnection {
+export interface WebSocketConnectionFoo {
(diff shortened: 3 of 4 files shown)


kettanaito:
@christoph-fricke, hi 👋 What do you think about this approach? Do you see any potential issues with this?

christoph-fricke (commented):
@kettanaito I packed it locally and tested it with `@msw/playwright`. With this, the existing protocol implementations work and the tests pass. Just had to add a wonky import to `node_modules/../../WebSocketHandler` to get to the `WebSocketConnectionFoo` type.

Not too sure what I think about the implementation. It feels a bit "bolted on", like trying to squeeze the protocol interfaces in afterwards. But that might be just me knowing the history...
Will it cause big observable differences between sources with `WebSocketConnectionFoo` and sources with `WebSocketConnectionData`?

Left an additional comment in #2707 that might make this PR obsolete.

kettanaito:
The dichotomy comes from the frames not guaranteeing to be created by the WebSocketInterceptor anymore. We have that assumption right now and it's rather strong (nothing wrong with that since `ws` works via the interceptor). But with custom network sources, they can yield whichever frames they'd like. Architecture-wise the `Foo` class makes perfect sense. 

The union with `Data` class is there only for posterity. The `ws` functionality, effectively, has to branch since now WebSocketFrame can have a connection that _wasn't_ created by the interceptor and, as a result, doesn't have `client.socket` and, potentially, other stuff. **This** is a bit troubling as I cannot predict whether our internal `ws` (and what developers expect from it) and user-defined WebSocketFrames won't conflict/miss in behaviors.

src/core/experimental/frames/websocket-frame.ts
```diff
@@ -16,8 +16,10 @@ import {
 import { devUtils } from '../../utils/internal/devUtils'
 import { HandlersController, AnyHandler } from '../handlers-controller'
 
-export interface WebSocketNetworkFrameOptions {
-  connection: WebSocketConnectionData
+export interface WebSocketNetworkFrameOptions<
```
christoph-fricke, line 19: As far as I understand, the `ConnectionProtocols` interfaces are the intended abstraction for decoupling MSW internals from WebSocket specifics, right? They are supposed to enable custom frames that extend can extend `WebSocketNetworkFrame`.
The previous `@msw/playwright` implementation already proofed that it is possible to implement `ConnectionProtocols` with something that is not a `WebSocketInterceptor`. For other alternative sources, I am toying with the idea of a `HttpServerSource` and `WebSocketServerSource`, i.e. sources that create an actual server. I have some use cases for that... I don't see blockers creating `ConnectionProtocols` there either.

**The biggest change I would make:** I think `ConnectionProtocols` as the abstraction target is better "highlighted" and "communicated" when the `WebSocketNetworkFrame` does not accept an generic connection. It feels like overhead and is just one more think to understand. To avoid more confusion, it might help if these interfaces do not need imports from `@mswjs/interceptors`, but that might be unfeasible. In the end, most people will likely just use existing sources and not implement their own sources and frames... Adjusted frame options:
```ts
interface WebSocketNetworkFrameOptions {
  connection: WebSocketConnectionFoo // <-- Type must be package exported
}

// I actually prefer it flattened (avoids having to find a name for "Foo"):
interface WebSocketNetworkFrameOptions {
  client: WebSocketClientConnectionProtocol
  server: WebSocketServerConnectionProtocol
  info: WebSocketConnectionData['info']
}
```

I quickly verify that removing the generic works. It only requires a small adjustment to the `InterceptorWebSocketNetworkFrame`:
```ts
class InterceptorWebSocketNetworkFrame extends WebSocketNetworkFrame {
  #client: WebSocketConnectionData['client']

  constructor(args: { connection: WebSocketConnectionData }) {
    super({ connection: args.connection })
    this.#client = args.connection.client
  }

  public errorWith(reason?: unknown): void {
     // ... collapsed code
     this.#client.socket.dispatchEvent(errorEvent)
  }

  public passthrough() {
    this.data.connection.server.connect()
  }
}
```

PR #2725, opened 2026-04-20
fix: reset generator state on `.resetHandlers()`/`.restoreHandlers()`

kettanaito:
## Problem

Calling `.resetHandlers()`/`.restoreHandlers()` does not reset the generator state if the response resolver is a generator function. That renders exhausted generators effectively useless. User expectation is that either of those methods will transition generator resolvers into a "fresh" state (`.resetHandlers()` because it's a full reset; `.restoreHandlers()` because `yield` is a substitute for one-time responses). 

## Todos

- [x] `.restoreHandlers()` must ONLY reset the generator state if the handler is a one-time handler. Otherwise, it has no effect on the generator state.
  - [x] Add two tests for this: restore + one-time, restore + regular handler.


kettanaito:
## Released: v2.13.5 🎉

This has been released in v2.13.5.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.13.5)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.13.5)

```
npm i msw@latest
```

---

PR #2728, opened 2026-04-29
fix(HttpResponse): forward cookies only when response is used

kettanaito:
Currently, we are forwarding the mocked response cookies onto `document.cookie` as a part of the `HttpResponse` constructor. That means that even constructing an `HttpResponse` instance will immediately forward its `Set-Cookie` onto `document.cookie`, even if that response never gets used (e.g. is conditional).

## Changes

- `HttpResponse` no longer forwards its `Set-Cookie` onto `document.cookie` in the constructor.
- Response cookie forwarding is moved down to `RequestHandler.run` and applies only to the mocked response getting used.


kettanaito:
## Released: v2.14.0 🎉

```
npm i msw@latest
```

---

PR #2729, opened 2026-04-29
feat: drop support for Node.js 18, 20

kettanaito:


kettanaito:
## Failing Node.js tests

Due to Interceptors not supporting Node.js 24 as of now. Requires https://github.com/mswjs/interceptors/pull/766 and also a fix regarding `response.url` as the internal urlList likely changed and we aren't setting the response urls correctly now.

PR #2731, opened 2026-04-29
fix: do not clone user-provided responses

kettanaito:
Right now, MSW always clones the user-provided mocked response. That disrupts the response's identity, which is relevant in transient response scenarios like on Cloudflare (e.g. providing custom response init options and preserving them). 

The issue is fundamental: instead of cloning a response for logs and using the original, we are cloning for the usage and using the original for logs. This is against what the rest of the architecture is doing (e.g. request clones are handled correctly).


kettanaito:
## Released: v2.14.0 🎉

```
npm i msw@latest
```

---

PR #2732, opened 2026-04-29
feat: support `ws.onUpgrade` for handling connection upgrades

### LINKED ISSUE #2436: Handle HTTP handshakes for WebSocket connections
### Scope

Improves an existing behavior

### Compatibility

### Feature description

- Context in #2414

## Summary

Some third-party WebSocket clients, like `ws`, may integrate a manual HTTP handshake for WebSocket connections. If such occur, Interceptors should handle them appropriately.

I am raising this in MSW because Interceptors may be too low-level of a library to handle this. Effectively, where we do this doesn't matter. Just has to be accounted for. 

## Proposal

The handshake handling itself isn't that complicated:

```ts
http.get(theSameUrlAsWsLink, ({ request }) => {
  if (request.headers.get('upgrade') === 'websocket') {
    return new HttpResponse(null, {
      status: 101,
      headers: {
        'Upgrade': 'websocket',
        'Connection': 'Upgrade',
        'Sec-WebSocket-Accept': 's3pPLMBiTxaQ9kYGzzhZRbK+xOo=',
      },
    })
  }
})
```

What's complicated is the ergonomics of a handler (ws link) that unfolds into multiple handlers (ws + http.get). MSW doesn't have a concept of a compound handler like that at the moment, and I would hate to ask people to spread `...ws.link()` in their handlers.

kettanaito:
- Fixes #2436

## Problem

We currently don't support WebSocket connections created by upgrading HTTP connections (`Upgrade` requests). 

## Solution

> [!IMPORTANT]
> Note that this will only cover upgrade requests in Node.js. In the browser, the `upgrade` request header is forbidden and cannot be read. Additionally, the clients should be using the WebSocket API instead. 

- `ws.onUpgrade` is a callback for any upgrade requests matching the defined WebSocket link URLs.
- Upgrade handling is applied _to the entire environment_. Having different upgrade mechanisms on a per-link basis is a no-op. Upgrade is always dictated by the environment. 
- The default `onUpgrade` produces a 101 mocked response in correspondence to the spec.

## Other changes

- Adds the `attachSiblingHandlers` and `getSiblingHandlers` utilities. A single handler can now have additional handlers attached alongside it.

## Todos

- [x] Add the upgrade listener _once_. It's per link, not per `.addEventListener()` connection handler. 
- [x] `Upgrade` and `Sec-*` request headers are forbidden. We cannot read them in the browser.


kettanaito:
## Released: v2.14.0 🎉

```
npm i msw@latest
```

---

PR #2736, opened 2026-05-03
fix: release HttpNetworkFrame listeners after response event

### LINKED ISSUE #2735: Memory leak: per-frame `Emitter` retained for the network's lifetime via shared `AbortSignal` (interceptor-source)
### Prerequisites

### Environment check

### Node.js version

v24.14.1 (also reproduces on v22.x)

### Reproduction repository

https://github.com/YH1715/msw-emitter-leak-repro

### Reproduction steps

```bash
git clone https://github.com/<your-handle>/msw-emitter-leak-repro
cd msw-emitter-leak-repro
npm install
npm test
```

The script (`repro.mjs`):

1. Sets up `setupServer()` with a single mocked endpoint
2. Fires 10,000 axios GET requests against it
3. Forces GC and takes heap snapshots before and after
4. Counts retained constructor instances and prints the delta

### Current behavior

Every intercepted request retains one rettime `Emitter` (plus its internal `LensList` / `WeakMap` / `WeakSet`) for the entire lifetime of the network — they are never released until `setupServer().close()`.

### Reproduction output

```
Firing 10000 requests at concurrency 20...

Constructor count after 10000 requests:

  name          baseline    settled       delta
  ------------------------------------------
  Emitter            202      10202     +10000
  Listener           204      10204     +10000
  LensList           404      20404     +20000
  WeakMap            822      40822     +40000
  WeakSet            206      10206     +10000

Result: LEAK DETECTED
```

The deltas are exact multiples of the request count. Each request retains exactly:

- 1 `Emitter` (the frame's `events` instance from `NetworkFrame`)
- 1 `Listener` (the registration record for the `"*"` listener)
- 2 `LensList` (`#listeners` and `#hookListeners` inside the Emitter)
- 4 `WeakMap` (`#listenerOptions`, `#listenerAbortCleanups`, `#hookListenerOptions`, `#hookListenerAbortCleanups`)
- 1 `WeakSet` (`#typelessListeners`)
- 2 `Map` (one inside each `LensList`)

This matches the entire internal field set of a single `rettime` `Emitter` per request.

### Root cause

`src/core/experimental/define-network.ts` (compiled to `lib/core/experimental/define-network.mjs`):

```js
enable() {
  listenersController = new AbortController();   // shared, lives for the whole network
  ...
  resolvedOptions.sources.map((source) => {
    NetworkSource.prototype.disable.call(source);
    source.on("frame", async ({ frame }) => {
      frame.events.on("*", (event) => events.emit(event), {
        signal: listenersController.signal       // every per-frame listener attaches an abort handler to the SAME signal
      });
      const handlers = frame.getHandlers(handlersController);
      await frame.resolve(...);
    });
    return source.enable();
  });
}
```

`rettime` registers an `abort` event listener on the provided `signal` for cleanup:

```js
// node_modules/rettime/build/index.mjs (Emitter)
if (options.signal) {
  const { signal } = options;
  const onAbort = () => {
    this.removeListener(type, listener);   // captures `this` = per-frame Emitter
  };
  signal.addEventListener("abort", onAbort, { once: true });
  this.#listenerAbortCleanups.set(listener, () => {
    signal.removeEventListener("abort", onAbort);
  });
}
```

Retainer chain after N requests:

```
frame.events (Emitter, holds LensList + 4 WeakMaps + WeakSet)
  ↑ captured in: onAbort closure  (one closure per request)
  ↑ stored on:   listenersController.signal  (one shared signal)
  ↑ owned by:    listenersController  (AbortController, network-lifetime)
  ↑ closure of:  defineNetwork's enable()
```

`onAbort` is registered with `{ once: true }`, so it would self-remove if the signal ever aborted — but the signal only aborts on `network.disable()`, which in a long-running server never happens. The retainer chain therefore holds the per-frame Emitter (and its internal collections) on the shared `AbortSignal` for the whole network lifetime.

### Possible fix

Release the per-frame listeners after the frame's last event (`response:mocked` / `response:bypass`) is emitted. This triggers `#listenerAbortCleanups`, which removes `onAbort` from `listenersController.signal`, breaking the retention chain. The cleanup runs **after** the response event has been dispatched, so user code listening on `network.events.on("response:mocked", ...)` is not affected.

`src/core/experimental/sources/interceptor-source.ts`:

```diff
   queueMicrotask(() => {
     httpFrame.events.emit(
       new ResponseEvent(
         isMockedResponse ? "response:mocked" : "response:bypass",
         {
           requestId,
           request,
           response
         }
       )
     );
+    httpFrame.events.removeAllListeners();
   });
```

The same one-liner applies to the matching `.mjs` / `.cjs` / iife bundles.

After applying the patch, the reproduction prints:

```
Constructor count after 10000 requests:

  name          baseline    settled       delta
  ------------------------------------------
  Emitter            202        202         +0
  Listener           204        204         +0
  LensList           404        404         +0
  WeakMap            822        822         +0
  WeakSet            206        206         +0

Result: OK
```

### Notes on lifecycle

- For request paths that complete normally (mocked or bypass-with-real-response), `#handleResponse` runs and the patch above releases the frame's listeners.
- For error paths where the interceptor never emits `response` (e.g., `errorWith()` aborts the request), the frame is also never removed from `this.#frames`, which is a smaller secondary leak. Probably worth handling, though I haven't reproduced it in the wild.
- The browser/iife bundle has the same code path; the same one-liner applies there too.

### Could this also be fixed in rettime instead?

Yes, but invasively. The cleanest rettime-side fix would be to hold the Emitter via a `WeakRef` inside the `onAbort` closure. That keeps `signal.removeEventListener` callable but stops `signal` from retaining the Emitter. The msw-side `removeAllListeners()` patch is much less risky and ships with one line of code.

### Expected behavior

After each request completes, the per-request `Emitter` (and its internal `LensList` / `WeakMap` / `WeakSet`) should be eligible for GC. Constructor counts in heap snapshots should not grow with request count — the `Emitter` delta after 10,000 requests should be ~0, not ~10,000.

Equivalently: `npm test` in the reproduction repo should print `Result: OK` (and exit code 0) instead of `Result: LEAK DETECTED` (exit code 1).

Zelys-DFKH:
Closes #2735

Hi! Jumping in with a fix for the per-frame listener retention described in #2735.

**Root cause**

`defineNetwork` attaches a wildcard listener to each `HttpNetworkFrame.events` with `{ signal: listenersController.signal }`. That signal lives for the full network lifetime and only aborts on `disable()`. Because `rettime` registers an `onAbort` closure on that signal for every listener attached, every completed request leaves one retained `Emitter` (and its `LensList`, `WeakMap`, `WeakSet` fields) pinned to the shared signal. After 10k requests that's 10k retained emitters; the heap snapshot in the issue shows deltas that are exact multiples of the request count.

**What this PR does**

One line: `httpFrame.events.removeAllListeners()` inside the existing `queueMicrotask` in `InterceptorSource.#handleResponse`, right after the `ResponseEvent` is emitted.

This triggers `rettime`'s `#listenerAbortCleanups` path, which detaches the `onAbort` closure from the shared signal and lets the frame's emitter be GC'd. Calling it after the emit (but still inside the microtask) means all `response:mocked`/`response:bypass` subscribers receive the event before cleanup. Existing code listening on either event is unaffected.

This fix works at the MSW layer regardless of rettime internals, so it holds across rettime versions.

**Test**

New unit test in `interceptor-source.test.ts`: intercepts a real `fetch()` via `FetchInterceptor`, responds synchronously from the frame listener, then asserts `removeAllListeners()` is called on the frame's `events` emitter after the microtask queue flushes. Fails on the pre-fix code, passes after.

**Out of scope**

The issue also notes a secondary leak for error paths where the interceptor never fires the `response` event. Happy to follow up on that separately if useful.

### DIFF
--- a/src/core/experimental/sources/interceptor-source.test.ts
+++ b/src/core/experimental/sources/interceptor-source.test.ts
@@ -1,0 +1,33 @@
+// @vitest-environment node
+import { FetchInterceptor } from '@mswjs/interceptors/fetch'
+import type { HttpNetworkFrame } from '../frames/http-frame'
+import { InterceptorSource } from './interceptor-source'
+
+afterEach(() => {
+  vi.restoreAllMocks()
+})
+
+it('removes frame event listeners after the response event is emitted', async () => {
+  const source = new InterceptorSource({
+    interceptors: [new FetchInterceptor()],
+  })
+
+  let capturedFrame: HttpNetworkFrame | undefined
+
+  source.on('frame', ({ frame }) => {
+    capturedFrame = frame as unknown as HttpNetworkFrame
+    capturedFrame.respondWith(new Response('ok'))
+  })
+
+  source.enable()
+
+  await fetch('http://localhost/test')
+
+  const removeAllListenersSpy = vi.spyOn(capturedFrame!.events, 'removeAllListeners')
+
+  await new Promise<void>((resolve) => queueMicrotask(resolve))
+
+  expect(removeAllListenersSpy).toHaveBeenCalledOnce()
+
+  source.disable()
+})
--- a/src/core/experimental/sources/interceptor-source.test.ts
+++ b/src/core/experimental/sources/interceptor-source.test.ts
@@ -29,5 +29,6 @@ it('removes frame event listeners after the response event is emitted', async ()
 
   await new Promise<void>((resolve) => queueMicrotask(resolve))
 
+  expect(removeAllListenersSpy).toBeDefined()
   expect(removeAllListenersSpy!).toHaveBeenCalledOnce()
 })
--- a/src/core/experimental/sources/interceptor-source.test.ts
+++ b/src/core/experimental/sources/interceptor-source.test.ts
@@ -3,31 +3,31 @@ import { FetchInterceptor } from '@mswjs/interceptors/fetch'
 import type { HttpNetworkFrame } from '../frames/http-frame'
 import { InterceptorSource } from './interceptor-source'
 
+let source: InterceptorSource
+
 afterEach(() => {
+  source?.disable()
   vi.restoreAllMocks()
 })
 
 it('removes frame event listeners after the response event is emitted', async () => {
-  const source = new InterceptorSource({
+  source = new InterceptorSource({
     interceptors: [new FetchInterceptor()],
   })
 
-  let capturedFrame: HttpNetworkFrame | undefined
+  let removeAllListenersSpy: ReturnType<typeof vi.spyOn> | undefined
 
   source.on('frame', ({ frame }) => {
-    capturedFrame = frame as unknown as HttpNetworkFrame
-    capturedFrame.respondWith(new Response('ok'))
+    const httpFrame = frame as unknown as HttpNetworkFrame
+    removeAllListenersSpy = vi.spyOn(httpFrame.events, 'removeAllListeners')
+    httpFrame.respondWith(new Response('ok'))
   })
 
   source.enable()
 
   await fetch('http://localhost/test')
 
-  const removeAllListenersSpy = vi.spyOn(capturedFrame!.events, 'removeAllListeners')
-
   await new Promise<void>((resolve) => queueMicrotask(resolve))
 
-  expect(removeAllListenersSpy).toHaveBeenCalledOnce()
-
-  source.disable()
+  expect(removeAllListenersSpy!).toHaveBeenCalledOnce()
 })
--- a/src/core/experimental/sources/interceptor-source.ts
+++ b/src/core/experimental/sources/interceptor-source.ts
@@ -97,6 +97,7 @@ export class InterceptorSource extends NetworkSource {
           },
         ),
       )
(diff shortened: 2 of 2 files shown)


Zelys-DFKH:
Good catches from the automated review — both issues were valid:

1. **Cleanup leak** — moved `source.disable()` into `afterEach` so it runs even if the assertion throws.
2. **Spy timing race** — moved the spy setup into the `'frame'` listener before `respondWith`. This guarantees the spy is in place before the microtask that calls `removeAllListeners` is queued, making the test deterministic regardless of how `FetchInterceptor` sequences the response event relative to the fetch promise resolution.

Both are addressed in the latest commit.

kettanaito:
## Released: v2.14.3 🎉

This has been released in v2.14.3.

- 📄 [**Release notes**](https://github.com/mswjs/msw/releases/tag/v2.14.3)
- 📦 [View on npm](https://www.npmjs.com/package/msw/v/2.14.3)

```
npm i msw@latest
```

---

Issue #2481, opened 2025-04-24
Transform WebSocket `connection` on the `ws.link` level


kettanaito:
### Discussed in https://github.com/mswjs/msw/discussions/2480

<div type='discussions-op-text'>

<sup>Originally posted by **colinbes** April 22, 2025</sup>
I battled coming up with a title to hint at what I am wanting to do.

In my handler I have various rest end point mock handlers as well as io socket listener for mocking of websockets.

I am wanting to mock my backend server behavior which emits status messages of application (running in backend).

For my mocks, as example, when endpoint PUT /app/generator/start I want to start an interval timer and every interval period emit a websocket message sending update of backend process.

I see that ws.link's return value has a broadcast member - would it be correct to use this to emit message to all connected clients?

I am not sure how I would send event plus message though - something like client.emit("send:status", "some valuable update")</div>

kettanaito:
## Motivation

If your WebSocket connection is through SocketIO, methods like `.broadcast()` won't work since they are scoped to the link and not to the connection listener. This undermines a significant chunk of WebSocket mocking and should be addressed. 

## Proposal

```ts
import { SocketIoTransport } from '@mswjs/socket.io-binding'

ws.link(url, {
  transport: new SocketIoTransport()
})
```

We can reuse the existing `WebSocketTransport` for this. But the transport modification is applied to the whole `connection` object _in-place_. We may preserve the raw `connection` under a different property in the event listener.

hamzaQuantiphi:
Noob question, but would this also work when apps connect to remote server using socket.io-client?
I can't imagine a scenario when apps WONT be using socket.io-client, and instead use WebSocket natively in production applications. I could be wrong. 

I have an Angular App and am unable to connect to MSW's socker handler because the app seems to want to connect over SocketIO's protocol. 

Example:

**handlers.ts**

```

import { toSocketIo } from "@mswjs/socket.io-binding";
import { delay, http, HttpResponse, ws } from "msw";

const socketIoServer = ws.link(
	`${environment.apiUrl.replace("https://", "wss://")}`
);

export const handlers = [
          socketIoServer.addEventListener("connection", (connection) => {
		console.log("outgoing WebSocket connection");

		const { client, server } = toSocketIo(connection);

		server.on("message", () => {
			console.log("in server");
		});

		client.on("message", (event) => {
			console.log("in:", event.data);
		});
	}),
        ...
]

```

**socket.service.ts**

```
import { Injectable } from "@angular/core";
[...]

PR #2742, opened 2026-05-11
feat: support ws.link connection transform

raashish1601:
Implemented optional `transform` callback support for `ws.link()` and wired it into `WebSocketHandler.run()` so listeners receive transformed `connection` objects. Added focused unit coverage for runtime behavior and typings.

### DIFF
--- a/src/core/handlers/WebSocketHandler.ts
+++ b/src/core/handlers/WebSocketHandler.ts
@@ -29,6 +29,10 @@ export interface WebSocketHandlerConnection {
   params: PathParams
 }
 
+export type WebSocketHandlerConnectionTransformer = (
+  connection: WebSocketHandlerConnection,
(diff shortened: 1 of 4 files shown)


kettanaito:
Hi @raashish1601. Thanks for opening this. I will keep this as prior art, but as of now, transforming WebSocket connections needs more thought. In particular:

- Transformed connections should be exposed on the `connection` life cycle event, too, for consistency. 
- WebSocket connections can originate cross-process. In that case, `transform` cannot accept the connection data itself but instead a serialized `{id,url}` representation of any connection.

I'm also unsure what's the primary use case for transforming connections. We have #2481, but it seems to chase a different use case than you here. The original discussion #2480 is likely a bug in the SocketIO transform, not a sign that we need new API.

PR #2768, opened 2026-07-22
chore: drop legacy forceFileExtensionsPlugin

Andarist:


kettanaito (approved):
Let's go! Always happy to remove code. Thank you!

PR #2769, opened 2026-07-22
chore: simplify tsdown config

Andarist:


kettanaito (approved):
Beautiful. Let me know if this is ready to be merged! And thank you.

Andarist:
I'll let you know in a couple of days. This has some *minor* differences and I need to reconsider how this fits your `.d.ts` generation needs.

config/polyfills-node.ts
```diff
@@ -1,8 +0,0 @@
-import { setTimeout as nodeSetTimeout } from 'timers'
-
-// Polyfill the global "setTimeout" so MSW could be used
-// with "jest.useFakeTimers()". MSW response handling
-// is wrapped in "setTimeout", and without this polyfill
-// you'd have to manually advance the timers for the response
-// to finally resolve.
-export const setTimeout = nodeSetTimeout
```
kettanaito, line 8: Not sure this is safe to remove. While we don't support Jest anymore, Vitest also relies on fake timers so removing this potentially reintroduces the "you have to manually advance the fake timers when mocking delay with MSW" issue.

kettanaito: I will check the tests we have, and if we don't have any specifically for this, add them and verify.

kettanaito: Confirmed: polyfill can be removed. Users should advance the fake timers normally if using delayed responses.

